PR #2069, opened 2018-02-09
Use waveform_transforms in pycbc_create_injections

soumide1102:
This PR : Allows `pycbc_create_injections` to use `waveform_transforms` from the config file. Also saves the additional parameters that are returned after applying the transforms, to the output hdf file.

### DIFF
--- a/bin/pycbc_create_injections
+++ b/bin/pycbc_create_injections
@@ -239,15 +239,19 @@ variable_args, static_args, constraints = \
                                      prior_section=opts.dist_section)
 
 if cp.has_section('replace_parameters'):
-    replace_args, out_args = option_utils.read_sampling_args_from_config(cp,
+    replace_args, replace_out_args = option_utils.read_sampling_args_from_config(cp,
                                section='replace_parameters')
     write_transforms = transforms.read_transforms_from_config(cp, 'transform')
     write_args = [arg for arg in variable_args if arg not in replace_args]
-    write_args += out_args
+    write_args += replace_out_args
 else:
     write_args = variable_args
     write_transforms = None
 
+if any(cp.get_subsections('waveform_transforms')):
+    waveform_transforms = transforms.read_transforms_from_config(cp, 'waveform_transforms')
+else:
+    waveform_transforms = None
 
 # get prior distribution for each variable parameter
 logging.info("Reading distributions")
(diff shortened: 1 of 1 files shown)


soumide1102:
@cdcapano Do you want the `[sampling_transforms-..]` section to be used both in create_injections as well as the PE code? If so, suppose we are running a PP workflow where the PE code is drawing the priors from say (mass1, mass2) and sampling in (mchirp, q), create_injections will also also draw the injection params from (mass1,mass2) and then save the (mchirp,q) values too in the injection output file. Whereas, in this case the waveform generator does not need (mchirp,q) to generate the injections. So, do you think we could put in an argument something like `--use-sampling-transform`  in create_injections to give the user a choice whether or not to read `[sampling_transforms-..]` from the config file.

cdcapano:
@soumide1102 No, `create_injections` should just ignore the `sampling_transforms` section. The sampling parameters are purely related to PE: they are used to make the sampler converge faster. That really has nothing to do with creating injections.

soumide1102:
@cdcapano So do you want the `[transform-..] section to be kept the same in create_injections, and just remove `replace_parameters` so that nothing is replaced, and *all* the variable args + the transformed ones are saved in the injection file? 
Sorry, I thought you wanted it to be the same as [sampling_transforms..] when you said "transforms section be renamed to be the same as in the PE code."

cdcapano:
@soumide1102 

> So do you want the [transform-..] section to be kept the same in create_injections

Well, I do want it renamed to `waveform_transforms`, as you've already done, so that it's the same as in the PE code.

> and just remove `replace_parameters` so that nothing is replaced, and all the variable args + the transformed ones are saved in the injection file?

Yes, that's right.

soumide1102:
Ah, okay. I see. I'll make the changes.

soumide1102:
@cdcapano I've removed the code for the replace_parameters section and made pycbc_create_injections to be able to use static args if waveforms_transforms is used to generate new parameters. Also modified the docs. Could you please take a look?

bin/pycbc_create_injections
```diff
@@ -22,8 +22,7 @@ Config file syntax
 ------------------
 The configuration file should have a [variable_args], a [static_args],
 and one or more [distribution] sections. Multiple [constraint] sections and
-a [replace_parameters] and one or more [transform] sections may also be
-provided. An example:
+one or more [waveforms_transforms] sections may also be provided. An example:
```
cdcapano: `[waveforms_transforms]` --> `[waveform_transforms]`

bin/pycbc_create_injections
```diff
@@ -89,11 +84,10 @@ arguments to initialize that distribution; see the distributions module for
 details.
 
 The variable_args need not be parameters understood by the waveform module.  In
-that case, a [replace_parameters] section and corresponding [transform]
-sections must be provided that map variable_args not understood by the waveform
-module to parameters that are understood. In the above example, spin1_a,
-spin1_azimuthal, and spin1_polar are converted to spin1x, spin1y, and spin1z
-before being written out. Any transform in the transforms module may be used to
+that case, a [waveforms_transforms] sections must be provided that map variable_args
```
cdcapano: `a [waveforms_transforms] sections` --> `one or more [waveform_transforms] sections`

soumide1102:
Thanks for spotting the typo! :)

PR #2074, opened 2018-02-12
Fix checking for required args in waveform

cdcapano:
By making required arguments have defaults of `None`, PR #2050 caused required arguments to not actually be checked for in `get_td_waveform` and `get_fd_waveform`. The result is that if you do not pass a required argument, you get unhelpful error messages about `None` not being a float. This fixes that by moving the check for required arguments to the `props` function, before the `input_params` are populated with default values.

This also adds `f_lower` to the list of required arguments, which we had forgotten about in PR #2050.


cdcapano:
Yeah, I tested it with `get_(f|t)d_waveform`, both providing the right number of parameters and not. Merging... I'll pursue a gold star in another PR.

PR #2080, opened 2018-02-19
Updated psd.estimate.welch to allow window as ndarray

duncanmmacleod:
This PR modifies `pycbc.psd.estimate.welch()` to accept `window` as a `numpy.ndarray`. This allows users to specify their own arbitrary window, without requiring pycbc to handle more name->function mappings.

The sanity checks have been updated to fail if

- `window` is given as an array that doesn't match `seg_len`
- `window` is not an array, and isn't in the `window_map`

### DIFF
--- a/pycbc/psd/estimate.py
+++ b/pycbc/psd/estimate.py
@@ -65,7 +65,8 @@ def welch(timeseries, seg_len=4096, seg_stride=2048, window='hann',
     seg_stride : int
         Separation between consecutive segments, in samples.
     window : {'hann'}
-        Function used to window segments before Fourier transforming.
+        Function used to window segments before Fourier transforming, or
+        a `numpy.ndarray` that specifies the window.
     avg_method : {'median', 'mean', 'median-mean'}
         Method used for averaging individual segment PSDs.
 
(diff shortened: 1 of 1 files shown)


ahnitz:
@duncanmmacleod I've added @titodalcanton as the reviewer, since he originated much of this code. Otherwise, looks like a nice feature.

pycbc/psd/estimate.py
```diff
@@ -65,7 +65,8 @@ def welch(timeseries, seg_len=4096, seg_stride=2048, window='hann',
     seg_stride : int
         Separation between consecutive segments, in samples.
     window : {'hann'}
-        Function used to window segments before Fourier transforming.
+        Function used to window segments before Fourier transforming, or
+        a `numpy.ndarray` that specifies the window.
```
tdent: I'm a bit uneasy about having a single argument that can be passed either a string or a ndarray and then needs a bunch of 'suck it and see' tests to work out what to do.  Personally would prefer to have two arguments, e.g. window_name with default 'hann' and window_array with default None - that would (IMO) make the logic easier to read.  (User should set window_name explicitly to None in order to use a ndarray window.)

duncanmmacleod: @tdent, I would argue that as long as the function documentation is up to scratch, having a multi-typed argument is fine, but then again, I'm not a maintainer of this package...

duncanmmacleod:
@titodalcanton, presuming you mean scientific tests (rather than unit tests of this change), yes, we have seen for GEO600 a big difference between hann and kaiser windows, as a result of the shape of the GEO spectrum.

titodalcanton:
Interesting, can you send a plot or reference on Slack?

PR #2079, opened 2018-02-19
PSD variation code

lauranuttall:
This pull request is for the PSD variation code I've been implementing along Alex's suggestions. 

This code calculates a PSD every 8 seconds with a 4 second overlap. It then calculates a 512 second PSD and divides the two, looking for variations between the longer and shorter PSD. Then by summing up the differences over a frequency range we care about (20-512Hz) we can gain a 'PSD variation statistic' which quantifies how different the PSDs are. This statistic value has been recorded for each trigger so that this value is saved at the output. The variable is called psd_var_val. 

I've tested this code with the usual pycbc_inspiral settings over a day's worth of data. Everything seems to be saved fine. I've also tested the code turning off these new settings and everything works fine. I'm currently just checking now how well this statistic picks up non-stationarity.

PR #2086, opened 2018-02-26
Revert "PSD variation code"

lauranuttall:
Reverts ligo-cbc/pycbc#2079 so it doesn't break pycbc_inference


soumide1102:
Can we merge this, please? :)

PR #2093, opened 2018-03-06
Fix Debian4 Einstein@Home/OSG build

bema-ligo:
These two commit fix issues with the PyCBC build for Einstein@Home on Debian 4:

- 'wget' links to the system's openssl library, which is too old for github's ssl (which the script uses to determine the 'blessed' version of LALSuite). There is a version of 'curl' built manually and linked with a newer version of openssl on that machine. Howver, 'curl' is not installed e.g. on the Cygwin build system. So I introduced a (shell function) wrapper that uses 'curl' when available and falls back to 'wget' if not.

- Recent versions e.g. of matplotlib require the Python module 'bz'. This requires the package `libbz2-dev` to be installed on the machine, so I added this to the catalog of packages to be installed when setting up the machine.

### DIFF
--- a/tools/einsteinathome/pycbc_build_eah.sh
+++ b/tools/einsteinathome/pycbc_build_eah.sh
@@ -325,6 +325,15 @@ wget_opts="-c --passive-ftp --no-check-certificate --tries=5 --timeout=30"
 export GIT_SSL_NO_VERIFY=true
 export PIP_TRUSTED_HOST="pypi.python.org github.com"
 
+function wwwget {
+  # use curl when available, else wget
+  if curl --version >/dev/null; then
+    curl -L "$@"
+  else
+    wget $wget_opts -O- "$@"
+  fi
+}
+
 # handle command-line arguments, possibly overriding above settings
 for i in $*; do
     case $i in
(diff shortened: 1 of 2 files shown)


duncan-brown:
@bema-ligo looks like the test of `pycbc_inspiral` is failing with
```
ImportError: No module named bz2
```

bema-ligo:
@duncan-brown that's funny, because this is the exact error I encountered and wrote this PR to fix. The point here is that 'bz' is used by matplotlib.cbook; possibly only in a recent version. And 'bz' is a standard Python module. If this error happens on a system where Python is compiled from sources, this means that also on this system `libbz-dev` needs to be installed and the Python recompiled to build the 'bz' module, as I did on the Debian 4 system. This PR can hardly introduce the dependency on 'bz', instead it fixes it for one system. This issue should come up on the same system with other branches / PRs as well; if it hasn't so far (possibly because the version of matplotlib was updated only recently), it sure will soon.

bema-ligo:
I'll address the Codacy complaint, though.

ahnitz:
@duncan-brown That was not introduced by any of our patches. It seems like a change to the build environment. Do you have any ideas?

bema-ligo:
The traceback is

```
  File "/pycbc/build/pycbc-build/environment/bin/pycbc_inspiral", line 23, in <module>
    from pycbc import vetoes, psd, waveform, strain, scheme, fft, DYN_RANGE_FAC, events
  File "/pycbc/build/pycbc-build/environment/lib/python2.7/site-packages/pycbc/strain.py", line 33, in <module>
    from pycbc.filter.zpk import filter_zpk
  File "/pycbc/build/pycbc-build/environment/lib/python2.7/site-packages/pycbc/filter/zpk.py", line 27, in <module>
    from pycbc import future
  File "/pycbc/build/pycbc-build/environment/lib/python2.7/site-packages/pycbc/future.py", line 1773, in <module>
    from matplotlib.colors import ListedColormap
  File "/pycbc/build/pycbc-build/environment/lib/python2.7/site-packages/matplotlib/__init__.py", line 127, in <module>
    from . import cbook
  File "/pycbc/build/pycbc-build/environment/lib/python2.7/site-packages/matplotlib/cbook/__init__.py", line 13, in <module>
    import bz2
ImportError: No module named bz2
```
so my bet would be on commit 4a39ba25 / #1369 (Jan 19), which introduced the import from `matplotlib.colors` and subsequently the use of `bz`.

bema-ligo:
@ahnitz @duncan-brown most likely this dependency was introduced with matplotlib 2.2.0, which was published on 2018-03-06 (on pypi). The previous version (2.1.2) probably didn't have it. The script gets the version to install from requirements.txt, which currently only lists "matplotlib>=1.3.1".

bema-ligo:
@duncan-brown Now that  as I suspect - the missing package has been installed and Python rebuilt on the previously faily system, is there anything left against merging this PR?

PR #2096, opened 2018-03-07
Updates to plot_inj_interval

cdcapano:
This patch does a few things to `pycbc_inference_plot_inj_interval`:
 * Drops the quantiles option, instead using `scipy.stats.percentileofscore` to calculate the x and y axes. This results in a more fine-grained line that always terminates at (1,1), as expected.
 * Adds a KS test, and reports the results in the legend.
 * Adds metadata to the plot so it'll appear in the results pages.
 * Casts the injection table to a `FieldArray` so that the same functions can be used on both the samples and the injection parameters (e.g., currently, you'll get an error if you try to use the `primary_mass` function).

Old plot:
https://www2.atlas.aei.uni-hannover.de/~work-cdcapano/projects/pycbc_inference/pp_tests/emcee_pt/nonspinning_phenomd_n100-extra_plots/mass1_pp.png

New plot:
https://www2.atlas.aei.uni-hannover.de/~work-cdcapano/projects/pycbc_inference/pp_tests/emcee_pt/nonspinning_phenomd_n100-extra_plots/mass1_new_pp.png

Compare to Fig. 10 of https://arxiv.org/abs/1409.7215.

### DIFF
--- a/bin/inference/pycbc_inference_plot_inj_intervals
+++ b/bin/inference/pycbc_inference_plot_inj_intervals
@@ -10,6 +10,7 @@ import matplotlib.colorbar as cbar
 import matplotlib.pyplot as plt
 import numpy
 import pycbc
+from scipy import stats
 from matplotlib import cm
 from pycbc import inject
 from pycbc import transforms
@@ -23,9 +24,6 @@ parser.add_argument("--output-file", required=True, type=str,
                     help="Path to save output plot.")
 parser.add_argument("--verbose", action="store_true",
                     help="Allows print statements.")
-parser.add_argument("--quantiles", nargs="+", type=float,
-                    default=numpy.arange(0.05, 1.05, 0.05),
-                    help="List of quantiles to plot.")
 parser.add_argument("--injection-hdf-group", default="H1/injections",
                     help="HDF group that contains injection values.")
 option_utils.add_inference_results_option_group(parser)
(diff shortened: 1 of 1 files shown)


cmbiwer:
Using ``percentileofscore`` isn't really the same thing though. Its more refined because you have more data points than was originally set by ``--quantiles``, though what I care more about is how it changes the compute time. If I have a 1000 injections drawn from distributions and run this code with the PR, its probably going to have ~1000 points to calculate the credible intervals. How long did it take to generate each plot?

cdcapano:
@cmbiwer I didn't notice any difference in run time. Using bash's `time` on ```pycbc_inference_plot_inj_intervals --parameters mass1 --output-file extra_plots/time_new.png --verbose --input-file `ls output/H1L1-INFERENCE_*hdf` ```, I get:

 * before patch:
```
real	0m4.575s
user	0m6.064s
sys	0m6.068s
```
 * after patch:
```
real	0m4.827s
user	0m5.236s
sys	0m5.884s
```
So, negligible difference in times. I should note that these timing tests were done after I had run the executable once. The first time running takes much longer because of reading the data. At least on atlas, after the first time you've read the data, the files are cached, so subsequent reads are much faster. Even so, reading the data is the dominant cost in this program, and that is unchanged by this patch.

PR #2103, opened 2018-03-12
Fix Lambda label

soumide1102:
According to the general convention in literature, \lambda is the deformability parameter which has units [gm cm^2 s^2], amd \Lambda is the dimensionless tidal deformability parameter. What we use in our calculations in pycbc is \Lambda. This label is also used in plotting PE posteriors, so using the correct symbol might help in avoiding any confusion.


soumide1102:
Thanks, @spxiwh .

PR #2109, opened 2018-03-15
Adds wrapper script for pycbc_inference

cdcapano:
This adds a bash script to side-step condor's issue of setting `OMP_NUM_THREADS`. The script manually sets `OMP_NUM_THREADS=1` before running `pycbc_inference`, thereby overriding the environment variable. This script should be used in workflows instead of `pycbc_inference` until the bug in condor can be fixed. (See [this ticket](https://htcondor-wiki.cs.wisc.edu/index.cgi/tktview?tn=6606,14) for details.)


ahnitz:
Change was trivial, and unfortunately, needed on ATLAS.

ahnitz:
@cdcapano Actually, you should add this to setup.py as well, so it is installed and in the user path.

cdcapano:
@ahnitz Ok, I added it to `setup.py`.

PR #2111, opened 2018-04-02
Fix bug in start index calculation and simplify inline linear; allow compress bank to run on compressed files

cdcapano:
This patch does two things:

1. Fixes a bug in the calculation of the starting index used for `inline_linear`. Currently, the floor of `f_lower / df` is used when figuring out what index to start the interpolation, whereas it should be the ceil. This fixes inconsistencies between `inline_linear` and scipy's `linear` for the starting index that's uses for some templates.

2. Allows `pycbc_compress_bank` to read bank files that have compressed waveforms in them. The old compressed waveforms are ignored.

### DIFF
--- a/bin/pycbc_compress_bank
+++ b/bin/pycbc_compress_bank
@@ -139,7 +139,8 @@ if args.tmplt_index is not None:
 
 # generate output file
 logging.info("writing template info to output")
-output = bank.write_to_hdf(args.output, force=args.force)
+output = bank.write_to_hdf(args.output, force=args.force,
+                           write_compressed_waveforms=False)
 
 # get the psd
 logging.info("getting psd")
--- a/bin/pycbc_compress_bank
+++ b/bin/pycbc_compress_bank
@@ -129,7 +129,8 @@ dtype = numpy.complex128
 
 bank = waveform.FilterBank(args.bank_file, N/2+1, df, dtype,
                            low_frequency_cutoff=args.low_frequency_cutoff,
-                           approximant=args.approximant)
+                           approximant=args.approximant,
+                           enable_compressed_waveforms=False)
 templates = bank.table
 if args.tmplt_index is not None:
     imin, imax = args.tmplt_index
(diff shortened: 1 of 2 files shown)


cdcapano:
@josh-willis Ok, I think I found a fix for the last point not being populated with the custom inline interpolation. See the last commit.

josh-willis:
So, it doesn't look to me like the Travis build failures are actually issues with the changes; is there a way to re-trigger the build?

cdcapano:
@josh-willis The second error being picked up by Travis was a bug due to my last commit. With that change, the interpolation code could go past the last sample in the output vector if the sample frequencies had larger frequencies in them, causing a segfault.

Anyway, I had enough with all the crappiness in my original interpolation code, so I largely re-wrote it... see the new commit. Instead of cycling over the compressed frequencies, trying to figure out where to write to the output, I now cycle over the output, and just update the interpolation when needed. This fixes both the issue with the inline code terminating too early, and insures that it can't go past either the sample frequencies nor the output vector. I tested on the template you had given me along with the template that caused a segfault in Travis, and both give nearly identical results to using scipy's interpolation.

josh-willis:
Sorry it's taken a while to review this.  I tested this out on the single template bank I'd been using, and even after recompressing that single template bank, I'm still seeing the compressed waveforms (whether `inline_linear` or just `linear`) ending one sample beyond the uncompressed waveform. Do you understand why this is?  Also, even though I compressed the bank at a tolerance of 1e-6, the actual relative difference in norms is about 1e-3 for this template; should that be possible?

Finally, a couple of questions about the code. On line 374 of the new version of ` pycbc/waveform/compress.py` you have the line (in the inline code):
```
memset(outptr, 0, sizeof(*outptr)*2*kmin);
```
and similarly in line 444:
```
memset(outptr, 0, sizeof(*outptr)*2*(hlen-kmax));
```
What is the purpose of the factor of 2 in these lines? Since `*outptr` is a double (cast from single complex), doesn't this take care of the factor of two?

josh-willis:
So, I looked over this some more and understand what you're doing with the pointer cast (which is not what I thought you were doing). The issue I'm seeing with relative difference in norms is attributable to using the actual norms (in linear algebra sense) not overlap; the overlaps are fine.  I still don't know why the ending index is not the same between the compressed and regular waveforms, but will approve.

PR #2113, opened 2018-04-03
Add uniform component masses distribution

cdcapano:
This addes a `UniformComponentMasses` distribution. This is mostly a wrapper around `Uniform`, but it requires that `mass1` and `mass2` be provided, and it ensures that `mass1 >= mass2` in the array returned by the `rvs` function.  This should fix annoying book-keeping issues when doing PP tests.

I stuck the distribution in a new module. If we add more mass-specific priors in the future, they can be added to this module.

### DIFF
--- a/examples/distributions/example.ini
+++ b/examples/distributions/example.ini
@@ -16,8 +16,8 @@ v7 =
 v8 =
 v9 =
 ;v10 =
-;mass1 =
-;mass2 =
+mass1 =
+mass2 =
 ;chi_eff =
 ;chi_a =
 ;xi1 =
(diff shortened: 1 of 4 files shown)


cdcapano:
@cmbiwer Ok, I removed the unused import. I also added a `param` optional argument to `rvs`, since codacy was complaining about it.

cmbiwer:
Hey @cdcapano looks like Travis is failing with:
```
Warning, treated as error:
/home/travis/build/pycbc-build/environment/lib/python2.7/site-packages/PyCBC-dbf531-py2.7.egg/pycbc/distributions/masses.py:docstring of pycbc.distributions.masses.UniformComponentMasses:8:Explicit markup ends without a blank line; unexpected unindent.
```
Not sure what its getting at but its causing Travis to fail.

cdcapano:
@cmbiwer Ok, I got it to pass all of the Travis tests.

PR #2116, opened 2018-04-04
allow injection of fourier domain approximants

ahnitz:
This allows injection of fourier domain approximants by exposing them all through the get_td_waveform interface. This is done in a way that ensures no time offset / phase offset is introduced in the conversion from fd to td. 

These are added with a new approximant name to differentiate.

The following all generated IMRPhenomD in the time domain now, but the get_td_waveform version will apply reasonable tapering at the start. 

In case anyone is wondering, injection in the time domain is still desired to avoid wraparound issues. An example would be where we want only part of the waveform to be injected since we are looking at a segment of strain that only contains a part of the signal. The waveform should *not* wrap around to the other end of the segment. Ensuring this is straightforward in the time domain.

```
hp, hc = get_td_waveform(approximant="IMRPhenomD_FD", mass1=10, mass2=10,
                                 f_lower=30, delta_t=1.0/4096)

rp, cc = get_td_waveform(approximant="IMRPhenomD", mass1=10, mass2=10,
                                 f_lower=30, delta_t=1.0/4096)

sp = get_fd_waveform(approximant="IMRPhenomD", mass1=10, mass2=10,
                                 f_lower=30, delta_f=hp.delta_f)[0].to_timeseries()
```
depends on #2115

### DIFF
--- a/pycbc/waveform/waveform.py
+++ b/pycbc/waveform/waveform.py
@@ -512,7 +512,8 @@ def get_td_waveform_from_fd(**params):
     Paramaters
     ----------
     params: dict
-        The parameters defining the waveform to generator. See `get_fd_waveform`.
+        The parameters defining the waveform to generator.
+        See `get_fd_waveform`.
 
     Returns
     -------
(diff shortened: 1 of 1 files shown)


pycbc/waveform/waveform.py
```diff
@@ -503,6 +503,63 @@ def get_fd_waveform(template=None, **kwargs):
...
+
+    if 'f_fref' not in nparams:
+        nparams['f_ref'] = params['f_lower']
+
+    # factor to ensure the vectors are all large enough. We don't need to
+    # completely trust our duration estimator in this case, at a small
+    # increase in computational cost
+    rwrap = 0.2
+    fudge_duration = (max(0, full_duration) + .1 + rwrap) * 1.5
+    nparams['delta_f'] = 1.0 / fudge_duration
```
cdcapano: I think you need to round `fudge_duration` to the nearest power of two here. Otherwise, the memory usage might grow a bunch if you used this multiple times with different parameters, due to the FFT-plan caching that silently happens in `to_timeseries`.

PR #2118, opened 2018-04-05
Use a keytab to access LIGO.ORG protected services

ahnitz:


ahnitz:
I'm somewhat incredulous that this actually worked... @duncan-brown Is it possible that credentials are partially saved between runs? Do I need to clear a cache to test this properly?

ahnitz:
Ah... I see the problem. It won't run the test in question unless on master due to needing the encrypted information.

duncan-brown:
@ahnitz ah, that's right. I'll approve this so you can merge, but you may need to fix master if it doesn't work.

PR #2123, opened 2018-04-09
Move functions for reading parameters from config to distributions module

cdcapano:
This moves `read_args_from_config` from `inference/option_utils.py` to `distributions/__init__.py`. It also moves the functions to add config arguments / parse them from `inference/option_utils.py` to methods of `WorkflowConfigParser`. This removes `pycbc_create_injections` dependency on the `inference` module.


cdcapano:
@ahnitz @cmbiwer Sorry this patch has grown so much... applying the initial moves uncovered some circular dependencies between the `distributions` and `transforms` modules. To fix them, I moved `distributions/boundaries.py` up to `pycbc/`, and moved `bounded.VARARGS_DELIM` into `pycbc/__init__.py`. Hopefully that will do it.

PR #2125, opened 2018-04-11
Modified ParameterList to use OrderedDict for dicts

duncanmmacleod:
This PR modifies `pycbc.waveform.ParameterList` to use `collections.OrderedDict` for all `ParameterList.xxx_dict` properties, to preserve the order of parameters.

### DIFF
--- a/pycbc/waveform/parameters.py
+++ b/pycbc/waveform/parameters.py
@@ -25,6 +25,7 @@
 """Classes to define common parameters used for waveform generation.
 """
 
+from collections import OrderedDict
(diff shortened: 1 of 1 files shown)


ahnitz:
@duncanmmacleod What's the purpose of the ordering here? Is it just to preserve style when printing out arguments? I just want to make sure we aren't really going to need to assume some ordering for something important.

duncanmmacleod:
@ahnitz, for my purposes the ordering is just for printing via the dict property, so not important.

PR #2129, opened 2018-04-11
Use string_types for string type checking

duncanmmacleod:
This PR updates string type checks for python3 by replacing references to unicode with `six.string_types` where appropriate.

**This adds a dependency on `six`** which has been added to `setup.py` and `requirements.txt`.

I deliberately didn't update `pycbc.ligolw` because `glue.ligolw` has already been patched.


ahnitz:
@duncanmmacleod Normally I like to avoid dependency creep, but six is essential for supporting python 2/3, so adding this is fine.

PR #2132, opened 2018-04-16
Update citation guide with Nitz+ 2017

titodalcanton:


titodalcanton:
@ahnitz do you know the meaning of the error? The docs build ok on my laptop. I tried restarting one of the Travis jobs and it still fails. Do you think it is safe to merge?

PR #2137, opened 2018-04-18
Document how Travis secure strings are constructed

couvares:


ahnitz:
@couvares @duncan-brown Poke

PR #2139, opened 2018-04-18
Fix bug in colored_noise function

micamu:
The colored noise function in the noise module has a default value for the low frequency cutoff. According to the documentation, that value us 10. Hz, but according to the function's arguments, that value is 1.0 Hz. 
This looks like a bug and it probably should be 10 Hz, so in this PR I am changing the function's argument to be low_frequency_cutoff=10.0

### DIFF
--- a/pycbc/noise/reproduceable.py
+++ b/pycbc/noise/reproduceable.py
@@ -79,7 +79,7 @@ def normal(start, end, seed=0):
     ts = TimeSeries(data, delta_t=1.0 / SAMPLE_RATE, epoch=start)
     return ts.time_slice(start, end)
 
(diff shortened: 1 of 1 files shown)


micamu:
@ahnitz That depends on what you want the default value to be. I assumed 10 Hz sounded more reasonable, but if you think the default should be 1 Hz then I'll change the documentation instead

ahnitz:
Well, it seems like a smaller change to leave the default the same and simply fix the documentation to indicate what it is *actually* doing. For 3g detectors for example, 1 Hz is not a bad default.

micamu:
ok, I'll change that

PR #2141, opened 2018-04-19
Switch PyCBC to using keytabs

ahnitz:
I need to test this from a branch on the gwastro repo to allow the secure key required portions to run.


duncan-brown:
@ahnitz is this still relevant?

ahnitz:
Should leave it for now, or remove our testing that requires ligo.org access and do that some other way.

duncan-brown:
@ahnitz I think we need to resurrect this as `ligo-proxy-init` seems to have changed and it looks like Travis is failing when it tries to apply the patch.

ahnitz:
obsoleted

PR #2143, opened 2018-04-20
Pygrb o2 dev

a-r-williamson:
These changes are to bring O2-related tweaks to the `pygrb` pipeline onto master. In particular:
* improve the behaviour when running with 'long' timeslides and fix a few result page bugs related to this
* improve integration of the gating of data into the workflow

This code was used to produce results for GW170817

These changes are made in conjunction with gwastro/pycbc-pylal#45

### DIFF
--- a/bin/pygrb/pycbc_make_grb_summary_page
+++ b/bin/pygrb/pycbc_make_grb_summary_page
@@ -79,6 +79,9 @@ def parse_command_line():
     parser.add_argument("-O", "--open-box", action="store_true",
                         help ="Use to show onsource results")
 
+    parser.add_argument("-S", "--time-slides", action="store_true",
+                        help ="If run featured time slides")
+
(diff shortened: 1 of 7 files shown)


ahnitz:
@a-r-williamson The two codacy issues are probably worth fixes. Looks pretty straightforward.

pannarale:
@a-r-williamson, Looks like [this line](https://github.com/gwastro/pycbc/pull/2143/files#diff-73de4c28a59ffb261bcdbc3dfff2cf8cR345) should say "Best NR" in the title.  Perhaps you could handle those two plots in a single loop.

Regarding this function, I think the various if/else's to determine out_dir are upsetting codebeat.  Dictionary?

pannarale:
@a-r-williamson let's merge and move on now.  But let's follow up on some points raised by codebeat in a separate PR.

PR #2144, opened 2018-04-20
Psd variation - adding freq weighting 

lauranuttall:
In this merge request I have added frequency weighting to the psd variation statistic. The psd variation timeseries also automatically picks up the precision of strain and copies it.


pycbc/psd/variation.py
```diff
@@ -74,11 +80,15 @@ def calc_psd_variation(strain, psd_short_segment, psd_long_segment,
         kmin = int(low_freq / psd_long.delta_f)
         kmax = int(high_freq / psd_long.delta_f)
         # Comapre the PSD of the short segment to the long segment
-        #diff = numpy.array([numpy.std((p_short[kmin:kmax] /
-        #           psd_long[kmin:kmax])) for p_short in psd_short])
-        diff = numpy.array([(p_short[kmin:kmax] / psd_long[kmin:kmax]).sum()
-                           for p_short in psd_short])
-        diff /= (kmax - kmin)
+        freqs = FrequencySeries(psd_long.sample_frequencies,
+                                delta_f = psd_long.delta_f,
```
tdent: for kwargs, don't use spaces around the '=' sign.

pycbc/psd/variation.py
```diff
@@ -74,11 +80,15 @@ def calc_psd_variation(strain, psd_short_segment, psd_long_segment,
...
+                                epoch = psd_long.epoch, dtype = fs_dtype)
+        weight = numpy.array(
```
tdent: I suggest adding an inline comment here about what is being done with this weighting and why. 
Then the sum(step) is integrating a weighted ratio of PSDs over frequency, correct?

lauranuttall: yep that's right. I've added in a comment so it's more obvious

tdent:
Merge when the green light shows.

pycbc/psd/variation.py
```diff
@@ -35,19 +35,26 @@ def calc_psd_variation(strain, psd_short_segment, psd_long_segment,
...
+
     # Find the times of the long segments
     times_long = numpy.arange(float(strain.start_time),
                               float(strain.end_time), psd_long_segment)
 
     # Set up the empty time series for the PSD variation estimate
-    psd_var = TimeSeries(zeros(int((strain.end_time - strain.start_time) /
-                  psd_short_segment)), delta_t=psd_short_segment,
-                  copy=False, epoch=strain.start_time)
+    psd_var = TimeSeries(zeros(int(math.ceil((strain.end_time -
```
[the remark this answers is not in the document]
lauranuttall: changed

pycbc/psd/variation.py
```diff
@@ -35,19 +35,26 @@ def calc_psd_variation(strain, psd_short_segment, psd_long_segment,
         Time series of the variability in the PSD estimation
     """
 
+    # Calculate strain precision
+    if strain.precision == 'single':
+        fs_dtype = numpy.float32
+    elif strain.precision == 'double':
+        fs_dtype = numpy.float64
+
```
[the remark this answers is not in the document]
lauranuttall: ok added this in, makes things more simple as you say

PR #2147, opened 2018-04-23
Only re-calculate burn-in if needed

cdcapano:
Currently, the burn-in iteration is re-calculated at every checkpoint, even if the sampler has already burned in. This makes it so that burn-in is only calculated if the sampler hasn't burned in yet.


soumide1102:
Travis failed though. Says : 
`KeyError: "Can't open attribute (can't locate attribute: 'is_burned_in')"`

PR #2152, opened 2018-04-24
Fix pycbc_single_template_plot error on headless systems

titodalcanton:
Matplotlib's agg backend needs to be set before importing `pycbc.results` (see #2150 ). This also passes the docstring to argparse.


ahnitz:
@titodalcanton This is a fine change, so no reason not to merger. As you said, it appears some other changes will be needed anyway though.

PR #2157, opened 2018-05-02
 don't import gwpy for colors

ahnitz:


ahnitz:
gwpy as a preinstall is requiring a different version of ligolw, and due to the import structure causes pycbc code to fail. This just removes the dependency and goes back to the fixed colormap.

PR #2160, opened 2018-05-10
Simplify ACL algorithm

cdcapano:
This switches the ACL algorithm to that used in emcee (see Daniel Foreman-Mackey's blog post [here](https://dfm.io/posts/autocorr/)), which comes from notes & papers by A. Sokal. This is very similar to the current algorithm we use, but uses the entire data set for computing the ACL instead of just the first half. The function is also simplified to make it easier to read.

### DIFF
--- a/pycbc/filter/autocorrelation.py
+++ b/pycbc/filter/autocorrelation.py
@@ -107,35 +107,48 @@ def calculate_acf(data, delta_t=1.0, unbiased=False):
     else:
         return acf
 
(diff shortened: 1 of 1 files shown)


cmbiwer:
Sorry, just saw the email notification. Will look later tonight. Have you done a comparison with the current implementation (changing ``K=1``)?

cdcapano:
Yeah, I did a comparison on a set of declination samples from an injection, and got the same result using the new function and the old one.

PR #2161, opened 2018-05-10
allow for the whitening psd to be returned

ahnitz:


tdent:
I guess it's not ideal to have a function that returns a different number of objects depending on its inputs, but the way this is done is unlikely to cause problems.

PR #2168, opened 2018-05-16
Allow  conversions to be used in PP dax

cdcapano:
The changes in #2136 and #2154 means that PP plots will not work anymore since mass1 will not necessarily be >= mass2 in the samples files. In order to get sane plots, we need to be able to pass `primary_mass(mass1, mass2)` from the workflow config file to `inj_interval` and `inj_recovery`. This patch does that by adding a `pp-plot-parameters` option to `workflow-inference` section. In this section you specify what parameters to plot for the PP and recovery plots. (If nothing is provided, it defaults to the variable args.) An example:

```
[workflow-inference]                                                            
; how the workflow generator should setup inference nodes                       
num-injections = 100                                                            
data-seconds-before-trigger = 6                                                 
data-seconds-after-trigger = 2                                                  
pp-plot-parameters = 'primary_mass(mass1,mass2):$m_1$'                          
                     'secondary_mass(mass1,mass2):$m_2$'                        
                     'primary_spin(mass1,mass2,spin1_a,spin2_a):$|\chi_1|$'     
                     'primary_spin(mass1,mass2,spin1_azimuthal,spin2_azimuthal):$\theta_1^\mathrm{azimuthal}$'
                     'primary_spin(mass1,mass2,spin1_polar,spin2_polar):$\theta_1^\mathrm{polar}$'
                     'secondary_spin(mass1,mass2,spin1_a,spin2_a):$\chi_2|$'    
                     'secondary_spin(mass1,mass2,spin1_azimuthal,spin2_azimuthal):$\theta_2^\mathrm{azimuthal}$'
                     'secondary_spin(mass1,mass2,spin1_polar,spin2_polar):$\theta_2^\mathrm{polar}$'
                     ra dec polarization distance inclination tc                
plot-group-all = ${pp-plot-parameters}                                          
plot-group-masses = 'primary_mass(mass1,mass2):$m_1$' 'secondary_mass(mass1,mass2):$m_2$' mchirp q
plot-group-spins = 'chi_p:$\chi_p$' chi_eff q  
```

To get the config file to pass through the functions appropriately, I had to use `shlex.split` for parsing the parameters. I also had to change the names of the output plot files. Previously they tried to stick the parameter in the file name, which causes issues if your parameter is `primary_mass(mass1, mass2):$m_1$`. Now the file names just have `param0`, `param1`, etc. 

[...]

PR #2169, opened 2018-05-16
Use ACL-per-walker and change the N acl burn-in

cdcapano:
Following https://dfm.io/posts/autocorr/, this changes the calculation of the ACL for the MCMC samplers. Instead of finding the ACL of the average of the walkers, an ACL is found for each walker, then the mean over the ACLs is taken. Since this takes more time, the ACL calculation is now parallelized over the inference pool.

Also, this changes the `n_acl` burn-in function. Now, the first half of the chains is ignored. An ACL is computed from the second half. If the second half contains 100*ACL, then the sampler is considered to be burned in at the half-way point. I changed this because I found that switching to the mean ACL over the walkers resulted in an ACL that was ~ a factor of 10 lower than the previous method. This might be fine, but using only 10ACLs (what the `n_acl` function currently does) resulted in the sampler thinking it was burned in much too early. The bit about excluding the first half of the chain is from talking to Will Farr. It also fixes an issue with the current implementation, where the sampler would have to run for a long time after burn-in to recalculate an ACL.

This PR depends on #2168.

I'm not really certain that this is the correct way to calculate an ACL, and that this new burn in test will work. So I'm marking as a work in progress for now, while I test it out in a PP test. If it looks ok (and #2168 is merged), I'll remove the label.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -444,7 +444,7 @@ with ctx:
             if opts.n_independent_samples is not None or end >= get_nsamples \
                     or not opts.checkpoint_fast:
                 logging.info("Computing acls")
-                acls = sampler.compute_acls(fp)
+                acls = sampler.compute_acls(fp, pool=sampler.pool)
                 sampler.write_acls(fp, acls)
 
         # write to backup
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -437,7 +437,8 @@ with ctx:
                 is_burned_in = False
             if not is_burned_in:
                 logging.info("Calculating burn in")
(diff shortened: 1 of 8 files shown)


cdcapano:
Using the mean of the ACLs over the walkers looks to badly underestimate the ACL for `emcee_pt`. See, for example, the posterior plots in [this PP test](https://www.atlas.aei.uni-hannover.de/~cdcapano/projects/pycbc_inference/pp_tests/emcee_pt/louder_precessing-per_walker_acl/). Some of the injections had ACLs as small as 9, with many duplicated positions.

It still may be worth using the mean of the ACLs for other samplers (or make it optional which one to choose), but that can be done in a separate PR. So, closing this one.

PR #2170, opened 2018-05-18
Update documentation on OSG running

josh-willis:
Add changes to the documentation to better describe running on the Open Science Grid, including:

 1. The need to specify a different `LIGO_DATAFIND_SERVER`, and
 2. Different arguments to `pycbc_submit_dax` that affect whether the inspiral jobs will run on a particular combination of the local cluster, the OSG, and possibly a specific XSEDE site.

### DIFF
--- a/docs/workflow/pycbc_make_coinc_search_workflow.rst
+++ b/docs/workflow/pycbc_make_coinc_search_workflow.rst
@@ -940,9 +940,22 @@ If you are running your own build of ``pycbc_inspiral``, you will need to give
 a path to a gsiftp URL and tell Pegasus that the executable is not installed
 on the OSG with the two ``--config-overrides`` options::
 
-    'executables:inspiral:gsiftp://server.hostname/path/to/pycbc_inspiral'
+    "executables:inspiral:gsiftp://server.hostname/path/to/pycbc_inspiral"
+    "pegasus_profile-inspiral:pycbc|installed:False"
 
 Make sure this executable is build following the instructions on the page :ref:`building_bundled_executables`.
+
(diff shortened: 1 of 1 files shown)


docs/workflow/pycbc_make_coinc_search_workflow.rst
```diff
@@ -940,10 +940,23 @@ If you are running your own build of ``pycbc_inspiral``, you will need to give
 a path to a gsiftp URL and tell Pegasus that the executable is not installed
 on the OSG with the two ``--config-overrides`` options::
 
-    'executables:inspiral:gsiftp://server.hostname/path/to/pycbc_inspiral'
+    "executables:inspiral:gsiftp://server.hostname/path/to/pycbc_inspiral"
+    "pegasus_profile-inspiral:pycbc|installed:False"
```
duncan-brown: @josh-willis why are you pulling the executable via gridftp, rather than using the bundle that is deployed in CVMFS?  For example:
```
"executables:inspiral:/cvmfs/oasis.opensciencegrid.org/ligo/sw/pycbc/x86_64_rhel_6/bundle/v1.9.4/pycbc_inspiral"
```

josh-willis: @duncan-brown Well, that part of the documentation I didn't change (the line you point to). I merely noticed that the text above it referenced two options and gave only the one, so my change was to add the second option.

I assume the original intent was for someone to run a bundled executable that they had built themselves, though I've never personally tested that and don't know whether it still works. But that's what the lines above the changed ones say.

But there are instructions right above those changes which do point the user toward the CVMFS executable; this is just a "secondary" approach, I suppose.

PR #2174, opened 2018-05-24
Updates to n acl burn-in

cdcapano:
This changes the `n_acl` burn-in function to do the following:

 1. Only the second half of the chain is used to compute ACL.
 2. If the number of iterations in the second half is > `n_acl * ACL`, then the sampler is considered burned in at the half way point.

The default `n_acl` is set to 25. If 200 walkers are used, with a desired number of independent samples of 5000, then this means the job will complete as soon as this burn-in test passes. This method seems to work best for `ecmee_pt`.

The way the ACL is computed is not changed. This PR supersedes #2169, which did something similar, but also switched to using the mean of ACLs instead of the ACL of the mean chain. Calculating ACL that way does not seem to work with `emcee_pt`, as the ACL was severely underestimated.

This also causes `BurnIn.update` to ignore what burn-in-iteration were previously saved in the file, as that was problematic for this test.

Depends on #2168; putting on hold until that is merged.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -119,7 +119,7 @@ parser.add_argument("--checkpoint-fast", action="store_true",
 parser.add_argument("--acl-interval", type=int, default=None,
                     help="Specify the interval on which to calculate ACL and "
                          "burn-in. If not provided, will use the "
-                         "checkpoint-interval. This is doubled after 10000 "
+                         "checkpoint-interval. This is doubled after 50000 "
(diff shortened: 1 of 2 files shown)


cdcapano:
@soumide1102 Ok, I've rebased to master.

For the record, I also added an option to this patch that allows you to specify what interval to compute ACLs / burn-in. This is automatically doubled after 50000 iterations, doubled again after 100000, and doubled again after 200000. I found I needed this for excessively long runs, when the acl calculation became the dominant cost.

PR #2185, opened 2018-05-30
Update Dockerfile to use v1.7-8cbd1b7

duncan-brown:
Update the Docker file image to use a lalsuite build that uses hash 8cbd1b7


duncan-brown:
Now redundant as the hash has been updated.

PR #2193, opened 2018-06-08
PyCBC Live: reorganize GraceDB I/O code

titodalcanton:
Several changes related to PyCBC Live and in particular the SingleCoincForGraceDB class.

* Make the class easily reusable by third party scripts as a simpler interface to GraceDB and BAYESTAR. In particular, the calculation of the SNR series is moved away from the `pycbc.io.live` module and into `pycbc_live` itself. The API is also simplified—at least in my opinion ;)
* Add a unit test for the class trying different combinations of detectors.
* Switch from having the event and PSDs in separate XML files to having them all in the same file. A copy of the single XML file is still uploaded as a separate `psd.xml` file, to avoid breaking things in GraceDB and follow-up processes. We can remove that redundant file once the rest of the chain is ready. Note that BAYESTAR is already able to take a single XML file as input (just with a slightly awkward command line).
* Fix an error related to module `six`.

Tested with simulated data and a few combinations of detectors.


bin/pycbc_live
```diff
@@ -77,16 +77,50 @@ class LiveEventManager(object):
...
-            event = SingleCoincForGraceDB(
-                    ifos, coinc_results, data_readers=data_readers,
-                    bank=bank, upload_snr_series=args.upload_snr_series,
-                    followup_ifos=followup_ifos,
-                    gracedb_server=args.gracedb_server)
+
+            event_kwa = {'bank': bank,
+                         'gracedb_server': args.gracedb_server}
+
+            if args.upload_snr_series:
```
ahnitz: Should we still allow it to be optional to upload the snr time series? Perhaps we should just require it and drop all the extra logic?

titodalcanton: I am trying to imagine a weird problem with a future, loud, multidetector event which forces us to disable the SNR time series and just use the trigger point estimates (we know how nature is). In that case however I think it should be BAYESTAR's responsibility to have the option to ignore the SNR series. So yes, I will just make it permanent from the PyCBC side.

PR #2201, opened 2018-06-18
pass down lalsuite hash

ahnitz:


ahnitz:
@duncan-brown How so? The script uses half a dozen variables which are already assumed to be pulled from the travis environment. It doesn't use options for any of them atm.

duncan-brown:
@ahnitz all the parameters (including the ones from the Travis env) that the docker container gets are passed as command line arguments to the script `docker_build_dist.sh`
```
bash /pycbc/tools/docker_build_dist.sh --pycbc-container=${PYCBC_CONTAINER} --pull-request=${TRAVIS_PULL_REQUEST} --commit=${TRAVIS_COMMIT} --secure=${TRAVIS_SECURE_ENV_VARS} --tag=${TRAVIS_TAG}
```
your patch would mix the method. Some would be command line arguments and some would be docker container environment variables. The docker container can't see *any* of the Travis variables, so they all need to be passed down.

ahnitz:
Ok, I see. I thought you were talking about the script that was being edited (i.e. docker build and test). There are too many similarly named scripts...

PR #2204, opened 2018-06-19
Added option to provide a filter string in pycbc_plot_singles_vs_params

tjma12:
This PR allows for custom trigger filtering in `pycbc_plot_singles_vs_params`.

The functionality of the `--min-snr` argument is retained by combining that option with `--filter-string` to make an overall trigger filter.

Code ran successfully when providing `--min-snr 5.` and `--filter-string "(self.snr < 100) & (self.mass1 > 30)"` as individual arguments, one at a time, and with neither provided.


tjma12:
@tdent That's effectively the current behavior. If the user doesn't provide a min-snr that is strictly > 0, the filter string is not constructed. The SNR filter variable is set to None and isn't picked up in the construction of `filter_func`.

tdent:
Yes, but I was thinking a specific numerical default value (0) is unnecessary.  Up to you.

PR #2219, opened 2018-06-27
https url for lalsuite

soumide1102:
The ssh method does not work for everybody, so mentioning the https option also.


ahnitz:
@soumide1102 Why not make it the default?

soumide1102:
@ahnitz Okay, done.

PR #2223, opened 2018-07-05
Add optimal detector functionality to conversions

sfield17:
Allows one to inject optimally oriented signals by adding existing optimal detector code to conversions.py file.

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -626,6 +626,94 @@ def _det_tc(detector_name, ra, dec, tc, ref_frame='geocentric'):
 
 det_tc = numpy.vectorize(_det_tc)
 
+def _detector_to_optimal_orientation(detector_name, tc):
+    """ Low-level function to be called from _optimal_dec_from_detector
+    and _optimal_ra_from_detector"""
+
+    d = Detector(detector_name)
+    ra, dec = d.optimal_orientation(tc)
+
+    return ra, dec
+
+
+def _optimal_dec_from_detector(detector_name, tc):
+    """For a given detector and GPS time, return the optimal orientation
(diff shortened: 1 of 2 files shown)


pycbc/conversions.py
```diff
@@ -626,6 +626,94 @@ def _det_tc(detector_name, ra, dec, tc, ref_frame='geocentric'):
 
 det_tc = numpy.vectorize(_det_tc)
 
+def _detector_to_optimal_orientation(detector_name, tc):
```
cdcapano: Just to be consistent with names in the rest of the module, change `_detector_to_optimal_orientation` to `_optimal_orientation_from_detector`.

pycbc/conversions.py
```diff
@@ -626,6 +626,94 @@ def _det_tc(detector_name, ra, dec, tc, ref_frame='geocentric'):
...
+    Example:
+
+    >>> pycbc_create_injections --config-files params.ini
+    >>> cat params.ini
+
+    [waveform_transforms-ra+dec]
+    name = custom
+    inputs = tc
+    ra = optimal_ra_from_detector("H1",tc)
+    dec = optimal_dec_from_detector("H1",tc)
```
cdcapano: Drop the example text, since that about how to run `pycbc_create_injections` (there is documentation about that in that program's help). Instead, give a definition of what "optimal" means: directly over head of the detector.

pycbc/conversions.py
```diff
@@ -626,6 +626,94 @@ def _det_tc(detector_name, ra, dec, tc, ref_frame='geocentric'):
...
```
cdcapano: Same comment as above.

PR #2226, opened 2018-07-06
add in array_cpu stuff

ahnitz:
This moves some of the hand written array operations over to cython (I'm going through the things that I had already written up to this point).

### DIFF
--- a/pycbc/types/array_cpu.pyx
+++ b/pycbc/types/array_cpu.pyx
@@ -28,10 +28,8 @@ import numpy as _np
 from pycbc.types.array import common_kind, complex128, float64
 from . import aligned as _algn
 from scipy.linalg import blas
-from pycbc.weave import inline
-from pycbc.opt import omp_libs, omp_flags
-from pycbc import WEAVE_FLAGS
 from pycbc.types import real_same_precision_as
+cimport cython, numpy
 
 def zeros(length, dtype=_np.float64):
     return _algn.zeros(length, dtype=dtype)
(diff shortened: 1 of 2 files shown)


pycbc/types/array_cpu.pyx
```diff
@@ -113,27 +84,47 @@ def weighted_inner(self, other, weight):
...
+    
+ctypedef fused COMPLEXTYPE:
+    float complex
+    double complex
+
+def abs_arg_max_complex(numpy.ndarray [COMPLEXTYPE, ndim=1] a):
+    cdef unsigned int xmax = a.shape[0]
+    cdef double mag
+    cdef double magmax = 0
+    cdef unsigned int idx = 0
```
josh-willis: Should `mag` and `magmax` be declared double, or can they also be `REALTYPE`? So that you don't impose a conversion every time you do the comparison if you're using single precision?

ahnitz: Actually, I think I couldn't figure out how to do that, so I went with higher precision. What is happening is that it is generating a kernel for every combination of COMPLEXTYPE, since that is in the interface. It doesn't do so for REALTYPE, unless there were an argument, I think....

ahnitz: Maybe to explain better. It has to know how to resolve fused types based on the signature. It can't really know how to do so for REALTYPE, as it has nothing in the input which could let it decide how to choose which type it should be. We know that we want float with float complex, but there doesn't appear to be a syntax to do this kind of 1-1 matching up.

josh-willis: OK, when Travis finishes  will approve, assuming no errors

PR #2229, opened 2018-07-06
allow pycbc to be installed in python3

ahnitz:
This disables weave when in a python3 environment. Many things will be broken, but it allows *something* to run in python 3. Waveform generation, for example, *does* work with this patch.

### DIFF
--- a/pycbc/weave.py
+++ b/pycbc/weave.py
@@ -22,9 +22,11 @@ import os.path, sys
 import logging
 import shutil, atexit, signal
 import fcntl
-import weave.inline_tools as inline_tools
 
-_compile_function = inline_tools.compile_function
+PY3 = sys.version_info[0] == 3
+if not PY3:
+    import weave.inline_tools as inline_tools
+    _compile_function = inline_tools.compile_function
 
 ## Blatently taken from weave to implement a crude file locking scheme
 def pycbc_compile_function(code,arg_names,local_dict,global_dict,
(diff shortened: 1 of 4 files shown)


ahnitz:
The main point of this is to enable testing of functionality within python3 while porting is ongoing. For example, the SPAtmplt code is ported, and I can run the unit test for that now within python3 and it passes.

pycbc/weave.py
```diff
@@ -61,8 +63,13 @@ def pycbc_compile_function(code,arg_names,local_dict,global_dict,
...
 
-inline_tools.compile_function = pycbc_compile_function
-from weave import inline
+if not PY3:
+    inline_tools.compile_function = pycbc_compile_function
+    from weave import inline
+else:
+    def inline(*args, **kwds):
+        raise RuntimeError("Oh no! You tried to use capabilities"
+                           "s we haven't ported to python3 yet")
```
titodalcanton: Small typo in the string.

setup.py
```diff
@@ -40,14 +40,15 @@
 from setuptools import Extension, setup, Command
 from setuptools.command.build_ext import build_ext as _build_ext
 
+PY3 = sys.version_info[0] == 3
```
titodalcanton: This line is repeated in two different files. Would it makes sense to set it once somewhere common and just import it where needed? Maybe `pycbc.PY3`?

PR #2235, opened 2018-07-10
Tell PyCBC that LOSC frames are available on the OSG

duncan-brown:
Since the current set of LIGO data find servers don't do this properly yet, PyCBC's datafind module needs to know if a URL that it gets is available at other (backup) locations. This implements two changes:

 1. Only the regular LIGO frames are available for gsiftp from UNL and CIT, so force a match on `file:///cvmfs/oasis.opensciencegrid.org/ligo/frames` to add those backup URLs.
 2. The LOSC frames in `file:///cvmfs/gwosc.osgstorage.org/` are available on the OSG, so if the URL matches, add them to the OSG site.


duncan-brown:
@josh-willis tagging you in the FYI.

PR #2239, opened 2018-07-11
Make section names for distributions keyword args

cdcapano:
This makes the sections to read the variable parameters, static parameters, and constraints keyword arguments in `distributions.read_args_from_config`. The defaults are what we currently use (`variable_args`, `static_args`, and `constraint`). Since these section names can now be specified explicitly, this also gets rid of the `section_group` argument.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -211,8 +211,8 @@ with ctx:
     cp = WorkflowConfigParser.from_cli(opts)
 
     # get the vairable and static arguments from the config file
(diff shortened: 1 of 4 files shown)


cdcapano:
I decided to break the `read_args_from_config` into two separate functions: one that reads the variable and static args, and one that reads the constraints.

ahnitz:
Is this needed before a release today?

cdcapano:
It would be good if it could get in; I'm making changes to gwin that would benefit from this. Then I can point people to the current release of pycbc at the f2f next week.

PR #2241, opened 2018-07-11
directly use ligolwsegmentlist constructor

ahnitz:


lauranuttall:
this update fixes the original problem and I can now make a workflow. Not sure if you want to address the codeclimate fix. Other than that this should be merged.

ahnitz:
@lauranuttall I think in this particular example, the message can be ignored.

ahnitz:
@lauranuttall Poke

PR #2254, opened 2018-07-17
timeseries.py: change to_frequencyseries to remove _nocomplex

oshaughn:
Enables twosided output, when TimeSeries object has complex form.
(The dtype allows the user to identify if a two-sided frequency series was used: this should only be called on complex input.)

### DIFF
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -722,8 +722,7 @@ class TimeSeries(Array):
         else:
             raise ValueError('Path must end with .npy, .txt or .hdf')
                 
-    @_nocomplex
-    def to_frequencyseries(self, delta_f=None):
+    def to_frequencyseries(self, delta_f=None,twosided=False):
         """ Return the Fourier transform of this time series
         
         Parameters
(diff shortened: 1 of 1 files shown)


cdcapano:
@ahnitz is on vacation, so also assigning to @josh-willis

ahnitz:
@oshaughn Thanks for putting in a PR! I think adding this kind of functionality is nice.

Do we need to add another function in addition to the keyword argument? Maybe just the extra argument is sufficient? If not, it would be better to have one call the other to avoid the code duplication. Take a gander at the static code analyzers and you should see that there is a problem with your function (np.complex128). You might want to replace this with self.kind as a check instead so you also catch complex64 dtype. Can you verify that calling this function works? 

Since I suppose you are with @cdcapano he may also point out a few smaller PEP8 (style) type things (i.e. space after commas) and get those corrected quickly.

oshaughn:
@ahnitz : Thanks, yes clearly development during a presentation wasn't effective.  New request forthcoming  to
  * delete second function as superfluous 
 * test swapped to self.kind == 'complex'
and otherwise fix minor issues.

FYI, similar changes coming to frequencyseries (get_sample_frequencies most notably)

oshaughn:
@ahnitz @cdcapano : Apparently the apostrophe in my name in the git commit causes problems in pycbc/version.py.

PR #2266, opened 2018-07-25
Add low-frequency cutoff option for every detector

soumide1102:
This PR adds the ability to read in different low-frequency cutoffs for every detector to pycbc inference. So one can provide for eg. : `--low-frequency-cutoff H1:20 L1:25` in the CLI. If a single low-frequency cutoff is provided, for eg. : `--low-frequency-cutoff 20`, that'll be used for all the detectors.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -200,7 +200,7 @@ if 'data' in likelihood_class.required_kwargs:
     logging.info("Loading data")
     strain_dict, stilde_dict, psd_dict = option_utils.data_from_cli(opts)
     low_frequency_cutoff_dict = option_utils.low_frequency_cutoff_from_cli(opts)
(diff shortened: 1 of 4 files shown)


soumide1102:
@cdcapano cut-off frequencies will have to be provided as`ifo:value` if they are different for different IFOs. But if they are the same for all IFOs, both for example`--low-frequency-cutoff H1:20 L1:20`  and `--low-frequency-cutoff 20`  would work with these changes. So it's not taking anything away from how it works now, but is just adding the option to read separately for different IFOs.

The pycbc inference documentation has `--low-frequency-cutoff ${F_MIN}` currently. So that will still work. If you think we should point out this feature in the documentation, perhaps we can add a line mentioning this somewhere under "BBH Software Injection example" or "GW150914 example"?  

& I've fixed the codeclimate issue.

PR #2272, opened 2018-08-01
simplify setup.py script 

ahnitz:


ahnitz:
@duncan-brown 
This removes some dependencies that were only required prior to python2.7, switches to finding scripts and packages auotmatically, and removes other cruft from the setup.py script. The convoluted test class was also removed as we don't use it in practice and it is nonstandard.

ahnitz:
I think codeclimate here is confused, so I'm ignoring.

PR #2274, opened 2018-08-04
improve detector module

ahnitz:
* Make several methods more pythonic and allow vector input
* Switch to python for several methods and rely on astropy for more accurate sidereal time
* Add unit tests for the detector module
* Add examples and associated docs

Note, that the injection path is not changed and will still call into lal functions for now. The antenna pattern functions will agree to some precision to the lal ones, but should be more accurate as the sidereal time calculation is less of an approximation when using astropy vs lalsuite.

### DIFF
--- a/docs/detector.rst
+++ b/docs/detector.rst
@@ -27,12 +27,10 @@ Time source gravitational-wave passes through detector
 .. literalinclude:: ../examples/detector/delay.py
 .. command-output:: python ../examples/detector/delay.py
 
-======================================================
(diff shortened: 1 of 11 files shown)


ahnitz:
@cdcapano The climate stuff should be ignored. I cannot change the order of imports here.

ahnitz:
@cdcapano Poke

ahnitz:
@cdcapano I think I've addressed the codacy issues in the most recent patch. Please re-review.

PR #2276, opened 2018-08-06
fixes to make test fftw unthreaded work in python3

ahnitz:
This allows the fft related unit tests to run under python3. I've checked only that they work on a machine with fftw installed.


ahnitz:
@josh-willis climate complaints are on pre-existing code, not the changes of this patch.

PR #2282, opened 2018-08-20
remove debug print statement

ahnitz:
@soumide1102 This was the source of the 'array_mkl' print statements you were seeing. Sorry, this was a holdover debug statement. This patch removes that message.


soumide1102:
@ahnitz Thanks for the fix. Codeclimate is reporting an issue I see. Probably that variable `e` in line 179 is not needed anymore now that the print statement has been removed.

PR #2285, opened 2018-08-21
Add a helper function for log histograms

duncan-brown:
Workaround for issues with matplotlib drawing contours on log plots.

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -786,6 +786,24 @@ def freq_from_final_mass_spin(final_mass, final_spin, l=2, m=2):
 def tau_from_final_mass_spin(final_mass, final_spin, l=2, m=2):
     return get_lm_f0tau(final_mass, final_spin, l, m, 1)[1][0]
 
+#
+# =============================================================================
+#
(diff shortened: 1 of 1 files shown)


cdcapano:
@duncan-brown I just noticed this... what was the reason for adding this function? Taking the `log10` of a parameter should already be supported via numpy.

duncan-brown:
@stevereyes01 should confirm, but numpy.log10 is not exposed in a way that the posterior plotting code can use. If it is, then this can be reverted.

cdcapano:
It should be exposed... the way you would do it is `--parameters log10(param)`. @stevereyes01 Did that not work for you? If that's the case, that would be a bug with `FieldArray`...

stevereyes01:
Ah, okay, sorry about that. You're correct. We can revert this. Thanks for that suggestion.

duncan-brown:
@stevereyes01 you probably can't revert this patch, as it will conflict with the changes in https://github.com/gwastro/pycbc/pull/2284 Just submit a new pull request (referencing this one) that removes the function.

PR #2288, opened 2018-08-23
Remove return_base_10

stevereyes01:
The function I used `return_base_10`, can be done via `log10(param)`. Simpler code which reverts https://github.com/gwastro/pycbc/pull/2285


stevereyes01:
Thanks for correcting me Collin.

PR #2291, opened 2018-08-23
Make ensurearray more numpy like

cdcapano:
This switches to using `numpy.broadcast_arrays` in `ensurearray`. That function both ensures that the inputs are arrays, and broadcasts all of them to have the same shape. As a result, `ensurearray` now can take multiple arguments at once. This simplifies calling, and removes the need to check for shapes.

This also makes all of the functions have more of a standard IO. Previously you'd have to make sure all arguments had the same shape for functions that used `ensurearray`, whereas other functions that did not could take a mixture of arrays and floats (as they would default to the using numpy's broadcasting rules). For example, `conversions.chi_eff(10., numpy.array([5., 10.]), 0.5, 0.5)` would work, but `conversions.chi_p(10., numpy.array([5., 10.]), 0.5, 0.5, 0.5, 0.5)` would raise an error.
Now both functions will work in the same way:
```
>>> conversions.chi_eff(10., numpy.array([5., 10.]), 0.5, 0.5)
array([0.5, 0.5])
>>> conversions.chi_p(10., numpy.array([5., 10.]), 0.5, 0.5, 0.5, 0.5)
array([0.70710678, 0.70710678])
```

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -45,7 +45,7 @@ from .coordinates import spherical_to_cartesian as _spherical_to_cartesian
 #
 def ensurearray(*args):
     """Apply numpy's broadcast rules to the given arguments.
-    
+
     This will ensure that all of the arguments are numpy arrays and that they
     all have the same shape. See ``numpy.broadcast_arrays`` for more details.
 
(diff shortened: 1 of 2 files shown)


cdcapano:
Two passed checks 'n a good review,
Yo-ho-ho and a bottle of rum.

PR #2295, opened 2018-08-24
Avoid MemoryErrors in SingleDetTriggers

titodalcanton:
When reading, filtering and clustering merged single-detector triggers using SingleDetTriggers, I noticed very large memory usages which occasionally lead to MemoryErrors from `numpy.in1d`, even on machines with several tens of gigabytes of memory. This change appears to eliminate the error, although I have not made precise measurements of before/after memory usage. It also simplifies the code a bit and possibly makes it slightly faster, since `in1d` seems to sort the array, which is not necessary in this case.


tdent:
If tested, this should be OK.

titodalcanton:
Yes, I tested this by comparing lists of extracted triggers from several boxes before and after the merge.

PR #2299, opened 2018-08-25
Update JSON status from within pycbc_live

titodalcanton:
Added an option to give the path to the JSON file. Added a check for excess latency which also sets the status to critical.

If you think this is sufficient, I can also remove the separate nagios_monitor script.

Not tested yet.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -465,7 +465,6 @@ with ctx:
 
     # get more data
     data_end = lambda: data_reader[data_reader.keys()[0]].end_time
-    i = 0
     while data_end() < args.end_time:
         t1 = time()
         logging.info('%s: Analyzing from %s', evnt.rank, data_end())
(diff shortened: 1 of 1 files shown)


bin/pycbc_live
```diff
@@ -543,9 +547,35 @@ with ctx:
...
+                                     'txt_status': 'OK: No reported problems',
+                                     'start_sec': 0},
+                                    {'num_status': 1,
+                                     'txt_status': 'WARNING: last report >2 min ago',
+                                     'start_sec': 120}]
+            status_intervals.append({'num_status': 2,
+                                     'txt_status': 'CRITICAL: last report >4 min ago',
+                                     'start_sec': 240})
+            status = {'author': 'Alexander Harvey Nitz',
+                      'email': 'alex.nitz@ligo.org',
```
ahnitz: Probably should update this info.

titodalcanton:
This has now been tested.

PR #2300, opened 2018-08-26
Status JSON should only be updated by root process

titodalcanton:


ahnitz:
Woops, yes, this should definitely be done.

PR #2311, opened 2018-09-04
Cache strain from catalog using astropy

lpsinger:


pycbc/catalog/__init__.py
```diff
@@ -83,18 +83,15 @@ def strain(self, ifo):
...
         channel = '%s:LOSC-STRAIN' % ifo
         url = self.data['frames'][ifo]
         f = tempfile.NamedTemporaryFile(suffix='.gwf')
-        r = requests.get(url, stream=True)
-
-        if r.status_code != 200:
-            raise ValueError("Could not download file, %s", r.status_code)
-
-        shutil.copyfileobj(r.raw, f)
+        with get_readable_fileobj(url, cache=True, encoding='binary') as r:
```
ahnitz: libframe, unfortunately needs a proper file path on disk, so it might be better to use the astropy download_file function directly. That way temp files are all in the astropy cache, which can be cleared in a straightforward manner. That should mean there is no more "leak" than a cache already entails.

Something like
```
fname = astropy.utils.data.download_data(url, cache=True)
return read_frame(fname, channel)
```

It looks like from the source that get_readable_fileobj will call this function internally anyway, so I don't think there is any loss. 

https://github.com/astropy/astropy/blob/v3.0.x/astropy/utils/data.py#L190

@lpsinger Does this change make sense to you as well?

lpsinger: Unfortunately, the cache construction code seems to expect a filename ending in .gwf, an assumption which is broken by the filename returned by `astropy.utils.data.download_file`.

ahnitz: Hmmm, I see. It doesn't look like they expose the extension option in NameTemporaryFile either. https://github.com/astropy/astropy/blob/v3.0.x/astropy/utils/data.py#L1046

ahnitz: @lpsinger I think this can be resolved in this way. https://github.com/lpsinger/pycbc/pull/1

ahnitz: @lpsinger Let me know if that works for you.

lpsinger: Actually, I'm trying out an option that looks for the magic string 'IGWD' at the beginning of a frame file.

lpsinger:
OK @ahnitz, check out the new version.

ahnitz:
@lpsinger I just tried it out on my laptop and it seems to behave as expected. I think your solution is a good one as well to keep the file type check.

ahnitz:
Once the automated tests pass, this may be merged.

PR #2316, opened 2018-09-06
add rwrap arg to get_td_waveform_from_fd

Cyberface:
See issue https://github.com/gwastro/pycbc/issues/2314


ahnitz:
@Cyberface At the moment, it will be overwritten, so it will do nothing. You might want to be a bit more clear in the description if you add a keyword argument. It's just a fudge factor to ensure that the entire time series is contiguous in the array and not wrapped around the end.

Cyberface:
@ahnitz thanks for finding that bug!

I've updated the docstring, inspired but what you said.

PR #2320, opened 2018-09-10
update to lalsuite 89a30fc and change base directory

duncan-brown:
This changes the docker image to use the 89a30fc lalsuite tag (it has been stuck at an old tag for a while) and changes the pycbc home directory to `/opt/pycbc` so the docker image can be converted to singularity.


lppekows:
Looks good to me.

duncan-brown:
@lppekows I think you hit the wrong button. The review needs approval then I can merge once Travis completes.

PR #2331, opened 2018-09-22
Add HPD credible interval plotting code for 1D histograms

soumide1102:
This PR adds code for plotting HPD credible intervals on 1D marginalized histograms to `scatter_histograms.py`. Also lets `pycbc_inference_plot_posterior` use this feature.  If the HPD credible interval option is to be plotted, `--plot-marginal-hpd-credible-interval` has to be provided to `pycbc_inference_plot_posterior`. The % HPD credible interval can be provided with `--marginal-hpd-percent`. If the latter is not provided, but `--plot-marginal-hpd-credible-interval` is `True`, will default to a 90% probability for the HPD CI . If `--plot-marginal-hpd-credible-interval` is not provided, will plot the regular symmetric credible interval.

### DIFF
--- a/bin/inference/pycbc_inference_plot_posterior
+++ b/bin/inference/pycbc_inference_plot_posterior
@@ -184,6 +184,9 @@ for i, (p, l, s) in enumerate(zip(parameters, labels, samples)):
                     p, s, labels=l, fig=fig, axis_dict=axis_dict,
                     plot_marginal=opts.plot_marginal,
                     marginal_percentiles=opts.marginal_percentiles,
+                    plot_marginal_hpd_credible_interval=
+                    opts.plot_marginal_hpd_credible_interval,
+                    marginal_hpd_percent=opts.marginal_hpd_percent,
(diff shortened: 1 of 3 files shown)


soumide1102:
gw150914 m1-m2 with HPD CI for 1D histograms https://sugwg-jobs.phy.syr.edu/~soumi.de/pycbc_inference/tests/hpd/gw150914_m1m2_hpdCI.png 
with the regular symmetric CI for comparison https://sugwg-jobs.phy.syr.edu/~soumi.de/pycbc_inference/tests/hpd/gw150914_m1m2_symmCI.png

cmbiwer:
Is there a case where you want both methods for plotting? Why not just reuse ``--marginal-percentiles`` and add options to ``--plot-marginal`` such as ``percentiles`` and ``hpd``, and leave the default as ``None`` for not including them? I'd consider this similar to the PSD estimation options where we do this with ``median`` on the command line, etc.

soumide1102:
Moving this to #2611.

PR #2351, opened 2018-09-27
jump

ahnitz:
* breaks emcee 2.X support. So maybe hold off merging for now. Demonstrates how to set up jump proposals however.


ahnitz:
@cdcapano I've rebased this from #2419, so that needs to be merged first.

PR #2361, opened 2018-09-28
Pycbc plot samples fix

stevereyes01:
Changing a few things around, and plotting the samples from gaussian example (x, y)

image

A future commit will reduce the number of time spend importing samples. It should just read all the samples and slice along the axis of walker. This should greatly reduce the time required to plot these samples.


stevereyes01:
I can remove the import to numpy. But a future commit will probably use it for slicing.

PR #2362, opened 2018-09-28
Update plot_inj_interval for fix imports for io mv

sum33it:
Fixed the import and functions in injection interval plots. I plotted following plot from 2 test injections and runs:

pycbc_inference_plot_inj_intervals --input-file run_files/inference1.hdf:1 run_files/inference2.hdf:2 --output-file test.png --parameters mass1 mass2

test

### DIFF
--- a/bin/inference/pycbc_inference_plot_inj_intervals
+++ b/bin/inference/pycbc_inference_plot_inj_intervals
@@ -17,13 +17,13 @@ from matplotlib import pyplot as plt
 
 import pycbc
 from pycbc.results import save_fig_with_metadata
+from pycbc.inference import (option_utils, io)
 
 from pycbc import __version__
-from pycbc.inference import option_utils
 
 # parse command line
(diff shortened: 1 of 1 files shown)


bin/inference/pycbc_inference_plot_inj_intervals
```diff
@@ -72,7 +70,7 @@ fig = plt.figure()
 ax = fig.add_subplot(111)
 
 # calculate the expected percentile for each injection and plot
-for param,label in zip(input_parameters, labels):
+for param,label in zip(parameters, labels):
```
cdcapano: `labels` is a dict now, not a list, so this should be:
```
for param in parameters:
    label = labels[param]
```

sum33it: Thanks for pointing this. Made necessary changes and initiated pull request again.

PR #2369, opened 2018-10-02
Fix incorrect docstring

titodalcanton:
This is what the function actually produces, so I am changing the docstring:
peanut


ahnitz:
@titodalcanton Looks good.

PR #2372, opened 2018-10-05
Changed multi ifo coinc statistic

tapaimarton:
Changed the multiifo newsnr statistic to @tdent 's suggestion.

### DIFF
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -146,7 +146,7 @@ class NewSNRStatistic(Stat):
         numpy.ndarray
             Array of coincident ranking statistic values
         """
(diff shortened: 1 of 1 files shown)


pycbc/events/stat.py
```diff
@@ -146,7 +146,7 @@ def coinc_multiifo(self, s, slide, step,
         numpy.ndarray
             Array of coincident ranking statistic values
         """
-        return (sum([s[i]**2 for i in s]))**0.5
+        return (sum([x**2 for x in s.values()]))**0.5
```
tdent: Looks good, but please use standard spacing etc. and better to be explicit that 2 is a float, i.e. `x ** 2.` and `(...) ** 0.5` with spaces.

PR #2381, opened 2018-10-12
Change workflow generator names to make except for temporary

tdent:
pycbc_create_offline_search_workflow is the work area for a multi-ifo replacement to pycbc_make_coinc_search_workflow, will be moved to a more 'consistent' name when ready


tdent:
ok, travis is failing since there are lots of other occurrences of 'pycbc_create_...workflow' hanging around.  will fix these ..

PR #2383, opened 2018-10-16
Implement multiifo coinc into workflow

tapaimarton:
This PR implements the multiifo coincidence code into the new pycbc_create_offline_search_workflow executable. The required config file changes are in github in a PR. With those changes I tested, and the workflow generates, and the coinc jobs are run.


bin/workflows/pycbc_create_offline_search_workflow
```diff
@@ -213,6 +212,37 @@ statfiles = []
 statfiles += wf.setup_trigger_fitting(workflow, insps, hdfbank,
                                       final_veto_file, final_veto_name)
 
+# setup 3 detector coinc for the filtering jobs
+ctags = [tag, 'full3d']
```
ahnitz: @tapaimarton One thing I think is not ideal here is that it hardcodes the 3-ifo combination and each two detector combination. It would be nice if we did not have to do this. 

Can I propose an alternate approach? We know which ifos have been requested to analyze. One can simply assume you want to analyze all of them (in the future we could provide flags to control this is needed). As such just take the list of ifos as input and do the complete combinatorics (i.e. nchoose2..n). The only thing you then need to control is *which* ifo for each of these is the pivot and the *fixed* ones. We could have that specified in the ini file as you've done, in which case, you can programmatically set the tags based on the ifo combination. Another solution which may be simpler and probably all we want anyway is to simply set a priority order. I.E. 
```
[workflow-coincidence]
timeslide-precedence = H1, L1, V1, K1, I1
```
This would translate to select the ifos being analyzed, and choose the first as "pivot" and the second as "fixed' preserving the order we give. 

@tapaimarton @spxiwh Does that sound reasonable?

tdent: @ahnitz @tapaimarton any reason I am not assigned/reviewing this given that I was doing so for the gitlab issue?

tapaimarton: @tdent Sorry, I forgot to add you here when I assigned the gitlab issue to you.

tapaimarton:
@ahnitz I think this is what you wanted, I ran a test with it, and the workflow generated and the jobs started. @tdent do you want me to push the config file changes required for this version to the github issue (most of the previous changes are cut and one option is added for this), or wait if there are more changes required here?

pycbc/workflow/coincidence.py
```diff
@@ -493,3 +497,42 @@ def setup_interval_coinc(workflow, hdfbank, trig_files, stat_files,
...
+    make_analysis_dir(out_dir)
+    logging.info('Setting up coincidence')
+
+    if len(hdfbank) != 1:
+        raise ValueError('Must use exactly 1 bank file for this coincidence '
+                         'method, I got %i !' % len(hdfbank))
+    hdfbank = hdfbank[0]
+
+    findcoinc_exe = PyCBCFindCoincExecutable(workflow.cp, 'multiifo_coinc',
+                                             ifos=workflow.ifos,
```
ahnitz: ifos maybe should just be the set of ifos used for this particular call, no?

ahnitz:
@tapaimarton Yeah, the structure basically looks right now.

tapaimarton:
@ahnitz  I tested the latest change with generating the workflow, it runs without issue for 3 detectors, and only the ifos used are displayed in the name. The 2det and 3det tags could maybe than be dropped, but I think it is easier to see and search those terms.

tdent:
@tapaimarton let us get code changes settled first then make the required config changes afterwards.

tdent:
Travis seems to be failing with a relevant error .. ?

tapaimarton:
Ah, I didn't test this, I modified the PyCBCFindCoincExecutable, and it now fails with the old input. I can just create a new class that handles the multiifo and change back the old one, or I can try to fix it, so it can handle both calls. Which option is better/recommended? Creating a new, and deleteing the old once we no longer need it seems easier to me, what do you think @ahnitz , @tdent ?

tdent:
@tapaimarton I have no idea what the best resolution is at the moment, I just see that Travis is failing and it should not be failing if the code is correct as a whole.

bin/workflows/pycbc_create_offline_search_workflow
```diff
@@ -213,6 +212,47 @@ statfiles = []
...
+final_bg_files = []
+ifoindeces = {}
+
+# Get the ifo precedent list for the multiifo coinc jobs
+ifo_precedence_list = workflow.cp.get_opt_tags('workflow-coincidence', 'timeslide-precedence', ctags)
+for ifo, file in zip(*insps.categorize_by_attr('ifo')):
+    ifoindeces[ifo] = ifo_precedence_list.index(ifo)
+
+for i in range(2,len(insps)+1):
+    combinations = itertools.combinations(ifoindeces.keys(),i)
```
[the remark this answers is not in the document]
tapaimarton: This line generates the detector combinations, at 2 it generates the 2 detector combination, if there are 3 trigger files, than it generates the 2 and 3 detector combination as well. Should there be 4, than it would create the 2, 3 and 4 combinations.

tapaimarton:
I am not sure how to resolve the 10 parameter complaint. We need to give the pivot and fixed ifos, and since we don't want to handle it in the config files, we have to give them in the workflow. The executable itself is run by more than 10 arguments, some of them are just hidden in the config files from codebeat.

tdent:
Well, one way of feeding lots of information to a function would be to pass it a list, or a dictionary, or some other more or less complicated object.  
Maybe that would require a bit more rewriting of the pipeline code, but on the other hand that might be a good thing (i.e. making functions with large numbers of arguments less common overall)

pycbc/workflow/coincidence.py
```diff
@@ -106,6 +106,33 @@ def create_node(self, trig_files, bank_file, stat_files, veto_file,
...
+    current_retention_level = Executable.ALL_TRIGGERS
+    file_input_options = ['--statistic-files']
+    def create_node(self, trig_files, bank_file, stat_files, veto_file,
+                    veto_name, template_str, pivot_ifo, fixed_ifo, tags=None):
+        if tags is None:
+            tags = []
+        segs = trig_files.get_times_covered_by_files()
+        seg = segments.segment(segs[0][0], segs[-1][1])
+        node = Node(self)
+        node.set_memory(10000)
```
ahnitz: Leave this for the config file.

pycbc/workflow/coincidence.py
```diff
@@ -106,6 +106,33 @@ def create_node(self, trig_files, bank_file, stat_files, veto_file,
...
+        node.add_input_opt('--template-bank', bank_file)
+        node.add_input_list_opt('--trigger-files', trig_files)
+        if len(stat_files) > 0:
+            node.add_input_list_opt('--statistic-files', stat_files)
+        if veto_file is not None:
+            node.add_input_opt('--veto-files', veto_file)
+            node.add_opt('--segment-name', veto_name)
+        if pivot_ifo is not None:
```
ahnitz: This isn't optional so no need for the if statements here (and the similar places)

ahnitz:
@tapaimarton Except for a couple comments I just posted which are really just minor cleanups, I'm mostly happy with this.

ahnitz:
@tapaimarton Please ignore the argument number comments from  Codacy
@tdent The issue with passing other options is that things like dicts or lists can be easily abused to hide what the true inputs are. Either really only makes sense if they are grouping the same kind of input.

tapaimarton:
@ahnitz I will delete the memory line (that was set there in the old coinc code) and I will fix the pivot and fixed ifo lines. However codacy is now complaining about line 539 in coincidence.py, where the file output from categorize_by_attr is not used. I don't see any built in way to access only the ifos from a file list. Is there?

If not I can iterate over the list and get the File.ifo maybe, that should work I think.

ahnitz:
@tapaimarton You just need a  list of the ifos to make the mapping, no? You can get that from a workflow instance property directly.

ahnitz:
@tapaimarton Also general python tip, there are ways to write that line of code without giving an object a name. You can do.

for a, _ in BLAH:
      pass

Where '_' eats the extra value.

tapaimarton:
@ahnitz Actually, I need the particular set of ifos passed inside to the executable (2 for 2 detectors, 3 for 3 detectors) on that line (that was a comment from you earlier to do it that way), so I can't get that from the workflow. I will use the tip you gave, thanks for that!

tapaimarton:
@ahnitz Can you look at that travis error? I do not think it is related to this PR.

ahnitz:
@tapaimarton If you rebase, the travis error should go away.

tapaimarton:
I rebased to current master so hopefully all checks will pass, @tdent asked us not to merge until he can check this over, so please hold off on merging until he approves.

tdent:
@tapaimarton @ahnitz I am now seeing 90 commits and 120 files changed in the diff, which makes it impossible to see what has actually been changed. 
Marton, did you do git rebase upstream/master in fact?  I don't think it should be showing all these 'extra' changes.

tapaimarton:
I fetched upstream into my master repository ('git fetch upstream'), checked out the branch, and did 'git rebase master', than I had to git pull, and than I could git push. Is this not the correct way to rebase to master?

tdent:
@tapaimarton Do 'git push origin HEAD -f' to force the remote branch history to be rewritten.  Just 'git push' on its own will mix up the previous sequential commits to the remote branch with the new rebased ones.

bin/workflows/pycbc_create_offline_search_workflow
```diff
@@ -213,6 +212,49 @@ statfiles = []
...
                                       final_veto_file, final_veto_name)
 
+# setup the detector combination coincs for the filtering jobs
+full_insps = insps
+ctags = [tag, 'full']
+bg_files = []
+final_bg_files = []
+ifo_ids = {}
+
+# Get the ifo precedent list for the multiifo coinc jobs
```
tdent: typo : 'precedence' not precedent

(I guess the description of what the precedence list *means* should be in the ini file, we'll deal with that later)

bin/workflows/pycbc_create_offline_search_workflow
```diff
@@ -213,6 +212,49 @@ statfiles = []
...
+for i in range(2, len(insps)+1):
+    combinations = itertools.combinations(ifo_ids.keys(), i)
+    for ifocomb in combinations:
+        combination_prec = {}
+        inspcomb = wf.core.FileList()
+        # create the precedence list for the detector combination
+        for ifo in ifocomb:
+            combination_prec[ifo] = ifo_ids[ifo]
+
+        # Use the inspiral files for the detector combination only.
```
tdent: edit comment: `Use only the inspiral files required by the detector combination`

bin/workflows/pycbc_create_offline_search_workflow
```diff
@@ -213,6 +212,49 @@ statfiles = []
...
+full_insps = insps
+ctags = [tag, 'full']
+bg_files = []
+final_bg_files = []
+ifo_ids = {}
+
+# Get the ifo precedent list for the multiifo coinc jobs
+ifo_precedence_list = workflow.cp.get_opt_tags('workflow-coincidence', 'timeslide-precedence', ctags)
+for ifo, file in zip(*insps.categorize_by_attr('ifo')):
+    ifo_ids[ifo] = ifo_precedence_list.index(ifo)
```
tdent: this is not immediately clear. does this just assign an object (does it have to be an integer??) to each ifo based on what is in the ini file under timeslide precedence?

bin/workflows/pycbc_create_offline_search_workflow
```diff
@@ -213,6 +212,49 @@ statfiles = []
...
+        for ifo in ifocomb:
+            combination_prec[ifo] = ifo_ids[ifo]
+
+        # Use the inspiral files for the detector combination only.
+        for ifo, file in zip(*insps.categorize_by_attr('ifo')):
+            if ifo in combination_prec:
+                inspcomb += file
+
+        # Determine the pivot-ifo and fixed-ifo based on the precedent list
+        pivot_ifo = min(combination_prec, key=combination_prec.get)
```
tdent: maybe need to explain this line .. I *think* it means pivot_ifo is assigned to the ifo that is the dict key in combination_prec such that its entry has the smallest value .. ? I'm not that familiar with iterating over a dict.
suggest comment text: `pivot_ifo is assigned to be the ifo with the smallest index value in the combination` (if that's correct)

tdent: (and same typo -> 'precedence')

bin/workflows/pycbc_create_offline_search_workflow
```diff
@@ -213,6 +212,49 @@ statfiles = []
...
+        del combination_prec[pivot_ifo]
+        fixed_ifo = min(combination_prec, key=combination_prec.get)
```
tdent: comment here: `fixed_ifo is assigned to be the ifo with the next smallest index value`

bin/workflows/pycbc_create_offline_search_workflow
```diff
@@ -213,6 +212,49 @@ statfiles = []
...
+
+        # Create coinc tag, and set up the coinc job for the combination
+        coinctag = '{}det_'.format(len(ifocomb))+pivot_ifo+fixed_ifo
```
tdent: put spaces between '+' please

tdent:
Thanks @tapaimarton !  I have some small requests on explaining the code and formatting. I'll see if I can deal with them by editing the pull directly

PR #2385, opened 2018-10-17
Add documentation for developers about the new sampler API

cdcapano:
Adds a page for the sampler classes and a page for the new IO classes. These are linked from the main inference help page, under a new section "For Developers".  Example is here:

[Sampler API](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/inference_dev-test_docs/html/inference/sampler_api.html)
[Inference IO](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/inference_dev-test_docs/html/inference/io.html)

@cmbiwer @soumide1102 Is this what you were thinking of?

### DIFF
--- a/docs/_include/inference_io_inheritance_diagrams.py
+++ b/docs/_include/inference_io_inheritance_diagrams.py
@@ -1,0 +1,34 @@
+# Creates RST for the sampler inheritance diagrams
+from __future__ import print_function
+import inspect
+from pycbc.inference.io import filetypes
+
+fname = 'inference_io_inheritance_diagrams.rst'
+
+def get_topclasses(cls):
+    """Gets the base classes that are in pycbc."""
+    bases = [c for c in inspect.getmro(cls)
+             if c.__module__.startswith('pycbc') and c != cls]
+    return ', '.join(['{}.{}'.format(c.__module__, c.__name__) for c in bases])
+
+tmplt = """.. _inheritance-io-{name}:
+
+* ``{name}``:
+
+.. inheritance-diagram:: {module}.{name}
+   :parts: 3
+   :top-classes: pycbc.inference.io.base_hdf.BaseInferenceFile
+
+|
+
+"""
+fp = open(fname, 'w')
+
+for ftname, cls in sorted(filetypes.items()):
+    # get the parents
+    topclasses = get_topclasses(cls)
+    out = tmplt.format(name=cls.__name__, module=cls.__module__)
+                       #topclasses=topclasses)
+    print(out, file=fp)
+
+fp.close()
--- a/docs/_include/inference_io_inheritance_diagrams.py
+++ b/docs/_include/inference_io_inheritance_diagrams.py
@@ -28,7 +28,6 @@ for ftname, cls in sorted(filetypes.items()):
     # get the parents
     topclasses = get_topclasses(cls)
     out = tmplt.format(name=cls.__name__, module=cls.__module__)
-                       #topclasses=topclasses)
     print(out, file=fp)
 
 fp.close()
--- a/docs/_include/inference_io_inheritance_diagrams.py
+++ b/docs/_include/inference_io_inheritance_diagrams.py
@@ -15,7 +15,7 @@ tmplt = """.. _inheritance-io-{name}:
 
 * ``{name}``:
 
-.. inheritance-diagram:: {module}.{name}
+.. inheritance-diagram:: {module}.{clsname}
    :parts: 3
    :top-classes: pycbc.inference.io.base_hdf.BaseInferenceFile
 
@@ -27,7 +27,8 @@ fp = open(fname, 'w')
 for ftname, cls in sorted(filetypes.items()):
     # get the parents
     topclasses = get_topclasses(cls)
-    out = tmplt.format(name=cls.__name__, module=cls.__module__)
+    out = tmplt.format(name=ftname, clsname=cls.__name__,
+                       module=cls.__module__)
     print(out, file=fp)
 
 fp.close()
--- a/docs/_include/sampler_inheritance_diagrams.py
+++ b/docs/_include/sampler_inheritance_diagrams.py
@@ -2,15 +2,13 @@
 from __future__ import print_function
 from pycbc.inference.sampler import samplers
 
-print("CREATING SAMPLER DIAGRAM")
-
 fname = 'sampler_inheritance_diagrams.rst'
 
-tmplt = """.. _inheritance-{}:
+tmplt = """.. _inheritance-{name}:
 
-* ``{}``:
+* ``{name}``:
 
-.. inheritance-diagram:: pycbc.inference.sampler.{}
+.. inheritance-diagram:: {module}.{name}
    :parts: 3
 
 |
@@ -19,6 +17,7 @@ tmplt = """.. _inheritance-{}:
 fp = open(fname, 'w')
 
 for sampler, cls in sorted(samplers.items()):
-    print(tmplt.format(sampler, sampler, cls.__name__), file=fp)
+    out = tmplt.format(name=cls.__name__, module=cls.__module__)
+    print(out, file=fp)
 
 fp.close()
--- a/docs/_include/sampler_inheritance_diagrams.py
+++ b/docs/_include/sampler_inheritance_diagrams.py
@@ -8,7 +8,7 @@ tmplt = """.. _inheritance-{name}:
 
 * ``{name}``:
 
-.. inheritance-diagram:: {module}.{name}
+.. inheritance-diagram:: {module}.{clsname}
    :parts: 3
 
 |
@@ -17,7 +17,8 @@ tmplt = """.. _inheritance-{name}:
 fp = open(fname, 'w')
 
 for sampler, cls in sorted(samplers.items()):
-    out = tmplt.format(name=cls.__name__, module=cls.__module__)
+    out = tmplt.format(name=sampler, clsname=cls.__name__,
+                       module=cls.__module__)
     print(out, file=fp)
 
 fp.close()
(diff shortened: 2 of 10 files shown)


cmbiwer:
I think this helps, but I actually had something simpler in mind.

But an explanation on how to add features is useful as long as we keep it updated, ie. if you're assigned a pull request then actually say if the documentation needs to be updated. Otherwise, this isn't that useful in the long run.

Just browsing our own documentation, I actually think the bases part of the class documentation is more what I had in mind, eg.: https://www.atlas.aei.uni-hannover.de/%7Ecdcapano/public_scratch/github_prs/inference_dev-test_docs/html/pycbc.inference.sampler.html#module-pycbc.inference.sampler.emcee_pt

It gets generated with the documentation for the class.

If the inheritance diagram could just be included with the class documentation, and then for functions, it says "``This function is defined in <class>``". That's all I was thinking of. Then, simply point someone to the documentation for the class and they have a roadmap of where all the functions are defined and inherited.

Also, sidenote, having multiple modules with the same name (even if you're using relative paths) is problematic in my opinion when communicating between people and not machines.

cdcapano:
Hi @cmbiwer 

> But an explanation on how to add features is useful as long as we keep it updated, ie. if you're assigned a pull request then actually say if the documentation needs to be updated. Otherwise, this isn't that useful in the long run.

If you're referring to the "How to add a sampler class" instructions, that shouldn't change any time soon. Adding a new sampler should not affect it. If you're referring to the inheritance diagrams, I agree that hard coding the diagrams for each sampler is a pain, since that will need to be updated every time someone adds a new class. I can add a python script that will automatically generate the reStructuredText (RST) for them. Will update this pr shortly with that.

> Just browsing our own documentation, I actually think the bases part of the class documentation is more what I had in mind, eg.: https://www.atlas.aei.uni-hannover.de/%7Ecdcapano/public_scratch/github_prs/inference_dev-test_docs/html/pycbc.inference.sampler.html#module-pycbc.inference.sampler.emcee_pt
>
> It gets generated with the documentation for the class.
>
> If the inheritance diagram could just be included with the class documentation, and then for functions, it says "This function is defined in <class>". That's all I was thinking of. Then, simply point someone to the documentation for the class and they have a roadmap of where all the functions are defined and inherited.

If I understand you right, you're basically asking for two things: (1) that the inheritance diagrams be included in the documentation with the individual classes; e.g., [here](https://www.atlas.aei.uni-hannover.de/%7Ecdcapano/public_scratch/github_prs/inference_dev-test_docs/html/pycbc.inference.sampler.html#pycbc.inference.sampler.emcee.EmceeEnsembleSampler). (2) That the inherited methods are included in that class documentation, so you have one place to go to to know everything that the class understands. Is that right?

I agree that both of those things would be nice, but there's a few issues with implementing that. The main problem is that the documentation pages for the classes are automatically generated by `sphinx-apidoc` using the classes' doc strings. You could get inheritance diagrams in there by adding the RST for them in the doc strings, but I'd rather not make people do that, and it will make the doc string more illegible.

It is possible to do (2) by setting some environment variables in `setup.py` that tells `sphinx-apidoc` to include the inherited members in class documentation. I gave that a try; here is the results for the emcee sampler class: [link](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/inference_dev-test_docs_inherited/html/pycbc.inference.sampler.html#pycbc.inference.sampler.emcee.EmceeEnsembleSampler). You can see that now all of the inherited methods are included (e.g., `compute_acf` is listed there, which came from `MCMCAutocorrSupport`). The problem is when you turn this on, sphinx will add docs for *all* inherited methods, even things outside of pycbc. So if you look at the documentation for one of the IO classes, which inherit from `h5py.File`, you end up with all of the methods from `h5py.File` too. For example, [here is the docs for EmceeFile](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/inference_dev-test_docs_inherited/html/pycbc.inference.io.html#pycbc.inference.io.emcee.EmceeFile). Also, making any sort of change to the auto-generated documentation will affect all of pycbc, not just the inference module. So, e.g., the [documentation for FieldArray](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/inference_dev-test_docs_inherited/html/pycbc.io.html#pycbc.io.record.FieldArray) ends up being ridiculously long because it now includes all of the methods from `numpy.record`, from which it inherits.

Another possibility is to use `shpinx-automodapi` to generate the module documentation rather than `autodoc`. This looks to make nicer documentation pages, and it will automatically generate inheritance diagrams for an entire module. Here's a test I did on the `inference.sampler` and `inference.io` modules: [link](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/inference_dev-test_docs_inherited/html/inference/automod_example.html). If you click on one of the classes, you can see that the documentation is easier to parse, as it includes tables of all of the attributes and methods. For example, [here](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/inference_dev-test_docs_inherited/html/api/pycbc.inference.sampler.EmceeEnsembleSampler.html#pycbc.inference.sampler.EmceeEnsembleSampler) is the documentation automodapi created for `EmceeEnsembleSampler`. It still has the issue that if you turn on included inherited methods, you end up with everything, even outside of `pycbc`. But maybe we could handle that on a module-by-module basis.

Anyway, this a long-winded way of saying that I agree with your suggestions, but implementing them will require retooling how the auto documentation is done pycbc wide. That should be something that is addressed on a PR off of master in which @spxiwh and @ahnitz take part. I'd rather not hold up merging the inference dev branch for that.

soumide1102:
Thanks @cdcapano for taking the time to put together the developer's guide. 
I think the two pages you added will be a useful introduction for people starting to develop in PyCBC Inference and pick up what's going on in the code overall faster. However, while the two pages look like will be useful for developers, I think the docstrings themselves in the code need the kind of details that Chris mentioned to make the code more transparent and provide better guidance for both developers and non-developers (people just using PyCBC Inference for a project maybe). Non-developers might not be interested in going the route of understanding the details of how the Sampler or IO modules work before they figure out where for example a specific function they're interested is called from. 

In that regard, I feel the whole `run` ->`run_mcmc` nested implementation can be hard to follow if not given a proper roadmap. Someone looking for documentation for that would not know that is explained in your Sampler intro page as an example. I was thinking it will be nice if we can add something more than just `"""Run the MCMC for the given number of iterations."""` in such docstrings for these kinds of methods. Maybe providing a comment there about "why it is an abstractmethod... or that it is overridden by the `run_mcmc` of the particular sampler being run" might help (should be very short and to-the-point kind of just serving the purpose of a road sign). Then for example the BaseSampler class says `"""Provides constructor classes and convenience functions for MCMC samplers."""`. Adding something like "..this is the class that specifies methods and properties in the form of its @abstractmethods and @abstractproperties that are required by all samplers..." will be more helpful. So in addition to providing the general overview of the code in the two pages you wrote, making the docstrings inside the code a little more verbose will help. I like your example of how the inherited members were listed in the documentation of the class.

Other than that, the two developer's guide pages have useful content like I said. Just a small comment : The "How an MCMC sampler is run ..." paragraph in the Sampler page, perhaps can be made a bit simpler in how it's written and the point made clear that `run` is the general method in BaseSampler for running any king of sampler module. That is overriden in BaseMCMC's `run` which runs blocks of iteration until convergence for MCMC samplers (like you wrote). BaseMCMC's `run` has it's own general method to call the underlying sampler `run_mcmc` which is overridden in the specific sampler class's `run_mcmc` customized according to how that sampler is run.
 
Apart from the documentation, I saw in the EmceePt class : 
```
@property
    def base_shape(self):
        return (self.ntemps, self.nwalkers,)
```
looks incomplete?

Also, I think I mentioned this earlier, but is there a specific reason to use the relative imports? When developing, I feel I can easily recall where for example a particular module X (say `conversions.py`) lives. But with the relative imports, I'll have to spend some time looking up where the code I'm developing lives and where relative to that the module `X` I'm importing lives, and do that for everything I import. Looks it's easy to get it wrong and add in the wrong number of `.`s.

cdcapano:
I've made it so that the inheritance diagrams are created automatically when the docs are built. This involved adding an `_include` directory to `docs` with the python scripts in them, and code to `docs/conf_std.py` to run them. Adding @ahnitz to reviewers to check that this makes sense.

ahnitz:
@soumide1102 @cmbiwer I think some of your requests are difficult to properly respond to as it is unclear what exactly you are looking for. 

I think this should *not* hold up merging this PR though, and the larger branch that is hanging over everything. From my view, we should merge this and and the larger branch once the examples are updated so that the documentation is at least in a self-consistent state and we can continue forward progress from there.

@soumide1102 It seems that you have a clearer view of what you are looking for. Could you open an issue and assign it to yourself. It would be really helpful here if you could provide some of the documentation you are thinking would be useful for others in a new PR. If you take on one of the classes for example, it will be more clear what you are thinking and others can follow suit if it makes sense.

cdcapano:
@soumide1102 I added a little more documentation to `BaseMCMC` about the abstract methods it introduces, since that seems to be the source of some of the confusion here. Updated docs are [here](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/inference_dev-test_auto_inherit/html/pycbc.inference.sampler.html#pycbc.inference.sampler.base_mcmc.BaseMCMC). Beyond this, I'm not sure what else to do. I agree with @ahnitz: I'd rather not hold this up any longer over this. If you have an idea of how to make the docs clearer, feel free to open a PR.

soumide1102:
Sure, will be happy to help with further improvements to the docstrings in a new PR.

PR #2387, opened 2018-10-17
remove cruft, lower overhead of event manager

ahnitz:
@josh-willis This reduces the overhead a bit of using concatenate repeatedly. I think you had pointed this out before, but I've run into a corner case where this becomes problematic. Please find a patch attached. I've also taken the liberty of removing some only partially working code.

### DIFF
--- a/bin/pycbc_inspiral
+++ b/bin/pycbc_inspiral
@@ -96,8 +96,6 @@ parser.add_argument("--cluster-function", choices=["findchirp", "symmetric"],
 parser.add_argument("--cluster-window", type=float, default = -1,
                     help="Length of clustering window in seconds."
                     " Set to 0 to disable clustering.")
-parser.add_argument("--maximization-interval", type=float, default=0,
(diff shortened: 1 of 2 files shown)


tdent:
Hi Alex, Josh - does this affect any current functionality? 
I'm not sure why maximization over the bank is classed as only partially working but I'd have appreciated more of a heads up that this was going to be nuked as an option.

ahnitz:
@tdent
The overall change does not affect any currently used functionality, at least not that I'm aware of since testing pycbc_inspiral within the ihope pipeline and using ethinca. 

I should have clarified that there still is a good option for this kind of clustering though, and we should use that if desired. I.E. the --loudest-keep-num, --loudest-keep-interval. If you want a similar setup, you can give 1 and .035 for similar results if testing an ethinca-like algorithm. 

It will dramatically speed up the case where the snr threshold is lowered however, with no change to the output results.

tdent:
OK, indeed it seemed to be mainly there for backwards compatibility which is now not really wanted.

PR #2391, opened 2018-10-19
Lower frequency cutoff specification in config file

stephanmohr:
The lower frequency bound at which to cut off the inner product used by the gaussian_noise model now needs to be specified in the config file. The command line argument is still used for the estimation of the noise.


cdcapano:
@stephanmohr Can you rebase this to master?  Than I can review it.

PR #2394, opened 2018-10-20
Add back likelihood calculation with different low-frequency-cutoff for each IFO features

soumide1102:
This PR should add back the changes from PR #2266 and PR #2325 for 
1. the likelihood computation with different low frequency cutoffs for each IFO
2. Save each IFO's low frequency cutoff used in PSD estimation as an attribute in that IFO's ${ifo}/psd group.
3. Save each IFO's low frequency cutoff used in likelihood computation as an attribute of the inference file

The low frequency cutoffs in 2. and 3. are the same, and looks like PR #2391 is adding a check so that that is true.

Looks like PR #2391 is adding the ability to read in the lower frequency for the likelihood calculation only ( and not the PSD calculation) from the config file. After PR 2391 is merged, this PR should be tested against that, then merged. So keeping this on hold till PR 2391 is merged.


cdcapano:
@soumide1102 Thanks for doing this. Yeah, I'd like to get Stefan's changes to load the low frequency cutoff for the likelihood from the config file #2391 in first, then have you update that to use different cutoffs for each IFO.

cdcapano:
@soumide1102 What's the status of this? This PR is so out of date, you might want to just kill it and start from a fresh checkout of master.

soumide1102:
Yes, will add this feature checking out from master.

PR #2395, opened 2018-10-22
Uniform intervals

stevereyes01:
I never got around to finishing this distribution. RVS works when you call it, but I can't get PDF to do anything sensible. I'm a bit clumsy with writing the distribution modules with the self / class structure so I haven't been able to really get this done.


stevereyes01:
uhh, I think this is the wrong list of commits. H
old on.

PR #2397, opened 2018-10-23
Use astropy.cosmology in the cosmology module

cdcapano:
Uses `astropy.cosmology` in the cosmology module. Currently, only a `distance_from_comoving_volume` function is added (which will allow for injections uniform in comoving volume), but more can be added later.

Here is a plot showing the luminosity distance returned by the function vs comoving volume: [link](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/use_astropy_cosmology/dist-comoving_volume.png)

This patch also replaces the current `redshift` function with equivalent function from `astropy`. Here is a plot comparing the redshift as a function of luminosity distance between the current redshift function and the new one: [link](https://www.atlas.aei.uni-hannover.de/~cdcapano/public_scratch/github_prs/use_astropy_cosmology/redshift_compare.png). Results are nearly identical.

There's a catch though... the `redshift` function using astropy is much slower than the current `redshift` function (which I've renamed `lalredshift`):
```
>>> distances = numpy.linspace(100, 10000, num=1000)

>>> %timeit redshifts = cosmology.redshift(distances)
1 loop, best of 3: 6.05 s per loop

>>> %timeit lal_redshifts = cosmology.lalredshift(distances)
The slowest run took 4.89 times longer than the fastest. This could mean that an intermediate result is being cached.
10000 loops, best of 3: 80 µs per loop
```
For this reason I didn't remove the current function. I'm not sure what the best long-term solution is though. ~6s for 1000 samples will make plotting source mass posteriors a pain. Typically we have O(10000) samples, so that would take a O(minutes) to do per parameter.

### DIFF
--- a/pycbc/cosmology.py
+++ b/pycbc/cosmology.py
@@ -45,8 +45,7 @@ def get_cosmology(cosmology=DEFAULT_COSMOLOGY):
     Parameters
     ----------
     cosmology : str, optional
-        The name of the cosmology to use. Must be the name of a cosmology
-        available in astropy.cosmology. For the list of options, see
+        The name of the cosmology to use. For the list of options, see
         :py:attr:`astropy.cosmology.parameters.available`. Default is
         ``'Planck15'``.
 
@@ -83,8 +82,7 @@ def redshift(distance, cosmology=DEFAULT_COSMOLOGY):
     distance : float
         The luminosity distance, in Mpc.
     cosmology : str, optional
-        The name of the cosmology to use. Must be the name of a cosmology
-        available in astropy.cosmology. For the list of options, see
+        The name of the cosmology to use. For the list of options, see
         :py:attr:`astropy.cosmology.parameters.available`. Default is
         ``'Planck15'``.
 
@@ -105,8 +103,7 @@ def distance_from_comoving_volume(vc, cosmology=DEFAULT_COSMOLOGY):
     vc : float
         The comoving volume, in units of cubed Mpc.
     cosmology : str, optional
-        The name of the cosmology to use. Must be the name of a cosmology
-        available in astropy.cosmology. For the list of options, see
+        The name of the cosmology to use. For the list of options, see
         :py:attr:`astropy.cosmology.parameters.available`. Default is
         ``'Planck15'``.
 
--- a/pycbc/cosmology.py
+++ b/pycbc/cosmology.py
@@ -29,19 +29,19 @@ Note: in all functions, ``distance`` is short hand for ``luminosity_distance``.
 Any other distance measure is explicitly named; e.g., ``comoving_distance``.
 """
 
-import numpy
 import logging
+import numpy
 from scipy import interpolate
 import astropy.cosmology
 from astropy import units
 from astropy.cosmology.core import CosmologyError
 from pycbc.conversions import ensurearray, formatreturn
 
-DEFAULT_COSMOLOGY='Planck15'
+DEFAULT_COSMOLOGY = 'Planck15'
 
 
 def get_cosmology(cosmology=None, **kwargs):
-    """Gets an astropy cosmology class.
+    r"""Gets an astropy cosmology class.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


pycbc/cosmology.py
```diff
@@ -23,14 +23,104 @@
...
 
+DEFAULT_COSMOLOGY='Planck15'
+
+def get_cosmology(cosmology=DEFAULT_COSMOLOGY):
+    """Gets an astropy cosmology class based on the given string.
+
+    Parameters
+    ----------
+    cosmology : str, optional
+        The name of the cosmology to use. Must be the name of a cosmology
```
tdent: remove 'Must be the name ..' sentence, think it's clear without (maybe just add 'For the list of *possible* options ..'

ahnitz:
@cdcapano Once @tdent is happy with this feel free to merge. I don't see any major issues.

cdcapano:
@tdent @ahnitz I've made several changes. I was worried that the speed of the `redshift` function really would be an issue since we use it often in PE runs to get the source masses. So, I kept my interpolation class around, but changed it to use astropy's redshift function instead of LAL. This has the added benefit that you can now use any of the cosmologies supported by astropy for calculating redshift based on luminosity distance. @tdent In the process, the docstrings changed from what they were.

I've also added `comoving_distance_from_comoving_volume` and `cosmological_quantity_from_redshift`. The later is just a wrapper around functions provided by the astropy cosmology you request.

Anyway, with this patch, to create a injections uniform in comoving volume, you'd put the following in the ini file for `create_injections`:
```
[variable_args]
comoving_volume =
ra =
dec =

[prior-ra+dec]
name = uniform_sky

[prior-comoving_volume]
name = uniform
min-comoving_volume = 1e3
max-comoving_volume = 1e10

[waveform_transforms-distance]
name = custom
inputs = comoving_volume
distance = distance_from_comoving_volume(comoving_volume)
```

ahnitz:
@cdcapano I think that is a reasonable compromise.

pycbc/cosmology.py
```diff
@@ -114,78 +254,145 @@ def get_redshift(self, dist):
...
+        select a cosmology. If none provided, will use
+        :py:attr:`DEFAULT_COSMOLOGY`.
+
+    Returns
+    -------
+    float :
+        The redshift at the given comoving volume.
+    """
+    cosmology = get_cosmology(**kwargs)
+    # first get the redshift associated with the given comoving volume
```
tdent: doesn't this comment belong in the next function instead?

pycbc/cosmology.py
```diff
@@ -23,88 +23,228 @@
...
+
+    Getting a z for a cosmological quantity involves numerically inverting
+    ``func``. The ``zmax`` argument sets how large of a z to guess (see
+    :py:func:`astropy.cosmology.z_at_value` for details). If a z is larger than
+    ``zmax``, this will try a larger zmax up to ``zmax * 10**5``. If that still
+    is not large enough, will just return ``numpy.inf``.
+
+    Parameters
+    ----------
+    func : function or method
```
tdent: i'm curious, what sort of method could this be? presumably something that could be called as e.g. `my_cosmology(z)` and produce a numerical value ..

pycbc/cosmology.py
```diff
@@ -23,88 +23,228 @@
...
+                  Tcmb0=2.725 K, Neff=3.05, m_nu=[0.   0.   0.06] eV,
+                  Ob0=0.0486)
+
+    Don't trust Planck? Use WMAP!
+
+    >>> get_cosmology("WMAP9")
+    FlatLambdaCDM(name="WMAP9", H0=69.3 km / (Mpc s), Om0=0.286, Tcmb0=2.725 K,
+                  Neff=3.04, m_nu=[0. 0. 0.] eV, Ob0=0.0463)
+
+    Don't trust anyone? Create your own!
```
tdent: maybe a bit less chirpy dialogue and a bit more explanation of what happens when you specify some cosmology parameters yourself but not others .. presumably astropy then chooses for you, which is what results in Neff = 3.04 coming out of 'nowhere' ?

pycbc/cosmology.py
cdcapano: Thanks catching that, removed.

pycbc/cosmology.py
cdcapano: Well, really, any method that takes redshift as input, as the doc says. You'd normally use it with one of the methods of `FlatLambdaCDM` though. This string was copied directly from astropy's documentation for `z_at_value`.

pycbc/cosmology.py
cdcapano: Ok, I modified it to be less cheeky. I also pointed the user to the docs for `FlatLambdaCDM` to see where the defaults come from.

PR #2412, opened 2018-10-30
Add final mass/spin from initial parameters to conversions

cdcapano:
This adds `final_mass_from_initial` and `final_spin_from_initial` to `conversions.py`. These functions take in the initial masses (and, optionally, spins) and returns the estimated final mass/spin. The fitting functions uses are the ones used by EOBNR. At some point it'd be good to add support for the PhenomD functions too, but this works for now.

Example:
```
>>> from pycbc import conversions
>>> conversions.final_mass_from_initial(35., 35.)
66.62499655233485
>>> conversions.final_spin_from_initial(35., 35.)
0.6864600000000001
```

These functions can be used when plotting posteriors.

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -1044,6 +1044,119 @@ def final_mass_from_f0_tau(f0, tau, l=2, m=2):
     return (a + b*(1-spin)**c)/(2*numpy.pi*f0*lal.MTSUN_SI)
 
 
+def get_final_from_initial(mass1, mass2, spin1x=0., spin1y=0., spin1z=0.,
+                            spin2x=0., spin2y=0., spin2z=0.,
+                            approximant='SEOBNRv4'):
+    """Estimates the final mass and spin from the given initial parameters.
+
+    Parameters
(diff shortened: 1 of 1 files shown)


micamu:
I like it 👍 
It looks good, but I am not really familiar with that lalsim function. According to the code
https://github.com/lscsoft/lalsuite/blob/master/lalsimulation/src/LALSimBlackHoleRingdown.c#L717
 the returned final mass is scaled by the original total mass, and the returned final spin is scaled by the final mass... So I am wondering if final_spin is really just fs? https://github.com/gwastro/pycbc/pull/2412/files#diff-13f55f699aa6c28aa2a559c96dde9c86R1102

cdcapano:
Forgot about this... @micamu I think it is as you thought: when they say "the final spin is scaled by the final mass", they just mean it's dimensionless spin. Otherwise, I would not have gotten 0.68 for the spin in the above example, which is the value I was expecting.

cdcapano:
@micamu poke

PR #2415, opened 2018-11-10
don't require kombine

ahnitz:
I don't think we need to require this anymore, given the status of kombine support.


cmbiwer:
I am out and will check tomorrow but I thought there was a kombine option in the scatter plot matrix code.

ahnitz:
@cmbiwer I think the point is just to make it optional and not force and install for the standard case. When I get a chance, I'll verify that these places don't cause a problem as long as they aren't called.

ahnitz:
@cmbiwer Since we don't support kombine atm, is the right solution to remove those places as well? They are already protected against kombine import and warn the user why if we leave it.

ahnitz:
Looking at the code it just uses the kombine kde, and does so in a way that is fairly clear what's gone wrong if a user tries to use it and it's not installed.

PR #2420, opened 2018-11-16
prime the pump

ahnitz:
@tdent I think the bug you ran into can be avoided by simply priming the list with an empty array.


tdent:
This is Gareth's run and I'm trying to get him added to gwastro / pycbc, should assign to him to test this eventually

PR #2421, opened 2018-11-19
turn on reporting of integrity error and other performance data

vahi:
AMQP reporting requires python package amqplib to be installed.
if not installed and enabled, then it can break monitord population to sqlite backend.
this is now fixed in pegasus 4.9.1dev also
https://jira.isi.edu/browse/PM-1333

Additionally, the 4.9.1dev RPM's do have python-amqplib listed as a dependency.

If not installing from RPM, then please install amqplib via pip


duncan-brown:
Ack, the Travis test doesn't actually send any information and this will fail without `amqplib` installed, so I fixed this in `requirements.txt`.

PR #2423, opened 2018-11-21
various fixes to marginalized likelihood

ahnitz:
When I was glancing at the marginalized likelihood class, I noticed several problems. I didn't want to forget about those, so I've attempted to fix the ones I know about. This is untested, and I am unlikely to be able to do my own testing anytime soon. Please either merge as is so the code is in a closer to correct state, or we should find someone to take it on to fix this module and test it. 

Various problems noticed and attempted to be corrected
* Time bounds not applied to the SNR times series. This means the entire SNR series would be intergrated, even portions which are padding
* Real part of SNR taken. This breaks phase marginalization while also using time marginalization. 
* Worse method of phase marginalization used. I've reverted this to our old method which is more numerically stable for large values.

### DIFF
--- a/pycbc/inference/models/marginalized_gaussian_noise.py
+++ b/pycbc/inference/models/marginalized_gaussian_noise.py
@@ -371,11 +371,9 @@ class MarginalizedGaussianNoise(GaussianNoise):
                                    for i in self._dist_array])
         if self._margtime:
             bounds = self._marg_prior["time"].bounds
-            self._time_array = numpy.linspace(bounds["time"].min,
-                                              bounds["time"].min, 10**4)
-            self.time_prior = numpy.array(
-                                  [self._marg_prior["time"].pdf(time=i) for
-                                   i in self._time_array])
+            self._time_min = bounds["time"].min
+            self._time_max = bounds["time"].max
+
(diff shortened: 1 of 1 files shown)


pycbc/inference/models/marginalized_gaussian_noise.py
```diff
@@ -371,11 +371,9 @@ def _setup_prior(self):
                                    for i in self._dist_array])
         if self._margtime:
             bounds = self._marg_prior["time"].bounds
-            self._time_array = numpy.linspace(bounds["time"].min,
-                                              bounds["time"].min, 10**4)
-            self.time_prior = numpy.array(
-                                  [self._marg_prior["time"].pdf(time=i) for
-                                   i in self._time_array])
```
cdcapano: Why did you remove the time prior bit? I think the idea with this was so you could use a non-uniform prior on the time. It looks like there was a bug there though in that `time_array` should've been between the min/max bounds not min/min.

cdcapano:
@ahnitz poke

cdcapano:
Superseded by #2524. The other marginalizations still need to be fixed, but that can be addressed in separate PRs.

PR #2432, opened 2018-12-07
Store smoothing parameters as output of fit_sngls_over_multiparam and _param

GarethDaviesGW:


tdent:
The main reason we are doing this is that some values (eg duration) take a lot of time to compute, so storing them simplifies any subsequent stages of plotting, diagnosis etc.

ahnitz:
@tdent Yeah, that seems reasonable. Whenever you are happy and this passes Travis feel free to merge.

PR #2433, opened 2018-12-07
Update triggers.py to use vectorized time delay

ahnitz:


ahnitz:
This should improve the performance to be on par with before.

tdent:
I will see if I can test via e.g. pycbc_page_recovery

tdent:
nb - this will be tested by running pycbc_page_recovery at the same time as #2441 .

tdent:
already addressed by #2441 - hence close

PR #2437, opened 2018-12-13
Plot vt ratio

GarethDaviesGW:


tdent:
@tjma12 note that I'm going to add you as reviewer once you get into the "pycbc maintainers" team.

bin/plotting/pycbc_plot_vt_ratio
```diff
@@ -0,0 +1,103 @@
...
+matplotlib.use('Agg')
+import matplotlib.pyplot as plt
+from collections import OrderedDict
+import numpy as np
+import argparse
+
+parser = argparse.ArgumentParser()
+parser = argparse.ArgumentParser(usage="",
+    description="Plot ratios of VTs calculated at various IFARs using pycbc_page_sensitivity for two analyses")
+parser.add_argument('--comp-1-file', required=True,
```
tdent: suggest to change file names to e.g. 'vt-file-one' and 'vt-file-two' .. numbers appearing in option names make me nervous

tdent:
TJ may want to comment on the aesthetics, but based on e.g. https://ldas-jobs.ligo-wa.caltech.edu/~gdavies/o3dev/bank/compare/compare_vt_single_multi_dimension_ogc_smoothing_split_mchirp_mi.png : 

- The Y axis label is hard to read.  Consider reworking as e.g. 'VT(multidim-ogc) / VT(singlesmooth)' ie. analogous to inline formula rather than displayed equation.

- Does the legend need to be outside the plot?  Couldn't it be located within the plot to fit between two of the vertical lines, thus avoiding the expanse of white around it?

tjma12:
@tdent Specifying `ncol=2` in the legend to flatten out the labels may help with the whitespace problem without needing to dynamically fit it in between two lines.

bin/plotting/pycbc_plot_vt_ratio
```diff
@@ -0,0 +1,103 @@
...
+        plt.xticks(xpos,labels,rotation='horizontal')
+
+
+# get the limit of the x axes, and draw a black line in order to highlight equal comparison
+xlimits = plt.xlim()
+plt.plot(xlimits,[1,1],'k', lw=3, zorder=0)
+plt.xlim(xlimits) # reassert the x limits so that the plot doesn't expand
+
+# do some stuff to make the plot pretty
+ax_mi.legend(bbox_to_anchor=(0.3,1))
```
tjma12: As mentioned in the conversation tab, specifying `ncol=2` here will avoid unnecessary white space on the top of the figure.

bin/plotting/pycbc_plot_vt_ratio
```diff
@@ -0,0 +1,103 @@
...
+        sets_2_errhigh = np.array([fsecond_set['errorhigh'][key][idv] for key in keys])
+        sets_2_errlow = np.array([fsecond_set['errorlow'][key][idv] for key in keys])
+
+	ys = np.divide(data, data_2)
+	yerr_errlow = np.multiply(np.sqrt(np.divide(sets_errlow,data)**2 +
+                np.divide(sets_2_errlow,data_2)**2),ys)
+	yerr_errhigh = np.multiply(np.sqrt(np.divide(sets_errhigh,data)**2 +
+                np.divide(sets_2_errhigh,data_2)**2),ys)
+	ax_mi.errorbar(xpos, ys, yerr=[yerr_errlow,yerr_errhigh],
+                fmt='o', markersize=7, alpha=0.6, linewidth=5,
```
tjma12: I'll leave this to your discretion, but I typically like to specify colors for each trace to avoid having matplotlib's defaults change out from under me. Something like:
```
colors = ['blue', 'orange']
for idv, c in zip(idxs, colors):
    do loop things and plot
```

bin/plotting/pycbc_plot_vt_ratio
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/python
+
+import h5py
+import matplotlib
+matplotlib.use('Agg')
+import matplotlib.pyplot as plt
+from collections import OrderedDict
```
tjma12: Is `OrderedDict` used at any point? I don't see it being called, it might be leftover from the old version.

bin/plotting/pycbc_plot_vt_ratio
tjma12: I agree that option names without numbers would be preferable

bin/plotting/pycbc_plot_vt_ratio
GarethDaviesGW: Correct, will remove

bin/plotting/pycbc_plot_vt_ratio
GarethDaviesGW: makes sense, this will be updated

bin/plotting/pycbc_plot_vt_ratio
GarethDaviesGW: sounds good - I'm quite keen in general on using colorblind-friendly schemes, so have put in one that has been suggested [here](https://github.com/matplotlib/matplotlib/issues/9460#issuecomment-377312565).

Because of the update to allow a generic number of ifar values, I've not used the zipping suggestion, rather using a counter which loops back to the start of the colors list when it gets to that point

bin/plotting/pycbc_plot_vt_ratio
GarethDaviesGW: applied, but using ncols = number of ifars, as this allows for >2 values (though it looks like the actual number of columns maxes out at 3 anyway)

GarethDaviesGW:
> TJ may also want to comment on the aesthetics, but based on e.g. https://ldas-jobs.ligo-wa.caltech.edu/~gdavies/o3dev/bank/compare/compare_vt_single_multi_dimension_ogc_smoothing_split_mchirp_mi.png I'd suggest :
> 
> * The Y axis label is hard to read.  Consider reworking as e.g. 'VT(multidim-ogc) / VT(singlesmooth)' ie. analogous to inline formula rather than displayed equation.
> * Does the legend need to be outside the plot?  Couldn't it be located within the plot to fit between two of the vertical lines, thus avoiding the expanse of white around it?

For comparison, the latest version has an output here:
https://ldas-jobs.ligo-wa.caltech.edu/~gdavies/o3dev/bank/compare/compare_vt_single_multi_dimension_smoothing_split_eta_test.png

note that I purposely used a 3-ifar comparison so that the differences can be seen

tdent:
Nice!  
Now I think about it, how about introducing small fictional x offsets to the plotted FAR values to avoid collisions of error bars?
Also, are the FAR values plotted and labelled those requested by the user, or the 'best matches' from the VT file? 
@tjma12 wouldn't we prefer to show the actual FARs for which VT is calculated?

GarethDaviesGW:
> Nice!
> Now I think about it, how about introducing small fictional x offsets to the plotted FAR values to avoid collisions of error bars?
If I were to do this then it would make sense for me to implement the "TODO?" that I'd put in regarding ensuring equal spaces of the x-positions so that even more things don't clash. I wasn't sure if this would be a good idea or if the current xpos values would suffice?
> Also, are the FAR values plotted and labelled those requested by the user, or the 'best matches' from the VT file?
These are plotted and labelled according to the 'best matches' from the VT file - this is something that should probably be in the help description though so I will add it on the next update
> @tjma12 wouldn't we prefer to show the actual FARs for which VT is calculated?

tdent:
There doesn't seem to be a basic problem with the existing x-axis positioning (IMO). 
Equal-log-spaced x values should be dealt with by having a log scale on the axis .. but let's not get overcomplicated unless needed.

PR #2439, opened 2018-12-16
A couple fixes/improvements to PyCBC Live from ER13

titodalcanton:
* Fix a silly sign error which prevented the lag check from working.
* Write channel names into event XML files, so they show up in GraceDB.
* Tag some log messages with their interferometer.


titodalcanton:
PS I did some testing but let's wait for completion of a more thorough one before merging.

ahnitz:
It certainly look OK. I'll do another look over in the morning and wait to hear back from more thorough testing.

PR #2441, opened 2018-12-17
add tau_0 and tau_3 as options into trigger_fits.get_param

GarethDaviesGW:
(also update fit_sngls_param and fit_sngls_multiparam instructions as a result)

### DIFF
--- a/bin/hdfcoinc/pycbc_page_recovery
+++ b/bin/hdfcoinc/pycbc_page_recovery
@@ -46,7 +46,7 @@ parser.add_argument("--gradient-far", action="store_true",
                     help="Show FAR of found injections as a color gradient")
 parser.add_argument("--f-lower", type=int,
                     help="lower frequency value for e.g. calculation of "
-                         " tempate duration. Only needed for certain parameters")
+                         "template duration. Only needed for certain parameters")
 parser.add_argument("--output-file", required=True)
 args = parser.parse_args()
 
--- a/bin/hdfcoinc/pycbc_page_recovery
+++ b/bin/hdfcoinc/pycbc_page_recovery
@@ -43,7 +43,7 @@ parser.add_argument("--plot-type", choices=["scatter", "err", "fracerr",
                     help="Type of plot to show accuracy of recovery. "
                          "Default='scatter'")
 parser.add_argument("--gradient-far", action="store_true",
-                    help="Show FAR of found injections as a gradient")
+                    help="Show FAR of found injections as a color gradient")
 parser.add_argument("--f-lower", type=int,
                     help="lower frequency value for e.g. calculation of "
                          " tempate duration. Only needed for certain parameters")
@@ -99,7 +99,7 @@ if args.x_param:
         xparam = "end_time_"+site
     if args.x_param == "effective_distance":
         xparam = "eff_dist_"+site
-    inj_xparams = triggers.get_inj_param(injs, xparam, args.ifo)[found]
+    inj_xparams = triggers.get_inj_param(injs, xparam, args.ifo, args)[found]
 # need a string value when defining axis labels even if not used
 args.x_param = args.x_param or "badger"
 
--- a/bin/hdfcoinc/pycbc_page_recovery
+++ b/bin/hdfcoinc/pycbc_page_recovery
@@ -44,6 +44,9 @@ parser.add_argument("--plot-type", choices=["scatter", "err", "fracerr",
                          "Default='scatter'")
 parser.add_argument("--gradient-far", action="store_true",
                     help="Show FAR of found injections as a gradient")
+parser.add_argument("--f-lower", type=int,
+                    help="lower frequency value for e.g. calculation of "
+                         " tempate duration. Only needed for certain parameters")
 parser.add_argument("--output-file", required=True)
 args = parser.parse_args()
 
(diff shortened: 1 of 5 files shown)


GarethDaviesGW:
Cancel this - not enough testing done before opening request. Sorry

tdent:
ok, you can still "push -f" more commits to this branch and reopen the request

GarethDaviesGW:
Will do, just wanted to mark as "do not review" for now

tdent:
should this also work directly within the injection parameter accuracy plot (pycbc_page_recovery ?)
ie can you now produce tau_0 accuracy plots

ahnitz:
@GarethDaviesGW Small comment about the naming. Probably better to use tau0 / tau3 instead of tau_0 / tau3 since that's the terminology we use elsewhere.

tdent:
ok, i agree - 'tau0' is much more frequent than 'tau_0'

GarethDaviesGW:
okay, will alter that

GarethDaviesGW:
The CodeClimate review is coming up with a problem of repeated code (this was repeated in the same way before) for the following sections: 
from get_found_param

        found_param = trstats.get_param(param, args, b['mass1'][:],
                                            b['mass2'][:],
                                            b['spin1z'][:],
                                            b['spin2z'][:])[foundtmp]

and get_inj_param

        inj_param = trstats.get_param(param, args, inj['mass1'][:],
                                            inj['mass2'][:],
                                            inj['spin1z'][:],
                                            inj['spin2z'][:])

the way I can see to sort this is to call get_inj_params but using the bank file as the injection file in the get_found_param function. 

However this would not make as much sense to a reader as doing the above

The thing I am asking is whether we can turn off codeclimate for this issue, or whether we should use the workaround I said with an explanatory comment? both seem quite hacky

tdent:
well, we should check that the code works first - can you demonstrate with a few accuracy plots of recovery of tau0 / recovery of tau3 for injections?

GarethDaviesGW:
Okay so some changes were needed to get pycbc_page_recovery working (getting f_lower into this was something that I overlooked)

image

pycbc/events/trigger_fits.py
```diff
@@ -261,14 +261,18 @@ def get_param(par, args, m1, m2, s1z, s2z):
         parvals = m1 + m2
     elif par =='eta':
         parvals = conversions.eta_from_mass1_mass2(m1, m2)
-    elif par == 'chi_eff':
+    elif par == 'chi_eff' or par =='effective_duration':
```
tdent: do you mean 'effective_spin' ?

GarethDaviesGW: of course i do... brain fart

tdent:
the plot looks ok (though shouldn't there be a colorbar?) - can you check for tau3 as well?

GarethDaviesGW:
Tau3: 

image

plotting tau3 error vs tau0 showed a problem with the `inj_xparams  = ...` line

Once that was fixed (by the commit i will make in a second):

image

GarethDaviesGW:
I will put in a colorbar as well

GarethDaviesGW:
Colorbar issue is due to not using the gradient-far flag

pycbc/events/trigger_fits.py
```diff
@@ -261,14 +261,18 @@ def get_param(par, args, m1, m2, s1z, s2z):
         parvals = m1 + m2
     elif par =='eta':
         parvals = conversions.eta_from_mass1_mass2(m1, m2)
-    elif par == 'chi_eff':
+    elif par == 'chi_eff' or par =='effective_spin':
```
tdent: better to use `elif par in ['chi_eff', 'effective_spin']` here

bin/hdfcoinc/pycbc_page_recovery
```diff
@@ -43,7 +43,10 @@ parser.add_argument("--plot-type", choices=["scatter", "err", "fracerr",
                     help="Type of plot to show accuracy of recovery. "
                          "Default='scatter'")
 parser.add_argument("--gradient-far", action="store_true",
-                    help="Show FAR of found injections as a gradient")
+                    help="Show FAR of found injections as a color gradient")
+parser.add_argument("--f-lower", type=int,
+                    help="lower frequency value for e.g. calculation of "
+                         " tempate duration. Only needed for certain parameters")
```
tdent: typo - 'tempate' .. also don't need the initial space on this line

pycbc/events/triggers.py
```diff
@@ -270,21 +273,14 @@ def get_found_param(injfile, bankfile, trigfile, param, ifo):
...
-
-    return found_param_dict[param]
-
-def get_inj_param(injfile, param, ifo):
+        found_param = trstats.get_param(param, args, b['mass1'][:],
+                                        b['mass2'][:],
+                                        b['spin1z'][:],
+                                        b['spin2z'][:])[foundtmp]
+
+    return found_param
```
tdent: i think it's preferable to directly `return trstats.get_param( .. )` rather than having to define a new variable

pycbc/events/triggers.py
```diff
@@ -297,32 +293,28 @@ def get_inj_param(injfile, param, ifo):
...
-                                               inj['spin2z'][:]),
-        "end_time_"+ifo[0].lower() :
-            inj['end_time'][:] + time_delay(inj['longitude'][:],
-                                            inj['latitude'][:],
-                                            inj['end_time'][:]),
-    }
-    return inj_param_dict[param]
+
+    if param == "end_time_"+ifo[0].lower():
+        inj_param = inj['end_time'][:] + det.time_delay_from_earth_center(
```
tdent: same as above: can just immediately return this value rather than defining inj_param and them returning it

GarethDaviesGW:
issues noted above fixed in latest commit

tdent:
will explicitly ignore the codeclimate 'repeated lines' complaint - other analyses are either not critical or don't seem to be finishing in time.

PR #2453, opened 2019-01-09
Merge cuda decompress

josh-willis:
This request merges the changes made over the last year and a half for GPU support.  It has not yet changed *cpu* versions of functions to not use weave, which I plan to do in a separate request after this is merged.


ahnitz:
@josh-willis I've taken a look through and the general structure all looks fine. If you rebase, we can take a look at some of the automated code quality checks. Some of them *do* seem valid and point to a couple unused variables / imports that should be removed before this is merged.

josh-willis:
@ahnitz I've fixed the one new Codacy unused import warning---note that there was a merge request before your previous comment that fixed a bunch of these.  I'm not actually clear when Codacy runs, as opposed to the other checks; there was no Codacy report for quite a long time after other merge requests.

Important note: this needs to be merged before any changes to remove weave from `compress.py` are made, as it will create major merge conflicts otherwise.

ahnitz:
@josh-willis I don't think codacy runs reliably, but the other two seem two. Can you take a pass through codeclimate? There are some warnings about unused variables in particular. Please check if these are meaningful in our case.

josh-willis:
@ahnitz I've just pushed a commit that fixes the codeclimate complaints that I considered both legitimate and worth fixing. Assuming that the travis tests succeed, I think this should then be merged.

PR #2454, opened 2019-01-10
Removed unused imports of matplotlib.pyplot

duncanmmacleod:
This PR removes some unused/unnecessary imports of `matplotlib.pyplot` in the scripts, which can cause issues where `DISPLAY` is not set, etc etc.

### DIFF
--- a/bin/population/pycbc_population_rates
+++ b/bin/population/pycbc_population_rates
@@ -37,7 +37,6 @@ import pycbc.version
 import h5py
 import numpy as np
 import scipy.stats as ss
-from matplotlib import cm
 from pycbc.population import scale_injections as si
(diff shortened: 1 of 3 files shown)


tdent:
@duncanmmacleod this needs to be assigned a reviewer - which I could do ; after recent changes to the travis setup it also needs to be rebased against current master.

bin/population/pycbc_population_rates
```diff
@@ -37,7 +37,7 @@ import pycbc.version
 import h5py
 import numpy as np
 import scipy.stats as ss
-from matplotlib.pyplot import cm
+from matplotlib import cm
```
tdent: This import doesn't seem to be used anyway.

PR #2456, opened 2019-01-10
Lowfreq

stephanmohr:
Lower frequency cutoff for inner product must be specified in config file.

In the section [model] of the config file, lower_frequency_cutoff must be specified.
f_lower, which is supplied on the command line, might (!?!) still be used for estimating the noise. 
If the two differ, a log is created, but the program is not interrupted.


bin/inference/pycbc_inference
```diff
@@ -151,6 +151,15 @@ with ctx:
     # construct class that will return the natural logarithm of likelihood
     model = models.read_from_config(cp, **model_args)
 
+    # Check if lower frequency cutoff on the command line and in config file are the same
+    if model._f_lower !=     low_frequency_cutoff_dict.values()[0]:
+        warnmsg = """Lower frequency cutoff {} specified in config file 
+        does not equal specification {} in command line argument""".format(
+            model._f_lower,     low_frequency_cutoff_dict.values()[0])
```
cdcapano: Formatting the string like this will cause a newline and large space to be printed. However, if you wrap a string with parentheses, you can do new line continuation by just ending the quote and opening a new quote on the following line. So, change this to:
```
    warnmsg = ("Lower frequency cutoff {} specified in config file does not "
               "equal specification {} in command-line argument"
               .format(model._f_lower, low_frequency_cutoff_dict.values()[0])
```

pycbc/inference/models/base.py
```diff
@@ -351,7 +351,7 @@ class BaseModel(object):
     name = None
 
     def __init__(self, variable_params, static_params=None, prior=None,
-                 sampling_transforms=None):
+                 sampling_transforms=None,**kwargs):
```
cdcapano: Why did you add kwargs here?

pycbc/inference/models/base.py
```diff
@@ -738,7 +738,8 @@ def _init_args_from_config(cls, cp):
         args['sampling_transforms'] = sampling_transforms
         # get any other keyword arguments provided
         args.update(cls.extra_args_from_config(cp, section,
-                                               skip_args=['name']))
+                                               skip_args=['name'],
+                                               dtypes={'lower_frequency_cutoff':float}))
```
cdcapano: Ok, I know it was easiest to just stick this here, but there's a couple of issues with this. First, it doesn't make a lot of sense to have something about lower frequency cutoff in BaseModel, since BaseModel may be used by models that don't involve data. I realize that having that there doesn't affect the running of the code, but the location is obscured. It'd be tough for someone to figure out where in the code the lower frequency cutoff was being set. Second, @soumide1102  is going to adding more functionality after this patch is merged to use different cutoff frequencies for different ifos, so will need some place to write extra lines of code.

So, I suggest removing that, and instead adding a an `_init_args_from_config` function to the `GaussianNoise` model, similar to what's done [here](https://github.com/gwastro/pycbc/blob/c989c1e5a73745e55708feec0c70299248f1f802/pycbc/inference/models/base_data.py#L168-L174). After you call `super(GaussianNoise, self)._init_args_from_config(cp)`, add a couple of lines to read `low-frequency-cutoff` from the `cp` file and add it to the `args`. Later, @soumide1102 can add the extra lines there to parse the low frequency cutoff option for different ifos.

pycbc/inference/models/gaussian_noise.py
```diff
@@ -90,8 +90,9 @@ class GaussianNoise(BaseDataModel):
         values are the data (assumed to be unwhitened). The list of keys must
         match the waveform generator's detectors keys, and the epoch of every
         data set must be the same as the waveform generator's epoch.
-    f_lower : float
+    lower_frequenc_cutoff : float
```
cdcapano: `frequenc` -> `frequency`

Also, call it `low_frequency_cutoff` rather than `lower`.

pycbc/inference/models/gaussian_noise.py
```diff
@@ -214,7 +215,8 @@ class GaussianNoise(BaseDataModel):
     name = 'gaussian_noise'
 
     def __init__(self, variable_params, data, waveform_generator,
-                 f_lower, psds=None, f_upper=None, norm=None,
+                 lower_frequency_cutoff=None, psds=None, 
+                 f_upper=None, norm=None,
```
cdcapano: Don't make `low_frequency_cutoff` a keyword argument, as you did here. Just rename `f_lower` -> `low_frequency_cutoff`. Then you don't need the block of code below to do all of the checking.

pycbc/inference/models/gaussian_noise.py
```diff
@@ -240,10 +242,20 @@ def __init__(self, variable_params, data, waveform_generator,
...
+            lower_frequency_cutoff = float(lower_frequency_cutoff)
+            self._f_lower = lower_frequency_cutoff
+        except (ValueError, TypeError):
+            if lower_frequency_cutoff is None:
+                raise KeyError("lower frequency cutoff "
+                               "needs to be specified in config (.ini) file")
+            else:
+                raise ValueError("lower frequency cutoff must be convertable "
+                                 "to float but type is "
+                                 "{}".format(type(lower_frequency_cutoff)))
```
cdcapano: You don't need this block of code. `low_frequency_cutoff` should already be a float when it's passed to the `__init__` function; if it isn't, that's a bug in the way the class is being initialized. Also, by not making `low_frequency_cutoff` a keyword arg, it will have to be provided.

pycbc/inference/models/gaussian_noise.py
```diff
@@ -456,4 +468,4 @@ def write_metadata(self, fp):
             attrs = fp[fp.samples_group].attrs
         attrs['lognl'] = self.lognl
         for det in self.detectors:
-            attrs['{}_lognl'.format(det)] = self.det_lognl(det)
+            attrs['{}_lognl'.format(det)] = self.det_lognl(det)
```
cdcapano: Not sure what happened here.

bin/inference/pycbc_inference
stephanmohr: Wow, what happened there?!? My version looks damn ugly. Going to change that in a second.
If you do it like this, will the .format apply to both strings?

cdcapano: Yes, the format applies to both.

pycbc/inference/models/base.py
stephanmohr: Ok

pycbc/inference/models/gaussian_noise.py
stephanmohr: No idea

stephanmohr:
Made the changes you suggested. 
However it is unclear to me if the command line argument is used at all, and at which point the noise estimation takes place. I remember someone saying that the separate command line argument is necessary for this, but I can't see where it is used. If noise estimation is already done with the low_frequency_cutoff argument upon initializing the model, then we can take out the command line argument and the corresponding warning completely.

bin/inference/pycbc_inference
```diff
@@ -151,6 +151,15 @@ with ctx:
     # construct class that will return the natural logarithm of likelihood
     model = models.read_from_config(cp, **model_args)
 
+    # Check if lower frequency cutoff on the command line and in config file are the same
+    if model._f_lower != low_frequency_cutoff_dict.values()[0]:
```
soumide1102: @cdcapano @stephanmohr Was looking at the changes in this PR. Is a low-frequency-cutoff being specified in the command line as well as the config file? What's the difference between the two?

cdcapano:
@soumide1102 @stephanmohr Yes, the command-line low frequency cutoff is needed by the strain module to generate fake noise; see [strain.from_cli](https://github.com/gwastro/pycbc/blob/master/pycbc/strain/strain.py#L185). This is called by [inference.option_utils.data_from_cli](https://github.com/gwastro/pycbc/blob/master/pycbc/inference/option_utils.py#L119-L135), which is called by`pycbc_inference` at [line 122](https://github.com/gwastro/pycbc/blob/master/bin/inference/pycbc_inference#L122).

I agree having two different low frequency cutoffs isn't the best, especially since they have the same name. There's a couple things we could do:

1. Rename the option to `fake-strain-low-frequency-cutoff` to make it clear what that option is for. You'd just have to make sure that that got saved to "low_frequency_cutoff" in the parsed arguments before calling strain.from_cli (that can be done by using the "dest" argument in the argument parser).

2. The Correct Solution™: move data reading options from the command line to the config file, then have `GaussianNoise` load the data in its `from_config` function. Since that function knows what the low frequency cutoff for the filter is from the config file, it could just set the cutoff for fake noise to the same value before calling strain from cli.

Option 2 would mean writing a function that read options from a section called `[data]`, then turn those into a string that a fake argument parser could parse and pass to `strain.from_cli`. That function would then be called by `GaussianNoise.from_cli`. This is what I want to do ultimately, but is obviously more work. Julian had started working on it a couple of months ago, but I'm not sure he made any progress. If either of you would like to take it on, feel free.

Otherwise, we can just do option 1 as a stop gap for now. To do that, just change [option_utils.add_low_frequency_cutoff_opt](https://github.com/gwastro/pycbc/blob/master/pycbc/inference/option_utils.py#L75-L76) to:
```
parser.add_argument("--fake-strain-low-frequency-cutoff", type=float,
                    dest="low_frequency_cutoff",
                    help="Low frequency cutoff for fake strain.")
```

soumide1102:
@cdcapano Ok, I see the reason now. I can look at implementing this in the PR to add the separate low-frequency-cutoffs for IFOs. Thanks.

soumide1102:
@cdcapano Thinking about your option 2 again, you said you want to move all the data reading options eventually to the config file? So move everything like `gps-start-time, gps-end-time, strain-high-pass, pad-data, sample-rate, channel-name` etc to the config file? In that case we will not be take advantage of `insert_strain_option_group` (and also `insert_psd_option_group`?). If I understand this correctly, on doing this `pycbc_inference --help` will not return anything useful, in which case it will be difficult to figure out which options are required for reading in the data or generating the psd, and get the code running easily.

cdcapano:
@stephanmohr Yes, please rebase so this can be merged.

cdcapano:
> @cdcapano Thinking about your option 2 again, you said you want to move all the data reading options eventually to the config file? So move everything like `gps-start-time, gps-end-time, strain-high-pass, pad-data, sample-rate, channel-name` etc to the config file? In that case we will not be taking advantage of `insert_strain_option_group` (and also `insert_psd_option_group`?). If I understand this correctly, on doing this `pycbc_inference --help` will not return anything useful, in which case it will be difficult to figure out which options are required for reading in the data or generating the psd, and get the code running easily.

@soumide1102 Yes, all of those options would go into the config file. It's true you won't see their options if you ran `--help`, but I don't think the current help is all that useful anyway because there are so *many* options, many of which we never use. I agree this would require more documentation, but we'd just have to make sure to put clear explanations in the inference docs, with a reference to that page in the `pycbc_inference --help` message. You need to read those docs in any event to know how to set up the prior, model, and sampler anyway.

The other issue with the data options on the command line is that data isn't even used by some of the models, such as the analytic ones. And when @Cyberface finishes adding ROQ support, they also won't be used by the ROQ model. Altogether, you have a large set of options which are only used for one or two models (granted, they are models we use most often).

soumide1102:
Yeah proper documentation has to be provided with option 2. Using some of the documentation from insert_strain(psd)_option_group will be useful. For example, some of the stuff there like the names of the models provided as `CHOICE`s in `--fake-strain IFO:CHOICE [IFO:CHOICE ...]` or `MODEL`s in `--psd-model IFO:MODEL [IFO:MODEL ...]` are useful information.

cdcapano:
@soumide1102 Merging this now. Stephan didn't implement option 1, so the command-line argument for the fake strain is still "low-frequency-cutoff". If you'd like to implement that in your patch, please go ahead.

PR #2459, opened 2019-01-14
Correct sampling_parameters to sampling_params in documentation

stephanmohr:
The documentation used sampling_parameters twice even though sampling_params would be correct. 
Following this section of the documentation and using 
[sampling_parameters]
instead of 
[sampling_params]
in the config file will make SamplingTransforms.fromConfig throw an exception (~ line 200 in inference/models/base.py) but this exception is caught in the process of setting up the model and sampling_transforms will be set to None. This will lead to incorrect behavior of the sampler when using sampling transforms is intended.


cdcapano:
I can't merge until Travis is fixed.

cdcapano:
@stephanmohr Please rebase this one too so that I can merge.

PR #2462, opened 2019-01-15
Pass MKL DFT descriptor by reference to desctructor

### LINKED ISSUE #2202: memory leak in mkl fft useage.
https://github.com/gwastro/pycbc/blob/master/pycbc/fft/mkl.py#L112

It should have been fixed by the above, but this seems not to be the case for mkl. Not sure what is causing the leak at this point, but it does not occur with FFTW.

This was observed in pycbc live testing.

josh-willis:
This PR should fix #2202 . The problem was that the "descriptor" used by MKL's FFT was erroneously being passed by value to the destructor when it should have been passed by reference in the ctypes wrapping. This is now fixed for the function API; the class-based API has never attempted to free the descriptor (that would require a `__del__` method for the classes).

### DIFF
--- a/pycbc/fft/mkl.py
+++ b/pycbc/fft/mkl.py
@@ -100,7 +100,7 @@ def fft(invec, outvec, prec, itype, otype):
     f = lib.DftiComputeForward
     f.argtypes = [ctypes.c_void_p, ctypes.c_void_p, ctypes.c_void_p]
     status = f(descr, invec.ptr, outvec.ptr)
-    lib.DftiFreeDescriptor(descr)
+    lib.DftiFreeDescriptor(ctypes.byref(descr))
     check_status(status)
 
 def ifft(invec, outvec, prec, itype, otype):
@@ -109,7 +109,7 @@ def ifft(invec, outvec, prec, itype, otype):
     f = lib.DftiComputeBackward
     f.argtypes = [ctypes.c_void_p, ctypes.c_void_p, ctypes.c_void_p]
     status = f(descr, invec.ptr, outvec.ptr)
-    lib.DftiFreeDescriptor(descr)
+    lib.DftiFreeDescriptor(ctypes.byref(descr))
     check_status(status)
 
 # Class based API


ahnitz:
@josh-willis Is there any chance that fftw has a similar problem in our code? I.E. here? https://github.com/gwastro/pycbc/blob/65f16a8f00b08089244046734a70eb95002c67d9/pycbc/fft/fftw.py#L333

josh-willis:
> @josh-willis Is there any chance that fftw has a similar problem in our code? I.E. here?
> 
> [pycbc/pycbc/fft/fftw.py](https://github.com/gwastro/pycbc/blob/65f16a8f00b08089244046734a70eb95002c67d9/pycbc/fft/fftw.py#L333)
> 
> Line 333 in [65f16a8](/gwastro/pycbc/commit/65f16a8f00b08089244046734a70eb95002c67d9)
>  destroy(theplan)

FFTW does not document that function as requiring a pointer to a plan, just the plan itself (likewise for plan creation, which simply returns the plan).  But MKL wants all of its functions to return a 'status' which is then checked, and so you can't just pass a pointer, but rather pointer-to-pointer, for things that are by reference. Hence the need for the constructor/destructor in MKL to be passed a pointer to something that is itself a pointer (for both FFTW and MKL, the opaque plan types are really just pointers). But for FFTW we should have been OK.

ahnitz:
@josh-willis So should we change these lines too? https://github.com/gwastro/pycbc/blob/65f16a8f00b08089244046734a70eb95002c67d9/pycbc/fft/mkl.py#L102

josh-willis:
> @josh-willis So should we change these lines too?
> 
> [pycbc/pycbc/fft/mkl.py](https://github.com/gwastro/pycbc/blob/65f16a8f00b08089244046734a70eb95002c67d9/pycbc/fft/mkl.py#L102)
> 
> Line 102 in [65f16a8](/gwastro/pycbc/commit/65f16a8f00b08089244046734a70eb95002c67d9)
>  status = f(descr, invec.ptr, outvec.ptr)

No. Those are not calls to the constructor/destructor; they require the DFTI descriptor, not a pointer to it.

I ran the unit-tests by hand before opening the pull-request, so I don't think there's any major bug that would cause everything to just stop working.  This went uncaught because the only symptom was memory that wasn't deallocated, rather than the actual creation or execution of an FFT.

ahnitz:
@josh-willis Ok, thank!

PR #2466, opened 2019-01-22
Multiifo inj coinc

tapaimarton:
This PR adds the injinj, injfull and fullinj coinc jobs to the workflow. The workflow generates, but the jobs are not included when ran with map reuse. Might be a config issue. Please comment if I did something obvious wrong.

### DIFF
--- a/bin/workflows/pycbc_create_offline_search_workflow
+++ b/bin/workflows/pycbc_create_offline_search_workflow
@@ -369,7 +369,7 @@ for inj_file, tag in zip(inj_files, inj_tags):
             # Create coinc tag, and set up the coinc job for the combination
             coinctag = '{}det_'.format(len(ifocomb)) + pivot_ifo + fixed_ifo
             ctagcomb = [tag, 'inj', coinctag]
-            inj_coinc += wf.setup_multiifo_interval_coinc_inj(workflow, hdfbank,
+            inj_coincs += wf.setup_multiifo_interval_coinc_inj(workflow, hdfbank,
                                     full_insps, insps, statfiles,
                                     final_veto_file[0], final_veto_name[0],
                                     output_dir, pivot_ifo, fixed_ifo, tags=ctagcomb)
--- a/bin/workflows/pycbc_create_offline_search_workflow
+++ b/bin/workflows/pycbc_create_offline_search_workflow
@@ -356,9 +356,9 @@ for inj_file, tag in zip(inj_files, inj_tags):
 
             # Use only the inspiral files required for the detector combination
             inspcomb = wf.core.FileList()
-            for ifo, f in zip(*insps.categorize_by_attr('ifo')):
+            for ifo, fi in zip(*insps.categorize_by_attr('ifo')):
                 if ifo in combination_prec:
-                    inspcomb += f
+                    inspcomb += fi
 
             # pivot_ifo is the ifo in combination_prec with the smallest precedence value
             pivot_ifo = min(combination_prec, key=combination_prec.get)
--- a/bin/workflows/pycbc_create_offline_search_workflow
+++ b/bin/workflows/pycbc_create_offline_search_workflow
@@ -239,9 +239,9 @@ for i in range(2, len(insps)+1):
 
         # Use only the inspiral files required for the detector combination
         inspcomb = wf.core.FileList()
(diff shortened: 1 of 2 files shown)


bin/workflows/pycbc_create_offline_search_workflow
```diff
@@ -345,7 +345,34 @@ for inj_file, tag in zip(inj_files, inj_tags):
...
+        combinations = itertools.combinations(ifo_ids.keys(), i)
+        for ifocomb in combinations:
+            # combination_prec stores precedence info for the detectors in the combination
+            combination_prec = {}
+            for ifo in ifocomb:
+                combination_prec[ifo] = ifo_ids[ifo]
+
+            # Use only the inspiral files required for the detector combination
+            inspcomb = wf.core.FileList()
+            for ifo, file in zip(*insps.categorize_by_attr('ifo')):
```
tdent: 'file' is a python builtin.  so do e.g. `for ifo, f in ...`

tdent: (this also ought to be fixed on line 242-244)

pycbc/workflow/coincidence.py
```diff
@@ -546,6 +546,68 @@ def setup_interval_coinc(workflow, hdfbank, trig_files, stat_files,
...
+        raise ValueError('Must use exactly 1 bank file for this coincidence '
+                         'method, I got %i !' % len(hdfbank))
+    hdfbank = hdfbank[0]
+    # Wall time knob and memory knob
+    factor = int(workflow.cp.get_opt_tags('workflow-coincidence', 'parallelization-factor', tags))
+
+    ffiles = {}
+    ifiles = {}
+    ifos, files = full_data_trig_files.categorize_by_attr('ifo')
+    for ifo, file in zip(ifos, files):
```
tdent: same remarks as for previous iterator over `ifo, file` : don't use 'file' as a variable name. 
Also I think you can do, as was done in the workflow code, `zip(*full_data_trig_files.categorize_by_attr('ifo'))` 
& save a few lines of code

pycbc/workflow/coincidence.py
```diff
@@ -546,6 +546,68 @@ def setup_interval_coinc(workflow, hdfbank, trig_files, stat_files,
...
+        ffiles[ifo] = file[0]
+    ifos, files = inj_trig_files.categorize_by_attr('ifo')
+    for ifo, file in zip(ifos, files):
+        ifiles[ifo] = file[0]
+
+    injinjFileList = FileList()
+    injfullFileList = FileList()
+    fullinjFileList = FileList()
+    for ifo in ifos:
+        if ifo == pivot_ifo:
```
tdent: perhaps add a comment here describing the logic, ie separating 'pivot ifo' on one side from all the others

pycbc/workflow/coincidence.py
```diff
@@ -546,6 +546,68 @@ def setup_interval_coinc(workflow, hdfbank, trig_files, stat_files,
...
+    ffiles = {}
+    ifiles = {}
+    ifos, files = full_data_trig_files.categorize_by_attr('ifo')
+    for ifo, file in zip(ifos, files):
+        ffiles[ifo] = file[0]
+    ifos, files = inj_trig_files.categorize_by_attr('ifo')
+    for ifo, file in zip(ifos, files):
+        ifiles[ifo] = file[0]
+
+    injinjFileList = FileList()
```
tdent: prefer snakecase naming for variables, eg `injinj_files`.

tapaimarton:
@tdent I think I addressed all your comments, and I corrected what was missing, the workflow now generates the jobs.

tdent:
@tapaimarton I'm not sure why doing `zip(*full_data_trig_files.categorize_by_attr('ifo'))` produces an error as it seems to be completely analogous to `zip(*insps.categorize_by_attr('ifo'))`. Can you see what `full_data_trig_files.categorize_by_attr('ifo')` actually evaluates to?

pycbc/workflow/coincidence.py
tdent: ok, ian pointed out what was the problem here, I've added a commit directly which should correctly compress these into a single assignment

PR #2468, opened 2019-01-24
Move waveform dependence out of BaseDataModel

cdcapano:
This moves stuff about waveform generation out of `BaseDataModel`. Now `BaseDataModel` only introduces attributes for data, and the concepts of log likelihood ratio and log likelihood noise. This should make it easier inheriting for things like the ROQ model.

Setting up the waveform generator is now moved to within `GaussianNoise.__init__`, with the actual generator setup farmed out to stand alone functions. This means that `GaussianNoise` no longer takes the generator as input, but instead uses the other inputs (which provide enough information) it already needed to set up the generator. This makes it simpler to initialize `GaussianNoise`.

Finally, I moved the stuff about loading low frequency cutoff from the config file that Stephan added to a stand alone function, which is now called directly from a reimplementation of `from_config` for `GaussianNoise`, thereby getting rid of a long spaghetti train of `_init_args_from_config`. I also added a function to read the high frequency cutoff from the config file, so that functionality may also be used when doing runs.

### DIFF
--- a/pycbc/inference/models/base.py
+++ b/pycbc/inference/models/base.py
@@ -709,7 +709,8 @@ class BaseModel(object):
         """Helper function for loading parameters.
 
         This retrieves the prior, variable parameters, static parameterss,
-        constraints, and sampling transforms.
+        constraints, sampling transforms, and waveform transforms
+        (if provided).
 
         Parameters
         ----------
@@ -720,7 +721,9 @@ class BaseModel(object):
         -------
         dict :
             Dictionary of the arguments. Has keys ``variable_params``,
-            ``static_params``, ``prior``, and ``sampling_transforms``.
+            ``static_params``, ``prior``, and ``sampling_transforms``. If
+            waveform transforms are in the config file, will also have
+            ``waveform_transforms``.
         """
         section = "model"
         prior_section = "prior"
--- a/pycbc/inference/models/base.py
+++ b/pycbc/inference/models/base.py
@@ -330,6 +330,13 @@ class BaseModel(object):
     sampling_transforms : list, optional
         List of transforms to use to go between the ``variable_params`` and the
         sampling parameters. Required if ``sampling_params`` is not None.
+    waveform_transforms : list, optional
+        A list of transforms to convert the ``variable_params`` into something
+        understood by the likelihood model. This is useful if the prior is
+        more easily parameterized in parameters that are different than what
+        the likelihood is most easily defined in. Since these are used solely
+        for converting parameters, and not for rescaling the parameter space,
+        a Jacobian is not required for these transforms.
 
     Properties
     ----------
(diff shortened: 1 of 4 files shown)


soumide1102:
Reading and applying the waveform_transforms have been moved to `BaseModel`. But wasn't that supposed to be the base class for all models including the analytic ones? Since there is no waveform involved in the analytic models, having a transform with the word `waveform` in `base.py` instead of somewhere like `base_data.py` looks a bit confusing.

cdcapano:
@soumide1102 Yeah, I moved waveform transforms to BaseModel. I struggled with where to put those.  It is weird to have something called `waveform_transforms` in BaseModel, when BaseModel knows nothing about waveforms. Generally, the waveform transforms offer the ability to convert parameters that the prior is specified in to parameters that the likelihood model understands. That is functionality that you could want with any model, which is why I decided to stick them there. It'd make more sense if the name was different, like `parameter_transforms` instead of `waveform_transforms`, but changing that would require a bunch of changes that I didn't want to touch in this patch, and that I'm not sure would be worth the effort.

ahnitz:
@cdcapano Please verify with a test that this doesn't break a run that uses the waveform transforms.
@cdcapano @soumide1102 I actually think the base model is the right place for the waveform transforms. The name is not quite right, so perhaps should consider changing the name at some point, but it is really a generic reparameterization functionality independent of actually using a waveform model. I'm fine to leave as is for now.

ahnitz:
@cdcapano Can you check on the backslash comment in the codeclimate?

pycbc/inference/models/base.py
```diff
@@ -736,9 +749,12 @@ def _init_args_from_config(cls, cp):
         except ValueError:
             sampling_transforms = None
         args['sampling_transforms'] = sampling_transforms
-        # get any other keyword arguments provided
-        args.update(
-                cls.extra_args_from_config(cp, section, skip_args=['name']))
+        # get any waveform transforms
+        if any(cp.get_subsections('waveform_transforms')):
+            logging.info("Loading waveform transforms")
+            args['waveform_transforms'] = \
```
soumide1102: The docstring for this function above (lines 712 and 723) would also need to be updated saying waveform transforms are also being retrieved.

cdcapano: Ok, fixed.

cdcapano:
@ahnitz @soumide1102 I just checked that waveform transforms work by using the chirp distance example in the docs, as well as specifying the prior in mchirp q, then relying on the waveform transforms to go to mass1 mass2. Both tests passed.

@ahnitz I also fixed the issue with the doc string code climate caught. As for the other issue with marginalized gaussian noise... that class is a hot mess right now anyway, so I'm ignoring for now.

Can one of you approve?

cdcapano:
Nevermind, just realized @soumide1102 already approved. I'll merge this as soon as Travis passes.

PR #2469, opened 2019-01-25
speed up astropy detector calls

ahnitz:


pycbc/detector.py
```diff
@@ -62,14 +63,45 @@ def get_available_detectors():
...
-    def __init__(self, detector_name):
+    def __init__(self, detector_name, reference_time=1126259462):
+        """ Create class representing a gravitational-wave detector
+
+        Parameters
+        ----------
+        detector_name: str
+            The two character detector string, i.e. H1, L1, V1, K1, I1
+        reference_time: str
+            Default is . In this case, the earth's rotation
```
tdent: Default is what?

ahnitz: It's set to the time of GW150914. Comment updated.

pycbc/detector.py
```diff
@@ -62,14 +67,40 @@ def get_available_detectors():
...
+
+        Parameters
+        ----------
+        detector_name: str
+            The two character detector string, i.e. H1, L1, V1, K1, I1
+        reference_time: str
+            Default is time of GW150914. In this case, the earth's rotation
+        will be estimated from a reference time. If 'None', we will
+        calculate the time for each gps time requested explicitly
+        using a slower but higher precision method.
```
cdcapano: `reference_time` should be `int`, no? Also, the lines after the first should be indented.

PR #2472, opened 2019-01-25
Multiifo combine statmap [WIP]

GarethDaviesGW:
Adding multiifo capability to the combine_statmap code

lots of bookkeeping - might need significant thought about how to combine different parts of the files which are no longer the same

To make it easier to see the differences between combine_statmap and multiifo_combine_statmap, here is a diff file:
[combine_statmap_diff.txt](https://github.com/gwastro/pycbc/files/2796820/combine_statmap_diff.txt)

I will try to update this following any significant commits - though it looks like it may have to be significantly different so a diff will essentially be the whole thing


GarethDaviesGW:
Going to close this and reopen a PR which has been created from a clean git branch on local machine

PR #2474, opened 2019-01-29
Multiifo inj statmap

tapaimarton:
This PR adds statmap for the multiifo injection coinc files.

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
@@ -46,10 +46,10 @@ zdata = pycbc.io.MultiifoStatmapData(files=args.zero_lag_coincs, ifos = args.ifo
 zdata = zdata.cluster(window)
 
 logging.info("Loading coinc full inj triggers")
-fidata = pycbc.io.MultiifoStatmapData(files=mixed_coincs_full_inj, ifos = args.ifos).cluster(window)
+fidata = pycbc.io.MultiifoStatmapData(files=args.mixed_coincs_full_inj, ifos = args.ifos).cluster(window)
 
 logging.info("Loading coinc inj full triggers")
-ifdata = pycbc.io.MultiifoStatmapData(files=mixed_coincs_inj_full, ifos = args.ifos).cluster(window)
+ifdata = pycbc.io.MultiifoStatmapData(files=args.mixed_coincs_inj_full, ifos = args.ifos).cluster(window)
 
 f = h5py.File(args.output_file, "w")
 
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
@@ -98,10 +98,10 @@ if len(zdata) > 0:
     logging.info('calculating injection backgrounds')
     fore_time = {}
     for ifo in args.ifos:
-        fore_time[ifo] = zdata.data['%s/time' % ifo][fore_locs]
+        fore_time[ifo] = zdata.data['%s/time' % ifo]
     ftimes = 0
     for ifo in args.ifos:
-        ftimes += fore_time[ifo] / len(ifos)
+        ftimes += fore_time[ifo] / len(args.ifos)
     start, end = ftimes - args.veto_window, ftimes + args.veto_window
 
     fnlouder = numpy.zeros(len(ftimes), dtype=numpy.float32)
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
@@ -96,7 +96,7 @@ if len(zdata) > 0:
     f['foreground/fap_exc'] = fap_exc
 
     logging.info('calculating injection backgrounds')
-    ifotimes = zip(zdata.data['%s/time' % ifo] for ifo in ifos)
+    ifotimes = zip(zdata.data['%s/time' % ifo] for ifo in args.ifos)
     ftimes = numpy.concatenate(ifotimes).mean(axis=1)
     start, end = ftimes - args.veto_window, ftimes + args.veto_window
 
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
@@ -112,12 +112,12 @@ if len(zdata) > 0:
     # this is determined
     # by the argument order to pycbc_coinc_findtrigs
     pivot_ifo = zdata.attrs['pivot']
-    ifsort = ifdata['%s/time' % pivot_ifo].argsort()
-    ifsorted = ifdata['%s/time' % pivot_ifo][ifsort]
+    ifsort = ifdata.data['%s/time' % pivot_ifo].argsort()
+    ifsorted = ifdata.data['%s/time' % pivot_ifo][ifsort]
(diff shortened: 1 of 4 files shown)


pycbc/workflow/coincidence.py
```diff
@@ -161,6 +161,25 @@ def create_node(self, coinc_files, ifos, tags=None):
         node.new_output_file_opt(seg, '.hdf', '--output-file', tags=tags)
         return node
 
+class PyCBCMultiifoStatMapInjExecutable(Executable):
```
ahnitz: I think this need not be duplicated as the options seem to be same. Is there any reason not to use the existing class, instead of adding a new one? If you are just wanting to add an alias, you can do. 
```
class PyCBCMultiifoStatMapExecutable(Executable)
    pass
```

tapaimarton: Actually, if you check carefully, as all multiifo executables, I take the combination of ifos as an argument, (ifos) that is extra compared to the previous ones.

bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
```diff
@@ -0,0 +1,157 @@
...
+f.attrs['background_time_exc'] = background_time
+f.attrs['foreground_time_exc'] = coinc_time
+f.attrs['background_time'] = background_time
+f.attrs['foreground_time'] = coinc_time
+
+if len(zdata) > 0:
+    fnlouder_exc = coinc.calculate_n_louder(back_stat, zdata.stat, dec_fac, skip_background=True)
+    ifar_exc = background_time / (fnlouder_exc + 1)
+    fap_exc = 1 - numpy.exp(- coinc_time / ifar_exc)
+    f['foreground/ifar_exc'] = conv.sec_to_year(ifar_exc)
```
tdent: Could you also change pycbc_coinc_statmap_inj to use this function?

bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
```diff
@@ -0,0 +1,157 @@
...
+    f['foreground/fap_exc'] = fap_exc
+
+    logging.info('calculating injection backgrounds')
+    fore_time = {}
```
tdent: I think these lines could be made more concise .. looks like you are just calculating an average trigger time? 
then can do it with eg 
`ifotimes = zip((zdata.data['%s/time' % ifo] for ifo in ifos))
ftimes = numpy.array(times).mean(axis=1)`
(or something similar depending on what type exactly the zdata field is)

bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
```diff
@@ -0,0 +1,157 @@
...
+    ftimes = 0
+    for ifo in args.ifos:
+        ftimes += fore_time[ifo] / len(args.ifos)
+    start, end = ftimes - args.veto_window, ftimes + args.veto_window
+
+    fnlouder = numpy.zeros(len(ftimes), dtype=numpy.float32)
+    ifar = numpy.zeros(len(ftimes), dtype=numpy.float32)
+    fap = numpy.zeros(len(ftimes), dtype=numpy.float32)
+
+    # We are relying on the injection data set to be the first one,
```
tdent: I don't know what these comments mean - have asked @ahnitz to think about if they are meaningful or needed

pycbc/workflow/coincidence.py
```diff
@@ -655,3 +688,31 @@ def setup_multiifo_interval_coinc(workflow, hdfbank, trig_files, stat_files,
 
     logging.info('...leaving coincidence ')
     return statmap_files
+
+def get_ordered_ifo_list(ifocomb, ifo_ids, insps):
+    # combination_prec stores precedence info for the detectors in the combination
```
tdent: please add a docstring that describes what this function does

pycbc/workflow/coincidence.py
```diff
@@ -655,3 +688,31 @@ def setup_multiifo_interval_coinc(workflow, hdfbank, trig_files, stat_files,
...
+    combination_prec = {}
+    for ifo in ifocomb:
+        combination_prec[ifo] = ifo_ids[ifo]
+
+    # Use only the inspiral files required for the detector combination
```
tdent: I don't see why this block is in this function, it doesn't seem to have much to do with ifo ordering or precedence. Can it be made a separate function, named eg `select_files_by_ifo_combination()` that just returns inspcomb?

bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
ahnitz: @tdent It means that the injection file and the full data file must be in a particular order. This is also indicated by the naming convention (i.e. if vs fi) and the corresponding argument names. Otherwise, I'm not sure the sorting will work out corectly.

bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
tapaimarton: Done (also removed some trailing spaces)

bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
tapaimarton: So anything decided about this?

bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
tapaimarton: Done (though need to test if it works correctly)

pycbc/workflow/coincidence.py
```diff
@@ -655,3 +688,31 @@ def setup_multiifo_interval_coinc(workflow, hdfbank, trig_files, stat_files,
...
+    inspcomb = FileList()
+    for ifo, fi in zip(*insps.categorize_by_attr('ifo')):
+        if ifo in combination_prec:
+            inspcomb += fi
+
+    # pivot_ifo is the ifo in combination_prec with the smallest precedence value
```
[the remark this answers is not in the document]
tapaimarton: actually, I can create it in one line using the dictionary, just didn't see it before

pycbc/workflow/coincidence.py
tapaimarton: done

pycbc/workflow/coincidence.py
tapaimarton: done

bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj
tdent: I think I know what this is referring to .. so I would change this comment to something like "we expect the injfull file to contain injection data as pivot and fullinj to contain full data as pivot" 
I don't think anything should depend on the _ordering_ of input files to multiifo coinc findtrigs !

pycbc/workflow/coincidence.py
```diff
@@ -655,3 +690,32 @@ def setup_multiifo_interval_coinc(workflow, hdfbank, trig_files, stat_files,
...
+    return inspcomb
+
+def get_ordered_ifo_list(ifocomb, ifo_ids):
+    """
+    This function sorts the combination of ifos (ifocomb) based on the given
+    precedence list (ifo_ids dictionary) and returns the first ifo as pivot
+    the second ifo as fixed, and the ordered list joined as a string.
+    """
+    # combination_prec stores precedence info for the detectors in the combination
+    combination_prec = {}
```
tdent: Think these 3 lines could be compressed into 1 via `combination_prec = {ifo : ifo_ids[ifo] for ifo in ifocomb}` (dictionary comprehension)

PR #2477, opened 2019-02-01
Add unit test for apply_fd_time_shift

cdcapano:
Does what it says. A sine wave is shifted by +/- pi/2 in the frequency domain, then transformed back to the time domain and compared to a cosine. Next, a shift by a non-integer step of the delta t is done and compared to just calling `sin()` on the time stamps shifted by the same amount.

Both the C code and the numpy code are tested.

### DIFF
--- a/test/test_waveform_utils.py
+++ b/test/test_waveform_utils.py
@@ -1,0 +1,90 @@
+import unittest
+import numpy
+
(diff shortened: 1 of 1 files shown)


ahnitz:
:+1:  Always great to see a new unittest.

PR #2479, opened 2019-02-02
Add unittest for pycbc_inference run on gw150914

soumide1102:
@ahnitz @cdcapano Were you thinking of something like this? Tested the `inference_gw150914_test.sh ` script on the 4kHz-4096s GWOSC data and ran in 49s. I've made the script get the GW150914 frame files (downloaded from GWOSC) from `pycbc/test/data`. Let me know if that's where those can be put. I'll then push the frames.


cdcapano:
@soumide1102 I'm not sure about adding the GW150914 GWOSC files to pycbc. They are fairly large (~100MB each), which will slow down pulling. Also, I'm not sure we should be re-releasing GWOSC data. Is it possible to just download them on the fly? @ahnitz any suggestions?

Alternatively, we could just do a unit test on the BBH example. We wouldn't be testing the ability to analyze real data, but that's more a test of the strain module anyway.

For the script itself, could you use the inference.ini file that is in the examples directory? That way we simultaneously test what we tell people to do in the docs, and it'll minimize the number of files people have to change whenever a new feature is added. I know the settings in that ini file are not conducive to being run quickly. You can just modify that in the test suite bash script with sed. In other words, copy run.sh and inference.ini from the examples to the current working directory, then use sed to modify the sampler section to be what you have, so that it runs quickly.

soumide1102:
@cdcapano I've now made changes so that the GW150914 frames are downloaded from GWOSC, the GW150914 inference scripts from the examples directory are copied over, edited with sed, and run for a few iterations/walkers/temperatures. Please can you take a look?

I see that the inspiral test uses a `--frame-type` test https://github.com/gwastro/pycbc/blob/master/examples/inspiral/run.sh#L8 . I'm wondering if we should also do `--frame-type` in the inference test instead of doing the download and using that in `--frame-files`.

cdcapano:
@soumide1102 Actually, yeah, if `frame-type` works, that might be better. @ahnitz can you comment? I don't know how this is done within travis.

tools/pycbc_test_suite.sh
```diff
@@ -69,6 +70,24 @@ else
 fi
 popd
 
+## Run inference on GW150914 data
```
ahnitz: Wouldn't it be clearer to have a dedicated script instead of running sed? This seems like it might be fragile to me.

soumide1102: @ahnitz I had a script to run this unit test on in the first commit of this PR. But @cdcapano wanted to edit and use the scripts in the examples directory so that those are tested as well when this runs. @cdcapano can you comment? 

If a dedicated script is desired let me know the changes you want in what I had initially.

soumide1102:
@cdcapano Okay. Will do that. & what about the way to read the frames. Is wget-ing it from GWOSC and running on those fine or should we do `frame-type`?

soumide1102:
If `--frame-type` is used for the GW150914 inspiral unittest, should also also work for the pycbc_inference unittest. I've updated with the required changes for `--frame-type`.

soumide1102:
I figured the `--config-delete` option still can't be used with the changes in #2606 because `from_cli` in `WorkflowConfigParser` that pycbc_inference is calling was not applying the config deletes. Updated `configuration.py` to include that.

soumide1102:
Okay, `--frame-type` and the other changes worked. Travis passed. @cdcapano Please can you take a look?

(Code climate's complaint is weird and I don't know what I can do about it.)

PR #2480, opened 2019-02-02
Cythonize apply fd time shift

cdcapano:
Switches to using Cython for `apply_fd_time_shift`. This also adds a test of the single precision code to the unit tests.

For dealing with single/double precision, I gave up on trying to be fancy with the fused types, and instead just made two copies of the function, with the types explicitly declared.

I plotted the results of the unit tests and they pass visual inspection. I also did some timing tests on this compared to the current weave version, and the performance is basically the same.
With:
```
fs = FrequencySeries(numpy.random.uniform(size=4096*256+1).astype(numpy.complex), delta_f=1./512)
fs32 = fs.astype(numpy.complex64)
```
Weave:
```
>>> %timeit apply_fseries_time_shift(fs, 4.1, copy=False)
100 loops, best of 3: 4.11 ms per loop
>>> %timeit apply_fseries_time_shift(fs32, 4.1, copy=False)
100 loops, best of 3: 4.27 ms per loop
```
Cython:
```
>>> %timeit apply_fseries_time_shift(fs, 4.1, copy=False)
100 loops, best of 3: 4.1 ms per loop
>>> %timeit apply_fseries_time_shift(fs32, 4.1, copy=False)
100 loops, best of 3: 4.07 ms per loop
```
Aside: I was surprised the single precision has basically the same performance as the double. If I set `copy=True`, then the double precision version is about 30% slower due to having to write the extra bytes. This was done on my laptop (Macbook Pro 2015 with dual-core Core i7), so maybe my processor has some optimizations for double precision operations.


ahnitz:
@cdcapano Why do you separate the functions into a separate module so that now utils has only a single function? You can have python functions in a pyx file and leave it all together.

cdcapano:
@ahnitz Because I didn't know you could do that... see the updated patch.

PR #2481, opened 2019-02-03
p3 env in travis

ahnitz:


ahnitz:
Add python3 environment to the travis build matrix. It is optional for now (i.e. allowed to fail) and will likely continue to do so for a bit, but we should start having this available for people to see.

PR #2488, opened 2019-02-04
Removing frequency lines

micamu:
First steps to incorporate a module that removes frequency lines from the data without using any auxiliary channels. For now, the functions take a list of frequencies and, for each frequency, take the inner product of the data with a line model  to obtain the amplitude and phase of the line. This is done taking the median over several time segments, to remove outliers that could originate from the presence of a signal.
The procedure works very nicely for calibration lines, which are stable and whose frequencies are well known. For other lines, which are time varying, the functions take also chunks of data and calculate the inner product in each chunk separately. Those lines do not get completely removed, but at least they reduced a few orders of magnitude. See for instance:
https://www.atlas.aei.uni-hannover.de/~miriam.cabero/area_theorem/GW150914/cleaned_data.png
which shows the PSD from the original data and after applying the line-cleaning functions (the vertical dashed lines are the frequencies I applied the line-cleaning functions to).

Caveat: if two lines are very close to each other, it becomes difficult to measure the amplitude and phase of each one separately and new artifacts might be introduced. For now these module should be used with care, improvements might follow in the future.

### DIFF
--- a/pycbc/strain/lines.py
+++ b/pycbc/strain/lines.py
@@ -1,0 +1,125 @@
+# Copyright (C) 2019 Miriam Cabero
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 3 of the License, or (at your
+# option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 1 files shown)


ahnitz:
@micamu No major comments, but it would be good to address some of the codeclimate issues.

micamu:
I added the docstrings you requested

micamu:
I don't know why codeclimate wants two blank lines in every function (I guess it is before the `def` line?), but I looked at a few other modules and they have only one blank line...

PR #2492, opened 2019-02-06
Thin inference samples on the fly

cdcapano:
This introduces the ability to thin inference samples before writing to disk (for MCMC samplers), which should substantially cut down on checkpoint file sizes.

Two options for thinning are provided, which are set in the sampler section of the config file. You can either:
 * Set `thin-interval` In that case, samples will be thinned by the given interval before being written to disk on each checkpoint. File sizes will still grow unbounded, but at a slower rate.
 * Set `max-samples-per-chain`. In that case, the maximum number of samples that will be stored per walker and per temperature will be capped at the given value by increasing the thinning interval as the run goes on. This caps the file size to ~ `ntemps x nwalkers x nparameters x max-samples-per-chain x 8` bytes.

For `max-samples-per-chain`, whenever the thinning interval is incremented, current samples on disk are thinned again to make room. To keep the interval between samples the same, the thinning interval can only be incremented in powers of 2. To ensure that enough samples exist to calculate an ACL, `max-samples-per-chain` cannot be set to less than 100.

For long running jobs, using `max-samples-per-chain` can result in the thin interval exceeding the checkpoint interval. In that case, no samples are written on some checkpoints. This can be fixed by having the last sample always written to a different group in the file, and have resumed runs start from there, rather than using the last sample in `samples`. But that can be addressed in a future patch.

In a number of places (like the burn-in module) it's assumed that iteration = index of samples on disk. To fix this, I the burn in class now has a `burn_in_iteration` and a `burn_in_index`. The former is the iteration of the sampler at which burn in occurred; the latter is the index of the samples at which that point occurs.  The burn in iteration is still stored in the file, as this is useful for comparing the performance of different samplers/methods. It's also independent of any thinning that's done.

Along the same lines, the `BaseMCMC` class now has an autocorrelation time (`acts`) attribute in addition to `acls`. The definition I adopted is ACLs gives the number of samples you need to skip of the thinned chains, while ACTs gives the number of iterations between independent samples. I think this is consistent with what's in the literature.

A ``thinned_by`` attribute is added to the samples file giving the thinning interval of the samples. This can be used to convert from burn in iteration to index and ACL to ACT.

I'm running a test on the BBH injection and GW150914 now. Will remove the WIP tag when it's done. I'll also add some documentation about this to the inference docs page.

### DIFF
--- a/docs/inference.rst
+++ b/docs/inference.rst
@@ -155,7 +155,8 @@ the ``[sampler]`` section of the configuration file. They are:
    samples will be thinned on disk, and new samples will be thinned to match.
    The thinning interval will grow with longer runs as a result. To ensure
    that enough samples exist to determine burn in and to measure an
-   autocorrelation length, ``max-samples-per-chain`` must be $\geq$ 100.
+   autocorrelation length, ``max-samples-per-chain`` must be greater than
+   or equal to 100.
 
 ^^^^^^^^^^^^^^^^^^^^^
 Configuring the prior
--- a/docs/inference.rst
+++ b/docs/inference.rst
@@ -127,6 +127,36 @@ max_postrior`` would instead consider the sampler to be burned in when either
 the ``nacl`` *or* ``max_posterior`` tests were satisfied. For more information
 on what tests are available, see the :py:mod:`pycbc.inference.burn_in` module.
 
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Thinning samples (MCMC only)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The default behavior for the MCMC samplers (``emcee``, ``emcee_pt``) is to save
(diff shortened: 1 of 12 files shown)


farr:
Hi Collin,

No comment on the details of the patch, but I'm glad to see this
functionality added to the sampler---thanks!

Will

cdcapano:
@ahnitz @duncan-brown Ok, this is ready for review now. I tested on GW150914 with `thin-interval = 10`, `max-samples-per-chain = 100`, and `max-samples-per-chain = 1000`. Using `max-samples-per-chain = 100` tended to give larger ACLs due to measurement error, so I set `max-samples-per-chain = 1000` in the example config files. That gave comparable results to using no thinning / using a thin interval of 10, and capped the file size at ~800MB.

cdcapano:
I think @duncan-brown is busy, so switching to @dfinstad for review.

cdcapano:
@ahnitz poke

PR #2496, opened 2019-02-08
Add support for multinest

dfinstad:
This will add support for the multinest sampler.

### DIFF
--- a/pycbc/distributions/angular.py
+++ b/pycbc/distributions/angular.py
@@ -406,8 +406,10 @@ class UniformSolidAngle(bounded.BoundedDist):
         return self._azimuthal_angle
 
     def cdfinv(self, param, value):
-        raise NotImplementedError("This distribution has not been "
-                                  "implemented for multinest.")
+        if param == self.polar_angle:
+            return self._polardist.cdfinv(param, value)
+        elif param == self.azimuthal_angle:
+            return self._azimuthaldist.cdfinv(param, value)
 
     def apply_boundary_conditions(self, **kwargs):
         """Maps the given values to be within the domain of the azimuthal and
--- a/pycbc/distributions/angular.py
+++ b/pycbc/distributions/angular.py
@@ -225,7 +225,7 @@ class SinAngle(UniformAngle):
             for bnd in self._bounds.values()])
         self._norm = numpy.exp(self._lognorm)
 
-    def _cdfinv(self, arg, value):
+    def cdfinv(self, arg, value):
         """Return inverse of cdf for mapping unit interval to parameter bounds.
         """
         scale = numpy.cos(self._bounds[arg][0]) - \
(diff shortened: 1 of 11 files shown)


pycbc/inference/sampler/multinest.py
```diff
@@ -0,0 +1,427 @@
...
+		    return -1e100
+		return l
+	    except Exception as e:
+		import sys
+		sys.stderr.write('ERROR in loglikelihood: %s\n' % e)
+		sys.exit(1)
+
+	kwargs['LogLikelihood'] = SafeLoglikelihood
+	kwargs['Prior'] = SafePrior
+	pymultinest.run(**kwargs)
```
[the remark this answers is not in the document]
cdcapano: @dfinstad On second thought, I'm fine to keep the `ns_likelihood` and prior functions as class methods of the sampler, rather than defining a new class. So ignore my last statement. I would prefer that functions not be defined in functions though, so move those definitions outside of the functions.

Do fix the logposterior -> loglikelihood issue though.

cdcapano:
@dfinstad poke

cdcapano:
@dfinstad poke

pycbc/inference/io/multinest.py
```diff
@@ -0,0 +1,163 @@
...
+        """Writes the given number of iterations to the sampler group."""
+        self[self.sampler_group].attrs['niterations'] = niterations
+
+    def write_sampler_metadata(self, sampler):
+        """Writes the sampler's metadata."""
+        self.attrs['sampler'] = sampler.name
+        if self.sampler_group not in self.keys():
+            # create the sampler group
+            self.create_group(self.sampler_group)
+        self[self.sampler_group].attrs['nwalkers'] = sampler.nlivepoints
```
cdcapano: Call it `nlivepoints`, not `nwalkers`.

pycbc/distributions/angular.py
```diff
@@ -388,6 +405,9 @@ def polar_angle(self):
     def azimuthal_angle(self):
         return self._azimuthal_angle
 
+    def cdfinv(self, param, value):
+        raise NotImplementedError("This distribution has not been "
+                                  "implemented for multinest.")
```
cdcapano: You have `cdfinv` defined for `SinAngle` and `UniformAngle`, so you should be able to return this. If `param` is the `polar_angle`, then  you just return `self._polardist.cdfinv(param, value)`; if it's the azimuthal angle you return `self._azimuthaldist.cdfinv(param, value)`.

pycbc/distributions/masses.py
```diff
@@ -75,6 +75,10 @@ def __init__(self, mass1=None, mass2=None):
             raise ValueError("must provide limits for mass2")
         super(UniformComponentMasses, self).__init__(mass1=mass2, mass2=mass2)
 
+    def cdfinv(self, param, value):
+        raise NotImplementedError("This distribution has not been "
+                                  "implemented for multinest.")
```
cdcapano: Change the message here to: `An inverse CDF function has not been implemented for this distribution.` This isn't multinest specific.

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +31,9 @@ class UniformSky(angular.UniformSolidAngle):
     _default_polar_angle = 'dec'
     _default_azimuthal_angle = 'ra'
 
+    def cdfinv(self, param, value):
+        raise NotImplementedError("This distribution has not been "
+                                  "implemented for multinest.")
+
```
cdcapano: You can remove this. You'll just inherit the `cdfinv` from `UniformSolidAngle`.

PR #2499, opened 2019-02-12
Multiifo hdf injfind

GarethDaviesGW:
hdf injfind code altered to allow for multiifo-style input. The way to differentiate between original-style and multi-ifo-style input is in the input hdf file

Difference in output:
 - `optimal_snr_N` where `N` is 1 or 2 becomes `optimal_snr_I1` for `I1` as detector Identifier
 - `found/timeN` --> `found/I1/time`, similarly for `trigger_id` instead of `time` and `found_after_vetoes` instead of `found`

Note that some file-reading parts may change depending on throughput options chosen earlier in the pipeline

### DIFF
--- a/bin/hdfcoinc/pycbc_coinc_hdfinjfind
+++ b/bin/hdfcoinc/pycbc_coinc_hdfinjfind
@@ -9,6 +9,7 @@ from glue.ligolw import ligolw, table, lsctables, utils as ligolw_utils
 from glue import segments
 import os.path, numpy
 from pycbc.events import indices_within_segments as veto_indices
+import pycbc.events
 from pycbc.types import MultiDetOptionAction
 import pycbc.version
 
@@ -24,7 +25,7 @@ def hdf_append(f, key, value):
         f[key] = tmp
     else:
         f[key] = value
-       
+
 h5py.File.append = types.MethodType(hdf_append, h5py.File)
 
 def keep_ind(times, start, end):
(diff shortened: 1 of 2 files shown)


bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -74,113 +76,197 @@ if args.verbose:
...
+             time_dict[ifo] = f['foreground/%s/time' % ifo][:]
+             trig_dict[ifo] = f['foreground/%s/trigger_id' % ifo][:]
+        time = numpy.array([pycbc.events.mean_if_greater_than_zero(vals)[0]
+                            for vals in zip(*ifo_times)])
+        # The injections missed because of 'off' time will be in
+        # 'no-coincident-detector' time, so need to combine segments
+        # to give the time when any two detectors are on
+        any_seg = segments.segmentlist([])
+        for key in f:
+            if key == 'foreground':
```
tdent: can't you just iterate over keys in f['segments'] ?

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -74,113 +76,197 @@ if args.verbose:
...
+                continue
+            elif key == 'segments':
+                for k2 in f['segments']:
+                    starts = f['segments/' + k2 + '/start'][:]
+                    ends = f['segments/' + k2 + '/end'][:]
+                    any_seg += pycbc.events.start_end_to_segments(starts,ends)
+            else:
```
tdent: if one of these is the 'old' behaviour then add a comment saying so

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -74,113 +76,197 @@ if args.verbose:
...
-                                    segment_name=args.segment_name)
-    i2, v2 = veto_indices(inj_time, [args.veto_file], ifo='L1', 
-                                     segment_name=args.segment_name)
-    vi = numpy.concatenate([i1, i2])
+    if multi_ifo_style:
+        vi = numpy.array([])
+        for ifo in ifo_list:
+            vi = numpy.append(vi, veto_indices(inj_time, [args.veto_file],
+                                    ifo=ifo, segment_name=args.segment_name))
+    else:
```
tdent: is this not equivalent to the new style code with ifo list ['H1', 'L1'] ?  If so then probably not worth having 2 separate blocks here. 
Main reason to separate old/new behaviour is for different data formats.

bin/hdfcoinc/pycbc_coinc_hdfinjfind
GarethDaviesGW: I forgot that this bit was there - it shouldn't actually have the segments key if it is using the new style file - as it will not have the segments group. The group will now be the different detector combos, and so I can remove this elif statement safely

bin/hdfcoinc/pycbc_coinc_hdfinjfind
GarethDaviesGW: Yes it is - I will update to have ifo_list set as ['H1','L1'] given old style files

GarethDaviesGW:
For clarity (it is in a commit message but should be highlighted as it is in the description) - the --ifos option is no longer needed and the new style/old style file is determined by checking for the /foreground/time1 key

tdent:
@GarethDaviesGW you can edit the description here if you want. in any case, the only things we are reviewing are the code & its comments, don't spend your time cleaning up something that will be thrown away and not seen by anyone

GarethDaviesGW:
> Looks mostly straightforward, if tested, only a couple of remarks.
> 
> There are several occurrences of old style/ new style code. These are in different orders through the code and introduced by different if/then logic. Can it always be the same logic please?
> I'd personally prefer 'if multi_ifo_style / else' since going forwards we will generally be using the multi ifo case & keep the old one just for backwards compatibility

This has now been implemented

GarethDaviesGW:
See links from #2516 for the foundmissed plots to confirm that the injections are being found in the same way

tdent:
Currently waiting on travis to be fixed

PR #2501, opened 2019-02-12
Add zero likelihood analytic model

dfinstad:
This adds the analytic zero-likelihood model, whose `loglikelihood` method just returns the `logprior`.

### DIFF
--- a/pycbc/inference/models/analytic.py
+++ b/pycbc/inference/models/analytic.py
@@ -190,6 +190,7 @@ class TestVolcano(BaseModel):
             numpy.exp(-r/35) + 1 / (sigma * numpy.sqrt(2 * numpy.pi)) *
             numpy.exp(-0.5 * ((r - mu) / sigma) ** 2))
 
+
 class TestZeroLikelihood(BaseModel):
     r"""The test distribution is zero likelihood.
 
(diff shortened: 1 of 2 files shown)


pycbc/inference/models/analytic.py
```diff
@@ -189,3 +189,25 @@ def _loglikelihood(self):
         return 25 * (
             numpy.exp(-r/35) + 1 / (sigma * numpy.sqrt(2 * numpy.pi)) *
             numpy.exp(-0.5 * ((r - mu) / sigma) ** 2))
+
+class TestZeroLikelihood(BaseModel):
```
cdcapano: Call it "TestPrior".

pycbc/inference/models/analytic.py
```diff
@@ -189,3 +189,25 @@ def _loglikelihood(self):
...
+    name = "test_zero_likelihood"
+
+    def __init__(self, variable_params, **kwargs):
+        # set up base likelihood parameters
+        super(TestZeroLikelihood, self).__init__(variable_params, **kwargs)
+
+    def _loglikelihood(self):
+        """Returns the log prior.
+        """
+        return self.logprior
```
cdcapano: This should return 0, not logprior. If you return logprior for the `loglikelihood`, then you end up with `prior^2` as the log posterior (since `logprior` will also be called by `emcee_pt`, and via `logposterior` when using `emcee`).

cdcapano:
@dfinstad Oh, fix the code climate issue too; that's serious.

PR #2505, opened 2019-02-13
Made hardware injections accept flexible list of detectors

tjma12:
Moved detector selection into a command line argument so that `--ifos H1 L1 V1` can be invoked rather than a hard-coded list. Tested using an injection XML.

### DIFF
--- a/bin/hwinj/pycbc_generate_hwinj_from_xml
+++ b/bin/hwinj/pycbc_generate_hwinj_from_xml
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
 # Copyright (C) 2015 Christopher M. Biwer
 #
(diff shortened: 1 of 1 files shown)


cmbiwer:
This looks fine.

One suggestion could be:
```
from pycbc import detectors
...
parser.add_argument(..., choices=zip(*detector.get_available_detectors())[0], ...)
```

Which gives a list of all IFOs available in PyCBC, currently:
```
('G1', 'A1', 'H1', 'T1', 'V1', 'E3', 'C1', 'P1', 'K1', 'E2', 'E0', 'I1', 'E1', 'O1', 'B1', 'H2', 'L1', 'N1')
```

To prevent incorrect options.

tjma12:
Good point, added a new commit and checked that the choices are being checked/displayed in the help string

PR #2507, opened 2019-02-19
Minor updates to pycbc.psd.estimate for python3

duncanmmacleod:
This PR fixes a few issues in `pycbc.psd.estimate` to make it (the `welch` function) work under python3. Two categories of change:

- use `int` arguments when creating arrays of fixed size
- use `six.moves.range` instead of `xrange`

### DIFF
--- a/pycbc/psd/estimate.py
+++ b/pycbc/psd/estimate.py
@@ -16,6 +16,8 @@
 """Utilites to estimate PSDs from data.
 """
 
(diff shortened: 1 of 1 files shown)


ahnitz:
@duncanmmacleod Thanks a lot for this! I really appreciate the help getting pycbc moving towards python3 support. I see no issues with this patch.

PR #2508, opened 2019-02-19
Some Python 3 compatibility fixes

mattpitkin:
This PR just (similarly to #2507) just fixes some divisions to make sure they produce integers when the output values are being used as array indices.

### DIFF
--- a/pycbc/noise/gaussian.py
+++ b/pycbc/noise/gaussian.py
@@ -102,8 +102,8 @@ def noise_from_psd(length, delta_t, psd, seed=None):
     randomness = lal.gsl_rng("ranlux", seed)
 
     N = int (1.0 / delta_t / psd.delta_f)
-    n = N/2+1
-    stride = N/2
+    n = N//2+1
+    stride = N//2
 
     if n > len(psd):
         raise ValueError("PSD not compatible with requested delta_t")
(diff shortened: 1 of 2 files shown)


mattpitkin:
Thanks @ahnitz. If I spot any more Python 3 compatibility issues I'll let you know (I'm using PyCBC for some undergraduate lab exercises and it's really helpful, but has thrown up these issues as the students are all using Python 3!)

ahnitz:
@mattpitkin Ah, thanks! That would be much appreciated. Python3 support is still be a iffy. I apologize for the trouble that may cause you and your students in the short term. 

I'd be very curious to know what kind of lab exercises you are doing at some point. If they are broadly useful, it might be good to share them around, or at them the the set of tutorials. I assume you know how to get to the pycbc chat. Feel free to drop us a line and let us know how it went!

mattpitkin:
@ahnitz The projects that the students are doing are fairly open and defined by them, so I've not been taking them through specific exercises other than pointing them to the PyCBC documentation and open data workshop tutorials. A lot of them are using PyCBC for signal generation, calculating matches, and running basic matched filtering on the open data. If I find any common questions that come up I can pass them on to help improve the documentation.

ahnitz:
Thanks @mattpitkin would be glad to hear what problems they run into.

PR #2513, opened 2019-02-19
Quick fix for SNR in plot_posterior

micamu:
Currently, the `--z-arg snr` option in `pycbc_inference_plot_posterior` is broken.
This PR is a quick fix to be able to plot the SNR again.


micamu:
This will still not work if the input-file has been generated using `pycbc_inference_extract_samples`, because the attributes are not stored in that case (and to calculate the SNR we need the attribute 'lognl'). I have another branch to work on this issue, but that requires a bit more digging into Inference's types and formats and some thought on how to store the 'lognl', so I made this quick fix to at least be able to use it on the results files for now

micamu:
@cdcapano is this PR ok to merge?

bin/inference/pycbc_inference_plot_posterior
```diff
@@ -77,11 +78,15 @@ samples = samples if isinstance(samples, list) else [samples]
 # if a z-arg is specified, load samples for it
 if opts.z_arg is not None:
     logging.info("Getting samples for colorbar")
+    z_arg = 'loglikelihood' if opts.z_arg == 'snr' else opts.z_arg
```
cdcapano: You can simplify this. Just do:
```
if opts.z_arg == 'snr':
    opts.z_arg = 'snr_from_loglikelihood(loglikelihood-lognl)'
```
Then you don't need the changes below.

bin/inference/pycbc_inference_plot_posterior
```diff
@@ -77,11 +78,15 @@ samples = samples if isinstance(samples, list) else [samples]
...
     zlbl = opts.z_arg_labels[opts.z_arg]
     zvals = []
     for fp in fps:
-        zsamples = fp.samples_from_cli(opts, parameters=opts.z_arg)
-        zvals.append(zsamples[opts.z_arg])
+        zsamples = fp.samples_from_cli(opts, parameters=z_arg)
+        if opts.z_arg == 'snr':
+            loglr = zsamples[z_arg] - zsamples.lognl
+            zsamples[z_arg] = conversions.snr_from_loglr(loglr)
+        zvals.append(zsamples[z_arg])
```
cdcapano: Make the change suggested above, then remove these changes.

pycbc/inference/io/base_hdf.py
```diff
@@ -181,7 +181,8 @@ def read_samples(self, parameters, array_class=None, **kwargs):
         # convert to FieldArray
         samples = array_class.from_kwargs(**samples)
         # add the static params
-        for (p, val) in self.static_params.items():
+        for (p, val) in self.static_params.items() + \
+                        self[self.samples_group].attrs.items():
```
cdcapano: (PEP8) This is over indented (see the code climate report). I think reducing the indent to two tabs past the `for` might do it. Alternatively, create a variable above, say, `addattrs = ...`, then do `for (p, val) in addattrs:`.

micamu: I did see the overindent complaint, but isn't this just a matter of style? I like the indentation the way I did it, rather than the way CodeClimate wants me to do it!  I will go for the option of creating a variable above.

bin/inference/pycbc_inference_plot_posterior
micamu: that won't work, `read_samples`  doesn't know about `snr_from_loglikelihood`. simplifying here requires other changes in other modules, this is the quick easy way to fix the bug

pycbc/inference/io/base_hdf.py
micamu: It still complains about the indentation, is that an issue that can be ignored?

bin/inference/pycbc_inference_plot_posterior
micamu: actually, `read_samples` does know about `snr_from_loglr`, but `read_raw_samples` does not know about the attributes, so it complains when requesting for `lognl`. There are `read_raw_samples` scattered all over the place: in the posterior module
https://github.com/gwastro/pycbc/blob/master/pycbc/inference/io/posterior.py#L35
in the base_mcmc module
https://github.com/gwastro/pycbc/blob/master/pycbc/inference/io/base_mcmc.py#L356
in the base_multitemper module
https://github.com/gwastro/pycbc/blob/master/pycbc/inference/io/base_multitemper.py#L162
...
And probably every time a new sampler is added, then a new `read_raw_samples` function will appear. I find it much cleaner to just do it this way, rather than having all the `read_raw_samples` functions return attributes (especially since to add the attribute to the samples recordarray, one would have to add the same constant value to each item of the array)

cdcapano: Ah, I forgot that read raw samples will interpret that as a parameter it needs to get. Ok, this is fine then.

pycbc/inference/io/base_hdf.py
cdcapano: It's because when doing new line continuation, you should just use a single tab indent from the previous line:
```
addattrs = self.static_params.items() + \
    self[self.samples_group].attrs.items()
```

Alternatively,  if you want the second line to line up, stick in a parenthesis. Then you can also get rid of the line continuation character:
```
addattrs = (self.static_params.items() +
            self[self.samples_group].attrs.items())
```

This isn't just a style set by Code Climate, it's the Python standard style guide ([PEP-8](https://www.python.org/dev/peps/pep-0008/)). I'd like to stick to that (and update old code as we move along), so please change it to one of those two.

micamu: Ok, I chose the parenthesis option and CodeClimate seems to like it now

micamu:
@cdcapano Are you happy with the indentation now? If so, can we merge this?

cdcapano:
@micamu Nevermind, I was able to resolve the merge conflict through github.

PR #2517, opened 2019-02-20
Fix storage format of coinc segments in coinc code

tapaimarton:
Changes segment handling to only store the segments once, keyed by the detector combination.

Since the StatmapData and MultiifoStatmap data has self.seg = f['segments'] I left it like this, but kept the order change in pycbc_multiifo_combine_statmap.

### DIFF
--- a/bin/hdfcoinc/pycbc_coinc_hdfinjfind
+++ b/bin/hdfcoinc/pycbc_coinc_hdfinjfind
@@ -110,12 +110,12 @@ for trigger_file, injection_file in zip(args.trigger_files,
         # coincident event, thus combine segments over all combinations
         # of coincident detectors to determine which times to keep
         any_seg = segments.segmentlist([])
-        for key in f:
+        for key in f['segments']:
             if key == 'foreground':
                 continue
             else:
-                starts = f[key + '/segments/start'][:]
-                ends = f[key + '/segments/end'][:]
+                starts = f['/segments/%s/start' % key][:]
+                ends = f['/segments/%s/end' % key][:]
                 any_seg += events.start_end_to_segments(starts,ends)
         ana_start, ana_end = events.segments_to_start_end(any_seg)
     else:
--- a/bin/hdfcoinc/pycbc_coinc_hdfinjfind
+++ b/bin/hdfcoinc/pycbc_coinc_hdfinjfind
@@ -211,9 +211,9 @@ for trigger_file, injection_file in zip(args.trigger_files,
                 continue
             if key not in fo:
                  fo.create_group(key)
-            fo[key].attrs['pivot'] = f['segments/%s' % key].attrs['pivot']
-            fo[key].attrs['fixed'] = f['segments/%s' % key].attrs['fixed']
-            fo[key].attrs['foreground_time_exc'] = f['segments/%s' % key].attrs['foreground_time_exc']
+            fo[key].attrs['pivot'] = f[key].attrs['pivot']
+            fo[key].attrs['fixed'] = f[key].attrs['fixed']
+            fo[key].attrs['foreground_time_exc'] = f[key].attrs['foreground_time_exc']
     else:
         fo.attrs['detector_1'] = f.attrs['detector_1']
         fo.attrs['detector_2'] = f.attrs['detector_2']
--- a/bin/hdfcoinc/pycbc_coinc_hdfinjfind
+++ b/bin/hdfcoinc/pycbc_coinc_hdfinjfind
@@ -206,14 +206,14 @@ for trigger_file, injection_file in zip(args.trigger_files,
                    sim_table.get_column(args.redshift_column))
 
     if multi_ifo_style:
-        for key in f.keys():
+        for key in f['segments'].keys():
             if key == 'foreground':
                 continue
             if key not in fo:
                  fo.create_group(key)
-            fo[key].attrs['pivot'] = f[key].attrs['pivot']
-            fo[key].attrs['fixed'] = f[key].attrs['fixed']
-            fo[key].attrs['foreground_time_exc'] = f[key].attrs['foreground_time_exc']
+            fo[key].attrs['pivot'] = f['segments/%s' % key].attrs['pivot']
+            fo[key].attrs['fixed'] = f['segments/%s' % key].attrs['fixed']
+            fo[key].attrs['foreground_time_exc'] = f['segments/%s' % key].attrs['foreground_time_exc']
     else:
         fo.attrs['detector_1'] = f.attrs['detector_1']
         fo.attrs['detector_2'] = f.attrs['detector_2']
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
@@ -330,8 +330,6 @@ if len(data['stat']) > 0:
                          compression_opts=9,
                          shuffle=True)
 
-f['segments/coinc/start'], f['segments/coinc/end'] = veto.segments_to_start_end(coinc_segs)
-
 f['segments/%s/start' % ''.join(sorted(detectors))], f['segments/%s/end' % ''.join(sorted(detectors))] = trigs[args.pivot_ifo].valid
 
 f.attrs['timeslide_interval'] = args.timeslide_interval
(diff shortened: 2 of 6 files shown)


bin/hdfcoinc/pycbc_multiifo_combine_statmap
```diff
@@ -26,25 +26,21 @@ f = h5py.File(args.output_file, "w")
...
-    ifosstring = ''.join(ifos_in_file)
-    group_to_save = '{}/segments'.format(ifosstring)
-    f[group_to_save + '/end'] = fi[ifo_to_get + '/end'][:]
-    f[group_to_save + '/start'] = fi[ifo_to_get + '/start'][:]
-    # Use this ID string to save the attributes from the input files
+    if len(fi['segments'].keys()) > 1:
+        logging.info('Incorrect segment format.')
+
+    key = fi['segments'].keys()[0]
+    f['%s/segments/end' % key] = fi['segments/%s/end' % key][:]
```
GarethDaviesGW: This is a point where consistency would be better - we change from segments/key to key/segments

This change was made because of wanting the attributes associated with the groups for each detector combination. These attributes are not necessarily associated with the segments and as a result we went for the combination/segments ordering

I think that having it as combinations/segments instead throughout would make more sense, (but I wouldn't be too upset if others disagree)

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
```diff
@@ -330,10 +330,8 @@ if len(data['stat']) > 0:
                          compression_opts=9,
                          shuffle=True)
 
-f['segments/coinc/start'], f['segments/coinc/end'] = veto.segments_to_start_end(coinc_segs)
-
-for t in trigs:
-    f['segments/%s/start' % trigs[t].ifo], f['segments/%s/end' % trigs[t].ifo] = trigs[t].valid
+# Store coinc segments keyed by detector combination
+f['segments/%s/start' % ''.join(sorted(detectors))], f['segments/%s/end' % ''.join(sorted(detectors))] = trigs[args.pivot_ifo].valid
```
GarethDaviesGW: Does detectors need to be sorted? or should it come from the order on the command line, (the order of which comes from the precedence list I think)

GarethDaviesGW: Having spoken with Tom, this should be kept as sorted alphabetically, in case the pivot/fixed and precedence scheme changes

bin/hdfcoinc/pycbc_multiifo_combine_statmap
GarethDaviesGW: Please change to use combinations/segments

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
tdent: btw this line is too long, please split (PEP8!)

tapaimarton: Done

bin/hdfcoinc/pycbc_multiifo_combine_statmap
tapaimarton: Changed to segments/combinations everywhere

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -206,14 +206,14 @@ for trigger_file, injection_file in zip(args.trigger_files,
                    sim_table.get_column(args.redshift_column))
 
     if multi_ifo_style:
-        for key in f.keys():
+        for key in f['segments'].keys():
             if key == 'foreground':
                 continue
```
GarethDaviesGW: Is this bit needed now?

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -206,14 +206,14 @@ for trigger_file, injection_file in zip(args.trigger_files,
...
             if key not in fo:
                  fo.create_group(key)
-            fo[key].attrs['pivot'] = f[key].attrs['pivot']
-            fo[key].attrs['fixed'] = f[key].attrs['fixed']
-            fo[key].attrs['foreground_time_exc'] = f[key].attrs['foreground_time_exc']
+            fo[key].attrs['pivot'] = f['segments/%s' % key].attrs['pivot']
```
GarethDaviesGW: does this make a group called key which just has the attributes hung to it and nothing else? I think this is okay but wanted to be sure

bin/hdfcoinc/pycbc_multiifo_combine_statmap
```diff
@@ -26,25 +26,21 @@ f = h5py.File(args.output_file, "w")
...
-    ifosstring = ''.join(ifos_in_file)
-    group_to_save = '{}/segments'.format(ifosstring)
-    f[group_to_save + '/end'] = fi[ifo_to_get + '/end'][:]
-    f[group_to_save + '/start'] = fi[ifo_to_get + '/start'][:]
-    # Use this ID string to save the attributes from the input files
+    if len(fi['segments'].keys()) > 1:
+        logging.info('Incorrect segment format.')
+
+    key = fi['segments'].keys()[0]
+    f['segments/%s/end' % key] = fi['segments/%s/end' % key][:]
```
GarethDaviesGW: It would be good to have a check that none of the detector combinations have been used before (`if key in f['segments']`) and then give an error if they have

I should have included it before, but hadn't thought of it

bin/hdfcoinc/pycbc_multiifo_combine_statmap
```diff
@@ -26,25 +26,21 @@ f = h5py.File(args.output_file, "w")
...
+    f['segments/%s/start' % key] = fi['segments/%s/start' % key][:]
     for attr_name in fi.attrs:
-        f[ifosstring].attrs[attr_name] = fi.attrs[attr_name]
+        f['segments/%s' % key].attrs[attr_name] = fi.attrs[attr_name]
```
GarethDaviesGW: As we are creating a coinc group for attributes above, then we should do the same here have a `key` group in this file as well (i.e. `f[key].attrs[attr_name]`)

tapaimarton: We are not creating it, it already exists, the segments are saved in there for the specific combination, hence there is no point in creating a new group.

GarethDaviesGW: Sorry I wasn't clear, I was suggesting that we _do_ create a `f[key]` group, even though `f['segments/%s' % key]` exists, for the attributes which are not related to the segments

bin/hdfcoinc/pycbc_coinc_hdfinjfind
tapaimarton: yes we do, see line 161-165 of pycbc_multiifo_coinc_statmap

bin/hdfcoinc/pycbc_coinc_hdfinjfind
tapaimarton: If the key does not exists, it is created.

bin/hdfcoinc/pycbc_multiifo_combine_statmap
tapaimarton: That can only occur if the same file is in the input twice, that should never happen, at least within a workflow. The input is 1 file per detector combination. I do not think this is needed.

bin/hdfcoinc/pycbc_coinc_hdfinjfind
tapaimarton: yes to me it seems only the attributes are there specific to the detector combination

bin/hdfcoinc/pycbc_multiifo_combine_statmap
tapaimarton: ah there is an error in here, as in there can be a segments/foreground from coinc_statmap. This needs to change for the same check as it is later in hdfinjfind. Though the key is the detector combination, so I am not sure what attribute you want to save not specific to detector combination, key-ed by detector combination

tapaimarton: It should be correct now.

GarethDaviesGW: so the pivot and fixed ifos are not segment related, but are combination related, so my suggestion is to have a different group, just called `f[key]` and not have the 'segments/'

tapaimarton: changed as requested

PR #2509, opened 2019-02-19
fit_sngls_by_template analysis time output

GarethDaviesGW:
Adding `analysis_time` output to `fit_sngls_by_template`

This will be needed for calculation of the new multiiforanking statistic (noise rate for each template = number of triggers/analysis time)

The analysis time is calculated through the trigger file segments and vetos as used in the fitting already

PR #2519, opened 2019-02-21
multi-ifo changes to pycbc_page_segtable

GarethDaviesGW:
Making it so that H1 & L1 are not longer hardcoded, and that the table output now has all possible detector combinations

### DIFF
--- a/bin/hdfcoinc/pycbc_page_segtable
+++ b/bin/hdfcoinc/pycbc_page_segtable
@@ -54,7 +54,7 @@ parser.add_argument('--title-text', type=str, required=False,
 parser.add_argument('--output-file', type=str,
                         help='Path of the output HTML file.')
 parser.add_argument('--ifos', nargs='+',default=['H1','L1'],
-                        help='(Optional) List of IFOs, default is H1, L1.')
+                        help='(Optional) List of IFOs: default is H1, L1.')
 opts = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


tdent:
Can you post links to some tests Gareth?

GarethDaviesGW:
Segtable with 3 IFOs and multiple rows to check:

https://ldas-jobs.ligo.caltech.edu/~gdavies/testoutput/test_segtable_output.html

I'm not sure how this will look on the actual results page though, as it is now a much wider table

tdent:
@spxiwh OK on the output from my point of view, however the functions to iterate over ifo combinations (`ifo_combinations` and `powerset_ifos`) seem like they might be more generally useful or maybe already exist somewhere - we're guessing you may know more about this

tdent:
@spxiwh can you look at this/review briefly? the one issue I had is whether the powerset_ifos function either exists elsewhere already or belongs in a library module, eg in the workflow module ?

GarethDaviesGW:
looking at https://ldas-jobs.ligo.caltech.edu/~gdavies/testoutput/H1L1-PAGE_SEGTABLE_VETO_SUMMARY-1187312718-428100.html in #2509 has made me want to have a look at this again at some point, need to understand if this is doing this right

I'm not sure if V1 segments are in a dfferent format and therefore need special treatment

PR #2527, opened 2019-02-27
Priors on 1D marginal posterior plots

cdcapano:
This adds the ability to plot the prior on the 1D marginal plots in `plot_posterior`. The prior is plotted as a dotted line (in black for single files; for multiple file inputs, the next color in the sequence is used). Examples can be seen in Figure 4 of https://arxiv.org/abs/1902.09496.

The prior is loaded from the config file. I moved the lines of code that do this in `plot_prior` to a function in `option_utils`, then have both `plot_posterior` and `plot_prior` call the same function.

This also fixes a couple of bugs I noticed:
 1. The color of the confidence interval lines in the 1D marginal plots was being set to the contour colors. This meant that if you set the contour color to white (as you need to do when plotting density), the 1D marginal percentile lines wouldn't be visible.
 2. The quoted confidence intervals in the plot titles were being calculated from the location of the 1D marginal lines, not from the samples. That's ok as long as the requested percentiles to plot were symmetric (e.g., 5, 50, 95), but will give wrong results if not (e.g, 50). Even when the percentiles are symmetric, it could result in the precision in the title being worse than it actually is. I fixed this by getting getting the percentile from the samples instead. This means that they are hard coded to always quote the middle 90% confidence interval.

### DIFF
--- a/bin/inference/pycbc_inference_plot_posterior
+++ b/bin/inference/pycbc_inference_plot_posterior
@@ -120,8 +120,9 @@ if opts.plot_prior is not None:
                          "--plot-prior")
     logging.info("Loading prior")
     cp = WorkflowConfigParser(opts.plot_prior)
-    prior_samples = option_utils.prior_samples_from_config(
-        cp, nsamples=opts.prior_nsamples)
+    prior = option_utils.prior_samples_from_config(cp)
+    logging.info("Drawing samples from prior")
+    prior_samples = prior.rvs(opts.prior_nsamples)
     _, ts = transforms.get_common_cbc_transforms(
         opts.parameters, prior_samples.fieldnames)
     # add parameters not included in file
--- a/bin/inference/pycbc_inference_plot_posterior
+++ b/bin/inference/pycbc_inference_plot_posterior
@@ -120,7 +120,7 @@ if opts.plot_prior is not None:
                          "--plot-prior")
     logging.info("Loading prior")
     cp = WorkflowConfigParser(opts.plot_prior)
-    prior = option_utils.prior_samples_from_config(cp)
+    prior = option_utils.prior_from_config(cp)
     logging.info("Drawing samples from prior")
     prior_samples = prior.rvs(opts.prior_nsamples)
     _, ts = transforms.get_common_cbc_transforms(
--- a/bin/inference/pycbc_inference_plot_posterior
+++ b/bin/inference/pycbc_inference_plot_posterior
@@ -41,6 +41,7 @@ from pycbc.results import metadata
 from pycbc import __version__
 from pycbc.workflow import WorkflowConfigParser
 from pycbc.inference import (option_utils, io)
+from pycbc import transforms
 
 from pycbc.results.scatter_histograms import create_multidim_plot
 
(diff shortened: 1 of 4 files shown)


bin/inference/pycbc_inference_plot_posterior
```diff
@@ -54,6 +56,12 @@ parser.add_argument("--output-file", type=str, required=True,
                     help="Output plot path.")
 parser.add_argument("--verbose", action="store_true", default=False,
                     help="Be verbose")
+parser.add_argument("--plot-prior", nargs="+", type=str,
```
soumide1102: The name isn't making it clear that it's for inputting a config file. Something like `--prior-config-file` could make it clear.

bin/inference/pycbc_inference_plot_prior
```diff
@@ -66,12 +67,12 @@ parser.add_argument("--parameters", nargs="+", action=ParseParametersArg,
                          "provided, will plot all of the parameters in the "
                          "[variable_params] section of the config file.")
 parser.add_argument("--sections", type=str, nargs="+", default=["prior"],
-                    help="Name of section plus subsection with distribution "
-                         "configurations, eg. prior-mass1.")
+                    help="Name of section plus with distribution "
```
soumide1102: Looks like a typo here?

pycbc/results/scatter_histograms.py
```diff
@@ -359,8 +364,8 @@ def create_marginalized_hist(ax, values, label, percentiles=None,
             ax.axvline(expected_value, color=expected_color, lw=1.5, zorder=2)
     if title:
         values_med = numpy.median(values)
-        values_min = values.min()
-        values_max = values.max()
+        values_min = numpy.percentile(values, 5.)
+        values_max = numpy.percentile(values, 95.)
```
soumide1102: So this is hard-coded to quote the middle 90% confidence interval always? If someone wants to read off the 68% confidence interval (a common % quoted by astronomers) that can't be done, right? Also, the percentiles plotted will not always be related to the numbers quoted in the title; if someone plots the 16 and 84 percentile, they will read the 90% credible interval not the 68% CI. That can be confusing.

bin/inference/pycbc_inference_plot_prior
cdcapano: Oops, fixed!

bin/inference/pycbc_inference_plot_posterior
cdcapano: It says in the help message that you need to provide one or more config file(s) for this option.

pycbc/results/scatter_histograms.py
cdcapano: Ok, I've now changed it so that it quotes the percentile halfway between the max/min of the `percentiles` +/- max/min of the `percentiles`. It will default to 50 +/- 95/5 if no percentiles were given. I chose average of the percentiles for the median value because just quoting the median will break if you gave percentiles that did not encompass the median.

Aside: you say in the astronomy they like to quote the 68% confidence interval. But do they quote the median +/- 84/16th percentile or the mean +/- 1 sigma? If the latter, you still can't use the quoted values in the title, since the mean is not a percentile.

soumide1102: You're right, they quote 68% credible interval as the mean +/- 1.0 sigma. Thanks for correcting that. & the new changes relating the percentile lines plotted with the quoted values look good to me.

micamu:
@cdcapano you didn't add my requested changes before merging... they are not crucial for the performance of the executable, but they could add clarity (especially if someone fails to give a config-file and there is an error they can easily understand, rather than a cryptic python error that you have to track down)

PR #2531, opened 2019-02-27
Allow psd group to be specified in plot_psd

cdcapano:
This adds an `hdf-group` option to `plot_psd` to specify the group to look in for the PSD. This allows `plot_psd` to be run on inference hdf files, in which the psd is stored in the `data` group. If the `hdf-group` option isn't given, it will default to the top level group, as it does now.

This also makes it optional to plot a reference PSD. Right now, if you don't specify one of `psd-file`, `asd-file` or `psd-model`, you'll get an error.

Finally, I added a `low-frequency-cutoff` option so that this can be specified in case it isn't stored in the hdf file, as is (currently) the case for inference hdf files.


titodalcanton:
Waiting for #2532 before merging.

PR #2533, opened 2019-02-27
Add (optional) Kerr factor to ringdown mass/spin approximant

micamu:
Currently the mass/spin ringdown approximant is a hybrid between a Kerr ringdown and an agnostic freq/tau damped sinusoid. It uses the NR formulae to map (mass,spin) -> (freq,tau), but it does not contain the M/d factor that appears in perturbation theory for a Kerr black hole. 
We did not introduce this before because it is just a constant factor, but it might easier to just include this factor when we want to understand how the SNR of the ringdown relates to the mass of the black hole and its distance. I have made the distance an optional argument, so if one does not give a distance the returned ringdown will be the same as before. I also noted in the documentation that when including this factor, the A_220 amplitude has a completely different order of magnitude (with a reference to a paper that does give example values for A_220).  

The rest of the changes in the file are just clean up, mainly `if taper is None:` -> `if not taper`  and `if taper is not None` -> `if taper`

P.S: I have not yet tested how the PE will behave when trying to use this factor, but it seems to have worked fine for other groups in the literature.

### DIFF
--- a/bin/pycbc_ringinj
+++ b/bin/pycbc_ringinj
@@ -66,6 +66,9 @@ parser.add_argument('--final-mass', type=float,
 parser.add_argument('--final-spin', type=float,
                     help='Spin of the final black hole for '
                          'MassSpin approximants.')
+parser.add_argument('--distance', type=float,
+                    help='Luminosity distance of the final black hole in Mpc '
+                         'for MassSpin approximants (optional).')
 parser.add_argument('--freqs-taus', nargs='+',
                     help='Central frequency (Hz) and damping time (s) of the '
                          'ringdown for each mode (FreqTau approximants).'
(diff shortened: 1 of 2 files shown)


micamu:
@JulianWesterweck I seem not to be able to add you as a reviewer, maybe you want to have a look at these change and talk to @cdcapano about approving the review or not

cdcapano:
@JulianWesterweck I've added you as a maintainer; once you accept the invitation, you can be added as a reviewer.

micamu:
Thanks, I'll merge when codebeat finishes if everything is in order

micamu:
I did do a test run on zero noise using the distance as a variable parameter. It doesn't measure the distance very well, but the measurement of ringdown parameters does not vary much when including this factor:
https://www.atlas.aei.uni-hannover.de/~miriam.cabero/ringdown/test_Kerr_factor.png
For comparison, the same run without using the distance (not exactly the same SNR but close):
https://www.atlas.aei.uni-hannover.de/~miriam.cabero/ringdown/methods_paper/ringdown_injection/zero_noise/330_0p3/mass_spin_snr_20.png

micamu:
@ahnitz @cdcapano codebeat has been running for days, could it be that it got stuck? the 2 issues it claims have been introduced are just function too long and cyclomatic complexity too high, but this would not be introduced by this PR (it was already like that before).
Can I merge this?

PR #2534, opened 2019-02-28
Use automatically determined number of sig. figs in posterior titles

cdcapano:
Currently, the titles in `plot_posterior` force the quoted values to two decimal places in the uncertainty. This causes the precision of some poorly constrained parameters to be over quoted, while other well constrained parameters to be under quoted. For example, here is a plot of the [intrinsic](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/quoted_intervals/orig_intrinsic_posterior.png) and [extrinsic](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/quoted_intervals/orig_extrinsic_posterior.png) parameters of a BNS run. The chirp mass is quoted as '1.35 + 0.00 - 0.00' because the uncertainty is < 0.01, while effective spin is quoted as '-0.0874 + 0.1371 - 0.1251', which is ridiculous since the error is at the 100% level.

The `format_value` function can automatically determine the number of sig. figs to quote based on the uncertainty, it's just been turned off until now. This patch turns that on. I've also adjusted the number of significant figures to quote. Currently, `format_value` rounds the estimated value such that only 1 sig. fig is quoted in the uncertainty. This is a bit too imprecise (e.g., doing so leads the effective spin above to be quoted as -0.1 + 0.1 - 0.1). So I've changed it so that it now rounds such that two significant figures are quoted in the uncertainty. Here are the same plots as above with this patch: [intrinsic](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/quoted_intervals/intrinsic_posterior.png), [extrinsic](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/quoted_intervals/extrinsic_posterior.png). Note that now, chirp mass is quoted as '1.3492 + 0.0029 - 0.0040' while effective spin is quoted as '-0.09 + 0.14 - 0.13'.

### DIFF
--- a/pycbc/results/scatter_histograms.py
+++ b/pycbc/results/scatter_histograms.py
@@ -364,7 +364,7 @@ def create_marginalized_hist(ax, values, label, percentiles=None,
         negerror = values_med - values_min
         poserror = values_max - values_med
         fmt = '${0}$'.format(str_utils.format_value(
-            values_med, negerror, plus_error=poserror, ndecs=2))
+            values_med, negerror, plus_error=poserror))
 
         if rotated:
             ax.yaxis.set_label_position("right")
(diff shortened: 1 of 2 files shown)


micamu:
Ok, I am going to approve the review, but it looks like Travis doesn't like it... You might need to fix something.

PR #2538, opened 2019-03-01
Improve PyCBC Live logging, help and metadata

titodalcanton:
* Save PyCBC version and command line into every output HDF file.
* Remove or clarify some log messages.
* Remove some unused code and options.
* Make `--enable-profiling` option more flexible.
* Add/fix CLI help.
* Record the number of live detectors into every output HDF file.
* ~~Record the command line arguments into the event XML files.~~ Multidetector arguments break LIGOLW :(

Should be merged after #2523 and after a quick test.


titodalcanton:
Tested. I would actually like this merged next as the additional saved metadata will be useful for testing/review.

PR #2540, opened 2019-03-04
Change to p-value combination method for Live

tdent:
Change to a method which 1) uses a nominal livetime specified on the command line (eg 0.1yr) for p-value combination 2) does max ifar between the original and combined p-value and applies a trials factor of 2.  See https://github.com/gwastro/pycbc/wiki/pvalue-combining-tests for more detail

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -28,23 +28,27 @@ import pycbc.waveform.bank
 
 
 def ptof(p, ft):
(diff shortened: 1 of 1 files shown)


titodalcanton:
Looks like this may conflict with #2538. Once that is merged please rebase.

tdent:
Rebased, all checks passing

PR #2548, opened 2019-03-05
Add back ability to provide beta values to emcee_pt from external file

soumide1102:
Adding feature to read an hdf file storing beta values for the emcee_pt sampler from the config file.

### DIFF
--- a/pycbc/inference/sampler/emcee_pt.py
+++ b/pycbc/inference/sampler/emcee_pt.py
@@ -119,8 +119,8 @@ class EmceePTSampler(MultiTemperedAutocorrSupport, MultiTemperedSupport,
         Loads the sampler from the given config file.
 
         For generating the temperature ladder to be used by emcee_pt, either
-        the number of temperatures (provided under a subsection 'ntemps'),
-        or the path to a file storing inverse temperature values (provided 
+        the number of temperatures (provided by the option 'ntemps'),
+        or the path to a file storing inverse temperature values (provided
         under a subsection inverse-temperatures-file) can be loaded from the
         config file. If the latter, the file should be of hdf format, having
         an attribute named 'betas' storing the list of inverse temperature
@@ -139,9 +139,8 @@ class EmceePTSampler(MultiTemperedAutocorrSupport, MultiTemperedSupport,
                              "inverse-temperatures-file, not both.")
         if cp.has_option(section, "inverse-temperatures-file"):
             # get the path of the file containing inverse temperatures values.
-            inverse_temperatures_file = cp.get(
-                                         section,
-                                         "inverse-temperatures-file")
+            inverse_temperatures_file = cp.get(section,
+                                               "inverse-temperatures-file")
             with h5py.File(inverse_temperatures_file, "r") as fp:
                 try:
                     betas = numpy.array(fp.attrs['betas'])
--- a/pycbc/inference/sampler/emcee_pt.py
+++ b/pycbc/inference/sampler/emcee_pt.py
@@ -58,8 +58,9 @@ class EmceePTSampler(MultiTemperedAutocorrSupport, MultiTemperedSupport,
     _io = EmceePTFile
     burn_in_class = MultiTemperedMCMCBurnInTests
 
-    def __init__(self, model, ntemps, nwalkers, checkpoint_interval=None,
-                 loglikelihood_function=None, nprocesses=1, use_mpi=False):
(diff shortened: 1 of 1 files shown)


pycbc/inference/sampler/emcee_pt.py
```diff
@@ -115,12 +117,29 @@ def from_config(cls, cp, model, nprocesses=1, use_mpi=False):
...
-        ntemps = int(cp.get(section, "ntemps"))
+        import h5py
+        if cp.has_option(section, "ntemps") and \
+                cp.has_option(section, "inverse_temperatures_input_file"):
+            raise ValueError("Must specify either ntemps or "
+                             "inverse-temperatures-input-file, not both.")
+        if cp.has_option(section, "inverse_temperatures_input_file"):
+            inverse_temperatures_input_file = cp.get(
+                                         section,
+                                         "inverse_temperatures_input_file")
```
cdcapano: There's some inconsistency here; should it be `inverse_temperature_input_file` or `inverse-temperature-input-file`? Personally, I'd prefer the later. And why not just call it `betas-file`? That's easier to type.

pycbc/inference/sampler/emcee_pt.py
```diff
@@ -115,12 +117,29 @@ def from_config(cls, cp, model, nprocesses=1, use_mpi=False):
         # get the number of walkers to use
         nwalkers = int(cp.get(section, "nwalkers"))
         # get the number of temps
-        ntemps = int(cp.get(section, "ntemps"))
+        import h5py
```
cdcapano: Put the import at the top of the file.

pycbc/inference/sampler/emcee_pt.py
soumide1102: I had `betas-file` initially when I made this patch last time. But Duncan suggested this more verbose name as betas can be used for something other than inverse temperatures too.

soumide1102: I've now made it `inverse-temperatures-input-file`.

soumide1102:
@cdcapano I've made the changes you suggested.

pycbc/inference/sampler/emcee_pt.py
```diff
@@ -114,13 +122,34 @@ def from_config(cls, cp, model, nprocesses=1, use_mpi=False):
...
-        ntemps = int(cp.get(section, "ntemps"))
+        if cp.has_option(section, "ntemps") and \
+                cp.has_option(section, "inverse-temperatures-input-file"):
+            raise ValueError("Must specify either ntemps or "
+                             "inverse-temperatures-input-file, not both.")
+        if cp.has_option(section, "inverse-temperatures-input-file"):
+            # get the path of the file containing inverse temperatures values
+            # to be provided to emcee_pt for the temperature ladder. This
+            # should be an hdf file having an attribute named 'betas' storing
+            # the list of inverse temperature values.
```
cdcapano: Thanks for adding this comment, but I you should add it to the doc string for the function, so it shows up in the online documentation.

pycbc/inference/sampler/emcee_pt.py
cdcapano: Oh ok... but how about dropping the 'input' (`inverse-temperatures-file`)?

soumide1102:
@cdcapano Changed to `inverse-temperatures-file` and updated the docstring. Let me know if that looks okay.
On another note, I saw that the docstring for emcee_pt and emcee sampler classes need to be updated. For example emcee_pt has documentation for `pool` which is no more an input parameter in `__init__`. `checkpoint_interval`, loglikelihood_function`, nprocesses`, and `use_mpi` are not mentioned in the list of parameters.

pycbc/inference/sampler/emcee_pt.py
```diff
@@ -107,20 +115,48 @@ def betas(self):
...
     @classmethod
     def from_config(cls, cp, model, nprocesses=1, use_mpi=False):
-        """Loads the sampler from the given config file."""
+        """
+        Loads the sampler from the given config file.
+
+        For generating the temperature ladder to be used by emcee_pt, either
+        the number of temperatures (provided under a subsection 'ntemps'),
+        or the path to a file storing inverse temperature values (provided 
+        under a subsection inverse-temperatures-file) can be loaded from the
```
cdcapano: Change "provided under a subsection" to "provided by the option" here and at line 122.

Issue #2434, opened 2018-12-07
PyCBC catalog missing events


geoffrey4444:
~~~~
from pycbc import catalog
for name in catalog.Catalog():
  print name
~~~~

still outputs
~~~~
LVT151012
GW170608
GW150914
GW170814
GW170817
GW170104
GW151226
~~~~

Is it straightforward to update the catalog to include all of the GWs from GWTC-1? More generally, how will the catalog get updated in the future as LIGO/Virgo release more detections?

ahnitz:
It's pretty straightforward. The thing I've been thinking about though is that since there are also multiple catalogs now, it should be possible to switch sources. In the meantime though, we can add in the additional events.

geoffrey4444:
That would be great, adding in the additional events! Switching catalogs would also be nice. Thanks for the reply @ahnitz! I'm happy to test as soon as it's ready. :)

ahnitz:
@geoffrey4444 Sorry for the very long delay here. I finally got around to addressing this. Hopefully, the pull request #2552 should address the issue.

PR #2552, opened 2019-03-05
Use GWOSC to directly provide the catalog information in pycbc

ahnitz:


ahnitz:
Note that due to differences in the GWOSC format, some changes to the tutorials will need to coincide to the next release.

ahnitz:
fixes #2434

ahnitz:
@cdcapano Poke

PR #2558, opened 2019-03-07
pycbc/results fixes for Python 3

titodalcanton:
This change allowed me to use `pycbc.results` and `ligo.skymap` in the same env.


ahnitz:
:+1: for helping with python 3 support.

PR #2561, opened 2019-03-08
Fix various errors in PyCBC Live

titodalcanton:
Investigating and fixing errors with SNR time series etc.


titodalcanton:
I think we should merge this since it fixes two very annoying errors. I will comment the assert's out before.

PR #2562, opened 2019-03-08
Load constraints when reading prior from config for plotting

soumide1102:
This PR adds the ability to load constraints from the config file in the `prior_from_config` function called for plotting priors.

@cdcapano Can we not just use 
```
variable_params, static_params = distributions.read_params_from_config(...)
dists = distributions.read_distributions_from_config(...)
```
instead of this block of code https://github.com/gwastro/pycbc/blob/master/pycbc/inference/option_utils.py#L488-L504 ?


soumide1102:
@cdcapano I removed that block of code. Do the new changes look fine to you?

cdcapano:
@soumide1102 Thanks, looks good!

PR #2568, opened 2019-03-12
Foreground censor multiifodev

tapaimarton:
updateing foreground censor to workflow for multiifo use and adding it to the workflow

### DIFF
--- a/bin/hdfcoinc/pycbc_foreground_censor
+++ b/bin/hdfcoinc/pycbc_foreground_censor
@@ -8,17 +8,17 @@ from pycbc.workflow import SegFile
 parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument('--version', action='version', version=pycbc.version.git_verbose_msg)
 parser.add_argument('--verbose', action='store_true')
-parser.add_argument('--foreground-triggers', 
+parser.add_argument('--foreground-triggers',
                     help="HDF file containing the zerolag foreground triggers "
                          "from the analysis")
 parser.add_argument('--veto-file',
                     help="Baseline veto information that is added to the outptut")
-parser.add_argument('--segment-name', 
+parser.add_argument('--segment-name',
                     help="Segment name to use from the input veto file")
 parser.add_argument('--output-file', help='Name of the output segment file')
 parser.add_argument('--strict-coinc-time', action='store_true',
                     help="Veto any time that is vetoed in either detector")
-parser.add_argument('--output-segment-name',    
+parser.add_argument('--output-segment-name',
                     help="(optional), Name of output segment file list",
                     default="censor_foreground")
 args = parser.parse_args()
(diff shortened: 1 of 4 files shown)


bin/hdfcoinc/pycbc_foreground_censor
```diff
@@ -28,26 +28,61 @@ pycbc.init_logging(args.verbose)
...
-
-if args.strict_coinc_time:
-    fsegs[0] = (fsegs[0].coalesce() + fsegs[1].coalesce()).coalesce()
-    fsegs[1] = fsegs[0]
-
-# FIXME: Don't use URL once Larne supports taking file paths directly in the
-#        workflow File class
-file_url = urlparse.urlunparse(['file', 'localhost',
+
+if 'segments/foreground_veto/start' in f:
```
tdent: can you add a comment saying what case this is (ie why we need to test for this)

bin/hdfcoinc/pycbc_foreground_censor
```diff
@@ -28,26 +28,61 @@ pycbc.init_logging(args.verbose)
...
+        fsegs[1] = fsegs[0]
+
+    # FIXME: Don't use URL once Larne supports taking file paths directly in the
+    #        workflow File class
+    file_url = urlparse.urlunparse(['file', 'localhost',
                           os.path.abspath(args.output_file), None, None, None])
-SegFile.from_multi_segment_list('UNUSED', fsegs, names, ifos, file_url=file_url)
-logging.info('done')
+    SegFile.from_multi_segment_list('UNUSED', fsegs, names, ifos, file_url=file_url)
+else:
```
tdent: similarly, add comment explaining the other case of the if/else switch.

bin/hdfcoinc/pycbc_foreground_censor
```diff
@@ -28,26 +28,61 @@ pycbc.init_logging(args.verbose)
...
+        names += [args.output_segment_name]
+        ifos += [ifo]
+
+    if args.strict_coinc_time:
+        fsegs[0] = (fsegs[0].coalesce() + fsegs[1].coalesce()).coalesce()
+        fsegs[1] = fsegs[0]
+
+    # FIXME: Don't use URL once Larne supports taking file paths directly in the
+    #        workflow File class
+    file_url = urlparse.urlunparse(['file', 'localhost',
```
tdent: do these lines need to be inside the if/else switch?  they look identical in both cases.

bin/hdfcoinc/pycbc_foreground_censor
```diff
@@ -28,26 +28,61 @@ pycbc.init_logging(args.verbose)
...
+        names += [args.output_segment_name]
+        ifos += [ifo]
+
+    if args.strict_coinc_time:
+        combfseg = fsegs[0]
+        for fseg in fsegs:
+            combfseg = (combfseg & fseg).coalesce()
+
+        for fseg in fsegs:
+            fseg = combfseg
```
GarethDaviesGW: This is very confusing to read and understand straight away - can you add some comments to explain what is happening (what the if case is here)

Also for legibility, rename `for fseg in fsegs` to `for fseg_ifo in fsegs` (or similar) to make it so that it is easily identifiable as fseg for each ifo

bin/hdfcoinc/pycbc_foreground_censor
```diff
@@ -28,26 +28,61 @@ pycbc.init_logging(args.verbose)
...
+
+    file_url = urlparse.urlunparse(['file', 'localhost',
+                          os.path.abspath(args.output_file), None, None, None])
+    SegFile.from_multi_segment_list('UNUSED', fsegs, names, ifos, file_url=file_url)
+
+logging.info('done')
```
GarethDaviesGW: add newline at end of file (PEP8)

bin/hdfcoinc/pycbc_foreground_censor
```diff
@@ -28,26 +28,61 @@ pycbc.init_logging(args.verbose)
...
+    if args.strict_coinc_time:
+        combfseg = fsegs[0]
+        for fseg in fsegs:
+            combfseg = (combfseg & fseg).coalesce()
+
+        for fseg in fsegs:
+            fseg = combfseg
+
+    file_url = urlparse.urlunparse(['file', 'localhost',
+                          os.path.abspath(args.output_file), None, None, None])
```
GarethDaviesGW: next line within brackets should start one space after under the bracket (PEP8)

bin/hdfcoinc/pycbc_multiifo_coinc_statmap
```diff
@@ -151,6 +151,7 @@ f = fw(args.output_file)
 f.attrs['num_of_ifos'] = all_trigs.attrs['num_of_ifos']
 f.attrs['pivot'] = all_trigs.attrs['pivot']
 f.attrs['fixed'] = all_trigs.attrs['fixed']
+f.attrs['ifos'] = all_trigs.attrs['ifos']
```
GarethDaviesGW: this needs to have the same `if 'ifos' in all_trigs.attrs` check as above

GarethDaviesGW: check that the `ifos` attribute is staying similar throughout (does it always have spaces or not?)

bin/hdfcoinc/pycbc_foreground_censor
tdent: @GarethDaviesGW not if the resulting line would be >79 characters. Marton has done it entirely correctly here, I will delete this comment

bin/hdfcoinc/pycbc_foreground_censor
tdent: this is done automatically by github anyway.

bin/hdfcoinc/pycbc_multiifo_coinc_statmap
tapaimarton: It is given with spaces in the coinc jobs. This takes a multiifo coinc job for the given detector combination, it will not run into error within a workflow.

bin/hdfcoinc/pycbc_foreground_censor
tapaimarton: Done

bin/hdfcoinc/pycbc_foreground_censor
tapaimarton: Done

bin/hdfcoinc/pycbc_foreground_censor
tapaimarton: Moved them outside of the if/else switch

bin/hdfcoinc/pycbc_foreground_censor
tapaimarton: The strict_coinc_time option is wrong here for the multiifo case, so I removed it alltogether.

bin/hdfcoinc/pycbc_multiifo_coinc_statmap
tapaimarton: Actually to be precise the 'ifos' attribute is always given by ' '.join(sorted(args.ifos)) to be exactly reproducible.

GarethDaviesGW: > this needs to have the same `if 'ifos' in all_trigs.attrs` check as above

Just a reminder for this - I think that as the two comments were for the same line, this got added to the same conversation as the other question on github

tapaimarton: I put in the check but it is unnecessary, this code should only get a coinc file from a pycbc_multiifo_coinc_findtrig job, which will always have an 'ifos' attribute. I think both checks should never return false.

tdent:
I will check formatting, comments etc. and merge later this evening.

PR #2572, opened 2019-03-13
add option to limit which coincs to follow-up with multiifos

ahnitz:


ahnitz:
This requires adding to the config file the following option. 
```
--ifar-double-followup-threshold XXXX
```
Units in years like the standard ifar-upload-threshold option. 
The option is *required*

ahnitz:
@titodalcanton Ok, I think this works in my test in the sense that I see coincs above this threshold being output to xml files and below I see it in the hdf files with a 'NO_FOLLOWUP' key.

tdent:
Does this need any further testing?

PR #2575, opened 2019-03-13
broadcast start time for pycbc live

ahnitz:


ahnitz:
@titodalcanton Untested, I'll report back when I have a chance to try out on ATLAS.

ahnitz:
Ok, I've tested on ATLAS that this seems to do as advertised. This should keep everyone in sync which may address #2547 and should also make it easier to bring up an analysis with rotating frame files.

PR #2580, opened 2019-03-17
PyCBC Live creates SNR time series with slightly different sample period

bhooshan-gadre:
Just corrected delta_t used in followup_snr_series using round(N*delta_f).
Add assert to check delta_ts for live detectors.


bhooshan-gadre:
To fix #2563

bhooshan-gadre:
Also, round is required at_time method to get the exact trigger time. Still, trigger and series snrs are slightly different but the peak times are now same

bin/pycbc_live
```diff
@@ -181,6 +181,9 @@ class LiveEventManager(object):
                                              data_readers, bank, followup_ifos)
 
             live_ifos = [ifo for ifo in ifos if 'snr_series' in fud[ifo]]
+            for ifo1, ifo2 in itertools.combinations(live_ifos, 2):
```
titodalcanton: I recommend using a one-liner with set comprehension here:
```python
assert len({fud[ifo]['snr_series'].delta_t for ifo in live_ifos}) == 1
```

pycbc/filter/matchedfilter.py
```diff
@@ -1247,7 +1247,7 @@ def matched_filter_core(template, data, psd=None, low_frequency_cutoff=None,
         h_norm = sigmasq(htilde, psd, low_frequency_cutoff, high_frequency_cutoff)
 
     norm = (4.0 * stilde.delta_f) / sqrt( h_norm)
-    delta_t = 1.0 / (N * stilde.delta_f)
+    delta_t = 1.0 / round(N * stilde.delta_f)
```
titodalcanton: Was this the source of the issue with the SNR series at the end? @ahnitz are you ok with this change in the matched filter core?

pycbc/types/timeseries.py
```diff
@@ -203,7 +203,7 @@ def at_time(self, time, nearest_sample=False):
         """
         if nearest_sample:
             time += self.delta_t / 2.0
-        return self[int((time-self.start_time)*self.sample_rate)]
+        return self[int(round((time-self.start_time)*self.sample_rate))]
```
titodalcanton: Is this issue related to the SNR series delta_t being different?

Anyway, I think you can achieve the same result by calling `at_time` with `nearest_sample=True`, without having to do this change.

ahnitz: @bhooshan-gadre This should not be added. You should given the 'nearest_sample' option if you want this behavior which is already implemented.

pycbc/filter/matchedfilter.py
ahnitz: @bhooshan-gadre Would using data.delta_t work in stead?

pycbc/types/timeseries.py
bhooshan-gadre: ah! I didn't use `nearest_sample` before assert. It is working fine.

pycbc/filter/matchedfilter.py
bhooshan-gadre: @ahnitz data.delta_t is working. I will change it so to
```suggestion
    delta_t = stilde.delta_t
```

bin/pycbc_live
bhooshan-gadre: Yes, this willl do.
```suggestion
           assert len({fud[ifo]['snr_series'].delta_t for ifo in live_ifos}) == 1
```

bhooshan-gadre:
Have put all the required things together in the recent commit.

PR #2583, opened 2019-03-19
add clustering option to pycbc inspiral to cluster triggers in chirp bins.

ahnitz:
This helps smooth out the triggers a bit as a function of chirp mass more than the raw bins in chirp mass made by splitting the template bank. 

Adds *optional* argument to pycbc inspiral

--keep-loudest-log-chirp-window which allows binning in the log chirp mass of the template being analyzed. Clustering is then done in each bin separately over time.

### DIFF
--- a/bin/pycbc_inspiral
+++ b/bin/pycbc_inspiral
@@ -161,6 +161,8 @@ parser.add_argument("--user-tag", type=str, metavar="TAG", help="""
                     This is used to identify FULL_DATA jobs for
                     compatibility with pipedown post-processing.
                     Option will be removed when no longer needed.""")
+parser.add_argument("--keep-loudest-log-chirp-window", type=float,
+                    help="Keep loudest triggers within log chirp mass window")
 parser.add_argument("--keep-loudest-interval", type=float,
                     help="Window in seconds to maximize triggers over bank")
 parser.add_argument("--keep-loudest-num", type=int,
@@ -214,10 +216,10 @@ with ctx:
 
     # Read specified user-provided wisdom files
     if opt.fftw_input_float_wisdom_file is not None:
-        import_single_wisdom_from_filename(opt.fftw_input_float_wisdom_file)        
+        import_single_wisdom_from_filename(opt.fftw_input_float_wisdom_file)
 
     if opt.fftw_input_double_wisdom_file is not None:
-        import_double_wisdom_from_filename(opt.fftw_input_double_wisdom_file)    
+        import_double_wisdom_from_filename(opt.fftw_input_double_wisdom_file)
 
     flow = opt.low_frequency_cutoff
     flen = strain_segments.freq_len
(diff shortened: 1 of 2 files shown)


tdent:
I'm not sure whether the new option is supposed to do more or less clustering .. from the description above, it seems like less ?

ahnitz:
@tdent I guess you could consider it less. It makes it more flexible with some options that are now directly tied to splitting the template bank for example.

pycbc/events/events.py
```diff
@@ -319,11 +319,29 @@ def keep_loudest_in_interval(self, window, num_keep):
...
 
+        if log_chirp_width:
+            from pycbc.conversions import mchirp_from_mass1_mass2
+            m1 = numpy.array([p['tmplt'].mass1 for p in self.template_params])
+            m2 = numpy.array([p['tmplt'].mass2 for p in self.template_params])
+
+            # chirp mass of each template
+            mc = mchirp_from_mass1_mass2(m1, m2)[e['template_id']]
+            lmc = numpy.log(mc)
+            imc = (lmc / log_chirp_width).astype(numpy.int32)
```
tdent: can you add a comment to this line to tell me what it does?  also perhaps can also combine with the previous line, ie `imc = (numpy.log(mc) / log_chirp_width).astype(numpy.int32)` as 'lmc' is not used in the rest of the code

pycbc/events/events.py
```diff
@@ -308,22 +308,42 @@ def keep_near_injection(self, window, injections):
...
+        # convert time to time bin index
         wtime = (time / window).astype(numpy.int32)
         bins = numpy.unique(wtime)
 
+        if log_chirp_width:
+            from pycbc.conversions import mchirp_from_mass1_mass2
+            m1 = numpy.array([p['tmplt'].mass1 for p in self.template_params])
+            m2 = numpy.array([p['tmplt'].mass2 for p in self.template_params])
+
+            # chirp mass of each template
```
tdent: this comment line & blank newline could be removed, calculation of chirp mass is self-documenting.

tdent:
Has this been tested (eg on a chirp-mass-split bank file) and shown to produce different trigger counts as expected from the scaling of the log mass window

bin/pycbc_inspiral
```diff
@@ -161,6 +161,8 @@ parser.add_argument("--user-tag", type=str, metavar="TAG", help="""
                     This is used to identify FULL_DATA jobs for
                     compatibility with pipedown post-processing.
                     Option will be removed when no longer needed.""")
+parser.add_argument("--keep-loudest-log-chirp-window", type=float,
+                    help="Keep loudest triggers within log chirp mass window")
```
tdent: to be unambiguous, should probably say 'ln chirp mass' in help (ie *not* log10)

ahnitz:
@tdent Yes, I've tested by hand, and also checked that if you don't use chirp bins it behaves as normal.

PR #2589, opened 2019-03-24
Pp workflow input injections

JulianWesterweck:
This should enable the workflow to read injections from a given file, split them into single injection files and use these for each inference run. 
Setting the value of a fixed parameter in the inference config file to 'injection' inserts the value from the injection file for that run. 
The workflow seems to be running through up to the plotting scripts for the recovery of the parameters and the intervals. 
The error from the most recent run says that lal cannot be imported, although I can load that module in python in the virtualenv I am using. So probably I just messed up the PYTHONPATH somehow?


cdcapano:
Superceded by #2738.

cdcapano:
Superseded by #2738.

PR #2590, opened 2019-03-25
Modified pycbc_plot_singles_vs_params to make plotting neat by removi…

sum33it:
Allow any single detector ranking statistic to plot using pycbc_plot_singles_vs_params

### DIFF
--- a/bin/hdfcoinc/pycbc_plot_singles_vs_params
+++ b/bin/hdfcoinc/pycbc_plot_singles_vs_params
@@ -35,6 +35,7 @@ import pycbc.results
 import pycbc.io
 import sys
 import pycbc.version
+from pycbc.events.stat import sngl_statistic_dict
 
 parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument("--version", action="version",
@@ -62,8 +63,7 @@ parser.add_argument('--y-var', required=True,
                     choices=pycbc.io.SingleDetTriggers.get_param_names(),
                     help='Parameter to plot on the y-axis. Required')
 parser.add_argument('--z-var', required=True,
-                    help='Quantity to plot on the color scale. Required',
-                    choices=['density', 'max(snr)', 'max(newsnr)', 'max(newsnr_sgveto)', 'max(newsnr_sgveto_psdvar)'])
+                    help='Quantity to plot on the color scale. Required')
 parser.add_argument('--detector', required=True,
                     help='Detector. Required')
 parser.add_argument('--grid-size', type=int, default=80,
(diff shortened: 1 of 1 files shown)


bin/hdfcoinc/pycbc_plot_singles_vs_params
```diff
@@ -127,16 +127,20 @@ if opts.z_var == 'density':
...
-        z = trigs.snr
-    elif opts.z_var == 'max(newsnr)':
-        z = trigs.newsnr
-    elif opts.z_var == 'max(newsnr_sgveto)':
-        z = trigs.newsnr_sgveto
-    elif opts.z_var == 'max(newsnr_sgveto_psdvar)':
-        z = trigs.newsnr_sgveto_psdvar
+    z = trigs.'%s'%opts.z_var    
+
+#elif opts.z_var in ['max(snr)', 'max(newsnr)', 'max(newsnr_sgveto)', 'max(newsnr_sgveto_psdvar)']:
```
ahnitz: Hey, let's delete these comments.

bin/hdfcoinc/pycbc_plot_singles_vs_params
```diff
@@ -127,16 +127,20 @@ if opts.z_var == 'density':
...
     cb_style = {}
-    if opts.z_var == 'max(snr)':
-        z = trigs.snr
-    elif opts.z_var == 'max(newsnr)':
-        z = trigs.newsnr
-    elif opts.z_var == 'max(newsnr_sgveto)':
-        z = trigs.newsnr_sgveto
-    elif opts.z_var == 'max(newsnr_sgveto_psdvar)':
-        z = trigs.newsnr_sgveto_psdvar
+    z = trigs.'%s'%opts.z_var
```
ahnitz: z = getattr(trigs, opts.z_var)

ahnitz:
@sum33it Has tested he can generate by hand plots from this patch. 

This patch requires changes to config files which will call the updated version of this code.

the sngl-stat option now uses the same name as the dict in the pycbc/events/stat module instead of things like "max(snr)" you give just "snr". 

The config file in https://github.com/gwastro/pycbc-config/blob/master/test/plotting.ini should be updated in a separate PR against the pycbc-config repo.

tdent:
@ahnitz @sum33it I was completely unaware of changes - which silently break most of the plots unless the input option for z-var is changed - and none of the O2 or O3 plotting ini files seem to have been updated to include this change.  The change is also not evident in the help message. 

How were we supposed to know that we should change the ini files for ongoing analysis and development?

tdent:
Also, there is no fail-through for the z-var value:
`if opts.z_var == 'density':
    norm = LogNorm()
    hb = ax.hexbin(x, y, norm=norm, vmin=1, **hexbin_style)
    fig.colorbar(hb, ticks=LogLocator(subs=range(10)))
elif opts.z_var in sngl_statistic_dict:
`

if z_var is *not* in sngl_statistic_dict, it just does nothing and produces a mysterious empty plot - whereas it should report an error.

ahnitz:
@tdent You should track changes to the ini files. They have been updated here 
https://github.com/gwastro/pycbc-config/tree/master/test
This is the location which should always work with master.

PR #2592, opened 2019-03-25
Force PyCBC Live's SNR series to have identical sampling

bhooshan-gadre:


bhooshan-gadre:
A quick-fix for #2563

bin/pycbc_live
```diff
@@ -181,6 +181,14 @@ class LiveEventManager(object):
                                              data_readers, bank, followup_ifos)
 
             live_ifos = [ifo for ifo in ifos if 'snr_series' in fud[ifo]]
+
+            ## FIXME: Assertion is not working with the real data.
+            # assert len({fud[ifo]['snr_series'].delta_t for ifo in live_ifos}) == 1
+            ## Temparary fix:
+            fix_delta_t = fud[live_ifos[-1]]['snr_series'].delta_t
+            for ifo in live_ifos:
+                fud[ifo]['snr_series']._delta_t = fix_delta_t
```
titodalcanton: A few comments on this block:
* Remove the commented assert (in general we should not leave bits of commented code around).
* There is nothing special in real data that requires this hack (I got the error from the O2 replay), and unfortunately we do not know how temporary this hack will be ;) So let's use a single comment line like this instead: `# FIXME delta_t can vary due to rounding errors, force them to be identical`
* Is there a reason for choosing the last ifo? If not, I would pick 0 instead of -1.

pycbc/filter/matchedfilter.py
```diff
@@ -1247,10 +1247,9 @@ def matched_filter_core(template, data, psd=None, low_frequency_cutoff=None,
         h_norm = sigmasq(htilde, psd, low_frequency_cutoff, high_frequency_cutoff)
 
     norm = (4.0 * stilde.delta_f) / sqrt( h_norm)
-    delta_t = 1.0 / (N * stilde.delta_f)
 
-    return (TimeSeries(_q, epoch=stilde._epoch, delta_t=delta_t, copy=False),
-           FrequencySeries(qtilde, epoch=stilde._epoch, delta_f=htilde.delta_f, copy=False),
+    return (TimeSeries(_q, epoch=stilde._epoch, delta_t=stilde.delta_t, copy=False),
```
titodalcanton: Isn't this change already on master? Maybe you should rebase this PR on master?

bin/pycbc_live
bhooshan-gadre: - I have rebased the PR against master.
- I have seen that the coinc IFOs have the same dt but followup is the one which has slightly different dt for all the cases I have seen in the simulated data. Hence I have chosen -1 instead of 0 as later is the followup detector

titodalcanton: Ok, but then I would still prefer the delta_t of the coinc IFOs, because that is going to be always the same across different triggers, while the delta_t of the followup ifo is going to be essentially slightly random.

bhooshan-gadre: Yes. Hence I have chosen live_ifos[-1] which is one of the coinc IFO.

titodalcanton: Going through the code again, I think there is actually no guarantee that the order of `live_ifos` means anything. The order of `live_ifos` is the same as `ifos`, which comes from the order of keys in a dictionary, and that is undefined in Python 2.7. So this should be more robust (and readable):
```python
fix_delta_t = fud[coinc_ifos[0]]['snr_series'].delta_t
```

PR #2599, opened 2019-03-28
fix double coinc workflow for api change to foreground table

ahnitz:


ahnitz:
@duncan-brown Let's see what Travis says now, and then add a re-enable for the docker tests if this passes. (the workflow test passes locally on my machine).

PR #2601, opened 2019-03-28
update gw150914 inf example to point to gwosc

Cyberface:
Points to GWOSC frames and Channel names

(also some white space from my text editor sorry)


ahnitz:
@Cyberface Please merge when travis has passed.

PR #2602, opened 2019-03-28
Stat refactor

tdent:
reorganizing ranking stat calculating functions etc


tdent:
(to address issue #2593)

tdent:
Looks like just removing unwanted imports, moving functions/modules to a more 'logical' place to not duplicate names between package/module, and the resulting deluge of CodeClimates, was enough for one PR.  Removing the 'enhancement' and 'WIP' labels .. so far, no functionality should be affected

tdent:
Yes, don't I know it .. most of eventmgr.py (former events.py) seems to have been written before PEP8 was a thing.  It could still be used for a tutorial on 'how to refactor in python' but that is for another day.

PR #2606, opened 2019-03-29
Use config argument group in pycbc inference

cdcapano:
Currently, `pycbc_inference` used it's own config option group and parser (in `inference/option_utils`). This is redundant with the argument group and `from_cli` option that exists in `workflow/configuration`. It also lacks the `config-delete` option. This patch has `inference` use the functions in the workflow configuration module instead, and removes the redundant functions from `inference/options_utils`.

I also changes the heading and description on the configuration argument group, to make it less workflow-specific.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -137,7 +137,7 @@ if stilde_dict:
 with ctx:
 
     # read configuration file
-    cp = configuration.WorkflowConfigParser.from_clie(opts)
+    cp = configuration.WorkflowConfigParser.from_cli(opts)
 
     # create an empty checkpoint file, if needed
     condor_ckpt = cp.has_option('sampler', 'checkpoint-signal')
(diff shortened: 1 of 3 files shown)


ahnitz:
:+1: Thanks for reducing code duplication.

bin/inference/pycbc_inference
```diff
@@ -136,7 +137,7 @@ if stilde_dict:
 with ctx:
 
     # read configuration file
-    cp = option_utils.config_parser_from_cli(opts)
+    cp = configuration.WorkflowConfigParser.from_clie(opts)
```
soumide1102: @cdcapano Noticed this typo. `from_clie` --> `from_cli`

ahnitz: :+1:

cdcapano: Thanks for catching @soumide1102; fixed.

PR #2610, opened 2019-03-31
remove expired elements only when trying to read data, no log with sl…

ahnitz:
…ow methods


titodalcanton:
Seems promising after several minutes of running. I will let it run for several hours.

Could you remove the commented code? I think that particular logging is no longer necessary, now that memory does not appear to be an issue anymore.

PR #2618, opened 2019-04-03
Update ringdown ini file in inference workflow docs

micamu:
Updating documentation


micamu:
@cdcapano I moved the ringdown ini file to the examples and removed it from the rst, as you requested. I did the same with the full IMR ini file, pointing to the ini file that is currently in examples (from Sumit's PR). Do you want me to do the same with the workflow ini? It is a bit different than the `make_inj_workflow` ini file, so I would create a new workflow ini file in the examples and point to that

cdcapano:
@micamu Thanks. Yes, please also create a workflow ini file in the examples and point to that.

micamu:
@cdcapano I have now moved the workflow config and the run workflow to the examples. Since the examples `inference_inj` folder seemed specific for the `pycbc_make_inference_inj_workflow`, I have renamed it to just `inference`, put all the files for both inference workflows there, and adjusted accordingly the inj_workflow docs as well. You might want to check if the workflow ini file `examples/workflow/inference/workflow_config.ini` and the workflow options in `examples/workflow/inference/run_pycbc_make_inference_workflow.sh` are up-to-date (a few things are different than in inference_inj_workflow, but I am not sure how much you want to change)

micamu:
@cdcapano The "work in progress" label was just because I was waiting for your answers, which might have required more changes. If you are happy with this as is, I am done

PR #2619, opened 2019-04-03
spread coinc ops over multiple processors

ahnitz:


ahnitz:
@titodalcanton I'll test this on ATLAS and report back if it seems to be doing the right thing.

ahnitz:
Ok, I've tested this runs on Gaussian noise and seems to produce sane output. I have not tested if it handles hwinj appropriately. This change should mean that they still get flagged properly, but they will pollute the background until their triggers get cycled out of the buffer. I don't know if that's a behavior we can live with or not. (depends how often hwinj are done).

titodalcanton:
Testing.

PR #2623, opened 2019-04-03
Allow astropy version > 3 if using Python 3

### LINKED ISSUE #2612: Remove upper requirement on the astropy version
Is there any reason for having the upper version bound on astropy (`astropy>=2.0.3,<3.0.0`) in the requirements file? Is this just that PyCBC is not yet completely Python 3 compatible yet? From looking at the locations where astropy is used in the code, and testing them with astropy v3.1.2 installed, I don't seem to see any problems.

If the issue is that not having the upper bound makes the test suite try and download a non-Python 2 compatible astropy version, you could edit the requirements file to, e.g.,

```
astropy>=2.0.3; python_version > '3.4'
```

mattpitkin:
Edit the `requirements.txt` and `setup.py` files to allow people using Python 3 to have astropy versions that are > 3.

Closes #2612


ahnitz:
@mattpitkin As always thank you!

PR #2625, opened 2019-04-04
add ligo-lw to requirements for virtualenv builds

ahnitz:


ahnitz:
@spxiwh Ok, this seems to work. Hopefully it works on master too 🤞

PR #2626, opened 2019-04-04
Re-enable using different low frequency cutoffs for different detectors

soumide1102:
As title says.

@cdcapano  The PR makes the low frequency cutoffs for likelihood calculation to be read from the config file as 
```
h1-low-frequency-cutoff = 20
l1-low-frequency-cutoff = 30
```
etc. These will be saved in fp.attrs as `'{}_likelihood_low_freq'.format(det)`. 

The low frequencies for multi ifos for data conditioning is read in the command line with the option `--data-conditioning-low-freq`. This is then saved to `opts.low_frequency_cutoff` for passing to strain.from_cli_multi_ifos and psd.from_cli_multi_ifos. I've removed the `low_frequency_cutoff_from_cli` function as it should not be needed anymore.

I don't see how I can save the data conditioning low freqs in the output file though, as the metadata is being written now in the models module which which does not have access to the data conditioning settings, ie. the `opts` in `pycbc_inference`.

### DIFF
--- a/pycbc/inference/option_utils.py
+++ b/pycbc/inference/option_utils.py
@@ -24,6 +24,7 @@ from pycbc.psd import from_cli_multi_ifos as psd_from_cli_multi_ifos
 from pycbc.strain import from_cli_multi_ifos as strain_from_cli_multi_ifos
 from pycbc.strain import (gates_from_cli, psd_gates_from_cli,
                           apply_gates_to_td, apply_gates_to_fd)
+from pycbc.types import MultiDetOptionAction
 from pycbc import waveform
 from pycbc import distributions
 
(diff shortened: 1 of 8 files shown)


soumide1102:
@cdcapano Updated with ability to read separate high frequency cutoffs for different IFOs from config file. If a high frequency cutoff (IFO-high-frequency-cutoff) for an IFO is provided, min(IFO-high-frequency-cutoff, template_f_upper) will be used for likelihood calculation and the inputted IFO-high-frequency-cutoff will be stored as an attribute in the Inference file. If a high freq is not provided for an IFO min(nyquist, template_f_upper) will be used and no high frequency cutoff attribute will be saved for that IFO.

If the changes look okay, I'll update the examples and documentation.

soumide1102:
Now updated the examples too, since those were related to the low frequency cutoff changes and Collin said those look okay to him.

soumide1102:
@cdcapano this is ready for reviewing again.

pycbc/inference/option_utils.py
```diff
@@ -37,27 +38,10 @@
...
-    """
-    # FIXME: this just uses the same frequency cutoff for every instrument for
-    # now. We should allow for different frequency cutoffs to be used; that
-    # will require (minor) changes to the Likelihood class
-    instruments = opts.instruments if opts.instruments is not None else []
-    return {ifo: opts.low_frequency_cutoff for ifo in instruments}
+    parser.add_argument("--data-conditioning-low-freq", type=float,
+                        nargs="+", action=MultiDetOptionAction,
+                        metavar='IFO:FLOW', dest="low_frequency_cutoff",
+                        help="Low frequency cutoff for fake strain.")
```
cdcapano: This is also needed for PSD estimation (as I've found out the hard way). Change to `Low frequency cutoff of the data. Needed for PSD estimation and when creating fake strain.`)

cdcapano:
@soumide1102 It looks like there's also a merge conflict. Please rebase to master and resolve it.

PR #2627, opened 2019-04-05
Adding autogating to PyCBC Live

litingxiao:
Adds autogating functionality to the low latency searches: 
- pycbc_live argument `autogating-threshold` is default to `None`, and gating won't be triggered. Specifying an autogating-threshold value will trigger the autogating function. 
- `Clustering time`, `gating padding`, `gating window`, `corrupt time` could be specified in pycbc_live arguments.

[NOTE: Perhaps should add arguments in pycbc_live to specify psd duration and overlap for detect_loud_glitches(). For now it's hardcoded in to be (1s, 0.5s).]

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -351,7 +351,7 @@ parser.add_argument('--trim-padding', type=float, default=0.25,
 parser.add_argument("--enable-bank-start-frequency", action='store_true',
                     help="Read the starting frequency of template waveforms"
                          " from the template bank")
-parser.add_argument('--autogating-threshold', type=float, default=100)
+parser.add_argument('--autogating-threshold', type=float, default=None)
 parser.add_argument('--autogating-pad', type=float, default=.25)
 parser.add_argument('--autogating-window', type=float, default=0.5)
 parser.add_argument('--autogating-cluster', type=float, default=.25)
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -351,7 +351,7 @@ parser.add_argument('--trim-padding', type=float, default=0.25,
 parser.add_argument("--enable-bank-start-frequency", action='store_true',
                     help="Read the starting frequency of template waveforms"
                          " from the template bank")
-parser.add_argument('--autogating-threshold', type=float, default=None)
+parser.add_argument('--autogating-threshold', type=float)
 parser.add_argument('--autogating-pad', type=float, default=.25)
 parser.add_argument('--autogating-window', type=float, default=0.5)
 parser.add_argument('--autogating-cluster', type=float, default=.25)
--- a/pycbc/strain/strain.py
+++ b/pycbc/strain/strain.py
@@ -1201,7 +1201,7 @@ class StrainBuffer(pycbc.frame.DataBuffer):
                  psd_segment_length=4,
                  psd_inverse_length=3.5,
                  trim_padding=0.25,
-                 autogating_threshold=100,
+                 autogating_threshold=None,
                  autogating_cluster=0.25,
                  autogating_window=0.5,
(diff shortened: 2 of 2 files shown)


bin/pycbc_live
```diff
@@ -337,7 +337,7 @@ parser.add_argument('--trim-padding', type=float, default=0.25,
 parser.add_argument("--enable-bank-start-frequency", action='store_true',
                     help="Read the starting frequency of template waveforms"
                          " from the template bank")
-parser.add_argument('--autogating-threshold', type=float, default=100)
+parser.add_argument('--autogating-threshold', type=float, default=None)
```
titodalcanton: Liting, you can remove the `default=None` here, since it is… the default :)

litingxiao: Yeah sure! Done :)

titodalcanton:
This just got a huge priority bump:

https://ldas-jobs.ligo.caltech.edu/~tito/pycbc/live/debug_trigger_gaps/production_20190514_2.png

tdent:
Where are we with this request? Formally we have "Review required/ At least 1 approving review is required".  Anything else?

titodalcanton:
Apparently this is now running on the O2 replay analysis.

titodalcanton:
So while debugging this patch, I found at least a problem of integer division causing a non-negligible error in the detected glitch time. I do not think that will be the end of the story, but I would like to rule that out before continuing. Please wait for #2823 to be merged, and rebase on master when it is.

pycbc/strain/strain.py
```diff
@@ -1597,12 +1597,28 @@ def advance(self, blocksize, timeout=10):
...
+                    cluster_window=self.autogating_cluster,
+                    low_freq_cutoff=self.highpass_frequency,
+                    high_freq_cutoff=self.sample_rate/2,
+                    corrupt_time=self.autogating_pad)
+            if len(glitch_times) > 0:
+                logging.info('Autogating at %s',
+                             ', '.join(['%.3f' % gt for gt in glitch_times]))
+                gate_params = [[gt, self.autogating_window, self.autogating_pad] \
+                               for gt in glitch_times]
+                strain = gate_data(strain, gate_params)
```
titodalcanton: I am a bit worried about boundary effects here. At this point, with the current configuration, the segment of strain is only 8 s long. So it is not unlikely that a gate will fall too close to the start or end of the segment, and the tapering of the gate will be "chopped off" by the boundary, effectively making the gate shorter and with an abrupt edge. We can easily solve half of this problem by moving the block starting with `if len(glitch_times) > 0:` past the line reading `self.strain.start_time += blocksize`, and applying `gate_data` to `self.strain` instead of `strain`. Since `self.strain` also contains a lot of past strain, any gate that intersects the start of `strain` will be applied correctly. I am still not sure what to do for the gates that intersect the *end* of `strain`, though. We would need to "remember the gate" and apply the missing part to the *beginning* of `strain` at the next segment. Does this make sense?

titodalcanton: There is also the lazy option to simply discard any gate that intersects the edges of `strain`, I guess…

titodalcanton:
I rebased this on master and set it to run on the O2 replay analysis.

titodalcanton:
Looking at the most recent autogates generated on replay data, they are all associated with glitches, so I am convinced the glitch-finding part is working fine at least.

titodalcanton:
I did a couple fixes to remove verbosity and avoid an unnecessary resampling, and rebased on master. This latest version is now running on replay data. Unless some problems show up there in the next ~hours, I propose we merge this. @spxiwh or @bhooshan-gadre maybe you should do the github review since both Liting and I contributed code.

PR #2629, opened 2019-04-08
fix import in fit_by_template

tdent:


tdent:
(strange that none of our code checkers worked out `triggers.get_param` was going to fail because `triggers` was not defined yet ..)

PR #2630, opened 2019-04-08
Write/read from hdf in inspiral test run in hwinj docs & update filenames

soumide1102:
`pycbc_inspiral` can't write to xml.gz files now. https://github.com/gwastro/pycbc/blob/master/bin/pycbc_inspiral#L259 calls https://github.com/gwastro/pycbc/blob/master/pycbc/events/eventmgr.py#L364-L372 . So made the `pycbc_inspiral` example output file hdf and print the time and SNR from that file.

Also updated the ASCII/xml.gz default filenames that I noticed have changed.


soumide1102:
@cmbiwer Just checking whether you've noticed this PR.

PR #2631, opened 2019-04-08
multiifo coinc rate module

GarethDaviesGW:
introduction of a coinc_rate module to include the estimation of noise coincidence rates from single trigger rates

### DIFF
--- a/pycbc/events/coinc_rate.py
+++ b/pycbc/events/coinc_rate.py
@@ -85,7 +85,8 @@ def multiifo_noise_coincident_area(ifos, slop):
         dets[ifo] = pycbc.detector.Detector(ifo)
     n_ifos = len(ifos)
     if n_ifos == 2:
-        allowed_area = 2 * (dets[ifos[0]].light_travel_time_to_detector(dets[ifos[1]]) + slop)
+        allowed_area = 2 * (dets[ifos[0]].light_travel_time_to_detector(
+                                                        dets[ifos[1]]) + slop)
     elif n_ifos == 3:
         tofs = numpy.zeros(n_ifos)
         ifo2_num = []
--- a/pycbc/events/coinc_rate.py
+++ b/pycbc/events/coinc_rate.py
@@ -5,7 +5,7 @@
 #
 # =============================================================================
 #
-""" This module contains functions for calculating expected rates of noise 
+""" This module contains functions for calculating expected rates of noise
     and signal coincidences.
 """
 
@@ -37,7 +37,6 @@ def multiifo_noise_coinc_rate(rates, slop):
     ifos = numpy.array(sorted(rates.keys()))
     rates_raw = list(rates[ifo] for ifo in ifos)
     expected_coinc_rates = {}
-    n_ifos = len(ifos)
     # Calculate coincidence for all-ifo combination
     # multiply the product of rates and by the overlap time
     allowed_area = multiifo_noise_coincident_area(ifos, slop)
@@ -47,11 +46,10 @@ def multiifo_noise_coinc_rate(rates, slop):
     expected_coinc_rates[ifostring] = allowed_area * numpy.array(rateprod)
     # if more than one possible coincidence type exists,
     # calculate coincidences for subsets through recursion
-    if n_ifos > 2:
+    if len(ifos) > 2:
         # Calculate rate for each 'miss-one-out' detector combination
-        subsets = itertools.combinations(ifos, n_ifos - 1)
+        subsets = itertools.combinations(ifos, len(ifos) - 1)
         for subset in subsets:
-            i_set = [numpy.nonzero(ifo == ifos)[0][0] for ifo in subset]
             rates_subset = {}
             for ifo in subset:
                 rates_subset[ifo] = rates[ifo]
--- a/pycbc/events/coinc_rate.py
+++ b/pycbc/events/coinc_rate.py
@@ -21,31 +21,32 @@ def multiifo_noise_coinc_rate(rates, slop):
     Parameters
     ----------
     rates: dict
-        Dictionary keyed on interferometer
-        Value is a vector of single-detector trigger rates
+        Dictionary keyed on ifo string
+        Value is a sequence of single-detector trigger rates, units assumed 
+        to be Hz
     slop: float
-        time to be added to time-of-flights between detectors to account
+        time added to maximum time-of-flight between detectors to account
         for timing error
 
     Returns
     -------
     expected_coinc_rates: dict
-        A dictionary of ifo combined rates keyed on the combination string
-        value is the expected coincidence rate in the combination given the
-        individual detector trigger rates
+        Dictionary keyed on the ifo combination string
+        Value is expected coincidence rate in the combination, units Hz
     """
     ifos = numpy.array(sorted(rates.keys()))
     rates_raw = list(rates[ifo] for ifo in ifos)
     expected_coinc_rates = {}
+    
     # Calculate coincidence for all-ifo combination
-    # multiply the product of rates and by the overlap time
+    # multiply product of trigger rates by the overlap time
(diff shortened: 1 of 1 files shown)


tdent:
@GarethDaviesGW - can you remove the work in progress tag when this is ready to review and assign to Alex N.

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
+#
+# =============================================================================
+#
+#                                   Preamble
+#
+# =============================================================================
+#
+""" This modules contains functions for calculating and manipulating
```
tdent: modules -> module 
i think 'calculating expected rates of noise and signal coincidences' would be more accurate

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+
+def multiifo_noise_coinc_rate(rates, ifos, slop):
+    """
+    Calculate the expected rate of noise coincidences for multiple detectors
+
+    Parameters
+    ----------
+    rates: list of lists/arrays
+        List of vector of single-detector trigger rates
+        in order of template id
```
tdent: do they have to be in order of anything specific?

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+    ifos: list of strings
+        list of ifos in e.g. 'H1' form
```
tdent: I'm guessing that it needs to get the list of ifos in the same order as the list of trigger rates.  if so, would it work better to use a dictionary as input (ie keyed on the ifo name) ?

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+        value is the expected coincidence rate in the combination given the
+        individual detector trigger rates
+    """
+    ifos = numpy.array(ifos)
+    rates = numpy.array(rates)
+    expected_coinc_rates = {}
+    n_ifos = len(ifos)
+    assert len(rates) == n_ifos
+    # Calculate coincidence for all-ifo combination
+    # multiply the two rates and by the overlap time
```
tdent: two or n?

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+    allowed_area = multiifo_noise_coincident_area(ifos, slop)
+    rateprod = [numpy.prod(rs) for rs in zip(*rates)]
+    ifostring = ''.join(ifos)
```
tdent: in other places i think we are using a different layout for sets of ifos, maybe want to be consistent here?

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+    expected_coinc_rates[ifostring] = allowed_area * numpy.array(rateprod)
+    # if more than one possible coicidence type exists,
+    # calculate coincidences for subsets
```
tdent: subsets -> 'subsets via recursion' & remove later comment about calculate through iteration

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+            # add these sub-coincidences to the overall dictionary
+            for sub_coinc in sub_coinc_rates:
+                expected_coinc_rates[sub_coinc] = sub_coinc_rates[sub_coinc]
+
+    return expected_coinc_rates
+
+
+def multiifo_noise_coincident_area(ifos, slop):
+    """
+    calculate the multiplicative factor of the individual detector trigger
```
tdent: prefer to say what calculation the function is doing. eg 'total extent of time offset between 2 detectors, or area of the 2d space of time offsets for 3 detectors, for which a coinc can be generated'

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+            ifo2_num.append(int(numpy.mod(i+1, n_ifos)))
+            det0 = dets[ifo]
+            det1 = dets[ifos[ifo2_num[i]]]
+            tofs[i] = det0.light_travel_time_to_detector(det1) + slop
+
+        # combine these to calculate allowed area
+        allowed_area = 0
+        for i, _ in enumerate(ifos):
+            allowed_area += 2*tofs[i]*tofs[ifo2_num[i]] - tofs[i]**2
+
```
tdent: probably want to put something like `else raise NotImplementedError("Not able to deal with more than 3 ifos")` here (and remove the 'todo' about this)

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+        tofs = numpy.zeros(n_ifos)
+        ifo2_num = []
+        # set up detector objects
+        for ifo in ifos:
+            dets[ifo] = pycbc.detector.Detector(ifo)
+
+        # calculate travel time between detectors (plus extra for timing error)
+        # TO DO: allow for different timing errors between different detectors
+        for i, ifo in enumerate(ifos):
+            ifo2_num.append(int(numpy.mod(i+1, n_ifos)))
```
tdent: PEP8, put spaces around operators. this also applies to line 110 (the actual area calculation)

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+        allowed_area = 2*det0.light_travel_time_to_detector(det1)
+    elif n_ifos == 3:
+        dets = {}
+        tofs = numpy.zeros(n_ifos)
+        ifo2_num = []
+        # set up detector objects
+        for ifo in ifos:
+            dets[ifo] = pycbc.detector.Detector(ifo)
+
+        # calculate travel time between detectors (plus extra for timing error)
```
tdent: does this include timing error? doesn't seem like it does.

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,156 @@
...
+        # TO DO: allow for different timing errors between different detectors
+        for i, ifo in enumerate(ifos):
+            ifo2_num.append(int(numpy.mod(i+1, n_ifos)))
+            det0 = dets[ifo]
+            det1 = dets[ifos[ifo2_num[i]]]
+            tofs[i] = det0.light_travel_time_to_detector(det1)
+
+        # combine these to calculate allowed area
+        phi_12 = numpy.arccos((tofs[0]**2+tofs[1]**2 -
```
tdent: spaces around operators please

pycbc/events/coinc_rate.py
```diff
@@ -0,0 +1,154 @@
...
+    for ifo in ifos:
+        dets[ifo] = pycbc.detector.Detector(ifo)
+    n_ifos = len(ifos)
+    if n_ifos == 2:
+        allowed_area = 2 * (dets[ifos[0]].light_travel_time_to_detector(
+                                                        dets[ifos[1]]) + slop)
+    elif n_ifos == 3:
+        tofs = numpy.zeros(n_ifos)
+        ifo2_num = []
+        # set up detector objects
```
tdent: did some code get eaten at this point?

GarethDaviesGW: I moved the creation of dets (and deprecated det1 and det0) to created outside of the if statement, which meant that the 2-ifo case could become a single line

tdent: ok, so fix the comment

pycbc/events/coinc_rate.py
GarethDaviesGW: updated

tdent:
@ahnitz I am OK with merging this - if you had any comments please speak soon.

Note that the formulae it implements will be (or are already - Gareth?) in the main text or appendix of the multi-ifo methods paper, comments pointing to them can be put in as and when published.

PR #2634, opened 2019-04-09
add multiifo method for expfitcombinedsnr

tdent:


pycbc/events/stat.py
```diff
@@ -456,7 +455,11 @@ def single(self, trigs):
 
     def coinc(self, s0, s1, slide, step): # pylint:disable=unused-argument
         # scale by 1/sqrt(2) to resemble network SNR
-        return (s0 + s1) / (2.**0.5)
+        return (s0 + s1) / 2.**0.5
+
+    def coinc_multiifo(self, s, slide, step): # pylint:disable=unused-argument
```
ahnitz: I don't think it's great practice to use the .values() method this way. Could you do instead

sum(s[k] for k in s)

Also, I *think* len(s) should work directly.

tdent:
Has been minimally tested by checking that the resulting statistic is similar to combined newsnr_sgveto (for doubles and triples).

PR #2636, opened 2019-04-09
Live: enable per-detector analysis flags

titodalcanton:
I do some refactoring of the data reader code in `pycbc_live`, and add the ability to use different analysis flags for different detectors. This is necessary for Virgo in O3.

Note that this patch will require a change in the current PyCBC Live configuration.

Needs testing.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -193,7 +193,7 @@ class LiveEventManager(object):
             fud = self.compute_followup_data(coinc_ifos, coinc_results,
                                              data_readers, bank, followup_ifos)
 
-            live_ifos = [ifo for ifo in ifos if 'snr_series' in fud[ifo]]
+            live_ifos = [ifo for ifo in fud if 'snr_series' in fud[ifo]]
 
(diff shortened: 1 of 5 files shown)


titodalcanton:
@ahnitz I still need to test this more carefully, but the approach is finalized, so ready for your review.

titodalcanton:
This is now tested with fake frame files carrying various combinations of states and DQ vetoes. @ahnitz ready for your review.

bin/pycbc_live
```diff
@@ -287,6 +285,54 @@ class LiveEventManager(object):
                 if store_psd[ifo] is not None:
                     store_psd[ifo].save(fname, group='%s/psd' % ifo)
 
+def init_data_reader(ifo, args, maxlen):
```
ahnitz: Should this be a class method of StrainBuffer?

ahnitz:
@titodalcanton One minor question, otherwise, this seems fine.

bin/pycbc_live
titodalcanton: Makes sense, moved to a `from_cli` class method.

PR #2640, opened 2019-04-12
make build more python3 compatible

ahnitz:


ahnitz:
@spxiwh Ok, I think this is ready for review.

 It changes the requirements so that packages that don't work in python 3 aren't installed. This allows the python3 travis build to continue. Also, we add to the travis.yml file the python2.7 version required for the "regular" build. This will become required in a couples days by travis. Finally, I remove the debian virtualenv build. We don't use it and from discussion in the chat it seems like the LVC isn't either.

PR #2641, opened 2019-04-12
Live: avoid human confusion with χ² not being calculated

titodalcanton:
People suggested that we rather not populate the χ² fields of the LIGOLW uploads for the followup detectors, rather than setting them to zero. I think this is reasonable and should have no side effects.

I also spotted and fixed an incorrect behavior: some detector-dependent trigger fields (namely `sg_chisq` and `sigmasq`) were being copied from one of the above-threshold detectors into the fields of followup detectors. I think the only important side effect was that the effective distance of the followup detectors in GraceDB was incorrect. I now take the correct `sigmasq` from the SNR series calculation.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -134,34 +134,40 @@ class LiveEventManager(object):
 
         # Determine if the other ifos can contribute to the coincident event
         for ifo in followup_ifos:
-            snr_series, ptime, pvalue = followup_event_significance(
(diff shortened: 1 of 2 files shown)


titodalcanton:
Hold on merging, HDF5 does not like that.

titodalcanton:
@ahnitz can you have another look? The patch is a bit more complex now.

ahnitz:
@titodalcanton This looks ok to me

PR #2642, opened 2019-04-14
Added injfilterrejector for multi ifo. 

idorrington92:
This just added the options for the multi ifo version, and makes a dictionary from running the old injfilterrejector on each ifo.

### DIFF
--- a/pycbc/inject/injfilterrejector.py
+++ b/pycbc/inject/injfilterrejector.py
@@ -108,6 +108,29 @@ def insert_injfilterrejector_option_group(parser):
     injfilterrejector_group.add_argument(curr_arg, type=int, default=None,
                                          help=_injfilterer_flower_help)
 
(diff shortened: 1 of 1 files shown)


spxiwh:
@idorrington92 See the CodeClimate report, some of this is formatting, but there's genuine bugs in there too.

Beyond that, this probably would need to be hooked into `pycbc_multi_inspiral` before testing. I'm happy to merge this (once CodeClimate issues are fixed) and then merge that separately.

idorrington92:
Hi Ian,
I'm on holiday for a couple of weeks, I'll look into this when I get a
moment. If it's significant changes I'll do it when I get back to work.

Thanks,
Iain

spxiwh:
Hi @idorrington92. No worries, I don't think this is too major, but ignore this until you're back from vacation!

idorrington92:
I've fixed the codeclimate issues. I  think this can be merged now. Thanks

PR #2644, opened 2019-04-14
pick max singles by network SNR

ahnitz:
For a coincident event use the loudest (by network SNR) singles information for followup.


tdent:
It looks like this defines 'network' SNR as the sum over 2 coinc triggers, ie not including additional detectors - yes?

When you say followup what does that mean, btw?

tdent:
(& Codeclimate seems pretty unhappy with this)

titodalcanton:
I am wondering if we want a command-line option to choose between the old and new behavior, since (if I understand correctly) the change affects the choice of which template is used for the p-value in the followup detectors, and thus the FAR/sensitivity of the search.

ahnitz:
I no longer think this is is the best approach so am closing for now.

PR #2648, opened 2019-04-15
Fixing some Python 3 errors in PyCBC Live

titodalcanton:


titodalcanton:
Yes it would be good for @cdcapano to double check.

titodalcanton:
I just verified that this continues to work under Python 2.7.

PR #2650, opened 2019-04-17
Fix make psd estimation workflow

bhooshan-gadre:
`segments` in used for `segments` module in ligo and for iterating over actual segments in the loop:
```python
from ligo import segments
# then ...
for ifo, segments in analyzable_segs.items():
# ..
    s = segments.segment(start_time, end_time) # later causing problems
```


titodalcanton:
@bhooshan-gadre does the entire workflow work after this fix, or do you think more changes are needed?

bhooshan-gadre:
It works. No more  changes needed

PR #2655, opened 2019-04-21
Multiifo coinc : discard 'extra' coincs from too-closely-spaced triggers

tdent:
Also lots of work on comments and readability, remove variable names that shadow other names etc

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
@@ -1,6 +1,5 @@
 #!/usr/bin/env python
 import h5py, argparse, logging, numpy, numpy.random
-from pycbc import events, detector
 from pycbc.events import veto, coinc, stat
 import pycbc.version
 from numpy.random import seed, shuffle
@@ -174,8 +173,8 @@ num_templates = len(h5py.File(args.template_bank, "r")['template_hash'])
 tmin, tmax = parse_template_range(num_templates, args.template_fraction_range)
 logging.info('Analyzing template %s - %s' % (tmin, tmax-1))
 
-# Create dictionary for trigger files
-trigs={}
+# Create dictionary for trigger files indexed on ifo name
+trigs = {}
 for i in range(len(args.trigger_files)):
     logging.info('Opening trigger file %s: %s' % (i,args.trigger_files[i]))
     reader = ReadByTemplate(args.trigger_files[i],
@@ -184,6 +183,7 @@ for i in range(len(args.trigger_files)):
                                                args.veto_files)
     trigs[reader.ifo] = reader
 
+# Coinc_segs contains only segments where all ifos are analyzed
 coinc_segs = trigs[args.pivot_ifo].segs
 for ifo in trigs:
     coinc_segs = (coinc_segs & trigs[ifo].segs).coalesce()
@@ -192,12 +192,8 @@ for ifo in trigs:
     trigs[ifo].segs = coinc_segs
     trigs[ifo].valid = veto.segments_to_start_end(trigs[ifo].segs)
 
-# initialize a Stat class instance to calculate the coinc ranking statistic
+# Stat class instance to calculate the coinc ranking statistic
 rank_method = stat.get_statistic(args.ranking_statistic)(args.statistic_files)
-
-detectors={}
-for ifo in trigs:  # what are these Detector objects used for?
-    detectors[ifo] = detector.Detector(trigs[ifo].ifo)
 
 # Sanity check, time slide interval should be larger than twice the
 # Earth crossing time, which is approximately 0.085 seconds.
@@ -210,8 +206,6 @@ if args.timeslide_interval is not None and args.timeslide_interval <= TWOEARTH:
 if args.timeslide_interval is None:
     args.timeslide_interval = 0
 
-data = {'stat':[], 'decimation_factor':[], 'timeslide_id':[], 'template_id':[]}
-
 if args.randomize_template_order:
     seed(0)
     template_ids = numpy.arange(0, num_templates)
@@ -223,7 +217,7 @@ else:
 # 'data' will store output of coinc finding
 # in addition to these lists of coinc info, will also store trigger times and
 # ids in each ifo
-data = {'stat':[], 'decimation_factor':[], 'timeslide_id':[], 'template_id':[]}
+data = {'stat': [], 'decimation_factor': [], 'timeslide_id': [], 'template_id': []}
 
 for tnum in template_ids:
     tids = {}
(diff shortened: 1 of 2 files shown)


tdent:
Not sure if just removing the clustered-away trigger is sufficient ..

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
```diff
@@ -209,18 +213,33 @@ if args.timeslide_interval is None:
...
-    times={}
+    times = {}
     for ifo in trigs:
-        times[ifo]=trigs[ifo]['end_time']
+        times[ifo] = trigs[ifo]['end_time']
+        snr = trigs[ifo]['snr']
+        # Coinc finding relies on triggers being spaced by more than coinc window
+        cid = coinc.cluster_over_time(snr, times[ifo], TWOEARTH)
+        if len(snr) - len(cid):
+            times[ifo] = times[ifo][cid]
```
tdent: I doubt this does the right thing in terms of preserving the single trig info.

pycbc/events/coinc.py
```diff
@@ -245,92 +249,106 @@ def time_coincidence(t1, t2, window, slide_step=0):
...
 
-        # Add this detector now to the cumulative set and proceed to the next
-        # ifo coincidence test
-        ids[ifo1] = sort[dep_ids]
-        ctimes[ifo1] = otime[ids[ifo1]]
+        # undo time sorting on indices of ifo1 triggers
+        ids[ifo1] = tsort[dep_ids]
+        # FIXME do we want to do this?? add ifo1 times to ctimes which will be
+        # further reduced if more than one additional detector is tested
+        #ctimes[ifo1] = otime[ids[ifo1]]
```
ahnitz: This should not be commented out. It is required if you have 4+ detectors that you check against all combinations.

tdent:
> This looks like it depends on #2635.

It incorporates one part of it, which got into my repo since I am using it already.  No actual dependence and should merge smoothly with #2635 itself

pycbc/events/coinc.py
tdent: Can you please describe the exact desired behaviour with 4+ detectors.  What tests should be passed by the resulting coinc triggers?  At the moment it seems to be that the 4th detector's triggers are tested only against the 'pivot' and 'fixed' ones

ahnitz: It's really the same behavior as for 2+ detectors. Every ifo must be coincident. You need to add any new detectors to this dict so that when another detector is checked for coincidences the past ones are *all* included. Before your patch this would be true, afterwards, it will just break I think, since it will still see an ifo in the ids dict but not have the coresponding ifo times.

tdent: The loop that checks a new extra ifo against existing ones is `for ifo2 in ids:` however the previously added extra ifos are not added to the 'ids' dictionary, just to 'ctimes'.  So that made it unclear what was supposed to be happening

pycbc/events/coinc.py
```diff
@@ -245,92 +249,106 @@ def time_coincidence(t1, t2, window, slide_step=0):
...
-            # (assumes triggers spaced > slide step)
+            left = numpy.searchsorted(time1, ctimes[ifo2] - w)
+            right = numpy.searchsorted(time1, ctimes[ifo2] + w)
+            # any times within time1 coincident with the time in ifo2 have
+            # indices between left and right
+            # nz indexes into times in ifo2 which have coincidences with ifo1
+            # times
             nz = (right - left).nonzero()
+            if len(nz[0]) and (right - left).max() > 1:
+                # we expect at most one coincident time in ifo1, assuming
```
tdent: The 'easiest' solution here would just be to ignore one of these coincident triggers in ifo1 - ie silently proceed with just taking the first trigger, which I think left[nz] already does.

pycbc/events/coinc.py
tdent: hmm, not sure why I thought that if the code has `ids[ifo1] = sort[dep_ids]`

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
tdent: now given up on doing this as it seems it would be excessively complicated to do the right book-keeping for a rare corner case

pycbc/events/coinc.py
tdent: see new diff.

tdent:
Has been minimally tested on a set of trigger files which threw the original error for too closely spaced triggers (it does indeed produce a warning).

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
```diff
@@ -234,7 +240,7 @@ for tnum in template_ids:
         logging.info('%s:%s' % (ifo, len(times_full[ifo])))
 
     logging.info('Calculating Single Detector Statistic')
-    sds_full = {ifo:rank_method.single(trigs[ifo]) for ifo in trigs}
+    sds_full = {ifo: rank_method.single(trigs[ifo]) for ifo in trigs}
```
ahnitz: Shouldn't be a space after 'ifo:'

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
```diff
@@ -219,13 +220,18 @@ if args.randomize_template_order:
 else:
     template_ids = range(tmin, tmax)
 
+# 'data' will store output of coinc finding
+# in addition to these lists of coinc info, will also store trigger times and
+# ids in each ifo
+data = {'stat':[], 'decimation_factor':[], 'timeslide_id':[], 'template_id':[]}
```
ahnitz: This appears to now be in two places.

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
tdent: Going by 
https://stackoverflow.com/questions/9253432/coding-guidelines-for-python-dictionaries

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
```diff
@@ -196,12 +196,13 @@ for ifo in trigs:
 rank_method = stat.get_statistic(args.ranking_statistic)(args.statistic_files)
 
 detectors={}
-for ifo in trigs:
+for ifo in trigs:  # what are these Detector objects used for?
```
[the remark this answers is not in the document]
tdent: Or don't use the Detector objects at all ?  The assignment of `detectors` seems needlessly complicated, unless I'm missing something.  All it is actually needed for is the ifo names (that I can see)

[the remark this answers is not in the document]
ahnitz: Yeah, I don't think they are used anymore here. They can be removed.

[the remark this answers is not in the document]
tdent: .. the only places it is used are `sorted(detectors)` which returns a sorted list of _keys_ which are just ifo names - or as an iterator `for ifo in detectors:` which again doesn't use the dict values.

tdent:
@ahnitz are you OK with the basic change being made here - ie to not enforce that exactly 1 trigger exists in the coinc time window, and in the corner case that 2 exist arbitrarily store the first and not the other

ahnitz:
@tdent Yeah, I'm ok with the overall change.

tdent:
@ahnitz can you put in an approval for this so that it can be merged

ahnitz:
@tdent Please fix the logging issue (i.e. the name should be changed to 'warning'). Then feel free to merger.

tdent:
Oh, I didn't realize the deprecation was just warning() vs. warn(). So, should be easy.

PR #2657, opened 2019-04-22
fix mistake in pr2499 which broken injection matching

ahnitz:
It looks like PR#2499 accidentally removed a critical line. This is causing injections to be counted as missed when in actuality they are not recoverable at all due to being in times you aren't analyzing. 

https://github.com/gwastro/pycbc/commit/426cae0cc9546cb16e2e98f0532a8178405c3e21#diff-83b130558a65c4e4349ba945877e1758

### DIFF
--- a/bin/hdfcoinc/pycbc_coinc_hdfinjfind
+++ b/bin/hdfcoinc/pycbc_coinc_hdfinjfind
@@ -33,6 +33,7 @@ def keep_ind(times, start, end):
     time_sorting = times.argsort()
     times = times[time_sorting]
     indices = numpy.array([], dtype=numpy.uint32)
+    left = numpy.searchsorted(times, start, side='left')
     right = numpy.searchsorted(times, end, side='right')
 
     for li, ri in zip(left, right):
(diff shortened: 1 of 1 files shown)


tdent:
Which line?

ahnitz:
@tdent Line 36 was deleted.

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -33,6 +33,7 @@ def keep_ind(times, start, end):
     time_sorting = times.argsort()
     times = times[time_sorting]
     indices = numpy.array([], dtype=numpy.uint32)
+    left = numpy.searchsorted(times, start, side='left')
```
tdent: can you call 'left' and 'right' something else to avoid a context mixup again ..

ahnitz: Ok, I've renamed them in the function.

ahnitz:
@tdent Can you take another look and approve if this is OK?

PR #2658, opened 2019-04-24
add in tiled plots pycbc_fit_sngls_split_binned

GarethDaviesGW:
splitting triggers into templates within certain boundary values and then binning and fitting as performed in pycbc_fit_sngls_binned

### DIFF
--- a/bin/hdfcoinc/pycbc_fit_sngls_split_binned
+++ b/bin/hdfcoinc/pycbc_fit_sngls_split_binned
@@ -91,9 +91,6 @@ parser.add_argument("--fit-function",
                     help="Functional form for the maximum likelihood fit")
 args = parser.parse_args()
 
-# explain or remove
-args.approximant = None
-
 if args.verbose:
     log_level = logging.DEBUG
 else:
@@ -132,7 +129,7 @@ for ex_p in extparams:
 
 # string formats for labels, logging etc.
 formats = {
-        "mchirp": '{:d}',
+        "mchirp": '{:.2f}',
         "eta": '{:.3f}',
         "chi_eff": '{:.3f}',
         'template_duration': '{:.0f}' 
--- a/bin/hdfcoinc/pycbc_fit_sngls_split_binned
+++ b/bin/hdfcoinc/pycbc_fit_sngls_split_binned
@@ -268,6 +268,7 @@ for x in range(args.split_one_nbins):
             fitted_cum_counts = len(vals_above_thresh) * \
                                 trstats.cum_fit(args.fit_function, fitrange,
                                                 alpha, args.stat_fit_threshold)
+            # upper and lower 1-sigma bounds on fit are not currently plotted
             fitted_cum_counts_plus = len(vals_above_thresh) * \
                                       trstats.cum_fit(args.fit_function,
                                                       fitrange, alpha + sig_alpha,
--- a/bin/hdfcoinc/pycbc_fit_sngls_split_binned
+++ b/bin/hdfcoinc/pycbc_fit_sngls_split_binned
@@ -20,7 +20,7 @@ import argparse, logging
 from matplotlib import use
 use('Agg')
 from matplotlib import pyplot as plt
-import copy, numpy as np
+import numpy as np
 
 from pycbc import events, bin_utils, results
 from pycbc.events import triggers as trigs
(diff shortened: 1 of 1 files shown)


bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -0,0 +1,354 @@
...
+
+import sys, h5py
+import argparse, logging
+
+from matplotlib import use
+use('Agg')
+from matplotlib import pyplot as plt
+import copy, numpy as np
+from pycbc import conversions as conv
+from scipy.stats import kstest
```
tdent: are these two imports used? if not remove

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -0,0 +1,354 @@
...
+parser.add_argument('--split-one-log', action='store_true', default=False,
+                    help="is split-param-one split in log-space?. Default: False")
+parser.add_argument('--split-two-log', action='store_true', default=False,
+                    help="is split-param-two split in log-space?. Default: False")
+parser.add_argument("--sngl-stat", default="new_snr",
+                    choices=sngl_statistic_dict.keys(),
+                    help="Function of SNR and chisq to perform fits with")
+parser.add_argument('--ifo', type=str, help="Interferometer")
+parser.add_argument('--plot-max-x',default=1000,
+                    help="Maximum stat value for use in plots, if this "
```
tdent: this seems overcomplicated /backwards in terms of how it is interpreted by the code. 
would it not work to 1) if no plot-max-x is given (ie None), scale plots to 1.05* largest stat 
2) if plot-max-x is given, use that value?

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -0,0 +1,354 @@
...
+        "spin1z" : full_bank['spin1z'][:],
+        "spin2z" : full_bank['spin2z'][:]
+}
+full_bank.close()
+
+# only calculate params that need calculation if they are actually used
+usedparams = [args.split_param_two, args.split_param_one, args.bin_param]
+extparams = [par for par in ['mchirp', 'eta', 'chi_eff', 'template_duration'] if par in usedparams]
+
+logging.info('calculating extra parameters {} as required'.format(extparams))
```
tdent: think this is not needed, given that every ex_p will be logged in the following loop.

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -0,0 +1,354 @@
...
+for i in range(args.split_one_nbins):
+    axes[i, 0].set_ylabel(args.split_param_one + ': ' + 
+                          (formats[args.split_param_one] + ' to ' + 
+                           formats[args.split_param_one]).format(sp_one_bounds.lower()[x],
+                                      sp_one_bounds.upper()[x]) + '\ncumulative number',
+                                      size="large")
+
+axes[0,0].add_artist(lgd)
+
+fig.savefig(args.output_file, bbox_extra_artists=(lgd,), bbox_inches='tight')
```
tdent: can you look at the existing fit_sngls_binned code https://github.com/gwastro/pycbc/blob/master/bin/hdfcoinc/pycbc_fit_sngls_binned#L394 which uses results.save_fig_with_metadata and see if you can use that (useful for putting plot onto results page)?

GarethDaviesGW: Done - added in, and in place [here (LIGO login required)](https://ldas-jobs.ligo.caltech.edu/~gdavies/analyses/multiifodev/test_output/background_fits/20190503_065633_H1_a22_split_chieff_eta_triggerhist_newsnr_sgveto_maxdur4_vetoed_singleexp_fitting.png)

bin/hdfcoinc/pycbc_fit_sngls_split_binned
GarethDaviesGW: Agree - I think this was the simpler option when I was performing the testing for some reason. sorted now

bin/hdfcoinc/pycbc_fit_sngls_split_binned
GarethDaviesGW: Removed - these were a hangover from copying from fit_sngls_binned

PR #2660, opened 2019-04-27
reduce memory useage of fit by template

ahnitz:
I've checked this produces the same number of triggers, etc. Reduces the memory useage of fits_sngl_by_template by a little over an order of magnitude when reading many triggers. 

In my case, this reduced peak memory useage from 78 GB -> 4 GB.

### DIFF
--- a/bin/hdfcoinc/pycbc_fit_sngls_by_template
+++ b/bin/hdfcoinc/pycbc_fit_sngls_by_template
@@ -37,12 +37,18 @@ def get_stat(statchoice, trigs, threshold):
     s = 0
     while s < size:
         e = s + chunk_size if (s + chunk_size) <= size else size
+
+        # read and format chunk of data so it can be read by key
+        # as the stat classes expect.
         chunk = {k: trigs[k][s:e] for k in trigs if len(trigs[k]) == size}
         chunk_stat = stat_instance.single(chunk)
         above = chunk_stat >= threshold
         stat.append(chunk_stat[above])
         select.append(above)
         s += chunk_size
+
+    # Return boolean area that selects the triggers above threshold
+    # along with the stat values above threshold
     return np.concatenate(select), np.concatenate(stat)
 #### MAIN ####
 
--- a/bin/hdfcoinc/pycbc_fit_sngls_by_template
+++ b/bin/hdfcoinc/pycbc_fit_sngls_by_template
@@ -26,12 +26,24 @@ import pycbc.version
 
 #### DEFINITIONS AND FUNCTIONS ####
 
-def get_stat(statchoice, trigs):
+def get_stat(statchoice, trigs, threshold):
     # Initialize statclass with an empty file list. In general could feed it
     # files here for statistics which need that.
     stat_instance = sngl_statistic_dict[statchoice]([])
-    return stat_instance.single(trigs)
-
+    chunk_size = 2**23
+    stat = []
+    select = []
+    size = len(trigs['end_time'])
+    s = 0
+    while s < size:
+        e = s + chunk_size if (s + chunk_size) <= size else size
+        chunk = {k: trigs[k][s:e] for k in trigs if len(trigs[k]) == size}
+        chunk_stat = stat_instance.single(chunk)
+        above = chunk_stat >= threshold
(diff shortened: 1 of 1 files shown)


tdent:
For information, what was the size of the single trigger file & how many triggers?

ahnitz:
30 GB and 1.2 billion

ahnitz:
I think the memory useage should probably have scaled before and after linearly with those numbers.

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -133,26 +145,23 @@ tb_hashorder = tb[hash_sort]
 # reorder template IDs in parallel to the boundary values
 tid_hashorder = tid[hash_sort]
 
-# Calculate the differences between the boundary indices to get the 
+# Calculate the differences between the boundary indices to get the
 # number in each template
 # adding on total number at the end to get number in the last template
-total_number = len(trigf[args.ifo + '/template_id'][:])
+total_number = len(trigf[args.ifo + '/template_id'])
```
tdent: I found the following on h5py docs: 
> On 32-bit platforms, len(dataset) will fail if the first axis is bigger than 2**32. It’s recommended to use Dataset.len() for large datasets 
Not at 2**32 yet, but we can't rule it out ..

ahnitz: Doesn't really matter actually. We can't even build pycbc on a purely 32bit platform. We require vector instructions that have only existed since the over a decade move to 64 bits. I think we are safe here.

tdent:
Suppose effectively all or nearly all of the triggers were used for fitting (ie low threshold) would this still help memory usage?

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -26,12 +26,24 @@ import pycbc.version
...
+    s = 0
+    while s < size:
+        e = s + chunk_size if (s + chunk_size) <= size else size
+        chunk = {k: trigs[k][s:e] for k in trigs if len(trigs[k]) == size}
+        chunk_stat = stat_instance.single(chunk)
+        above = chunk_stat >= threshold
+        stat.append(chunk_stat[above])
+        select.append(above)
+        s += chunk_size
+    return np.concatenate(select), np.concatenate(stat)
```
tdent: add a comment on what 'select' will be?  (it looks like a list of booleans to me .. ?)

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -26,12 +26,24 @@ import pycbc.version
...
-    return stat_instance.single(trigs)
-
+    chunk_size = 2**23
+    stat = []
+    select = []
+    size = len(trigs['end_time'])
+    s = 0
+    while s < size:
+        e = s + chunk_size if (s + chunk_size) <= size else size
+        chunk = {k: trigs[k][s:e] for k in trigs if len(trigs[k]) == size}
```
tdent: comment on this line, what is it supposed to do - I guess set up a 'trigs'-like object containing all possible trigger properties that could be used to calculate statistic?

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -26,12 +26,24 @@ import pycbc.version
...
+        chunk_stat = stat_instance.single(chunk)
+        above = chunk_stat >= threshold
+        stat.append(chunk_stat[above])
+        select.append(above)
```
tdent: ah, here it looks like `select` will be a list of arrays of Booleans - correct?

ahnitz:
@tdent If you wanted to process all singles, then of course, the memory useage would blow up in the way you are using them. You'd need to switch to a model where you only read in one triggers information at a time.

bin/hdfcoinc/pycbc_fit_sngls_by_template
ahnitz: Yes

ahnitz:
@tdent Is this now ok to merge?

tdent:
yes, should be.

PR #2662, opened 2019-04-29
change coinc options to allow multiple decimation levels

ahnitz:
Make the coinc code more flexible in terms of decimation levels. This *changes* the options to findtrigs

The new decimation option is 
```
--loudest-keep-values 9:10 8.5:100 8.0:100 7.5:10
```

Decimation is applied from left to right, so this can be read as (decimate all triggers < 9 by 10x) then (decimate remaining triggers < 8.5 by another 100x) and so on...

### DIFF
--- a/bin/hdfcoinc/pycbc_coinc_findtrigs
+++ b/bin/hdfcoinc/pycbc_coinc_findtrigs
@@ -33,8 +33,9 @@ parser.add_argument("--timeslide-interval", type=float,
                          "disabled if the option is omitted.")
 parser.add_argument("--decimation-factor", type=int,
                     help="The factor to reduce the background trigger rate.")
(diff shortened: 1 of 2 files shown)


ahnitz:
I'll prepare a config file update to correspond to this if it gets preliminary approval.

ahnitz:
Current Master:
decimation-factor:50000
loudest-keep-value 8.5 (you can see a slight break here)
background samples: 



After Patch
loudest-keep-values = 9:10 8.5:30 8:30 7.5:30

ahnitz:
@spxiwh Corresponding change required in pycbc-config https://github.com/gwastro/pycbc-config/pull/53

tdent:
@ahnitz btw, the 'after' plot is now 'not found'

PR #2663, opened 2019-04-30
add coinc follow up script

ahnitz:


ahnitz:
May reappear in different form, but not with this PR.

PR #2667, opened 2019-05-02
Warning message if sampling_params is not provided

soumide1102:
Currently the lines
```
if not cp.has_section('sampling_params'):
            raise ValueError("no sampling_params section found in config file")
```
in `from_config` in `models/base.py` is not letting the user know that a `sampling_params` section has not been provided, ie. nothing is printed to the err file.

This PR fixes this so that a warning is printed if `sampling_params` is not being used, so that the user is aware of that.

### DIFF
--- a/pycbc/inference/models/base.py
+++ b/pycbc/inference/models/base.py
@@ -27,9 +27,8 @@
 
 import numpy
 import logging
-
 from abc import (ABCMeta, abstractmethod)
-
+from ConfigParser import NoSectionError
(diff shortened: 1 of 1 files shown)


ahnitz:
@soumide1102 Was the ValuError being eaten by something? That should have been printed to the screen, no?

soumide1102:
@ahnitz The ValueError was not being being printed. https://github.com/gwastro/pycbc/blob/master/pycbc/inference/models/base.py#L752-L753 was eating it up.

PR #2668, opened 2019-05-03
Evidence calculations module

stevereyes01:
This module introduces 6 methods for estimating the evidence of a Bayesian Inference model.

The first method is the Arithmetic Mean Estimator. It is the simplest evidence estimator, and is done simply by drawing samples from the prior and calculating the log likelihood. This estimator converges to the true evidence in the limit that you sample the entire prior space. The calculation is done for log likelihoods and returns a log evidence.

The second method is the Harmonic Mean Estimator. It samples from the posterior distribution and is the harmonic mean of the log likelihood. This is a really bad estimator that should never be used, https://radfordneal.wordpress.com/2008/08/17/the-harmonic-mean-of-the-likelihood-worst-monte-carlo-method-ever/  . I've included it here with that warning. The calculation is done for log likelihoods and returns a log evidence.

The next four methods are variations of thermodynamic integration methods and a steppingstone algorithm method.

The first is doing T.I via a trapezoid rule. It also returns an estimate of the MC variation of that trapezoid rule estimate (See Evans, Annis 2019 [https://link.springer.com/article/10.3758/s13428-018-1172-y](url) eq. 36 and 37). Returns the log evidence and 1 standard deviation of the log evidence from Monte Carlo error.

Next method is doing T.I via trapezoid rule with the analytic gradient correction term added in to improve the estimate (See Friel et al 2014 [https://link.springer.com/article/10.1007%2Fs11222-013-9397-1](url) ). Returns the log evidence with the 1 standard deviation of the log evidence from Monte Carlo error from above. This 1 sigma error bar seems to match the previous method since we're just adding a constant term to the evidence and the variance is unaffected by adding a constant term.

Next method is doing T.I. via Simpson's rule from SciPy. Oddly enough there exist routines for doing Simpson's rule for unequally spaced data, and SciPy implements this. This routine seems to slightly outperform T.I with next order integration added in. Returns the log-evidence and an MC variation estimate.

Last method is using the Steppingstone algorithm, which instead of direct integration, uses importance sampling to do the integration for multiple temperatures. See Xie et al 2012 [https://www.ncbi.nlm.nih.gov/pmc/articles/PMC3038348/](url).

Lastly are just some simple convenience functions for calculating Bayes factors. There's a bit more that needs to be ported in, but this is a good enough start for people to start calculating evidences.

-------
I've tested these functions on a few analytic distributions as well as many actual GW events. They seem stable and the values they return make sense to me.

The module could be used as follows:
**PyCBC v1.12 series**
```
import h5py
from pycbc.inference import evidence

f = h5py.File("my_thinned_run.hdf", "r")
beta = f.attrs["betas"]
logl = f["likelihood_stats/loglr"][:] + f.attrs["lognl"]
f.close()

log_z_trapz_TI, dlog_z_trapz_TI = evidence.trapezoid_TI(logl, betas)
....
```

**PyCBC v1.13 series**
```
import h5py
from pycbc.inference import evidence

f = h5py.File("my_thinned_run.hdf", "r")
betas = f["sampler_info"].attrs["betas"]
logl = f["samples/loglikelihood"][:]
f.close()

log_z_trapz_TI, dlog_z_trapz_TI = evidence.trapezoid_TI(logl, betas)
...
```
If the data file isn't a thinned data file, one can just read in the log likelihoods with the appropriate settings.

I can provide plots and more information if necessary. It would be nice to have these routines completely replace the evidence calculation in PyCBC as it uses (I believe) T.I with Riemann sums, which is very inaccurate.

### DIFF
--- a/pycbc/inference/evidence.py
+++ b/pycbc/inference/evidence.py
@@ -112,10 +112,10 @@ def thermodynamic_integration(log_likelihood, betas,
 
     Returns
     -------
-    log_evidence   : float
+    log_evidence : float
         Estimation of the log of the evidence.
 
-    mcmc_std       : float
+    mcmc_std : float
         The standard deviation of the log evidence estimate from
         Monte-Carlo spread.
     """
--- a/pycbc/inference/evidence.py
+++ b/pycbc/inference/evidence.py
@@ -1,0 +1,252 @@
+# Copyright (C) 2019 Steven Reyes
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 3 of the License, or (at your
+# option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+This modules provides functions for estimating the marginal
+likelihood or evidence of a model.
+"""
+import numpy
+from scipy import integrate
+
+
+def arithmetic_mean_estimator(log_likelihood):
+    """Returns the log evidence via the prior arithmetic mean estimator (AME).
+
+    The logarithm form of AME is used. This is the most basic
+    evidence estimator, and often requires O(billions) of samples
+    from the prior.
+
+    Parameters
+    ----------
+    log_likelihood : 1d array of floats
+        The log likelihood of the data sampled from the prior
+        distribution.
+
+    Returns
+    -------
+    float :
+        Estimation of the log of the evidence.
+    """
+    num_samples = len(log_likelihood)
+    logl_max = numpy.max(log_likelihood)
+
+    log_evidence = 0.
+    for i, _ in enumerate(log_likelihood):
+        log_evidence += numpy.exp(log_likelihood[i] - logl_max)
+
+    log_evidence = numpy.log(log_evidence)
+    log_evidence += logl_max - numpy.log(num_samples)
+
+    return log_evidence
+
+
+def harmonic_mean_estimator(log_likelihood):
+    """Returns the log evidence via posterior harmonic mean estimator (HME).
+
+    The logarithm form of HME is used. This method is not
+    recommended for general use. It is very slow to converge,
+    formally, has infinite variance, and very error prone.
+
+    Not recommended for general use.
+
+    Parameters
+    ----------
+    log_likelihood : 1d array of floats
+        The log likelihood of the data sampled from the posterior
+        distribution.
+
+    Returns
+    -------
+    float :
+        Estimation of the log of the evidence.
+    """
+    num_samples = len(log_likelihood)
+    logl_max = numpy.max(-1.0*log_likelihood)
+
+    log_evidence = 0.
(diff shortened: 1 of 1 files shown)


pycbc/inference/evidence.py
```diff
@@ -0,0 +1,465 @@
...
+    # Eqs  (36) and (37)
+    for i in range(len(log_likelihood[0])):
+        log_likelihood_samps = [log_likelihood[x][i] for x in range(len(betas))]
+        TI_vec[i] = numpy.trapz(log_likelihood_samps, betas)
+
+    mcmc_std = numpy.std(TI_vec) / numpy.sqrt(float(len(log_likelihood[0])))
+
+    return log_evidence, mcmc_std
+
+def improved_trapezoid_TI(log_likelihood, betas):
```
ahnitz: Anything more descriptive than "improved"?

pycbc/inference/evidence.py
```diff
@@ -0,0 +1,465 @@
...
+        val_2 = numpy.sum(pre_fact ** 2)
+
+        mcmc_std += val_2
+
+    mcmc_std /= float(len(log_likelihood[0])) ** 2.0
+    mcmc_std = numpy.sqrt(mcmc_std)
+
+    return log_evidence, mcmc_std
+
+def calculate_log_bayes_factor(log_ev_1, log_ev_2):
```
ahnitz: Does this need it's own function to do a simple difference?

ahnitz:
@stevereyes01 Probably should address most of the codeclimate issues.

pycbc/inference/evidence.py
```diff
@@ -0,0 +1,465 @@
...
+
+    Returns
+    ----------
+    bayes_factor : float
+               The Bayes factor for model 1 vs model 2.
+    """
+    return numpy.exp(calculate_log_bayes_factor(log_ev_1,
+                                                log_ev_2))
+
+def sim_log_bayes_factor_distr(log_ev_1, std_log_ev_1,
```
duncan-brown: It might makes sense to put these functions in a separate module, as they are really test functions.

ahnitz: If they are really test functions, would it make sense to put them into a unit test?

pycbc/inference/evidence.py
stevereyes01: Yeah, I'll fix that.

higher_order_trapezoid maybe? https://en.wikipedia.org/wiki/Trapezoidal_rule#Error_analysis

It's adding in that constant term.

pycbc/inference/evidence.py
stevereyes01: Probably not. I did want to have some bayes factor convenience functions though. I can push them to another module.

pycbc/inference/evidence.py
stevereyes01: I've removed these functions. Not sure what's meant by them being test functions. They're just conversions from log B to B.

stevereyes01:
> Thanks for doing this @stevereyes01! I have a few requests related to the doc string formatting; see my comments. Also, are the errors that are returned `dZ` or `d logZ`? Make that clear in the doc strings.
> 
> Most of the remaining code climate issues can be ignored. However, please address the ones about more than 79 characters in a line.
> 
> It looks like the Travis failure was an unrelated issues. Hopefully it will resolve itself when you push a new commit.

The errors are in `d logz`. I'll fix the doc strings. A lot of the extra returns was me following codeclimates complaints.

stevereyes01:
I think the `codeclimate` has been reduced to what I can control without rewriting the code.

stevereyes01:
Okay, I think this has now been fixed.

cdcapano:
I think @stevereyes01 has addressed @duncan-brown 's review requests, so I'm merging.

PR #2670, opened 2019-05-03
add function to return 'filter-ready' psd, small cleanup of estimate method

ahnitz:


ahnitz:
I'm assigning this to @spxiwh as I think Tito won't be able to look at this soon.

PR #2672, opened 2019-05-03
Live: add ability to output background

titodalcanton:
Allows us to periodically save the coinc background and hence make plots like this:


titodalcanton:
I will have to test that more with the full bank, so I can hold off merging until that is done. In any case I am thinking of outputting snapshots every hour or so, maybe every 5 hours.

ahnitz:
This probably just takes a second or two would be my guess, so that's probably reasonable.

PR #2674, opened 2019-05-04
stop hiding output, remove lines I think we aren't using

ahnitz:


ahnitz:
@duncan-brown Please take a look at this when you canf you spot a problem, we can try to address, but in the meantime, I'm merging this to get other PRs moving again. 

This should just be a simplification (and removal of no longer used components). The major change is that lalsuite is not longer installed *twice*. It was manually installed here once and then installed again through pip. I have no idea if what effect that has, but it's probably asking for trouble. This PR resolves that issue. I

PR #2676, opened 2019-05-06
don't assume args.approximant exists

tdent:


pycbc/events/triggers.py
```diff
@@ -283,6 +283,8 @@ def get_param(par, args, m1, m2, s1z, s2z):
     elif par in ['chi_eff', 'effective_spin']:
         parvals = conversions.chi_eff(m1, m2, s1z, s2z)
     elif par == 'template_duration':
+        if not hasattr(args, 'approximant'):
+            args.approximant = None
```
GarethDaviesGW: should this not be defaulted to "SEOBNRv4" - as the next line ignores any None values anyway?

tdent: I don't want to have 'SEOBNRv4' hard-coded as default in two places. Can reorganize a bit to achieve that.

tdent:
will ignore travis timeouts ...

PR #2679, opened 2019-05-07
Fix bug spotted by Tom pycbc_multiifo_coinc_findtrigs

ahnitz:


tdent:
I suppose I should ask for this to be tested (ie checked at least that it doesn't crash) in a 3-ifo case .. if you have some such files handy

tdent:
@ahnitz what about `sds_full` vs `sds` - same issue?

tdent:
BTW, it turns out that `int(factor)` already throws an error if 'factor' is not the string representation of an integer.  Even `int('100.0')` is an error.  So will remove the assert and just write a comment on this

tdent:
@ahnitz it looks like you even forgot to increment start0 and start1 in the batching.  so the code never exits atm ..

ahnitz:
@tdent Ok, hold on a second, something seems wrong here. I feel like there is a commit or two missing...

ahnitz:
@tdent Well damn, that appears to be what happened. I did write a commit that had all these fixes already. https://github.com/gwastro/pycbc/commit/bc2ef75d1373f3c47c195dab49eaab702de7e53d. 

Somehow I didn't push this and it didn't end up getting merged.

PR #2681, opened 2019-05-07
fixes to multi-ifo version

ahnitz:
This was the commit that apparently I forgot to merge. Somewhat critical for the multi-ifo code to actually work!

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
@@ -232,27 +232,25 @@ for tnum in template_ids:
         if len(tids[ifo] == 0):
             continue
 
-    times_full={}
-    for ifo in trigs:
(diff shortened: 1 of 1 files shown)


bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
```diff
@@ -341,6 +338,10 @@ for tnum in template_ids:
             data['decimation_factor'] += [dec_fac]
             data['timeslide_id'] += [slide[ti]]
             data['template_id'] += [numpy.zeros(len(ti), dtype=numpy.uint32) + tnum]
+           
+            logging.info('{}-{}'.format(start0, start1))
```
tdent: do we want to keep this logging call?  seems like it would be very verbose.

tdent:
oh well, and I had the illusion i had been doing something useful for the last hour and a half

PR #2686, opened 2019-05-09
PyCBC Inference, TI integrand, "Gooseneck" plots

stevereyes01:
This executable makes plots of the thermodynamic integrand (TI). Which is the integrand of the following expression:

image

It is always helpful to examine the TI integrand to look for systematic error (undersampling) of the temperature ladder in TI. Good posteriors at `$\beta = 1$` are **not** representative of a converged integral or of a well-formed temperature ladder. See ```https://github.com/dfm/emcee/pull/154/files``` for more information on how and why to plot TI integrands in the following methods.

The options in this executable are:
```
$ pycbc_inference_plot_thermodynamic_integrand --help
usage: pycbc_inference_plot_thermodynamic_integrand [-h]
                                                    [--inference-file INFERENCE_FILE]
                                                    [--thin-start THIN_START]
                                                    [--thin-end THIN_END]
                                                    [--thin-interval THIN_INTERVAL]
                                                    [--beta-log-scale]
                                                    [--integrand-logarithmic]
                                                    [--output-file OUTPUT_FILE]

optional arguments:
  -h, --help            show this help message and exit
  --inference-file INFERENCE_FILE
                        The PyCBC multi-temper inference file.
  --thin-start THIN_START
                        MCMC iteration to begin drawing samples from.
  --thin-end THIN_END   MCMC iteration to end drawing samples from.
  --thin-interval THIN_INTERVAL
                        MCMC intervals to downselect samples between thin-start
                        and thin-end.
  --beta-log-scale      Put inverse-temperature axis in logarithm base 10
                        scale.
  --integrand-logarithmic
                        Plot the integrand as integrated in log beta.
  --output-file OUTPUT_FILE
```

I give the following example from running TI on a toy problem involving a likelihood function that is a 1-d gaussian (`test_normal`), with a prior 1-d that is also a Gaussian of same parameters. The means are 0, and the standard deviations are 1 for these Gaussians. For this pair, one can show analytically that the evidence is, for many samples, (`1/ sqrt(4*pi)`,  `see eq. 2 of  Xie et al. 2012 in Systematic Biology`).

```pycbc_inference_plot_thermodynamic_integrand --inference-file my_inference.hdf --thin-start 0 --thin-end -1 --thin-interval 1  --output-file logl_v_beta_lin_scale.png``` yields:

image

The "goose-neck" plot can be made by using `--beta-log-scale`:
```pycbc_inference_plot_thermodynamic_integrand --inference-file my_inference.hdf --thin-start 0 --thin-end -1 --thin-interval 1 --beta-log-scale  --output-file logl_v_beta_lin_scale.png``` yields:

image

The `logarithmic-integrand` plot, where the integrand is done in `d log beta` instead of `beta` can be done via:
```
pycbc_inference_plot_thermodynamic_integrand --inference-file my_inference.hdf --thin-start 0 --thin-end -1 --thin-interval 1 --integrand-logarithmic  --output-file beta_logl_v_log_beta_log_scale.png```
```

image

I have omitted the option to plot the `logarithmic-integrand` plot on a linear scale with `beta`, since I haven't really ever used it before, and it doesn't strike me as that useful to inspect given the other 3 plots. Here is what that would look like though:

image

-----

A good TI integrand will be smooth and monotonic, and without jagged edges [See Annis et al 2019 in Journal of Mathematical Psychology]. The entire idea of doing a good job with TI integration is limiting the systematic bias/error in numerical integration of the curve. Current numerical integration methods are through importance sampling (stepping stone); or trapezoid rule and Simpson's rule (at the moment). You have to think carefully about where these rules fail relative to the concavity of the TI integrand. The TI integrand looks different relative to different data sets, so making this plot is **mandatory**. Can also use the plot to place additional temperatures to mitigate not-enough-temperatures error.

-----

It is also sometimes useful to inspect the cumulative thermodynamic integral plot.


stevereyes01:
Just for those who are looking back, the reference on the toy problem from above is actually the wrong link.

http://www.tina-vision.net/docs/memos/2003-003.pdf

Gives the product of two Gaussians, and then you can integrate that product to get the two answers.

PR #2688, opened 2019-05-10
delete uniform component masses distribution

dfinstad:
This deletes the uniform component masses distribution since it is unnecessary.


dfinstad:
@cdcapano I found and fixed the offending config file, so this is ready to go pending your approval.

PR #2689, opened 2019-05-10
Multiifo ranking statistic

GarethDaviesGW:
Multiifo ranking statistic:

Currently calculates the expected rate of coincidences at a given snr value, and subtracts this from a mean coincidence rate (this is a rate for _all_ coincidences, not including the snr infromation) in that detector combination

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
@@ -32,12 +32,11 @@ parser.add_argument("--use-maxalpha", action="store_true")
 parser.add_argument("--coinc-threshold", type=float, default=0.0,
                     help="Seconds to add to time-of-flight coincidence window")
 parser.add_argument("--timeslide-interval", type=float,
-                    help="Interval between timeslides in seconds. Timeslides are"
-                         " disabled if the option is omitted.")
-parser.add_argument("--loudest-keep-values", type=str, nargs='*', default='6:1',
-                    help="Apply successive multiplicative levels of decimation"
-                         " to coincs with stat value below the given thresholds"
-                         " Ex. 9:10 8.5:30 8:30 7.5:30. Default: no decimation")
+                    help="Interval between timeslides in seconds. Timeslides are "
+                         "disabled if the option is omitted.")
+parser.add_argument("--loudest-keep-values", type=str, nargs='*',
+                     help="Apply a sequence of decimations at the given thresholds"
+                          "Ex. 8.5:10 8:10 7.5:10 7:10")
 parser.add_argument("--template-fraction-range", default="0/1",
                     help="Optional, analyze only part of template bank. Format"
                          " PART/NUM_PARTS")
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
@@ -32,11 +32,12 @@ parser.add_argument("--use-maxalpha", action="store_true")
 parser.add_argument("--coinc-threshold", type=float, default=0.0,
                     help="Seconds to add to time-of-flight coincidence window")
 parser.add_argument("--timeslide-interval", type=float,
-                    help="Interval between timeslides in seconds. Timeslides are "
-                         "disabled if the option is omitted.")
-parser.add_argument("--loudest-keep-values", type=str, nargs='*',
-                     help="Apply a sequence of decimations at the given thresholds"
-                          "Ex. 8.5:10 8:10 7.5:10 7:10")
+                    help="Interval between timeslides in seconds. Timeslides are"
+                         " disabled if the option is omitted.")
+parser.add_argument("--loudest-keep-values", type=str, nargs='*', default='6:1',
+                    help="Apply successive multiplicative levels of decimation"
+                         " to coincs with stat value below the given thresholds"
+                         " Ex. 9:10 8.5:30 8:30 7.5:30. Default: no decimation")
 parser.add_argument("--template-fraction-range", default="0/1",
                     help="Optional, analyze only part of template bank. Format"
                          " PART/NUM_PARTS")
@@ -50,7 +51,6 @@ parser.add_argument("--output-file",
                     help="File to store the coincident triggers")
 parser.add_argument("--batch-singles", default=5000, type=int,
                     help="Number of single triggers to process at once")
-
 args = parser.parse_args()
 
 # flatten the list of lists of filenames to a single list (may be empty)
@@ -267,7 +267,7 @@ for tnum in template_ids:
             logging.info('Calculating Multi-Detector Combined Statistic')
             sdswithids = {}
             for ifo in sds:
-                sdswithids[ifo]=sds[ifo][ids[ifo]]
+                sdswithids[ifo] = sds[ifo][ids[ifo]]
             cstat = rank_method.coinc_multiifo(sdswithids, slide,
                                            args.timeslide_interval,
                                            time_addition=args.coinc_threshold)
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
@@ -264,12 +264,11 @@ for tnum in template_ids:
                                                       args.pivot_ifo,
                                                       args.fixed_ifo)
             logging.info('Coincident Trigs: %s' % (len(ids[args.pivot_ifo])))
-            print(len(ids[args.pivot_ifo]))
             logging.info('Calculating Multi-Detector Combined Statistic')
             sdswithids = {}
             for ifo in sds:
                 sdswithids[ifo]=sds[ifo][ids[ifo]]
-            c = rank_method.coinc_multiifo(sdswithids, slide,
+            cstat = rank_method.coinc_multiifo(sdswithids, slide,
                                            args.timeslide_interval,
                                            time_addition=args.coinc_threshold)
 
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
@@ -50,6 +50,7 @@ parser.add_argument("--output-file",
                     help="File to store the coincident triggers")
 parser.add_argument("--batch-singles", default=5000, type=int,
                     help="Number of single triggers to process at once")
+
 args = parser.parse_args()
 
 # flatten the list of lists of filenames to a single list (may be empty)
@@ -263,13 +264,14 @@ for tnum in template_ids:
                                                       args.pivot_ifo,
                                                       args.fixed_ifo)
             logging.info('Coincident Trigs: %s' % (len(ids[args.pivot_ifo])))
-
+            print(len(ids[args.pivot_ifo]))
             logging.info('Calculating Multi-Detector Combined Statistic')
             sdswithids = {}
             for ifo in sds:
-                sdswithids[ifo] = sds[ifo][ids[ifo]]
-            cstat = rank_method.coinc_multiifo(sdswithids, slide,
-                                               args.timeslide_interval)
+                sdswithids[ifo]=sds[ifo][ids[ifo]]
+            c = rank_method.coinc_multiifo(sdswithids, slide,
+                                           args.timeslide_interval,
+                                           time_addition=args.coinc_threshold)
 
             # index values of the zerolag triggers
             fi = numpy.where(slide == 0)[0]
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
@@ -86,7 +86,7 @@ class ReadByTemplate(object):
         self.segs = veto.start_end_to_segments(s, e).coalesce()
         for vfile, name in zip(veto_files, segment_name):
             veto_segs = veto.select_segments_by_definer(vfile, ifo=self.ifo,
-                                                     segment_name=name)
+                                                        segment_name=name)
             self.segs = (self.segs - veto_segs).coalesce()
         self.valid = veto.segments_to_start_end(self.segs)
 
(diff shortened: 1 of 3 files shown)


pycbc/events/stat.py
```diff
@@ -567,6 +568,109 @@ def single(self, trigs):
...
+                  step, time_addition=0.005): # pylint:disable=unused-argument
+        """Calculate the final coinc ranking statistic"""
+        ifostr = ' '.join(sorted(self.ifos))
+
+        ratesprod = sum(x for x in s.values())
+        ifo_list = [ifo for ifo in self.fits_by_tid]
+        ln_coinc_area = numpy.log(coinc_rate.multiifo_noise_coincident_area(
+                                                     ifo_list, time_addition))
+
+        loglr = mean_log_coincidence_rates[ifostr] - ln_coinc_area - ratesprod
```
tdent: @GarethDaviesGW why are you normalizing by an ifo-dependent mean rate?

tdent: The statistic we want is the rate of noise coincidences (as a density over stat value). The rate of noise coincidences is dependent on the ifo combination. That is the point of calculating the rates for each ifo combinations separately, they *are* different. I suggested scaling out by a 'normal/standard' H1L1 rate simply in order to make the numbers manageable.

pycbc/events/stat.py
```diff
@@ -567,6 +568,109 @@ def single(self, trigs):
...
+        self.ifos = [at[0] for at in parsed_attrs if
+                     (len(at) == 2 and at[1] == 'fit_coeffs')]
+        if not len(self.ifos):
+            raise RuntimeError("None of the statistic files has the required "
+                               "attribute called {ifo}-fit_coeffs !")
+        self.fits_by_tid = {}
+        self.alphamax = {}
+        for i in self.ifos:
+            self.fits_by_tid[i] = self.assign_fits(i)
+            self.get_ref_vals(i)
```
tdent: Is this block any different from what is done in the existing expfit stat?

pycbc/events/stat.py
```diff
@@ -567,6 +568,109 @@ def single(self, trigs):
...
+        coeff_file = self.files[ifo+'-fit_coeffs']
+        template_id = coeff_file['template_id'][:]
+        alphas = coeff_file['fit_coeff'][:]
+        rates = coeff_file['count_in_template'][:]/\
+                float(coeff_file.attrs['analysis_time'])
+        # the template_ids and fit coeffs are stored in an arbitrary order
+        # create new arrays in template_id order for easier recall
+        tid_sort = numpy.argsort(template_id)
+        return {'alpha':alphas[tid_sort], 'rate':rates[tid_sort],
+                'thresh':coeff_file.attrs['stat_threshold']}
```
tdent: Is this block any different from the existing exp fit stat?

pycbc/events/stat.py
```diff
@@ -567,6 +568,109 @@ def single(self, trigs):
...
+            tnum = trigs['template_id']  # exists for SingleDetTriggers
+            # Should only be one ifo fit file provided
+            assert len(self.ifos) == 1
+            ifo = self.ifos[0]
+        # fits_by_tid is a dictionary of dictionaries of arrays
+        # indexed by ifo / coefficient name / template_id
+        alphai = self.fits_by_tid[ifo]['alpha'][tnum]
+        ratei = self.fits_by_tid[ifo]['rate'][tnum]
+        thresh = self.fits_by_tid[ifo]['thresh']
+        return alphai, ratei, thresh
```
tdent: Is this block different from the existing exp fit stat?

pycbc/events/stat.py
```diff
@@ -567,6 +568,109 @@ def single(self, trigs):
...
+                                - alphai * (newsnr - thresh),
+                                ndmin=1, dtype=numpy.float32)
+        return lognoisel
+
+    def single(self, trigs):
+        """
+        Single-detector statistic, here just equal to the log noise
+        rate density
+        """
+        return self.lognoiseratedensity(trigs)
```
tdent: Are these `lognoiseratedensity` and `single` methods different from the existing exp fit stat?

tdent:
@GarethDaviesGW in general, you should not be repeating code blocks from existing classes. You can inherit their methods by making the new class a subclass/child class of them, or (less preferable) by explicitly calling a method of an existing class.

pycbc/events/stat.py
```diff
@@ -567,6 +568,109 @@ def single(self, trigs):
...
+        if not len(self.ifos):
+            raise RuntimeError("None of the statistic files has the required "
+                               "attribute called {ifo}-fit_coeffs !")
+        self.fits_by_tid = {}
+        self.alphamax = {}
+        for i in self.ifos:
+            self.fits_by_tid[i] = self.assign_fits(i)
+            self.get_ref_vals(i)
+
+        self.get_newsnr = ranking.get_newsnr
```
tdent: as far as I can see this will end up using newsnr as the single-ifo stat, not newsnr_sgveto?

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
```diff
@@ -32,12 +32,11 @@ parser.add_argument("--use-maxalpha", action="store_true")
 parser.add_argument("--coinc-threshold", type=float, default=0.0,
                     help="Seconds to add to time-of-flight coincidence window")
 parser.add_argument("--timeslide-interval", type=float,
-                    help="Interval between timeslides in seconds. Timeslides are"
-                         " disabled if the option is omitted.")
-parser.add_argument("--loudest-keep-values", type=str, nargs='*', default='6:1',
-                    help="Apply successive multiplicative levels of decimation"
-                         " to coincs with stat value below the given thresholds"
-                         " Ex. 9:10 8.5:30 8:30 7.5:30. Default: no decimation")
+                    help="Interval between timeslides in seconds. Timeslides are "
```
tdent: This seems to be a revert of a change now on master. Can you rebase to upstream/master and push -f

pycbc/events/coinc_rate.py
```diff
@@ -62,6 +61,38 @@ def multiifo_noise_coinc_rate(rates, slop):
...
+        for timing error
+
+    Returns
+    -------
+    combo_coinc_rate: numpy array
+        Value is expected coincidence rate in the combination, units Hz
+    """
+    # extract ifo list and rates vectors from the dictionary
+    ifos = numpy.array(sorted(rates.keys()))
+    rates_raw = list(rates[ifo] for ifo in ifos)
```
tdent: isn't `[rates[i] for i in ifos]` simpler?  (being already a list)

pycbc/events/coinc_rate.py
```diff
@@ -62,6 +61,38 @@ def multiifo_noise_coinc_rate(rates, slop):
...
+        time added to maximum time-of-flight between detectors to account
+        for timing error
+
+    Returns
+    -------
+    combo_coinc_rate: numpy array
+        Value is expected coincidence rate in the combination, units Hz
+    """
+    # extract ifo list and rates vectors from the dictionary
+    ifos = numpy.array(sorted(rates.keys()))
```
tdent: i can't work out why this needs to be sorted. Presumably the calculation works OK if `ifos` were an unsorted array of ifo names?

pycbc/events/stat.py
```diff
@@ -567,6 +569,43 @@ def single(self, trigs):
...
+        tid_sort = numpy.argsort(template_id)
+        self.fits_by_tid[ifo]['rate'] = \
+            coeff_file['count_in_template'][:][tid_sort] / \
+            float(coeff_file.attrs['analysis_time'])
+
+    def coinc_multiifo(self, s, slide,
+                       step, **kwargs): # pylint:disable=unused-argument
+        """Calculate the final coinc ranking statistic"""
+        snglsprod = sum(x for x in s.values())
+        ifo_list = [ifo for ifo in self.fits_by_tid]
```
tdent: will `self.fits_by_tid.keys()` work?

pycbc/events/stat.py
```diff
@@ -567,6 +569,43 @@ def single(self, trigs):
...
+        ln_coinc_area = numpy.log(coinc_rate.multiifo_noise_coincident_area(
+                                  ifo_list, kwargs['time_addition']))
+        loglr = - ln_coinc_area + snglsprod - self.benchmark_lograte
```
tdent: can this use the `combination_noise_coinc_rate()` function in fact?

pycbc/events/stat.py
```diff
@@ -567,6 +569,43 @@ def single(self, trigs):
...
+    template over single-ifo newsnr values.
+    """
+
+    def __init__(self, files, benchmark_lograte=-14.6):
+        # benchmark_lograte is log of a representative noise trigger rate
+        # This comes from H1L1 (O2) and is 4.5e-7 Hz
+        super(CoincRateCalcStatistic, self).__init__(files)
+        self.benchmark_lograte = 0  # benchmark_lograte
+
+        for ifo in self.ifos:
```
tdent: add comment explaining why this is done

pycbc/events/stat.py
```diff
@@ -567,6 +569,43 @@ def single(self, trigs):
                            dtype=numpy.float32), ndmin=1, copy=False)
 
 
+class CoincRateCalcStatistic(ExpFitStatistic):
```
tdent: The name could use some work. Eg `ExpFitCoincRateStatistic` ?  Since it's based on expfit ..

Also, by making ExpFitStatistic the parent, the basic single-ifo SNR-like statistic that is being used is newsnr, not newsnr_sgveto.  I thought we wanted to do the latter .. so, make an init step that ensures this, or maybe make this a child of ExpFitSGCombinedSNR which does that already.  In that case the name should be eg `ExpFitSGCoincRateStatistic`.

pycbc/events/stat.py
```diff
@@ -582,7 +621,8 @@ def single(self, trigs):
     'phasetd_exp_fit_stat_sgveto': PhaseTDExpFitSGStatistic,
     'newsnr_sgveto': NewSNRSGStatistic,
     'newsnr_sgveto_psdvar': NewSNRSGPSDStatistic,
-    'phasetd_exp_fit_stat_sgveto_psdvar': PhaseTDExpFitSGPSDStatistic
+    'phasetd_exp_fit_stat_sgveto_psdvar': PhaseTDExpFitSGPSDStatistic,
+    'coinc_rate_calc': CoincRateCalcStatistic
```
tdent: revise name here as well - eg `exp_fit_coinc_rate`

GarethDaviesGW:
Comparison scatter plot of the exp_fit_sg_coinc_rate (x-axis) vs exp_fit_sg_csnr statistics in L1V1 O2 analysis 20, coloured by mchirp

image

Histograms of each of the statistics are similar:
exp_fit_sg_csnr:
image

exp_fit_sg_coinc_rate
image

Other detector combinations are similar, e.g. H1L1V1:
image

pycbc/events/stat.py
GarethDaviesGW: The `ExpFitStatistic` has the single function that we want, so have added the init step.

Did we want to use the `newsnr_sgveto_psdvar`?

tdent: No. Newsnr_sgveto is the basic statistic we used in O2 and is the current basis for comparison & not decided to replace it yet.

GarethDaviesGW:
Scatter vs exp_fit_sg_csnr:
image

histograms of the two statistics:
image

pycbc/events/coinc_rate.py
```diff
@@ -34,20 +35,18 @@ def multiifo_noise_coinc_rate(rates, slop):
...
-    allowed_area = multiifo_noise_coincident_area(ifos, slop)
-    rateprod = [numpy.prod(rs) for rs in zip(*rates_raw)]
     ifostring = ' '.join(ifos)
-    expected_coinc_rates[ifostring] = allowed_area * numpy.array(rateprod)
+    expected_coinc_rates[ifostring] = combination_noise_coinc_rate(rates, slop)
 
     # if more than one possible coincidence type exists,
     # calculate coincidence for subsets through recursion
     if len(ifos) > 2:
+        ifos = numpy.array(sorted(rates.keys()))
```
tdent: any reason why they need to be sorted?

pycbc/events/stat.py
```diff
@@ -567,6 +569,46 @@ def single(self, trigs):
...
+    template over single-ifo newsnr values.
+    """
+
+    def __init__(self, files, benchmark_lograte=-14.6):
+        # benchmark_lograte is log of a representative noise trigger rate
+        # This comes from H1L1 (O2) and is 4.5e-7 Hz
+        super(ExpFitSGCoincRateStatistic, self).__init__(files)
+        self.benchmark_lograte = benchmark_lograte
+        self.get_newsnr = ranking.get_newsnr_sgveto
+        # Reassign the rate as the we are now calculating total number in
```
tdent: rewrite comment - 'as the we' & I would just say that 'rate' data is now number per time, rather than an arbitrarily normalized number, don't say anything about total number or threshold here

pycbc/events/coinc_rate.py
```diff
@@ -62,6 +61,38 @@ def multiifo_noise_coinc_rate(rates, slop):
...
+        time added to maximum time-of-flight between detectors to account
+        for timing error
+
+    Returns
+    -------
+    combo_coinc_rate: numpy array
+        Value is expected coincidence rate in the combination, units Hz
+    """
+    # extract ifo list and rates vectors from the dictionary
+    ifos = numpy.array(rates.keys())
```
tdent: does ifos have to be a numpy array? the noise coincident area function says it takes a list of ifos. 

similarly, does it work to just use rates.values() for the rate vectors?

pycbc/events/stat.py
GarethDaviesGW: Comment updated

pycbc/events/coinc_rate.py
GarethDaviesGW: You're right, `ifos` doesn't have to be a numpy array

Aslo, `rates.values()` works and I have updated

Testing this on my laptop rather than the cluster (by accident) has shown that the behaviour of `dictionary.keys()` and `dictionary.values()` will change in python 3; in python 3 they give specific `dict_keys` and `dict_values` objects (tagging @spxiwh in this as he has been working on python 3 updates)

As such, for future-proofing, I have included a `list()` wrapper around the `.keys()` call, which does nothing in python 2 but changes them back to being a list in python 3
The `dictionary.values()` being a `dict_values` object rather than list doesn't matter in this case as it is being looped over and acts in exactly the same way

I have done the same on line 39 to convert the keys to a list

pycbc/events/coinc_rate.py
GarethDaviesGW: This line is not actually needed, as it is just repeating line 39

In that case, it _is_ needed to ensure consistency for the ordering of the `ifostring` dict key. Otherwise I think the ordering is not stable

The conversion to a numpy array however is not needed and is removed

pycbc/events/coinc_rate.py
tdent: OK. It looks like `list(mydict)` would be the same as `list(mydict.keys())` so that's one less pair of brackets.  I'll edit this in the browser and then I think we're finished for this iteration

tdent:
will merge assuming that 'final' tests work out.

PR #2693, opened 2019-05-11
Use naming convention params instead of args

dfinstad:
This changes to the convention of using `static_params` and `variable_params` in several places: the executables `pycbc_create_injections` and `pycbc_make_inference_inj_workflow`, and the `make_inference_summary_table` function in `pycbc.workflow.inference_followups`.

Note that `pycbc_make_inference_workflow` was already calling `make_inference_summary_table` with a `variable_params` arg.

### DIFF
--- a/bin/pycbc_create_injections
+++ b/bin/pycbc_create_injections
@@ -20,7 +20,7 @@
 
 Config file syntax
 ------------------
-The configuration file should have a [variable_args], a [static_args],
+The configuration file should have a [variable_params], a [static_params],
 and one or more [distribution] sections. Multiple [constraint] sections and
 one or more [waveform_transforms] sections may also be provided. An example:
 
--- a/bin/pycbc_create_injections
+++ b/bin/pycbc_create_injections
@@ -24,14 +24,14 @@ The configuration file should have a [variable_args], a [static_args],
 and one or more [distribution] sections. Multiple [constraint] sections and
 one or more [waveform_transforms] sections may also be provided. An example:
 
-    [variable_args]
+    [variable_params]
     mass1 =
     mass2 =
     spin1_a =
     spin1_azimuthal =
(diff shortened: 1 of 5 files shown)


bin/pycbc_create_injections
```diff
@@ -24,14 +24,14 @@ The configuration file should have a [variable_args], a [static_args],
 and one or more [distribution] sections. Multiple [constraint] sections and
 one or more [waveform_transforms] sections may also be provided. An example:
 
-    [variable_args]
+    [variable_params]
```
soumide1102: Looks like the docs in line 23 has to be updated too.

dfinstad: Whoops, good catch.

soumide1102:
@dfinstad I see you are editing `pycbc_make_inference_inj_workflow` as well. I think there are a few `args` --> `params` updates needed there, if you want to edit that.

dfinstad:
@soumide1102 yeah you're right, I do see a couple there so I'll make those changes. Thanks!

soumide1102:
@dfinstad Thanks for the updates! There is one here I noticed `https://github.com/gwastro/pycbc/blob/master/pycbc/inference/io/base_hdf.py#L304`. Though wherever `samples_from_cli` is called in pycbc, I see a parameters is provided. So currently none of the executables calling this fails. But if this function is called from outside, an Inference file is provided,  and no parameters are provided, this should fail.

dfinstad:
@soumide1102 good catch yet again! I've fixed all the ones you mentioned now, and the results from a simple test run of `pycbc_make_inference_inj_workflow` are here:
https://sugwg-jobs.phy.syr.edu/secure/daniel.finstad/test_workflow/

This has changed substantially since your approval so I'll also re-request review.

PR #2694, opened 2019-05-11
Fix extract samples

cdcapano:
This restores functionality to `pycbc_inference_extract_samples` so that you can specify what parameters to write to the output file, including renaming them in the output. Also, this just makes it so that `extract_samples` always writes `PosteriorFiles`, removing the the `--posterior-only` option and getting rid of the option. (It used to be that you could also use `extract_samples` to thin MCMC files, but with thinning on the fly, that's no longer needed. Plus, that hasn't worked since the inference dev merge anyway.)

It also restores all metadata to the output file, although you can control what groups are written using the `--skip-groups` option.

In doing this, I found that the `write_posterior` method was basically useless, and had led to a bunch of redundancies. Extracting posteriors is already what `read_samples` does by default, and it responds to the command-line options in `extract_samples`, which are meant to override the default behavior. So I've removed the `write_posterior` methods from the Inference file HDF definitions. If we find it necessary in the future, we can move the body of `extract_samples` to a method later.

I've also created a `BaseSamplerFile` class, which inherits from `BaseInferenceFile`. The `write_resume_point` and `write_sampler_metadata` methods have been moved to them. The sampler IO classes now inherit from `BaseSamplerFile`; `PosteriorFile` inherits directly from `BaseInferenceFile`. Truthfully, this bit wasn't really needed for this patch to work (it was something I did while I was trying to figure out what to do to get `write_posterior` to respect `extract_samples`'s options), but since it cleans up the `PosteriorFile` code a bit, I decided to leave it in.

Example usage:
```
pycbc_inference_extract_samples --input-file inference.hdf --output-file posterior_masses.hdf --verbose --parameters 'primary_mass(mass1, mass2):mass1' 'secondary_mass(mass1, mass2):mass2' 'mchirp_from_mass1_mass2(mass1, mass2):mchirp' loglikelihood logprior
```
Yields:
```
$ h5ls posterior_masses.hdf/samples
loglikelihood            Dataset {1400/Inf}
logprior                 Dataset {1400/Inf}
mass1                    Dataset {1400/Inf}
mass2                    Dataset {1400/Inf}
mchirp                   Dataset {1400/Inf}
```


ahnitz:
@cdcapano A couple of unused things indicated in codeclimate.

PR #2696, opened 2019-05-12
Fix inference workflow

dfinstad:
This cleans up and makes several fixes to `pycbc_make_inference_workflow` so that it works. Duplicated code is removed, imports are updated to be more consistent with `pycbc_make_inference_inj_workflow`, and several function calls are fixed that were failing due to incorrect or outdated arguments.

A simple test run with these changes produced [this output](https://sugwg-jobs.phy.syr.edu/secure/daniel.finstad/inference_workflow_test/).


cdcapano:
@dfinstad A to do for another PR: it’d be good to add an `extract_samples` node to the workflow to create a posterior file immediately after `pycbc_inference`. Then, you’d have as final output the same sort of file, regardless of the sampler. That node could also be used to add derived parameters to the posterior file, like mchirp, and to swap the masses and spins around so that mass1 > mass2. That’d simplify later plotting calls.

PR #2709, opened 2019-05-13
Fix py3 error when analytical PSD length is not int

titodalcanton:
I noticed a common error in the Python 3 unit test, seemingly due to passing non-int lengths to a LAL function taking a `size_t`.

### DIFF
--- a/pycbc/psd/analytical.py
+++ b/pycbc/psd/analytical.py
@@ -18,6 +18,7 @@
 """
 
 import warnings
+import numbers
 from pycbc.types import FrequencySeries, zeros
 import lal
 import lalsimulation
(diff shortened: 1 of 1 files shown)


pycbc/psd/analytical.py
```diff
@@ -96,6 +97,10 @@ def from_string(psd_name, length, delta_f, low_freq_cutoff):
         raise ValueError(psd_name + ' not found among analytical '
                          'PSD functions.')
 
+    if type(length) is not int:
```
ahnitz: Does this work if it's a numpy.int32 for example?

ahnitz: Maybe we don't care...

titodalcanton: With a 1-element int32 array it will still print the warning, but otherwise produce the correct result. An array longer than 1 will raise a TypeError.

titodalcanton: Hold on, there is a better way for the type check.

titodalcanton:
This should now silently work for `int`, `long` and `numpy.int*`, and print a warning in any other case (possibly followed by an exception if something crazy is being passed).

titodalcanton:
[Reference](https://stackoverflow.com/questions/3501382/checking-whether-a-variable-is-an-integer-or-not/48940855#48940855)

PR #2715, opened 2019-05-14
missing mask init removes all triggers

ahnitz:


tdent:
I don't really know what the 'premask' code stuff does or is supposed to do yet. The masking/vetoing functionality was a bit messy before recent changes but now imo (if these are 'fixing things properly') needs more documentation.

Just on this one change, if it's doing what I'm guessing it should be doing, you could use numpy.ones() and save a line of code ?

ahnitz:
Yes, numpy.ones would be better. I needed a way to set the mask *before* anything in init was called, as those functions themselves would then try to read in entire columns of data. In principle this could somehow be merged for my uses with something like a filter_func that acted before the vetoes, but I couldn't figure out a nice way to do that without completely rewriting the syntax of filter_func into something very different. This is just a way to say, I have a mask now, use that before applying additional things to the mask.

tdent:
I do wonder if this general approach of having an arbitrarily large hdf file and trying to only use small pieces of it at any given time is right, as opposed to e.g. distributing single triggers over several smaller files.  Maybe it is if we do the right h5py-ish thing.  
Too late in the day to think about it in detail atm.  Up till now we have mostly operated by converting a h5py dataset into a numpy array before going forward.

tdent:
Preferable to have an explanation in the code rather than in a disposable merge request thread, but we can leave that to another day.

tdent:
plot_singles_vs_params memory usage is pretty huge as well on even ~1 billion triggers, I doubt CIT will accommodate it indefinitely, so a way to cut down on that would be desirable too.

PR #2716, opened 2019-05-15
Fixing issue of PosteriorFile Class inheritance to CPNestFile class

sum33it:
This is a minor but necessary change to write posterior files.

### DIFF
--- a/pycbc/inference/io/cpnest.py
+++ b/pycbc/inference/io/cpnest.py
@@ -33,7 +33,7 @@ class CPNestFile(BaseSamplerFile):
     name = 'cpnest_file'
 
     def read_raw_samples(self, fields, **kwargs):
-        read_raw_samples_from_file(self, fields, **kwargs)
+        return read_raw_samples_from_file(self, fields, **kwargs)
 
     def write_resume_point(self):
         pass
--- a/pycbc/inference/io/cpnest.py
+++ b/pycbc/inference/io/cpnest.py
@@ -24,7 +24,7 @@
 """Provides IO for the emcee sampler.
 """
 from .base_sampler import BaseSamplerFile
-from .posterior import PosteriorFile
+from .posterior import *
 
 
 class CPNestFile(BaseSamplerFile):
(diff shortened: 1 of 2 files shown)


pycbc/inference/io/cpnest.py
```diff
@@ -27,7 +27,7 @@
 from .posterior import PosteriorFile
 
 
-class CPNestFile(BaseSamplerFile):
+class CPNestFile(BaseSamplerFile, PosteriorFile):
```
cdcapano: Don't do this. Keep the inheritance limited to `BaseSamplerFile`. Instead, add a `read_raw_samples` function to this class, and have it call `PosteriorFile.read_raw_samples`, similar to what I did with `write_posterior`.

pycbc/inference/io/posterior.py
```diff
@@ -33,48 +33,55 @@ class PosteriorFile(BaseInferenceFile):
...
+
+def write_samples_to_file(fp, samples, parameters=None):
+    """Writes samples to the given file.
+
+    Results are written to ``samples_group/{vararg}``, where ``{vararg}``
+    is the name of a model params. The samples are written as an
+    array of length ``niterations``.
+
+    Parameters
+    -----------
```
cdcapano: Add a doc string here for `fp `.

pycbc/inference/io/posterior.py
```diff
@@ -33,48 +33,55 @@ class PosteriorFile(BaseInferenceFile):
     name = 'posterior_file'
 
     def read_raw_samples(self, fields, **kwargs):
-        samples = self[self.samples_group]
-        return {field: samples[field][:] for field in fields}
+        return read_raw_samples_from_file(self, fields, **kwargs)
 
     def write_samples(self, samples, parameters=None):
-        """Writes samples to the given file.
+        return write_samples_to_file(self, samples, parameters=None)
```
cdcapano: You need to pass `parameters=parameters` here. As is, the parameters input would be ignored.

Issue #2702, opened 2019-05-13
decimation fails for non-time-slide analysis if no decimation option given


tdent:
In an injection coinc job I am seeing
`File &quot;/home/tdent/pycbc-aug18/bin/pycbc_multiifo_coinc_findtrigs&quot;, line 286, in &lt;module&gt;`
    `for decstr in args.loudest_keep_values:`
`TypeError: &apos;NoneType&apos; object is not iterable`

If the code is not doing time slides, it should not require decimation options (or try to do decimation anyway).  Alternatively if the decimation option is always needed it should be made a required option.

tdent:
Fixed by #2721

PR #2721, opened 2019-05-16
Multi coinc

tdent:
Apply some cleanups from the 2-ifo code to multi-ifo and vice versa. Also don't fail if the user doesn't specify a decimation option.


tdent:
Addresses #2702

tdent:
Fixes #2702

PR #2723, opened 2019-05-17
Fix injection accuracy plotting for multi-ifo format

tdent:
Recovery plot was only working for 2-ifo case, update for multi-ifo.


tdent:
Turns out I didn't realize that the '-1' value for injections not found in a given ifo needed some more fixes.  This version seems to work for both the 2-ifo and multi-ifo formats.

PR #2725, opened 2019-05-19
Multiifo sensitivity plot

tapaimarton:
Quick fix to make sensitivity plot to work.

### DIFF
--- a/bin/hdfcoinc/pycbc_coinc_hdfinjfind
+++ b/bin/hdfcoinc/pycbc_coinc_hdfinjfind
@@ -218,11 +218,11 @@ for trigger_file, injection_file in zip(args.trigger_files,
                  fo.create_group(key)
             fo[key].attrs['pivot'] = f[key].attrs['pivot']
             fo[key].attrs['fixed'] = f[key].attrs['fixed']
-            fo[key].attrs['foreground_time_exc'] = f[key].attrs['foreground_time_exc']
+        fo.attrs['foreground_time_exc'] = foreground_time_exc
     else:
         fo.attrs['detector_1'] = f.attrs['detector_1']
         fo.attrs['detector_2'] = f.attrs['detector_2']
-        fo.attrs['foreground_time_exc'] = foreground_time_exc
+        fo.attrs['foreground_time_exc'] = f.attrs['foreground_time_exc']
 
     hdf_append(fo, 'missed/all', missed + injection_index)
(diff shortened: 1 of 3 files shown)


bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -120,6 +121,8 @@ for trigger_file, injection_file in zip(args.trigger_files,
                 ends = f['/segments/%s/end' % key][:]
                 any_seg += events.start_end_to_segments(starts,ends)
         ana_start, ana_end = events.segments_to_start_end(any_seg)
+        any_seg.coalesce()
+        foreground_time_exc = abs(any_seg)
```
tdent: It's not clear to me whether this is the right calculation for 'foreground_time_exc' i.e. total analysis time excluding small windows around zerolag events.  If we don't know that this is right we need a 'FIXME' here saying so.

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -215,7 +218,7 @@ for trigger_file, injection_file in zip(args.trigger_files,
                  fo.create_group(key)
             fo[key].attrs['pivot'] = f[key].attrs['pivot']
             fo[key].attrs['fixed'] = f[key].attrs['fixed']
-            fo[key].attrs['foreground_time_exc'] = f[key].attrs['foreground_time_exc']
+        fo.attrs['foreground_time_exc'] = foreground_time_exc
```
tdent: I think the output file should continue to have the individual `foreground_time_exc` fields for the different types of coinc time.  Otherwise we are throwing information from the statmap files away.

tdent:
Once a solution to #2727 is found this change to hdfinjfind should not be needed.

tapaimarton:
Some change will be required, in order to give the value over to the injfind file, but I agree, better to solve in combine statmap, this was a quick fix to see if the sensitivity code only needed this working, I am ok with not merging this while we solve it in combine statmap.

tdent:
@tapaimarton I think the temporary fix to this plot is to not look for the livetime unless the plot type requested is VT.  I.e. if the plot is just volume or distance it should already work.

tapaimarton:
@tdent I reverted the hdfinjfind changes and added a check to only add to the t value if vt plot is requested.

PR #2733, opened 2019-05-25
Make PSD plotting compatible with inference and search workflows

dfinstad:
This fixes a bug introduced in #2696 that caused PSD plotting to fail when running `pycbc_make_coinc_search_workflow`. The plotting code will now look for a low frequency cutoff in file locations used by both the search and inference workflows.

### DIFF
--- a/bin/hdfcoinc/pycbc_plot_psd_file
+++ b/bin/hdfcoinc/pycbc_plot_psd_file
@@ -69,9 +69,9 @@ for psd_file in args.psd_files:
                     flow = base_file.attrs[
                                '{}_likelihood_low_freq'.format(ifo)]
             except KeyError:
-                raise ValueError("low_frequency_cutoff not found in hdf file; "
-                                 "please specify a value using the "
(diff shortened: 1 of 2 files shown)


bin/hdfcoinc/pycbc_plot_psd_file
```diff
@@ -68,8 +72,6 @@ for psd_file in args.psd_files:
                 raise ValueError("low_frequency_cutoff not found in hdf file; "
```
tdent: maybe you would want to rewrite this error message as it's now not quite accurate - eg `"hdf file %s does not contain an attribute giving low frequency cutoff" % psd_file' .

Also `ValueError` is not quite the right error class, the docs describe it as "Raised when an operation or function receives an argument that has the right type but an inappropriate value".  Use `RuntimeError` for a generic error encountered during execution.

dfinstad: thanks for the comments @tdent, I've made those changes.

PR #2734, opened 2019-05-26
Multiifo inf minifollowup

tapaimarton:
This PR enables injection minifollowups. Changes: give time = -1.0 to the plotting code when the injection is looked for an ifo that is not online. This is how combine statmap determines and foreground and single followup expects the codes.

Work in progress, some pycbc_single_template and pycbc_page_injinfo jobs still fail.

### DIFF
--- a/bin/minifollowups/pycbc_injection_minifollowup
+++ b/bin/minifollowups/pycbc_injection_minifollowup
@@ -146,7 +146,7 @@ for num_event in range(num_events):
     for single in single_triggers:
         ifo = single.ifo
         for seg in insp_data_seglists[single.ifo]:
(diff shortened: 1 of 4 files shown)


tdent:
the result looks good - except the title of the empty Q transform plots is "No triggers or this IFO"

presumably this should have been '.. for this IFO' - and maybe it should actually say no *data* for this IFO?  ie the plot can be made even if there are no triggers, and will be made for missed injections with no trigger

tapaimarton:
@tdent Done, that title was setup when I fixed the foreground minifollowup but the title was left like that since that PR. Changed now.

PR #2738, opened 2019-05-29
Injections from file for pp-workflow

JulianWesterweck:
This allows to use the pp_workflow with a file containing pre-made injections. 

pycbc_make_inference_inj_workflow takes files with multiple injections and splits these into files containing one injection each for inference to run on. 

Also allows to insert the keyword 'injection' in the static_params section of the inference.ini file. Static parameters set to this are replaced with the value found in the corresponding injection.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -35,6 +35,7 @@ from pycbc import inference
 from pycbc.inference import (models, burn_in, option_utils)
 from pycbc.strain.calibration import Recalibrate
 from pycbc.workflow import configuration
+from pycbc.inject import InjectionSet
 
 # command line usage
 parser = argparse.ArgumentParser(usage=__file__ + " [--options]",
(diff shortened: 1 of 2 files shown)


micamu:
@JulianWesterweck is this a substitute for #2589 or something different?

JulianWesterweck:
@micamu Yes, this is a substitute - we decided it would be easier to start over after the changes to unify CBC and ringdown injections.

bin/inference/pycbc_inference
```diff
@@ -147,7 +148,31 @@ with ctx:
...
+    if injection_file \
+        and "injection" in dict(cp.items('static_params')).values():
+        inj = InjectionSet(injection_file)
+        for opt in cp.options(section='static_params'):
+            if cp.get('static_params', opt) == "injection" \
+                                and opt not in inj.table.fieldnames:
+                raise ValueError("Static param placeholder \"injection\" \
+                            has no counterpart in injection file.")
+            elif cp.get('static_params', opt) == "injection":
+                cp.set('static_params', opt, str(inj.table[opt][0]))
```
micamu: Please see my comment in #2589 regarding `str(inj.table[opt][0])`. This is what causes the ringdown lmns parameter to be in the wrong format and it should be fixed here rather than having a complicated `format_lmns` function in the ringdown module because of this (@cdcapano has been disliking the shape that format_lmns was taking, and it is all because of this)

cdcapano:
@JulianWesterweck The history appears to be mixed up. Try rebasing to current master, keeping only the commits you've done. (`git rebase -i gwastro/master`)

micamu:
> @JulianWesterweck Please make the change @micamu requested.
> 
> Also, can you change the keyword used to substitute in the injection values to `FROM_INJECTION`? That way it's clear that it is a directive.

Does that mean now we have to write `FROM_INJECTION` in the ini file instead of just `injection`?

JulianWesterweck:
@micamu Exactly, the keyword for the ini file has changed and it should be able to deal with lists of parameters such as for the lmns now.

micamu:
> I am now worried about the `str` handling in python3, which came up in the conversation in #2768. We might have troubles here too once we switch to python3

As far as the lmns is concerned, I think I have found a way to fix the python3 string issues in the `format_lmns` function. So this might not need any further changes here.

PR #2741, opened 2019-05-31
pycbc_make_inference_workflow: remove file-path command line options

### LINKED ISSUE #2739: pycbc_make_inference_workflow : superfluous commandline options for auto-generated files
The three commandline options --output-map, --transformation-catalog and --output-file seem superfluous (as the executable could instead just make up filenames in the general --output-dir) and are actively confusing to new users (as evidenced by 2 students and me thinking for a long time whether we need to provide pre-generated map and catalog files).

After discussion with @spxiwh I or one of the students will make a pull request soon to remove these; or if there is some use case we haven't thought of where the user would want those files to live outside --output-dir, we could also just make them optional instead.

dbkeitel:
-user shouldn't need to worry about filenames of --output-map,
  --transformation-catalog and --output-file
 -will be put in default --output-dir with default names automatically
 -closes #2739

Or should they still be optional arguments?

### DIFF
--- a/bin/workflows/pycbc_make_inference_workflow
+++ b/bin/workflows/pycbc_make_inference_workflow
@@ -77,12 +77,6 @@ parser.add_argument("--tags", nargs="+", default=[],
                     help="Tags to apend in various places.")
 parser.add_argument("--output-dir", default=None,
                     help="Path to output directory.")
-parser.add_argument("--output-map", required=True,
-                    help="Path to output map file.")
-parser.add_argument("--transformation-catalog", required=True,
-                    help="Path to transformation catalog file.")
-parser.add_argument("--output-file", required=True,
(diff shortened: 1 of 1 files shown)


cdcapano:
Making @cmbiwer responsible as this is his code.

cmbiwer:
Please do not remove ``--output-file``. You can specify ``default=None`` and then ``if opts.output_file == None else opts.output_file`` below. The other two, I have no preference anymore if they are there or not.

cdcapano:
@dbkeitel Are you guys still working on this? If so, could you address @cmbiwer's comments?

dbkeitel:
Sorry, I'll get back to this next week.

dbkeitel:
I've opened a new #2855 instead.

PR #2767, opened 2019-06-11
Added new post-processing scripts for PyGRB

duncanmmacleod:
This PR adds new executables for the PyGRB workflow to go with @idorrington92's new `pycbc_multi_inspiral` code

- `pycbc_grb_trig_combiner`
- `pycbc_grb_trig_cluster`

### DIFF
--- a/bin/pygrb/pycbc_grb_inj_combiner
+++ b/bin/pygrb/pycbc_grb_inj_combiner
@@ -1,0 +1,269 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+# Copyright (C) 2019 Duncan Macleod
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 3 of the License, or (at your
+# option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+"""Combine triggers from a splitbank GRB run
+"""
+
+from __future__ import (division, print_function)
+
+import argparse
+import os
+import re
+from collections import defaultdict
+from math import (pi as PI, radians)
+
+import numpy
+
+import tqdm
+
+import h5py
+
+from astropy.constants import M_sun
+
+from pycbc import __version__
+
+MSUN_SI = M_sun.si.value
+
+__author__ = "Duncan Macleod <duncan.macleod@ligo.org>"
+
+TQDM_BAR_FORMAT = ("{desc}: |{bar}| "
+                   "{n_fmt}/{total_fmt} {unit} ({percentage:3.0f}%) "
+                   "[{elapsed} | ETA {remaining}]")
+TQDM_KW = {
+    "ascii": " -=#",
+    "bar_format": TQDM_BAR_FORMAT,
+    "smoothing": 0.05,
+}
+
+INJSTRING_REGEX = re.compile(r"_[A-Z]*(\d+)(INJ)_(FOUND|MISSED)")
+
+# -- utilties -----------------------------------
+
+
+def use_file(filename, inclination):
+    """Returns `True` if this file can be used to analysis the given inclination
+    """
+    match = INJSTRING_REGEX.search(filename)
+    if match and float(match.groups()[0]) >= inclination:
+        return True
+    return False
+
+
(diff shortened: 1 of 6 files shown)


spxiwh:
@idorrington92 and/or @pannarale should review this.

Please read through the code and indicate that you can understand the code that is written and follow what it is doing (and that what it is doing is right). Where it is not clear please ask for in-line comments (or whatever needed) to improve matters.

idorrington92:
Sure, I'll do it today.

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -0,0 +1,407 @@
...
+    required=True,
+    metavar="TRIGGER FILE",
+    help="read in listed trigger files",
+)
+
+args = parser.parse_args()
+
+vprint = print if args.verbose else str
+
+vprint("-- Hello")
```
idorrington92: Don't think we need 'hello'

bin/pygrb/pycbc_grb_trig_cluster
```diff
@@ -0,0 +1,256 @@
...
+                        compression_opts=old.compression_opts,
+                    )
+                    bar.update()
+            for ifo in ifos:
+                idx = ifo_index[ifo]
+                for old in h5in[ifo].values():
+                    if isinstance(old, h5py.Dataset):
+                        h5out.create_dataset(
+                            old.name,
+                            data=old[:][idx],
```
idorrington92: I think idx should be include.

duncanmmacleod: @idorrington92, no, `include` is the index-array of network events to keep, `idx` here is the equivalent index-array of single-detector events to keep, so this is correct.

bin/pygrb/pycbc_grb_trig_combiner
duncanmmacleod: There's nothing wrong with a polite welcome...

idorrington92:
Ok, I'm running on a different GRB, and this line is crashing. I'll keep
investigating.

spxiwh:
Travis is failing:
```
>> [Fri Jun 28 16:03:40 UTC 2019] running /home/travis/virtualenv/python2.7.15/bin/pycbc_grb_inj_combiner --help
    FAILED!
  File "/home/travis/virtualenv/python2.7.15/bin/pycbc_grb_inj_combiner", line 33, in <module>
    import tqdm
ImportError: No module named tqdm
```

bin/pygrb/pycbc_grb_inj_combiner
```diff
@@ -0,0 +1,269 @@
...
+    """Returns `True` if this file can be used to analysis the given inclination
+    """
+    match = INJSTRING_REGEX.search(filename)
+    if match and float(match.groups()[0]) >= inclination:
+        return True
+    return False
+
+
+def theta(h5injgroup):
+    """Calculate the theta angle column for an injection group
```
spxiwh: I think there's better versions of this function in LAL (including the non-leading terms in L). We might want to switch to a better function if there is one. Worth checking with Seb/Patricia/Harald, but doesn't hold up merging this.

idorrington92:
Inj finder returns the same injections repeatedly. Guessing there is a bug where it does not loop over the files correctly or something like that. I'll try to dig into it if I have time, otherwise it'll have to wait for @duncanmmacleod  to return from his holiday.

bin/pygrb/pycbc_grb_inj_combiner
spxiwh: Would be nice to have the conversion function in here though:
https://github.com/gwastro/pycbc/blob/master/pycbc/conversions.py

spxiwh:
The Travis failure is because a new dependency (tqdm) is being added, but it has not been declared. I should also ask why the new dependency is needed, but I'm not too worried about that (other than a minor concern that it makes the code hard to understand, because `tqdm.tqdm` gives me *no* idea what is going on)

bin/pygrb/pycbc_grb_inj_combiner
duncanmmacleod: I am not confident that I understand the function and its use cases properly enough to translate this into that module. I would rather leave it here for now, and pass the buck to other pycbc devs to move this upstream and modify it to fit appropriately.

duncanmmacleod:
> The Travis failure is because a new dependency (tqdm) is being added, but it has not been declared. I should also ask why the new dependency is needed, but I'm not too worried about that (other than a minor concern that it makes the code hard to understand, because `tqdm.tqdm` gives me _no_ idea what is going on)

@spxiwh, `tqdm` just provides the fancy progress bars, and is fairly ubiquitous in python libraries these days (it's already in the release virtualenvs under `/cvmfs`). Should this be a first-class dependency in `setup.py`, or just an optional extra in `requirements.txt`?

spxiwh:
@duncanmmacleod No strong preference on my part. Let's put it in both `setup.py` and `requirements.txt` for now.

duncanmmacleod:
ok, @spxiwh, I think this is ready for final review/merge, thanks.

PR #2773, opened 2019-06-13
Propagate analysis time

GarethDaviesGW:
Give `foreground_time` and `foreground_time_exc` as outputs to the combine_statmap code

This output time is the amount of time for which two or more detectors are functioning, as well as this time with foreground events removed.

Note that if an event happens within >2 detector time, but has not triggered in all detectors, this will not be removed with the foreground veto

This helps to address https://github.com/gwastro/pycbc/issues/2727

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_combine_statmap
+++ b/bin/hdfcoinc/pycbc_multiifo_combine_statmap
@@ -1,7 +1,8 @@
 #!/bin/env python
-""" Apply a trials factor based on the number of available detector combinations at the
-time of coincidence. This clusters to find the most significant foreground, but
-leaves the background triggers alone. """
+""" Apply a trials factor based on the number of available detector
+combinations at the time of coincidence. This clusters to find the most
+significant foreground, but leaves the background triggers alone.
+"""
 
 import h5py, numpy, argparse, logging, pycbc, pycbc.events, pycbc.io, lal
 import pycbc.version
@@ -25,7 +26,8 @@ files = [h5py.File(n, 'r') for n in args.statmap_files]
 f = h5py.File(args.output_file, "w")
 
 logging.info('Copying segments and attributes to %s' % args.output_file)
-# Move segments information into the final file - remove some duplication in earlier files
+# Move segments information into the final file - remove some duplication
+# in earlier files
 for fi in files:
     for key in fi['segments']:
         if key.startswith('foreground') or key.startswith('background'):
@@ -33,13 +35,16 @@ for fi in files:
         f['segments/%s/end' % key] = fi['segments/%s/end' % key][:]
         f['segments/%s/start' % key] = fi['segments/%s/start' % key][:]
         if 'segments/foreground_veto' in fi:
-            f['segments/%s/foreground_veto/end' % key] = fi['segments/foreground_veto/end'][:]
-            f['segments/%s/foreground_veto/start' % key] = fi['segments/foreground_veto/start'][:]
+            f['segments/%s/foreground_veto/end' % key] = \
+                                         fi['segments/foreground_veto/end'][:]
+            f['segments/%s/foreground_veto/start' % key] = \
+                                       fi['segments/foreground_veto/start'][:]
         for attr_name in fi.attrs:
             if key not in f:
                  f.create_group(key)
             f[key].attrs[attr_name] = fi.attrs[attr_name]
 
+logging.info('combining foreground and foreground excluded segments')
 # Set up dictionaries to contain segments from the individual statmap files
 indiv_segs = segments.segmentlistdict({})
 indiv_segs_fg_exc = segments.segmentlistdict({})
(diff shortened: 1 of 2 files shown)


bin/hdfcoinc/pycbc_multiifo_combine_statmap
```diff
@@ -24,23 +26,58 @@ files = [h5py.File(n, 'r') for n in args.statmap_files]
...
                  f.create_group(key)
             f[key].attrs[attr_name] = fi.attrs[attr_name]
 
+logging.info('combining foreground and foreground excluded segments')
+# Set up dictionaries to contain segments from the individual statmap files
+indiv_segs = segments.segmentlistdict({})
+indiv_segs_fg_exc = segments.segmentlistdict({})
+
+# obtain list of all ifos involved in the coinc_statmap files
+all_ifos = numpy.unique([ifo for fi in files for ifo in \
```
GarethDaviesGW: This line means that we can remove `--ifos` from the arguments, as it will read the interferometers from the input statmap files (this can be removed if it is not wanted/needed)

tdent: Maybe you can ask people about this in the multiifo channel. NB if --ifos is not needed as an argument then that implies some change/simplification in how the job is called in the workflow

tdent:
Is this ready for review Gareth? I didn't get a notification that you'd removed the wip label. Do you have some sample output numbers for sanity
(& ping @spxiwh as I'll be on planes today)

GarethDaviesGW:
Tested on one of the runs I had lying around:
Single IFO times:
H1 Time: 590857.0
L1 Time: 589332.0
V1 Time: 770957.0 (I may not have applied Virgo vetoes properly!)

Coincident analysis times for each type of coinc (from coinc_statmap):
H1L1 Time: 443456.0
H1V1 Time: 531338.0
L1V1 Time: 528177.0
H1L1V1 Time: 402737.0

More than 2 ifo time (i.e. the new number that I wanted checking): 697497.0

This number is plausible, and it makes sense as being somewhere between the 2-detector time and the largest single detector time. Does it make sense to others?

tdent:
The total time number is correct given the other numbers, you should be able to work out why.
I'm assuming these are not 'exclusive' time, are there numbers for that too?

GarethDaviesGW:
> The total time number is correct given the other numbers, you should be able to work out why.
> I'm assuming these are not 'exclusive' time, are there numbers for that too?

Yes, sorry I forgot to include it, `foreground_time_exc`, 697492.2537641525

bin/hdfcoinc/pycbc_multiifo_combine_statmap
```diff
@@ -24,23 +24,58 @@ files = [h5py.File(n, 'r') for n in args.statmap_files]
...
+    # get the foreground_vetoed times from individual statmap files
+    remove_start_time = fi['segments/foreground_veto/start'][:]
+    remove_end_time = fi['segments/foreground_veto/end'][:]
+    vetoed_segs = pycbc.events.veto.start_end_to_segments(remove_start_time,
+                                                          remove_end_time)
+    # remove from analysis segments
+    indiv_segs_fg_exc[key] = indiv_segs[key] - vetoed_segs
+
+# Add the segments to obtain '2 or more detectors on' time
+foreground_segs = numpy.sum(list(indiv_segs.values()), axis=0)
```
tdent: can you explain a bit how this addition works in terms of the structures (dicts, lists, segments, ...)

GarethDaviesGW: Is that a request for better inline comments or for detailed explanation here?

`indiv_segs` is a segmentlistdict, a dictionary containing the segmentlists from the individual statmap files, in this case it is keyed on detector combination

Using `list(indiv_segs.values())` we get a list of segmentlists

We then add these together using `numpy.sum` and by setting `axis=0` we end up getting `segmentlist[0] + segmentlist[1] + ...` rather than `sum(segmentlist[0]) + sum(segmentlist[1]) + ...` as is the `numpy.sum` default, and doesn't really make sense for segmentlists. As it is a segmentlist type, this addition then works in the natural segment addition way

`indiv_segs_fg_exc` is the same but with the foreground excluded. By adding these segmentlists together we get the foreground time, but with times removed which have triggers in *all currently 'on'* detectors. So a 2-detector coincidence in 3-detector time would *not* be removed unless it has an associated 3-detector coincidence. There is a question which can be asked here about the window size given the difference in how the different foreground exclusive windows are formed in the `coinc_statmap` stage but that may not be a necessary conversation

These segments then get turned into a total time by the `abs` function in the output lines

I hope this is clear, if not then ask

tdent: This sounds like a long enough explanation that some comments would be good, eg 

```
# Convert segmentlistdict to a list ('seglists') of segmentlists
# then numpy.sum(seglists, axis=0) does seglists[0] + seglists[1] + ...
# Note that 'exclusive' time only removes windows around coincs
# that involve *all* active ifos (2 in 2-ifo time, 3 in 3-ifo time ..)
```

GarethDaviesGW: comments added in

tdent:
@GarethDaviesGW can you edit the previous comment with the test livetime numbers to explain the arithmetic for why they're correct, then I think we're OK to merge

tdent:
Travis failure seems unrelated (grrr condor) so I am approving

PR #2790, opened 2019-06-19
Add support for subdominant modes in freq_tau distribution

micamu:
Currently, the functions `final_spin_from_f0_tau` and `final_mass_from_f0_tau` in the conversions module only work for the (l, m, n) = (2, 2, 0) mode. This PR adds the needed constants from Berti et al. to be able to use also the n=0, (l, m) = (3, 3), (4, 4), (2, 1) modes.

With that, we can use those modes in the `UniformF0Tau` distribution. I have added a few lines in the `qnm` distributions module to read which mode to use from the ringdown parameter, which will be in the format `f_lmn`. Default is `l = m = 2`, as it was before.


micamu:
There is only one codeclimate issue that makes sense and I have fixed, "standard import import re should be placed before import numpy".
I believe the issues "Exactly one space required after comma" would cause codebeat to complain, since codebeat always complains if there are white spaces at the end of a line...

micamu:
what should I do about the codeclimate issues? can these be ignored or are there any that should be implemented?

PR #2792, opened 2019-06-20
hdfinjfind: fix veto index bug, simplify some numpy logic

tdent:
It seems veto indices have not been being applied for some time due to an oversight in extracting the indices from segments. 

Also, `numpy.array([i for i in found_within_time if i not in vetoid])` is simpler than `numpy.delete(found_within_time, numpy.where(numpy.in1d(found_within_time, vetoid))[0])`


tdent:
@GarethDaviesGW heads up on the aspect of this that extracts veto indices

tdent:
@GarethDaviesGW either you can approve this and merge into your branch for doing the vetoes properly, or just take over analogous changes onto that branch.

PR #2794, opened 2019-06-21
Multiifo injfind

GarethDaviesGW:
hdfinjfind is removing some injections within valid time because of incorrect application of vetoes

This PR contains:
- updates to the veto removal logic to allow for vetos in one detector not to remove an injection which could be found coincident in others
- addition of a command line option which defines the minimum number of required ifos for an injection *not* to be vetoed
- a small update to foundmissed plots so that it isn't so 'blobby'

### DIFF
--- a/bin/hdfcoinc/pycbc_coinc_hdfinjfind
+++ b/bin/hdfcoinc/pycbc_coinc_hdfinjfind
@@ -88,8 +88,9 @@ for trigger_file, injection_file in zip(args.trigger_files,
         multi_ifo_style = False
         ifo_list = ['H1', 'L1']
         if args.min_required_ifos > 2:
-            raise NotImplementedError('min-required-ifos must be less than or '
-                                'equal to number of ifos involved in search')
+            raise RuntimeError('min-required-ifos (%s' % args.min_required_ifos + 
+                                ') must be less than or equal to the number ' +
+                                'of ifos involved in search (2)')
     else:
         multi_ifo_style = True
         # Get list of groups which contain subgroup 'time'
@@ -98,8 +99,10 @@ for trigger_file, injection_file in zip(args.trigger_files,
                     if 'time' in f['foreground/%s/' % key]]
         assert len(ifo_list) > 1
         if len(ifo_list) < args.min_required_ifos:
-            raise NotImplementedError('min-required-ifos must be less than or '
-                                'equal to number of ifos involved in search')
+            raise RuntimeError('min-required-ifos (%s' % args.min_required_ifos +
+                                ') must be less than or equal to the number ' +
+                                'of ifos involved in the search (%s)' %
+                                len(ifo_list))
     fo.attrs['ifos'] = ' '.join(sorted(ifo_list))
 
     template_id = f['foreground/template_id'][:]
@@ -178,8 +181,9 @@ for trigger_file, injection_file in zip(args.trigger_files,
 
     logging.info('Removing injections in vetoed time')
 
-    # Put individual detector vetoes into list of all vetoed indices, if it
-    # is vetoed in more than one detector, then it will be repeated
+    # Put individual detector vetoes into list of all vetoed indices, if an
+    # injection is vetoed in more than one ifo, then its index will appear
+    # more than once
     vetoid = numpy.array([])
     for ifo in ifo_list:
         vi, _ = indices_within_segments(inj_time, [args.veto_file], ifo=ifo,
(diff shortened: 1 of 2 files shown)


GarethDaviesGW:
The previous version (though this is not particularly previous as it only started working after https://github.com/gwastro/pycbc/pull/2792 !) removed some valid injections as it was removing if *any* IFO was vetoed at the time:
image

This has been updated to only remove the injection if the vetoes mean that *no* coincidences can be formed:
image

As an example, if an injection was during a vetoed V1 time, the 'old' version would have removed it, but a coincidence can still be formed from H1 and L1

GarethDaviesGW:
*Note that the first 3 or 4 commits can probably be disregarded as they were attempting to use a different method to verify the vetoes

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -84,11 +87,19 @@ for trigger_file, injection_file in zip(args.trigger_files,
     if 'foreground/time1' in f:
         multi_ifo_style = False
         ifo_list = ['H1', 'L1']
+        assert args.min_required_ifos > 2
```
tdent: I don't think this is what `assert` does. Read the documentation for it

GarethDaviesGW: ah yes - that was because I had been asserting, then I realised raising an error would be more descriptive. forgot to change to an if statement

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -84,11 +87,19 @@ for trigger_file, injection_file in zip(args.trigger_files,
...
+            raise NotImplementedError('min-required-ifos must be less than or '
```
tdent: Travis didn't like this (unexpected indent). I suppose what you want to do is if args.min_required_ifos > 2 then raise error. I think RuntimeError would be preferable since if the search is only 2-ifo it's logically impossible to require 3 ifos being on.  Also the error message should say the number of ifos in the search is 2 in this case.

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -84,11 +87,19 @@ for trigger_file, injection_file in zip(args.trigger_files,
...
-        # Get list of groups which contain subgroup 'time' - these will be the IFOs
-        ifo_list = [key for key in f['foreground'] if 'time' in f['foreground/%s/' % key]]
+        # Get list of groups which contain subgroup 'time'
+        # - these will be the IFOs
+        ifo_list = [key for key in f['foreground']
+                    if 'time' in f['foreground/%s/' % key]]
         assert len(ifo_list) > 1
+        if len(ifo_list) < args.min_required_ifos:
+            raise NotImplementedError('min-required-ifos must be less than or '
+                                'equal to number of ifos involved in search')
```
tdent: same here - make it RuntimeError and tell the user what the code finds the number of ifos to be

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -166,22 +177,37 @@ for trigger_file, injection_file in zip(args.trigger_files,
                  (len(found_within_time), len(missed_within_time)))
 
     logging.info('Removing injections in vetoed time')
+
+    # Put individual detector vetoes into list of all vetoed indices, if it
+    # is vetoed in more than one detector, then it will be repeated
```
tdent: -> 'if an injection is vetoed in more than one ifo its index will appear more than once'

bin/hdfcoinc/pycbc_coinc_hdfinjfind
```diff
@@ -166,22 +177,37 @@ for trigger_file, injection_file in zip(args.trigger_files,
...
         vetoid = numpy.append(vetoid, vi)
         del vi
+
+    # Find out which of these indices are unique, and how many times each
+    # vetoed index occurs - this is how many detectors it is vetoed in
+    vetoid_unique, count_vetoid = numpy.unique(vetoid, return_counts=True)
+
+    # Vetoed for all combinations are ones which do not allow *any*
+    # combination of min_required_ifos detectors, and so veto more than
+    # (len(ifo_list) - args.min_required_ifos) detectors
```
tdent: not sure why you call it 'vetoed for all combinations'. this line is deciding which injections to remove, yes? 
so i would say `remove injections where the number of unvetoed ifos is less than the minimum specified by user`

& imo arranging the arithmetic as (len(ifo_list) - count_vetoid) < args.min_required_ifos may be clearer

GarethDaviesGW: I think that my thought process was more along the lines of 'calculate the number of ifos which would need to be vetoed in order to remove, and then check' - the end result is the same but I've changed it to be clearer

tdent:
so I think you should now be able to reproduce the 'old' result using min required ifos =3, and also to 'uncover' more missed injections by setting it to =1 or =0 - does that check out?

GarethDaviesGW:
> so I think you should now be able to reproduce the 'old' result using min required ifos =3, and also to 'uncover' more missed injections by setting it to =1 or =0 - does that check out?

reproduction of the old-style result checks out
'uncovering' does not, as there doesn't seem to be any in that state in any of the examples I have

In both these cases, they've already been removed by the 'outside of analysed time' removal from earlier in the code

PR #2799, opened 2019-06-24
New PSD variation

Moz-Port:
Replacement pull request for PSD variation.

Changes with respect to gwastro/pycbc: 
New function calc_filt_psd_variation added to variation.py to compute the psd variation.
The new function implement a different approach to compute the PSD variation, filtering the strain rather then compare long and short PSD. This increase the efficiency and allow to easily remove short duration outliers.

calc_filt_psd_variation is now used in pycbc_inspiral to compute psd_var_val.

Ranking.py is modified to downrank data with psd_var>1.2 rather then 1.8.

### DIFF
--- a/pycbc/psd/__init__.py
+++ b/pycbc/psd/__init__.py
@@ -204,16 +204,19 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
                                   "to you, then it is probably safe to "
                                   "ignore this option.")
     # Options for PSD variation
+    psd_options.add_argument("--psdvar-segment", type=float,
+                             metavar="SECONDS", help="Length of segment "
+                             "when calculating the PSD variability.")
     psd_options.add_argument("--psdvar-short-segment", type=float,
                              metavar="SECONDS", help="Length of short segment "
                              "when calculating the PSD variability.")
     psd_options.add_argument("--psdvar-long-segment", type=float,
                              metavar="SECONDS", help="Length of long segment "
                              "when calculating the PSD variability.")
-    psd_options.add_argument("--psdvar-short-psd-duration", type=float,
+    psd_options.add_argument("--psdvar-psd-duration", type=float,
                              metavar="SECONDS", help="Duration of short "
                              "segments for PSD estimation.")
-    psd_options.add_argument("--psdvar-short-psd-stride", type=float,
+    psd_options.add_argument("--psdvar-psd-stride", type=float,
                              metavar="SECONDS", help="Separation between PSD "
                              "estimation segments.")
     psd_options.add_argument("--psdvar-low-freq", type=float, metavar="HERTZ",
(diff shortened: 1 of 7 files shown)


pycbc/psd/__init__.py
```diff
@@ -204,24 +204,28 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
                                   "to you, then it is probably safe to "
                                   "ignore this option.")
     # Options for PSD variation
+    psd_options.add_argument("--psdvar-segment", type=float,
+                             metavar="SECONDS", help="Length of segment "
+                             "when calculating the PSD variability.")
```
tdent: might be clearer to say eg 'Length of segment for mean square calculation of PSD variation' to link this explicitly to the mean-square method

pycbc/psd/variation.py
```diff
@@ -160,3 +165,214 @@ def find_trigger_value(psd_var, idx, start, sample_rate):
...
+
+    srate = int(data.sample_rate)
+
+    # Calculate mean square of data once per short stride and remove
+    # the outliers
+    data = numpy.array(data)
+    short_ms = numpy.mean(data.reshape(-1, int(srate * short_stride)) ** 2,
+                          axis=1)
+    ave = 0.5 * (short_ms[2:] + short_ms[:-2])
+    outliers = short_ms[1:-1] > (2 * ave)
```
tdent: add comment here explaining why the 2 * ave definition of 'outlier' is chosen - also maybe prefer `2.` ie explicitly a float.

pycbc/psd/variation.py
```diff
@@ -160,3 +165,214 @@ def find_trigger_value(psd_var, idx, start, sample_rate):
...
+    short_ms[1:-1][outliers] = ave[outliers]
```
tdent: here i think you're not /removing/ the outliers, you're replacing them with an averaged value. so maybe make a comment to explain what this does

pycbc/psd/variation.py
```diff
@@ -160,3 +165,214 @@ def find_trigger_value(psd_var, idx, start, sample_rate):
...
+    """
+
+    srate = int(data.sample_rate)
+
+    # Calculate mean square of data once per short stride and remove
+    # the outliers
+    data = numpy.array(data)
+    short_ms = numpy.mean(data.reshape(-1, int(srate * short_stride)) ** 2,
+                          axis=1)
+    ave = 0.5 * (short_ms[2:] + short_ms[:-2])
```
tdent: i'm not entirely sure why this quantity is calculated - can you describe a bit more what it should be doing?

pycbc/psd/variation.py
```diff
@@ -160,3 +165,214 @@ def find_trigger_value(psd_var, idx, start, sample_rate):
...
+            astrain = strain.time_slice(tlong, end_time)
+            plong = pycbc.psd.welch(
+                           strain.time_slice(end_time - psd_long_segment,
+                                             end_time),
+                           seg_len=int(psd_duration * strain.sample_rate),
+                           seg_stride=int(psd_stride * strain.sample_rate),
+                           avg_method=psd_avg_method)
+
+        # Make the weighting filter - bandpass, which weight by f^-7/6,
+        # and withen. The normalization is chosen so that the variance
```
tdent: typo -> 'whiten'

pycbc/psd/variation.py
Moz-Port: Every element of "ave" is the average of the closest elements of the corresponding index of "short_ms". Consider for example the element of the array "short_ms" with index n. The element in the n index of "ave" will be the average between the (n-1) and (n+1) elements of short_ms. This is then use to remove the outliers (or smoothing) short_ms

pycbc/psd/variation.py
Moz-Port: Yes, exactly. Ok, I will make a comment.

pycbc/psd/variation.py
tdent: ah, ok - then it is testing whether any element is an 'outlier' with respect to its two neighbours. perhaps then this should be commented together with the following line which defines an outlier as anything bigger than 2* the average of the 'neighbours'

Moz-Port: Is it fine if I describe it at the beginning of the function?

tdent: yes, commenting in one place may be easier to understand than many small comments on different lines

pycbc/psd/variation.py
```diff
@@ -160,3 +165,219 @@ def find_trigger_value(psd_var, idx, start, sample_rate):
...
     return vals
+
+
+def mean_square(data, delta_t, short_stride, stride):
+    """ Calculate mean square of given time series once per stride
+
+    First of all this function calculate the mean square of given time
+    series once per short_stride. This is used to find and remove
+    outliers due to short glitches. Here an outlier is defined as any
+    element which is grater than two times the average of its closest
```
tdent: typos -> 'greater' / 'corresponding'

PR #2811, opened 2019-06-27
adding add_statmap code

GarethDaviesGW:
`pycbc_multiifo_add_statmap` code will use backgrounds of any coincidences where that detector combination is 'on'.

For example, if the foreground coincidence comes from H1L1, but V1 was also operating at the time, the background used for ifar/fap/ifar_exc/fap_exc calculation will combine backgrounds from H1L1, H1V1, L1V1 and H1L1V1

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_add_statmap
+++ b/bin/hdfcoinc/pycbc_multiifo_add_statmap
@@ -151,7 +151,7 @@ del idx_within_segment
 # interferometer combinations are available at the time of coincidence
 available_combos =[' '.join(sorted([key for key in is_in_combo_time 
                                     if is_in_combo_time[key][i]])
-                           ).encode('utf8') for i in np.arange(n_triggers)]
+                            ).encode('utf8') for i in np.arange(n_triggers)]
 
 all_combo_types = np.unique(available_combos)
 idx = {ct:np.where(np.array(available_combos)==ct)[0]
@@ -180,17 +180,19 @@ fg_ifar_exc = np.zeros(n_triggers)
 for ct in all_combo_types:
     cts = ct.split(' ')
     if len(cts) == 1:
-        logging.info('IFAR is the same as previosuly calculated for triggers in {} time'.format(ct))
+        logging.info('IFAR is the same as previously calculated for triggers in {} time'.format(ct))
         # If only one combination is available, then the stat is the same
         # as previously calulated
         fg_ifar[idx[ct]] = f['foreground/ifar'][:][idx[ct]]
         fg_ifar_exc[idx[ct]] = f['foreground/ifar_exc'][:][idx[ct]]
-    else:
+    elif len(cts) > 1:
         logging.info('Recalculating ifar for coincidences which are in {} time'.format(ct))
         far_new = np.sum([far[ifo_combo_key][idx[ct]] for ifo_combo_key in cts], axis=0)
         far_new_exc = np.sum([far_exc[ifo_combo_key][idx[ct]] for ifo_combo_key in cts], axis=0)
         fg_ifar[idx[ct]] = conv.sec_to_year(1. / np.array(far_new))
         fg_ifar_exc[idx[ct]] = conv.sec_to_year(1. / np.array(far_new_exc))
+    else:
+        raise RuntimeError('Empty combo type string, something has gone wrong')
 
 f.attrs['foreground_time_exc'] = f.attrs['foreground_time']
 
--- a/bin/hdfcoinc/pycbc_multiifo_add_statmap
+++ b/bin/hdfcoinc/pycbc_multiifo_add_statmap
@@ -1,0 +1,287 @@
+#!/bin/env python
+""" This clusters to find the most
+significant foreground, but leaves the background triggers alone.
+"""
+
+import h5py, numpy as np, argparse, logging, pycbc, pycbc.events, pycbc.io, lal
+import pycbc.version
+import pycbc.conversions as conv
+from pycbc.events import coinc
+from ligo import segments
+import itertools
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--version", action="version", version=pycbc.version.git_verbose_msg)
+parser.add_argument('--verbose', action='store_true')
+parser.add_argument('--statmap-files', nargs='+',
+                    help="List of coinc files to be redistributed")
+parser.add_argument('--cluster-window', type=float)
+parser.add_argument('-ranking-statistic', type=str,
+                    help='Ranking statistic used to calculate stat and ifar '
+                          'in previous codes, this is used to test whether '
+                          'the background distributions can be safely '
+                          'combined. Default=Not safe')
+parser.add_argument('--output-file', help="name of output file")
+args = parser.parse_args()
+
+pycbc.init_logging(args.verbose)
+
+files = [h5py.File(n, 'r') for n in args.statmap_files]
+
+f = h5py.File(args.output_file, "w")
+
+logging.info('Copying segments and attributes to %s' % args.output_file)
+# Move segments information into the final file - remove some duplication
+# in earlier files
+for fi in files:
+    for key in fi['segments']:
+        if key.startswith('foreground') or key.startswith('background'):
+            continue
+        f['segments/%s/end' % key] = fi['segments/%s/end' % key][:]
+        f['segments/%s/start' % key] = fi['segments/%s/start' % key][:]
+        if 'segments/foreground_veto' in fi:
+            f['segments/%s/foreground_veto/end' % key] = \
+                                         fi['segments/foreground_veto/end'][:]
+            f['segments/%s/foreground_veto/start' % key] = \
+                                       fi['segments/foreground_veto/start'][:]
+        for attr_name in fi.attrs:
+            if key not in f:
+                 f.create_group(key)
+            f[key].attrs[attr_name] = fi.attrs[attr_name]
+
+logging.info('combining foreground and foreground excluded segments')
+# Set up dictionaries to contain segments from the individual statmap files
+indiv_segs = segments.segmentlistdict({})
+indiv_segs_fg_exc = segments.segmentlistdict({})
+
+any_fg_vetoes = False
+# loop through statmap files and put segments into segmentlistdicts
+for fi in files:
+    key = fi.attrs['ifos'].replace(' ','')
+    # get analysed segments from individual statmap files
+    starts = fi['segments/{}/start'.format(key)][:]
+    ends = fi['segments/{}/end'.format(key)][:]
+    indiv_segs[key] = pycbc.events.veto.start_end_to_segments(starts, ends)
+    # get the foreground_vetoed times from individual statmap files
+    if 'foreground_veto' in fi['segments']:
+        any_fg_vetoes = True
+        remove_start_time = fi['segments/foreground_veto/start'][:]
+        remove_end_time = fi['segments/foreground_veto/end'][:]
+        vetoed_segs = pycbc.events.veto.start_end_to_segments(remove_start_time,
+                                                          remove_end_time)
+        # remove from analysis segments
+        indiv_segs_fg_exc[key] = indiv_segs[key] - vetoed_segs
+
+del remove_end_time, remove_start_time, vetoed_segs
+
+# Convert segmentlistdict to a list ('seglists') of segmentlists
+# then np.sum(seglists, axis=0) does seglists[0] + seglists[1] + ...
+foreground_segs = np.sum(list(indiv_segs.values()), axis=0)
+# obtain list of all ifos involved in the coinc_statmap files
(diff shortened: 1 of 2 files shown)


GarethDaviesGW:
diff between this and the multiifo combine statmap may make the changes easier to see: 
[add_vs_combine_statmap_diff_output.txt](https://github.com/gwastro/pycbc/files/3335344/add_vs_combine_statmap_diff_output.txt)

tdent:
I think the appropriate test here is to print out or otherwise output the FAR values after clustering from the previous trials-factor (combine_statmap) method and from this one and check they are the same for double time and comparable (but not identical) for triple time

GarethDaviesGW:
The changes to the effective trials factor are as follows:
image

This is as expected for this chunk of data. The background reverse-cumulative histogram for all coincidences divided by that for only those in each coincidence is as follows (zoomed to match plot area of above), This has been divided by 4 to get the _changes_ to the effective trials factor:
image

We see that these agree well

tdent:
(Note that these plots are from an O2 chunk 21 run with a weird background statistic distribution due to a separate issue with the [uncombined] statmap code which is being addressed in parallel.)

GarethDaviesGW:
Updated diff of add_statmap and combine_statmap:

[add_vs_combine_statmap_diff_output_20190702.txt](https://github.com/gwastro/pycbc/files/3349603/add_vs_combine_statmap_diff_output_20190702.txt)

[on CIT](https://ldas-jobs.ligo.caltech.edu/~gdavies/testoutput/add_statmap/add_vs_combine_statmap_diff_output_20190702.txt)

tdent:
Noting that we want to initially cluster coincs in triple time by stat (not ifar) since we believe the stat should be a better ranking between different types of coinc. 
Also the assignment of the final FAR as the sum of the individual coinc type FARs _at the stat value of the candidate event_ is appropriate for clustering on stat not ifar.

GarethDaviesGW:
image

After latest updates, the ifar calculation is now equivalent to before (with a better comparison plot style now being used)

I believe this is ready for review now

tdent:
@ahnitz note this is no longer WIP.

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
+#!/bin/env python
+""" Reduce IFAR for coincidences in times with more than one ifo combination
+available. Cluster for best IFAR. This clusters to find the most
+significant foreground, but leaves the background triggers alone.
```
tdent: 'Reduce IFAR' -> 'Calculate total FAR based on statistic ranking' 
'Cluster to keep coincs with highest statistic'

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+"""
+
+import h5py, numpy as np, argparse, logging, pycbc, pycbc.events, pycbc.io, lal
```
tdent: lal is not needed, remove

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+import pycbc.version
+import pycbc.conversions as conv
+from pycbc.events import coinc
+from ligo import segments
+import itertools
```
tdent: I think itertools is not needed, remove

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--version", action="version", version=pycbc.version.git_verbose_msg)
+parser.add_argument('--verbose', action='store_true')
+parser.add_argument('--statmap-files', nargs='+',
+                    help="List of coinc files to be redistributed")
```
tdent: -> 'combined' as redistributed is unclear?

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+parser.add_argument('--censor-ifar-threshold', type=float, default=0.003,
+    help="If provided, only window out foreground triggers with IFAR (years)"
+         "above the threshold [default=0.003yr]")
+parser.add_argument('--veto-window', type=float, default=0.1,
+    help="Time around each zerolag trigger to window out [default=.1s]")
+parser.add_argument('--cluster-window', type=float)
```
tdent: wlt have a help message here - eg 'maximum time interval to cluster coinc events'

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+
+# Convert segmentlistdict to a list ('seglists') of segmentlists
+# then np.sum(seglists, axis=0) does seglists[0] + seglists[1] + ...
+foreground_segs = np.sum(list(indiv_segs.values()), axis=0)
+# obtain list of all ifos involved in the coinc_statmap files
+all_ifos = np.unique([ifo for fi in files
+                      for ifo in fi.attrs['ifos'].split(' ')])
+
+# output to file
+f.attrs['foreground_time'] = abs(foreground_segs)
```
tdent: this line belongs just after `foreground_segs` is calculated - also can remove the comment (it's obvious) and put in a blank line before the line that calculates 'all_ifos'

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+    f['foreground/{}/time'.format(ifo)] = all_trig_times[ifo]
+    f['foreground/{}/trigger_id'.format(ifo)] = all_trig_ids[ifo]
+
+n_triggers = f['foreground/ifar'].size
+logging.info('{} triggers'.format(n_triggers))
+
+# all_times is a tuple of trigger time arrays
+all_times = (f['foreground/%s/time' % ifo][:] for ifo in all_ifos)
+
+# Currently only clustering zerolag, i.e. foreground, so set all timeslide_ids
```
tdent: Begin comment with 'Cluster by statistic value' to be definite

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+# to zero
+cidx = pycbc.events.cluster_coincs_multiifo(f['foreground/stat'][:], all_times,
+                                            np.zeros(n_triggers), 0,
+                                            args.cluster_window, argmax=np.argmax)
```
tdent: np.argmax is already the default so don't need to assign it here.

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+    if key not in all_ifos:
+        filter_dataset(f, 'foreground/%s' % key, cidx)
+    else:  # key is an ifo
+        for k in f['foreground/%s' % key].keys():
+            filter_dataset(f, 'foreground/{}/{}'.format(key, k), cidx)
+
+n_triggers = f['foreground/ifar'].size
+
+times_tuple = (f['foreground/{}/time'.format(ifo)] for ifo in all_ifos)
+test_times = np.array([pycbc.events.mean_if_greater_than_zero(tc)[0]
```
tdent: perhaps put a comment before this block eg 'Calculating event times to determine which types of coinc are active'

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+        continue
+    end_times = np.array(f['segments/%s/end' % key][:])
+    start_times = np.array(f['segments/%s/start' % key][:])
+    idx_within_segment = pycbc.events.indices_within_times(test_times,
+                                                           start_times,
+                                                           end_times)
+    is_in_combo_time[key][idx_within_segment] = np.ones_like(idx_within_segment)
+del idx_within_segment
+
+available_combos =[' '.join(sorted([key for key in is_in_combo_time if is_in_combo_time[key][i]])).encode('utf8') for i in np.arange(n_triggers)]
```
tdent: split this up into (at least) 2 lines, may also be clearer to do an explicit iteration over the n_triggers over rather than a list comprehension. 
also, add a comment saying what 'available combos' is .. it appears to be an enormous string?

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+
+all_combo_types = np.unique(available_combos)
+idx = {ct:np.where(np.array(available_combos)==ct)[0]
+       for ct in all_combo_types}
+
+del available_combos
+
+logging.info('Calculating n_louder background triggers in each type for all foreground events')
```
tdent: slight change : 'Calculating total rate of false alarms over coinc types for foreground events'

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+        _, fnlouder = coinc.calculate_n_louder(f_in['background/stat'][:],
+                                               f['foreground/stat'][:],
+                                               f_in['background/decimation_factor'][:])
+        far[ifo_combo_key] = (fnlouder + 1) / f_in.attrs['background_time']
+        _, fnlouder_exc = coinc.calculate_n_louder(f_in['background_exc/stat'][:],
+                                                   f['foreground/stat'][:],
+                                                   f_in['background_exc/decimation_factor'][:])
+        far_exc[ifo_combo_key] = (fnlouder_exc + 1) / f_in.attrs['background_time_exc']
+
+logging.info('Recalculating ifar according to summed trigger distributions')
```
tdent: remove as each case below already has its own logging statement

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+
+fg_ifar = np.zeros_like(f['foreground/decimation_factor'][:])
+fg_ifar_exc = np.zeros_like(f['foreground/decimation_factor'][:])
+
+for ct in all_combo_types:
+    cts = ct.split(' ')
+    if len(cts) == 1:
+        logging.info('Copying over triggers which are only in {} time'.format(ct))
+        # If only one combination is available, then the stat is the same
+        # as previously calulated
```
tdent: -> 'the IFAR is the same as previously calculated'

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+        fg_ifar[idx[ct]] = f['foreground/ifar'][:][idx[ct]]
+        fg_ifar_exc[idx[ct]] = f['foreground/ifar_exc'][:][idx[ct]]
+        continue
```
tdent: would it work equally well to have an 'else' here, then wouldn't have to break out of the iteration?

tdent: or rather `elif len(cts) > 1:` then after that `else: raise RuntimeError('Empty combo type string, something has gone wrong')`

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+    logging.info('Recalculating ifar for coincidences which are in {} time'.format(ct))
+    far_new = np.sum([far[ifo_combo_key][idx[ct]] for ifo_combo_key in cts], axis=0)
+    far_new_exc = np.sum([far_exc[ifo_combo_key][idx[ct]] for ifo_combo_key in cts], axis=0)
+    ifar = 1 / np.array(far_new)
+    ifar_exc = 1 / np.array(far_new_exc)
```
tdent: use `1.` here so we never have to worry about float division

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -0,0 +1,214 @@
...
+    fg_ifar[idx[ct]] = conv.sec_to_year(ifar)
+    fg_ifar_exc[idx[ct]] = conv.sec_to_year(ifar_exc)
```
tdent: i think these lines could be combined with the previous 2, ie `fg_ifar[idx[ct]] = conv.sec_to_year(1. / np.array(far_new))`

tdent:
& you will want to fetch/rebase after addressing comments.

ahnitz:
@GarethDaviesGW I haven't tried running this myself, but I've scanned through the code. I don't have any major comments. It looks to have the pieces I'd have expected to be there for what I think it's supposed to do (i.e. rerank events with a combined background).

PR #2817, opened 2019-07-01
dont get rid of segments bit of combine statmap

GarethDaviesGW:
reverses overzealous deletions from #2815


tdent:
I won't be able to review this, sorry

tdent:
nah, well, it seems to be exactly what was there before minus the fg veto part, so if this can be tested then it seems trivial to approve

tdent:
looks like there are some lines of 'orphaned' comments and/or log messages now, probably best to remove them to avoid confusion.

PR #2823, opened 2019-07-02
Fix error in autogating & minor cleanup

titodalcanton:
There is an integer division problem in `detect_loud_glitches()` which causes the glitch detection algorithm to report wrong glitch times (by a large fraction of a second) when autogating is used in PyCBC Live. This was found while debugging #2627. In offline analyses the time series passed to `detect_loud_glitches()` is much longer, and the error becomes negligible. This patch fixes it and cleans up the autogating code a little bit in preparation for #2627.

### DIFF
--- a/pycbc/strain/strain.py
+++ b/pycbc/strain/strain.py
@@ -107,7 +107,7 @@ def detect_loud_glitches(strain, psd_duration=4., psd_stride=2.,
     w = numpy.arange(corrupt_length) / float(corrupt_length)
     strain[0:corrupt_length] *= pycbc.types.Array(w, dtype=strain.dtype)
     strain[(len(strain)-corrupt_length):] *= \
-            pycbc.types.Array(w[::-1], dtype=strain.dtype)
+        pycbc.types.Array(w[::-1], dtype=strain.dtype)
 
     if output_intermediates:
         strain.save_to_wav('strain_conditioned.wav')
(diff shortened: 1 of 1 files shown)


pycbc/strain/strain.py
```diff
@@ -100,26 +99,27 @@ def detect_loud_glitches(strain, psd_duration=4., psd_stride=2.,
...
     logging.info('Autogating: tapering strain')
-    taper_length = int(corrupt_time * strain.sample_rate)
-    w = numpy.arange(taper_length) / float(taper_length)
-    strain[0:taper_length] *= pycbc.types.Array(w, dtype=strain.dtype)
-    strain[(len(strain)-taper_length):] *= pycbc.types.Array(w[::-1],
-                                                             dtype=strain.dtype)
+    corrupt_length = int(corrupt_time * strain.sample_rate)
+    w = numpy.arange(corrupt_length) / float(corrupt_length)
+    strain[0:corrupt_length] *= pycbc.types.Array(w, dtype=strain.dtype)
+    strain[(len(strain)-corrupt_length):] *= \
```
tdent: may as well fix some PEP8 as well (spacing around `-`)

pycbc/strain/strain.py
```diff
@@ -100,26 +99,27 @@ def detect_loud_glitches(strain, psd_duration=4., psd_stride=2.,
...
-
     # zero-pad strain to a power-of-2 length
     strain_pad_length = next_power_of_2(len(strain))
     pad_start = int(strain_pad_length/2 - len(strain)/2)
     pad_end = pad_start + len(strain)
     strain_pad = pycbc.types.TimeSeries(
             pycbc.types.zeros(strain_pad_length, dtype=strain.dtype),
             delta_t=strain.delta_t, copy=False,
-            epoch=strain.start_time-pad_start/strain.sample_rate)
+            epoch=strain.start_time-pad_start/float(strain.sample_rate))
```
tdent: PEP8 here (spaces around `-` `/`)

titodalcanton: Does that apply also when the operation is used as a kwarg?

tdent: yes please!  (.. it would if you calculated the strain epoch on a separate line of code, which might actually be more readable)

PR #2825, opened 2019-07-03
Fix multiifo statmap to calculate time of zerolag coincs and remove them

tdent:


bin/hdfcoinc/pycbc_multiifo_coinc_statmap
```diff
@@ -323,68 +317,60 @@ while numpy.any(louder_foreground == 0):
...
-
     indices_to_rm = []
     for ifo in args.ifos:
         indices_to_rm = numpy.concatenate([indices_to_rm, ind_to_rm[ifo]])
-
     all_trigs = all_trigs.remove(indices_to_rm)
 
-    fore_locs = all_trigs.timeslide_id == 0
-    # The foreground trigger has been removed, continue with typical statmap operations.
+    # NB the values calculated below are not used!
```
ahnitz: Why not remove these lines if we aren't going to use them?

bin/hdfcoinc/pycbc_multiifo_coinc_statmap
```diff
@@ -109,37 +102,40 @@ else:
...
+# This gives exclusive background triggers
+exc_zero_trigs = all_trigs.remove([])  # Start by copying existing triggers
 for ifo in ifos:
-    veto_indeces[ifo] = veto.indices_within_times(all_trigs.data['%s/time' % ifo], remove_start_time,
-                                                  remove_end_time)
-    exc_zero_trigs = all_trigs.remove(veto_indeces[ifo])
+    fg_veto_ids = veto.indices_within_times(
+                       exc_zero_trigs.data['%s/time' % ifo],
+                       remove_start_time, remove_end_time)
+    exc_zero_trigs = exc_zero_trigs.remove(fg_veto_ids)
```
ahnitz: Was this the main bug? I.E. it was always starting out with the all_trigs and thus wouldn't be cumulatively removing?

ahnitz:
Travis error is an indentation problem. 
```
  File "/home/travis/virtualenv/python2.7.15/bin/pycbc_multiifo_coinc_statmap_inj", line 115
    numpy.searchsorted(ifsorted, end)
    ^
IndentationError: unexpected indent
```

bin/hdfcoinc/pycbc_multiifo_coinc_statmap
tdent: I don't know if it's worth thinking about whether we might use them .. I guess the exclusive background is what it is independent of any hierarchical removal, so indeed they were never useful.

bin/hdfcoinc/pycbc_multiifo_coinc_statmap
tdent: well also the wrong time calc for doubles in triple time, but this all_trigs.remove is the main bug, yes.

PR #2829, opened 2019-07-04
EM-Bright accessible from allow_em_bright_in_bankverifier

pannarale:
Adding a feature to allow user to only inject potentially EM-Bright signal when verifying a template bank with pycbc_make_bank_verifier_workflow.  This is very useful to verify the PyGRB template banks.

### DIFF
--- a/bin/workflows/pycbc_make_bank_verifier_workflow
+++ b/bin/workflows/pycbc_make_bank_verifier_workflow
@@ -33,7 +33,7 @@ import pycbc.version
 import pycbc.workflow as wf
 import pycbc.workflow.pegasus_workflow as pwf
 from pycbc.results import create_versioning_page, static_table, layout
-from pycbc.workflow import LalappsInspinjExecutable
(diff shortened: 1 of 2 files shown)


pannarale:
@a-r-williamson, I wanted you to look at the jobsetup.py change since you wrote that part up.  I am basically avoiding that we loose track of the tags in there and append a new tag for bright/dim.

@spxiwh, the rest is for you and I was as uninvasive as possible, so this does not break standard bank verifier usage.  For a very quick, click around [here](https://geo2.arcca.cf.ac.uk/~spxfp1/LVC/O3/BANKS/PyGRB/EarlyO3_PSD_maxNSpin_0p05_maxBHmass_25_IMRPhenomD_27-1000Hz_grb_uberbank/pycbc_bank_verify_short_test_bns_nsbh/2._broad_injection_sets/2.02_broadinjnsbh1/).  In particular, I linked to the nsbh injections which you will see are (potentially) EM-Bright only.

Thanks to both!

pannarale:
@a-r-williamson are you happy with the change that extends the tags?

PR #2830, opened 2019-07-04
add 2-ifo phase/time/amp consistency to coinc rate stat

tdent:


tdent:
@ahnitz this seems to work minimally in terms of not breaking previously working multi-ifo behaviour.

tdent:
@GarethDaviesGW I'm not looking for a full review here, just to check changes that affect the coinc-noise-rate stat and the network-sensitivity stat (sigma etc.)

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
```diff
@@ -175,27 +176,39 @@ num_templates = len(h5py.File(args.template_bank, "r")['template_hash'])
...
-                                               args.segment_name,
-                                               args.veto_files)
-    trigs[reader.ifo] = reader
+                            args.template_bank,
+                            args.segment_name,
+                            args.veto_files)
+    ifo = reader.ifo
+    trigs.ifos.append(ifo)
+    # time shift is subtracted from pivot ifo time
+    trigs.to_shift.append(-1 if ifo == args.pivot_ifo else 0)
```
ahnitz: Ok, so if I understand, this tells which way to shift a particular trigger time. i.e. time + DIRECTION * SLIDE. Where direction can be 0 or -1 in this case. Makes sense. (might also be extendable in the future which is also good).

pycbc/events/stat.py
```diff
@@ -198,52 +210,82 @@ def coinc(self, s0, s1, slide, step): # pylint:disable=unused-argument
...
-        # Normalize so that peak has no effect on newsnr
-        self.hist = self.hist / float(self.hist.max())
-        self.hist = numpy.log(self.hist)
 
-        # Bin boundaries are stored in the hdf file
-        self.tbins = self.files['phasetd_newsnr']['tbins'][:]
-        self.pbins = self.files['phasetd_newsnr']['pbins'][:]
-        self.sbins = self.files['phasetd_newsnr']['sbins'][:]
-        self.rbins = self.files['phasetd_newsnr']['rbins'][:]
+    def __init__(self, files, ifos):
```
ahnitz: ifos should still be optional here to not break the two-ifo workflow and be consistent with the Stat class definition.

pycbc/events/stat.py
```diff
@@ -198,52 +210,82 @@ def coinc(self, s0, s1, slide, step): # pylint:disable=unused-argument
...
+        NewSNRStatistic.__init__(self, files, ifos)
```
ahnitz: Should try to be consistent and use ifos=ifos here.

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
tdent: yes, or +1 (for some reason the 2-ifo coinc code seems to be written to positively shift ifo1, at least that's what I deduced from the way the time difference was done in the 2-ifo pta stat)

pycbc/events/stat.py
tdent: ok .. yes, this makes sense, I did test the 2-ifo case but probably some 'semi-random' combination of defaults saved it from erroring

GarethDaviesGW:
> @GarethDaviesGW I'm not looking for a full review here, just to check changes that affect the coinc-noise-rate stat and the network-sensitivity stat (sigma etc.)

They look good to me, and the list-of-tuples thing makes sense as an alternative to the dictionary for singles info. When I have tested before with the 'only use pta if using HL' bit, it worked and it doesn't look like that's changed, but I guess this will be changed anyway once the HV/LV files are made

tdent:
Codeclimate is only 'similar blocks of code' (which all refer to comment strings!!) or 'too many arguments' (for a method that was already present with the same number of arguments). 
@ahnitz @spxiwh I think this is ready to be merged, has been tested in a full workflow

tdent:
Note that the choice to use the HL signal hist for HLV coincs is not necessarily the best simple approximation, for instance one could evaluate all 3 histograms and take a minimum or geometric mean over them. HL only does not take advantage of the amplitude test between V and other ifos which can suppress the majority of background

PR #2831, opened 2019-07-05
exp_fit_sg_coinc_rate stat Inf issue

GarethDaviesGW:
performing exp of the log coinc rate (single statistic) in the coincident calculation (`coinc_multiifo` function) was causing numerical underflow with high-statistic singles

This PR changes the calculation to retain the log rates in order to prevent this

I do not know if we should remove the `combination_noise_coinc_rate` function from the `coinc_rate` module, or leave it, or rewrite to be a wrapper the log version (suggestions welcome)

### DIFF
--- a/pycbc/events/coinc_rate.py
+++ b/pycbc/events/coinc_rate.py
@@ -82,12 +82,11 @@ def combination_noise_coinc_rate(rates, slop):
         Value is expected coincidence rate in the combination, units Hz
     """
     # multiply product of trigger rates by the overlap time
-    allowed_area = multiifo_noise_coincident_area(list(rates), slop)
-    # list(dict.values()) is python-3-proof
-    rateprod = numpy.prod(list(rates.values()), axis=0)
-    combo_coinc_rate = allowed_area * rateprod
+    log_rates = {k:numpy.log(r) for (k,r) in rates.items()}
+    combo_coinc_rate = combination_noise_coinc_rate_log(log_rates, slop)
 
     return combo_coinc_rate
+
 
 def combination_noise_coinc_rate_log(log_rates, slop):
     """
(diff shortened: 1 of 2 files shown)


pycbc/events/coinc_rate.py
```diff
@@ -82,10 +82,38 @@ def combination_noise_coinc_rate(rates, slop):
...
+        time added to maximum time-of-flight between detectors to account
+        for timing error
+
+    Returns
+    -------
+    combo_coinc_rate: numpy array
+        Value is expected coincidence rate in the combination, units Hz
+    """
+    # multiply product of trigger rates by the overlap time
+    allowed_area = numpy.log(multiifo_noise_coincident_area(list(log_rates),
```
tdent: would prefer to calculate allowed area as before then take the log in the line that calculates combo_coinc_rate (and maybe call that lograte rather than rate).

pycbc/events/coinc_rate.py
```diff
@@ -82,10 +82,38 @@ def combination_noise_coinc_rate(rates, slop):
         Value is expected coincidence rate in the combination, units Hz
     """
     # multiply product of trigger rates by the overlap time
-    allowed_area = multiifo_noise_coincident_area(list(rates), slop)
+    log_rates = {k: numpy.log(r) for (k, r) in rates.items()}
+    combo_coinc_rate = combination_noise_coinc_rate_log(log_rates, slop)
```
tdent: don't you want to do numpy.exp on this? otherwise it will return the log of a rate.

pycbc/events/coinc_rate.py
GarethDaviesGW: Done, though instead of changing the name of `combo_coinc_rate` it now simply returns numpy.log(allowed_area) + rateprod

tdent:
ok, so since the non-log expected noise rate calc still exists but can be liable to underflows, should maybe add a logging warning to this function.

tdent:
i think we should also change `multiifo_noise_coinc_rate()` to use the log rate basic engine and to return log rates, in case it's needed in future

PR #2833, opened 2019-07-05
don't perform unnecessary call to astropy when a detector object is initialized

tdent:


pycbc/detector.py
```diff
@@ -82,21 +83,29 @@ def __init__(self, detector_name, reference_time=1126259462.0):
...
+            self.gmst_reference = gmst_accurate(self.reference_time)
+        else:
+            raise RuntimeError("Can't get accurate sidereal time without GPS "
+                               "reference time!")
 
     def gmst_estimate(self, gps_time):
         if self.reference_time is None:
             return gmst_accurate(gps_time)
 
+        self.set_gmst_reference()
```
ahnitz: This should only be called if the gmst_reference isn't already set. We don't want to call gmst_accurate more than once if the reference time is provided.

ahnitz: I think you may just need another if statment here that checks if the gmst_reference is already set or not, so that it doesn't reset it each time.

PR #2834, opened 2019-07-05
Add statmap inj

GarethDaviesGW:
Make modifications to add_statmap to allow for injection inputs

works by including coinc_statmap files from full_data as an extra input `--background-files`
These are then used for the background

some small changes to add_statmap also included, and an option to output the ifo_combination

This is not yet within the workflow
EDIT: this is now included in the workflow

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_add_statmap
+++ b/bin/hdfcoinc/pycbc_multiifo_add_statmap
@@ -98,17 +98,17 @@ for ifo in all_ifos:
 for f_in in files:
     for ifo in all_ifos:
         if ifo in f_in['foreground']:
-            all_trig_times[ifo] = np.concatenate([all_trig_times[ifo], \
+            all_trig_times[ifo] = np.concatenate([all_trig_times[ifo],
                                     f_in['foreground/{}/time'.format(ifo)][:]])
             all_trig_ids[ifo] = np.concatenate([all_trig_ids[ifo],
                               f_in['foreground/{}/trigger_id'.format(ifo)][:]])
         else:
             all_trig_times[ifo] = np.concatenate([all_trig_times[ifo],
-                                 -1*np.ones_like(f_in['foreground/fap'][:],
-                                                    dtype=np.uint32)])
+                                 -1 * np.ones_like(f_in['foreground/fap'][:],
+                                                   dtype=np.uint32)])
             all_trig_ids[ifo] = np.concatenate([all_trig_ids[ifo],
-                                 -1*np.ones_like(f_in['foreground/fap'][:],
-                                                    dtype=np.uint32)])
+                                 -1 * np.ones_like(f_in['foreground/fap'][:],
+                                                   dtype=np.uint32)])
 n_triggers = f['foreground/ifar'].size
 all_combo_types = []
 if args.output_combination_information:
@@ -148,6 +148,7 @@ def filter_dataset(h5file, name, idx):
     del h5file[name]
     h5file[name] = filtered_dset
 
+
 # Downsample the foreground columns to only the loudest ifar between the
 # multiple files
 for key in f['foreground'].keys():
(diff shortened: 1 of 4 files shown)


tdent:
I don't think we want to produce or store 'foreground veto' segments for injection events, so only do this in the 'full data' mode.

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -11,6 +11,17 @@ import pycbc.conversions as conv
 from pycbc.events import coinc
 from ligo import segments
 
+def get_ifo_attr(fi):
```
tdent: put 1 line of comment saying what it does, ie what the return value is supposed to be

tdent: also, since in one case it's not an attr may want to call this something else

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -11,6 +11,17 @@ import pycbc.conversions as conv
...
+    if 'ifos' in fi.attrs:
```
tdent: maybe this should be a try-except, if eventually we want to be in a position where every input file has the expected attrs entry and only obsolete ones need the workaround

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -93,12 +112,27 @@ for f_in in files:
             all_trig_ids[ifo] = np.concatenate([all_trig_ids[ifo],
                                  -1 * np.ones_like(f_in['foreground/fap'][:],
                                                    dtype=np.uint32)])
+n_triggers = f['foreground/ifar'].size
+all_combo_types = []
```
tdent: is `all_combo_types` used for anything?

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -23,7 +34,11 @@ parser.add_argument('--veto-window', type=float, default=0.1,
     help="Time around each zerolag trigger to window out [default=.1s]")
 parser.add_argument('--cluster-window', type=float,
     help="Maximum time interval to cluster coincident events")
+parser.add_argument('--output-combination-information', action='store_true')
```
tdent: need a help msg for this.

bin/hdfcoinc/pycbc_multiifo_add_statmap
GarethDaviesGW: This was a hangover from the version which did not use the mask

GarethDaviesGW:
> For the injection case, the 'background/stat' calculation now contains background from loud coincs in zerolag forming coincs in time slides (i.e. 'little dogs'), hence is influenced by the content of zerolag. It used to be that the injection 'inclusive FAR' calculation was from the background from _each injection_ slid against full data. Since the non-inclusive calculation here is not that, and since it indirectly contains zerolag signal info, I would prefer _not_ to calculate or store it. I.e. just calculate and store the `far_exc` for injections and nothing else. This may break some things downstream but they 'ought to' be broken rather than displaying incorrect 'inclusive background' for injections.

I have updated so that now only stores ifar_exc and fap_exc in injection cases

pycbc/workflow/coincidence.py
```diff
@@ -746,18 +763,29 @@ def setup_multiifo_combine_statmap(workflow, final_bg_file_list, out_dir, tags):
...
-                                            ifos=workflow.ifos,
-                                            tags=tags,
-                                            out_dir=out_dir)
+    cstat_exe_name = os.path.basename(workflow.cp.get("executables", "combine_statmap"))
+    if cstat_exe_name == 'pycbc_multiifo_combine_statmap':
+        cstat_class = PyCBCMultiifoCombineStatmap
+    elif cstat_exe_name == 'pycbc_multiifo_add_statmap':
+        cstat_class = PyCBCMultiifoAddStatmap
+    else:
+        raise NotImplementedError('executable should be combine_statmap or add_statmap')
```
tdent: this is a bit unclear, can you put the full executable names

PR #2837, opened 2019-07-15
Don't upimport pygrb results utils

ahnitz:
The pygrb plotting module is setting parts of rcparams and is also setting up logging directly in the import. The former has minor unintended affects on plots in other workflows, the latter hides the logging output from other programs entirely (i.e. the standard 2-ifo coinc workflow). I can't verify this completely, but it seems that many modules are directly importing this module anyway, so I don't think it's necessarily needed to upimport everything. Can someone on the PyGRB side verify? 

i.e. this wouldn't need to change.
https://github.com/gwastro/pycbc/blob/07c32dc38591b324cd34971beea55de2a0989377/bin/pygrb/pycbc_pygrb_plot_skygrid#L32

### DIFF
--- a/pycbc/results/__init__.py
+++ b/pycbc/results/__init__.py
@@ -7,4 +7,3 @@ from pycbc.results.legacy_grb import *
 from pycbc.results.layout import *
 from pycbc.results.dq import *
 from pycbc.results.str_utils import *
-from pycbc.results.pygrb_plotting_utils import *


pannarale:
Hi @ahnitz, instead of changing __init__.py, can I change pygrb_plotting_utils and the imports in the (4, I think) scripts that rely on it?

ahnitz:
@pannarale The imports in the scripts I think are fine and don't seem to require this to be uimported atm anyway. 

I don't mind which you want to go here, but I think what you'd have to change is to remove all uses of matplotlib not within a function and to avoid starting up any logging just by importing the module.

pannarale:
Okay, lines 60 and 62 can go in the single plotting functions, no problem.  Do 54 and 55 have to go, too?  They are similar to https://github.com/gwastro/pycbc/blob/3acd4dd8847182ed196c2eb07aba854f7a15da0d/pycbc/results/followup.py#L34, and in an if statement so perhaps these are not causing issues the way they are.

Regarding the logging, is https://github.com/gwastro/pycbc/blob/master/pycbc/results/versioning.py giving you problems as well?  It's doing the same logging import.

ahnitz:
@pannarale I'm not sure about that first matplotlib line, it may be fine. As for the logging, just importing the module is fine. It was the fact that it was being called from within the module outside a function. The issue is if this happens before it can be manually configured within a script. The script settings get overridden.

pannarale:
Gotcha, @ahnitz.  I'll fix this after some teaching work.

ahnitz:
@pannarale In the meantime, I think we'll mere this PR, but feel free to propose how to re-enable this with the suggested changes.

Issue #2835, opened 2019-07-08
pycbc_page_foundmissed assumes 'ifos' attr


tdent:
Running recent master version on an old 2-ifo injfind file I get 
`KeyError: "Can't open attribute (can't locate attribute: 'ifos')"`

PR #2838, opened 2019-07-15
add support for 2-ifo case in fm plot

ahnitz:
This may address #2835


ahnitz:
The old code took an ifos argument, which we no longer want to support. The old behavior was also to just assume H-L in this case, which is what this patch adds back in. It provides at least the most likely minimal support for the 2-ifo workflow now.

PR #2842, opened 2019-07-16
multiifo statmap setup: don't set up cluster window twice

tdent:
the cluster window setting is automatically copied over from the ini file section anyway, don't need to set it via explicit extra workflow code.


tdent:
(currently looking at the .sh file for add statmap jobs, the cluster-window option appears twice)

PR #2850, opened 2019-07-19
add in sigma variation into statistic

GarethDaviesGW:


pycbc/events/stat.py
```diff
@@ -599,12 +603,75 @@ def reassign_rate(self, ifo):
...
+                logging.info('Applying HL phase/time/amp consistency')
+            # makeshift factor to compensate HL(V) coincs which are penalized
+            # on average by p/t/a vs HV, LV which are not
+            logr_s = logr_s + 4.5
+        else:
+            logr_s = numpy.zeros_like(s['V1']['snr'])
+            if len(logr_s):
+                logging.info('HV or LV coinc, no phase/time-amp/consistency')
+        loglr = logr_s - ln_noise_rate + self.benchmark_lograte \
+                     + network_sigma - s[ifos[0]]['benchmark_sigma']
```
tdent: it's not clear to me why this is getting the 'benchmark_sigma' via the singles information - which looks as if it might be ifo-dependent - as far as I can see, benchmark_sigma is a function only of the template not the ifo. 
I guess if there is no way for the coinc() method to directly tell which template it is dealing with this is a reasonable workaround, but will need a bit of commenting.

pycbc/events/stat.py
```diff
@@ -599,12 +603,75 @@ def reassign_rate(self, ifo):
...
 
+class ExpFitSGFgBgRateStatistic(PhaseTDStatistic, ExpFitSGBgRateStatistic):
+    def __init__(self, files):
+        # read in background fit info and store it, also use newsnr_sgveto
+        ExpFitSGBgRateStatistic.__init__(self, files)
+        # Use PhaseTD statistic single.dtype
+        PhaseTDStatistic.__init__(self, files)
+        self.single_dtype.append(('benchmark_sigma', numpy.float32))
+        self.benchmark_sigma = 3.0 * numpy.log(numpy.amin([self.fits_by_tid[ifo]['median_sigma']
+                                   for ifo in ['H1', 'L1']], axis=0))
```
tdent: this needs to be unpacked a bit (split into 2 lines, eg the first one obtains the sigmas we want, second one converts them into log of estimated network sensitive volume) and commented
also, I would name it something like 'benchmark_logvol' since it's the log of a sensitive volume

pycbc/events/stat.py
```diff
@@ -599,12 +603,75 @@ def reassign_rate(self, ifo):
...
+    def coinc_multiifo(self, s, slide,
+                       step, **kwargs): # pylint:disable=unused-argument
+        sngl_rates = {ifo: sngl_data['snglstat'] for ifo, sngl_data in
+                      s.items()}
+        ln_noise_rate = coinc_rate.combination_noise_lograte(
+                                  sngl_rates, kwargs['time_addition'])
+
+        network_sigma = 1.5 * numpy.log(numpy.amin(
+                            [s[ifo]['sigmasq']
+                             for ifo in s.keys()], axis=0))
```
tdent: again, prefer to split into 2 lines for readability. first line obtains the minimum of sigma(sq) over ifos for the given triggers in each coinc, second converts it to the log of a sensitive volume -> name it 'network_logvol' or similar

pycbc/events/stat.py
GarethDaviesGW: The second one: this was the only way I could see to get the template ID into the `multiifo_coinc()` function. Comment has been added

pycbc/events/stat.py
```diff
@@ -599,10 +599,77 @@ def reassign_rate(self, ifo):
...
+    def __init__(self, files):
+        # read in background fit info and store it, also use newsnr_sgveto
+        ExpFitSGBgRateStatistic.__init__(self, files)
+        # Use PhaseTD statistic single.dtype
+        PhaseTDStatistic.__init__(self, files)
+        for ifo in self.ifos:
+            self.assign_median_sigma(ifo)
+        self.single_dtype.append(('benchmark_logvol', numpy.float32))
+        hl_net_med_sigma = numpy.amin([self.fits_by_tid[ifo]['median_sigma']
+                                       for ifo in ['H1', 'L1']], axis=0)
```
tdent: maybe explain that this is an array over template IDs, i.e. a template-dependent benchmark sensitivity

pycbc/events/stat.py
```diff
@@ -599,10 +599,77 @@ def reassign_rate(self, ifo):
...
+
+    def coinc_multiifo(self, s, slide,
+                       step, **kwargs): # pylint:disable=unused-argument
+        sngl_rates = {ifo: sngl_data['snglstat'] for ifo, sngl_data in
+                      s.items()}
+        ln_noise_rate = coinc_rate.combination_noise_lograte(
+                                  sngl_rates, kwargs['time_addition'])
+
+        network_sigmasq = numpy.amin([s[ifo]['sigmasq'] for ifo in s.keys()],
+                                     axis=0)
```
tdent: explain e.g. 'sensitivity for a given coinc type is approximately determined by the least sensitive ifo'

Issue #2824, opened 2019-07-02
Average time in multiifo statmap inj may not be correct


tdent:
See https://github.com/gwastro/pycbc/blob/master/bin/hdfcoinc/pycbc_multiifo_coinc_statmap_inj#L98 
is this correct in cases where there may be ifo-time = -1 sentinel values?

ahnitz:
@tdent I think you may be right. The effect this should have is that injections aren't really going to get their 'inclusive' background calculated correctly (i.e. not at all really). It may not be obvious in the current output results that anything is amiss, but my expectation is that some injections will have identical ifar and ifar_exc valuees.

GarethDaviesGW:
Are we looking at the specific linked line `ifar_exc = background_time / (fnlouder_exc + 1)` here or at the mean calculation a few lines after

The mean calculation later should not have any of the -1 values as the statmap_inj is only acting on the specific IFO combination, or am I missing something?

Otherwise I don't see how the linked line will be affected by the -1 times

tdent:
Gareth, the link was put there on Jul 3, for the version of the code at that point

tdent:
anyway, yes, of course it was referring to a calculation of mean times `ftimes = numpy.concatenate(ifotimes).mean(axis=0)`

tdent:
if you don't think this is a problem (i.e. you *know* that there will be no -1 values at this point) then can close this issue

GarethDaviesGW:
Okay, that's what I thought, but as I couldn't see any issue with the calculation (No -1s should be present at this point), I'd confused myself

PR #2851, opened 2019-07-22
Inj exc ifar

tdent:
Temporarily suspending the calculation of inclusive IFAR for injections in the multi-ifo pipeline while we work out what to do with full data/zerolag.  (Also since as noted in #2824 the calculation is probably already buggy with existing code.)

These changes are needed to not break the pipeline given the absence of inclusive IFAR, but also preserve existing 2-ifo behaviour.


tdent:
This is *needed* for the multi-ifo pipeline, please look and review.

PR #2852, opened 2019-07-23
select phasetd file based on ifos

ahnitz:
*Untested*

This should select the two-ifo phase td file based on the ifos used for the phasetd class.


ahnitz:
@tdent Ok, this now also pass through the list of active ifos and the pivot to the statistic class, which is needed for multi-ifo statistics.

tdent:
Subsumed in #2830

PR #2855, opened 2019-07-23
pycbc_make_inference_workflow: optional file-path options

### LINKED ISSUE #2739: pycbc_make_inference_workflow : superfluous commandline options for auto-generated files

dbkeitel:
This supersedes #2741 after feedback from @cmbiwer , not removing any of the 3 options, but making them optional instead. By default, --output-file will be set to workflow_name.dax, while auto-naming for the map and transform catalog are handled by the Workflow.save() method already so they can just stay "None" by default.

```
pycbc_make_inference_workflow: optional file-path options

 -normally, users shouldn't need to worry about filenames of
  --output-map, --transformation-catalog and --output-file
 -will be put in default --output-dir with default names automatically
 -but can still be manually set if required
 -closes #2739
```

### DIFF
--- a/bin/workflows/pycbc_make_inference_workflow
+++ b/bin/workflows/pycbc_make_inference_workflow
@@ -77,11 +77,11 @@ parser.add_argument("--tags", nargs="+", default=[],
                     help="Tags to apend in various places.")
 parser.add_argument("--output-dir", default=None,
                     help="Path to output directory.")
-parser.add_argument("--output-map", required=True,
+parser.add_argument("--output-map", default=None,
                     help="Path to output map file.")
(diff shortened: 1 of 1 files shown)


cmbiwer:
This looks good.

PR #2856, opened 2019-07-23
add --verbose option to pycbc_make_inference_workflow

dbkeitel:
-will simply be passed through to main inference executable

@cmbiwer @cdcapano another fairly simple change we found useful to make runs easier to debug. The extra output will end up in one of the files like workflow/submitdir/work/runname_inference-main_ID0000001.000/inference-H1L1_ID0_ID0000001.out.0xx if I remember correctly.

### DIFF
--- a/bin/workflows/pycbc_make_inference_workflow
+++ b/bin/workflows/pycbc_make_inference_workflow
@@ -113,6 +113,10 @@ parser.add_argument("--frame-files", nargs="+", default=None,
                     action=MultiDetOptionAppendAction,
                     help="GWF frame files to use.")
 
+# verbosity flag (to be passed through to main inference executable)
+parser.add_argument("--verbose", action="store_true",
+                    help="Turn on --verbose for main pycbc_inference executable.")
+
 # add option groups
 configuration.add_workflow_command_line_group(parser)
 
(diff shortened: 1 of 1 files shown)


cmbiwer:
Instead of making it part of the workflow generator. Why not just use in the configuration file:
```
[inference]
verbose =
```

dbkeitel:
Hmm, good question! Probably just because it seemed intuitive to me to expect it as a CLA... Let me check if what you suggest did for any obscure reason not work for us. If it works, I'll close the PR myself.

cmbiwer:
Or consider adding the option to an example configuration file, e.g.: https://github.com/gwastro/pycbc/blob/master/examples/workflow/inference/workflow_config.ini

ahnitz:
I agree with @cmbiwer here since it's a command line option to the inference job and not really the workflow generator.

ahnitz:
@dbkeitel p.s. if you are in the position of wanting logging on all the time, you may be interested to know that you can toggle on / off logging at any time with pycbc executables by sending the python program a SIGUSR1 signal i.e. kill -SIGUSR1 PID_OF_PYTHON_PROGRAM

dbkeitel:
@ahnitz Didn't know that, sounds quite useful, thanks!

PR #2887, opened 2019-08-02
Make normalization term optional and save data instead of whitened data

cdcapano:
This makes the normalization term that was introduced in #2795 optional, as it sometimes can cause numerical issues for nested samplers. By default, the normalization is set to False now. To turn it on, you set `normalize =` in the model part of the configuration file.

This also makes it so that the original data is stored to the output file, rather than the whitened data. It also fixes some documentation issues that were introduced in #2795.

### DIFF
--- a/pycbc/inference/models/gaussian_noise.py
+++ b/pycbc/inference/models/gaussian_noise.py
@@ -27,6 +27,7 @@ from pycbc.strain import gates_from_cli
 from pycbc.strain.calibration import Recalibrate
 from pycbc.inject import InjectionSet
 
+from .base import ModelStats
 from .base_data import BaseDataModel
 from .data_utils import (data_opts_from_config, data_from_cli,
                          fd_data_from_strain_dict, gate_overwhitened_data)
@@ -42,7 +43,7 @@ class GaussianNoise(BaseDataModel):
 
         \log p(d|\Theta, h) &=  \log\alpha -\frac{1}{2} \sum_i
             \left< d_i - h_i(\Theta) | d_i - h_i(\Theta) \right> \\
-        \log p(d|n) &= -\log\alpha -\frac{1}{2} \sum_i \left<d_i | d_i\right>
+        \log p(d|n) &= \log\alpha -\frac{1}{2} \sum_i \left<d_i | d_i\right>
 
     where the sum is over the number of detectors, :math:`d_i` is the data in
     each detector, and :math:`h_i(\Theta)` is the model signal in each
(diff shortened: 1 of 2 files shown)


cdcapano:
@sum33it poke

cdcapano:
@sum33it poke again

sum33it:
Hi @cdcapano ! Thanks for doing this. I was checking (un)normalized likelihood values from your PR with (previous) current master versions. Result files from each case are on following location:

https://www.atlas.aei.uni-hannover.de/work/sumit.kumar/inference/check_normalization/fast_runs/

See the following files:
inference_normalize_pr.hdf  
inference_unnormalize_pr.hdf
inference_unnormalize_gwastro_master.hdf  
inference_unnormalize_older_version.hdf  

It seems that though the template is same but likelihood values are different. Could you check it once more?

cdcapano:
@sum33it Thank you for checking this. There was indeed a bug in the patch. I had switched to storing the whitened data as `_whitened_data`, while `_data` was the normal data (with no whitening). However, I forgot to update the `MarginalizedPhaseGaussianNoise` model for this change. So, when the marginalized phase model was calculating the overlap of h with the data, it was actually doing whitened h with normal data, instead of whitened h with whitened data, as it should be.

I just pushed a commit to fix that. I tested on one of the points from your run to check that it gave the same loglikelihood as before and after the patch, and it did. Still, you may want to double check.

cdcapano:
Rebased to master to try to fix Travis issues.

cdcapano:
Getting around to finishing some old PRs... @sum33it, can you review this? Was there some other issue I had to fix?

cdcapano:
@sum33it Ok, I've updated this to current master. I looked into why you were getting different results for the normalized value before and after this patch. It turns out there's a bug in current master; the log of the noise likelihood (`lognl`) is wrong. Details are below. Long story short, when the `lognl` is initially calculated, it assumes that the stored data has been whitened. But the data isn't whitened until _after_ `lognl` is calculated. The result is that `lognl` is extremely small, making the log likelihood be off by a (large) constant. This is why the log likelihoods are not matching up before/after this patch. By storing whitened data as a separate attribute (when the psds are saved), this patch fixes that bug. (I've added a bug label to this PR for this reason.)

I've now tested this both with `GaussianNoise` and `MarginalizedPhaseGaussianNoise` on a few points using GW150914 data. I compared this patch to current master before this patch, and master from before the normalization term was added (and the bug above introduced) back in July. I get the same log likelihood ratio for all tests, as expected. I also get the same loglikelihood when I turn off normalization in the current patch as I did from the code in July. I think this is ready to be merged now.

------

Details of the bug:

The issue is that `lognl` is currently being calculated when the PSDs are set. That happens at [line 288](https://github.com/cdcapano/pycbc/blob/master/pycbc/inference/models/gaussian_noise.py#L288) in `GaussianNoise.__init__`. That calls the function at [line 361](https://github.com/cdcapano/pycbc/blob/master/pycbc/inference/models/gaussian_noise.py#L361) which eventually calls [det_lognl](https://github.com/cdcapano/pycbc/blob/master/pycbc/inference/models/gaussian_noise.py#L361), where the inner product is calculated ([line 559](https://github.com/cdcapano/pycbc/blob/master/pycbc/inference/models/gaussian_noise.py#L559)). Notice that the inner product is being calculated without applying any weight. This is because it was assumed that the data had been whitened at this point. But that happens at [line 290](https://github.com/cdcapano/pycbc/blob/master/pycbc/inference/models/gaussian_noise.py#L290), which is after the point that this entire block of code was called (by setting the PSDs).

That calls:

sum33it:
@cdcapano Looks good to me.

PR #2889, opened 2019-08-08
PyCBC Live autogating fixes

titodalcanton:
This makes PyCBC Live save autogates into the output HDF5 files for posterity. It also makes the autogating options consistent with the offline search code and documented.


titodalcanton:
This is running on replay data now.

PR #2897, opened 2019-08-16
Fix issue with invalid checkpoint file raising error

cdcapano:
PyCBC Inference has a `--force` option that is supposed to check if the output file exists, and raising an error if it does (unless the user supplies `--force`). However, currently, the check is done on the checkpoint file rather than the output file. This causes two bugs. First, the output file will be overwritten even if you have not supplied `--force`. Second, if an invalid checkpoint file exists, it will cause the code to quit unless `--force` is supplied. This isn't what we want. If an invalid checkpoint file exists, it should just be overwritten. This has caused issues when running in a dax. If the job fails before the first checkpoint, then on restart the job fails again because it finds a checkpoint file without anything in it (i.e., an invalid checkpoint) then raises an error if `--force` isn't supplied. It should always just create a fresh checkpoint file in that case.

This patch fixes this by moving the check for the output file into `setup_output` so that the output file is the file that is actually being checked for. It also results in invalid checkpoint files being overwritten, as desired.

### DIFF
--- a/pycbc/inference/sampler/base.py
+++ b/pycbc/inference/sampler/base.py
@@ -166,6 +166,13 @@ class BaseSampler(object):
         injection_file : str, optional
             If an injection was added to the data, write its information.
         """
+        # check that the output file doesn't already exist
+        if os.path.exists(output_file):
+            if force:
+                os.remove(output_file)
+            else:
+                raise OSError("output-file already exists; use force if you "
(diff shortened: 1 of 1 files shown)


sum33it:
Looks good to me. I have also checked it with a backup and checkpoint file present.

PR #2899, opened 2019-08-16
add Ian's autogating iterability feature, add command line option to …

ahnitz:
…invoke, store gating info


ahnitz:
@spxiwh This copies over your patch, but tries to fix the record of gating times and adds command line options for number of gating iterations. Also it will auto-stop if no more times are found to gate.

PR #2911, opened 2019-08-30
added casting to float

cieslar:
Without the cast, the cos can't process `lal.LIGOTimeGPS` type:
```
  File "/.../python3.7/site-packages/pycbc/detector.py", line 141, in antenna_pattern
    cosgha = cos(gha)
AttributeError: 'lal.LIGOTimeGPS' object has no attribute 'cos'
```
Error encountered working with python 3.7 conda based dist, pycbc installed from pip.

### DIFF
--- a/pycbc/detector.py
+++ b/pycbc/detector.py
@@ -146,7 +146,7 @@ class Detector(object):
         fcross: float or numpy.ndarray
             The cross polarization factor for this sky location / orientation
         """
-        if np.isscalar(t_gps):
+        if isinstance(t_gps, lal.LIGOTimeGPS):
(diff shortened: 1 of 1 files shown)


ahnitz:
@cieslar Thank you for this patch! I've made one minor change so that the float case is only around the gmst_time estimate. We support RA being a vector which was failing in the unittests. Assuming that now passes, we'd be happy to merge this.

ahnitz:
@cieslar I had to make some modifications so that your patch would respect the ability to pass in arrays in addition to single values. Let  us know if you run into any other issues, and thanks for the PR!

PR #2921, opened 2019-09-04
PyCBC Live: fix issue in how the detector state is reported

titodalcanton:
When (some of) the worker nodes think a detector is unusable, but the master node thinks the opposite, the detector is still reported as being active despite no triggers are produced from it. This patch ensures that the detector is reported as inactive when there are no (or partial) results from it.

Also fixes a harmless confusion between objects in different namespaces.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -176,7 +176,7 @@ class LiveEventManager(object):
         """
         #check for hardware injection
         for ifo in ifos:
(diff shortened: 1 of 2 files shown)


titodalcanton:
@tdent here I also remove the log info we discussed.

titodalcanton:
This is now tested.

PR #2922, opened 2019-09-04
fix trigger id bug in mult find trigs

ahnitz:


ahnitz:
The issue is that when a single template produces many triggers it handles them in chunks. The trigger ids weren't be properly calculated in this case for the multi-ifo coinc case. It was being handled correctly for the two-ifo code.

tdent:
was the bug just the (lack of) assignments of tids in the batch processing? 

Under what exact conditions does it come up - if there are >5000 triggers in a given template? 

That sort of count would give a total trigger merge file size of about 2e9 triggers, which is indeed larger than the ones we have with settings used for O2 multi-ifo dev runs.

ahnitz:
Yes, if a template had > 5000 triggers (or whatever the single batch size was set to) it would run into this problem for everything past the first batch. The issue was that it was always indexing the full array of trigger ids which may be offset from the batch itself.

PR #2931, opened 2019-09-11
fix to get new opt snr datasets

tdent:


tdent:
has been minimally tested (run by hand & inspection of resulting opt SNRs)

PR #2935, opened 2019-09-11
fix inj snrchisq plot

tdent:
Plot was broken because some chisq values were zero or negative (?) 
Fix and do a bit of tweaking of symbol sizes and transparency, test output here 
image


tdent:
all 3 FIXMEs are for multi-ifo compatibility, which this is not yet addressing.

tdent:
adding more reviewers since this has been up for >10 days

Issue #2944, opened 2019-09-16
python3 "KeyError: b'posterior_file'" from pycbc.inference.io.loadfile()


dbkeitel:
I have a pycbc posterior file created under 1.13.8. Using 1.14.1 with python2, it still loads fine. However, with python3:
```
from pycbc.inference.io import loadfile
fp = loadfile('test.hdf','r')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/cvmfs/ligo-containers.opensciencegrid.org/lscsoft/conda/latest/envs/ligo-py36/lib/python3.6/site-packages/pycbc/inference/io/__init__.py", line 105, in loadfile
    fileclass = get_file_type(path)
  File "/cvmfs/ligo-containers.opensciencegrid.org/lscsoft/conda/latest/envs/ligo-py36/lib/python3.6/site-packages/pycbc/inference/io/__init__.py", line 70, in get_file_type
    return filetypes[filetype]
KeyError: b'posterior_file'
```

Looks like one of those pesky byte string conversion issues to me: it seems to want to use a read-from-file byte string as a key to a dict which expects regular strings.

dbkeitel:
Generalised the title of this issue, as after trying to fix the first KeyError, at least two more byte-string-related errors crop up later in loadfile() and read_samples().

dbkeitel:
@spxiwh I'm absolutely with you on the hating! But thanks for the details, and I now also found a "support for python 3 in almost all core modules (inference and tmpltbank not included yet)" warning in the [1.14.0 notes](https://zenodo.org/record/3247679) which unfortunately wasn't repeated explicitly for 1.14.1... ;)

PR #2945, opened 2019-09-16
inference: decode bytes->str for filetype from hdf

dbkeitel:
@spxiwh @cdcapano This seems sufficient to allow me to read posteriors with python3, but not sure if it's "the right way (tm)" to fix the issue.

Note also when cloning from current master and trying to import from within the source tree, I ran into several unrelated import issues, so I haven't tested this in any more detail yet than the simple isolated two-liner from #2944.

### DIFF
--- a/pycbc/inference/io/__init__.py
+++ b/pycbc/inference/io/__init__.py
@@ -69,7 +69,7 @@ def get_file_type(filename):
         if filename.endswith(ext):
             with _h5py.File(filename, 'r') as fp:
                 filetype = fp.attrs['filetype']
-            return filetypes[filetype]
+            return filetypes[filetype.decode()]
     for ext in txt_extensions:
         if filename.endswith(ext):
             return InferenceTXTFile
--- a/pycbc/inference/io/base_hdf.py
+++ b/pycbc/inference/io/base_hdf.py
@@ -74,7 +74,7 @@ class BaseInferenceFile(h5py.File):
                 self.attrs['filetype'] = filetype
             else:
                 filetype = None
-        if filetype != self.name:
+        if filetype.decode() != self.name:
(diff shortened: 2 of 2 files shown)


dbkeitel:
Well ok, CI failed:
```
2019-09-16 17:12:20,995 Creating file inference.hdf.checkpoint

  File "/home/travis/virtualenv/python3.6.7/bin/pycbc_inference", line 210, in <module>

    injection_file=injection_file)

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pycbc/inference/sampler/base.py", line 189, in setup_output

    injection_file=injection_file)

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pycbc/inference/sampler/base.py", line 234, in create_new_output_file

    with sampler.io(filename, "w") as fp:

  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pycbc/inference/io/base_hdf.py", line 77, in __init__

    if filetype.decode() != self.name:

AttributeError: 'str' object has no attribute 'decode'

    FAILED!
```

Will need to take a closer look at that.

Also doing my own further testing, I found that the current "solution" actually only fixed some semi-manual file reading that I did first, but not the use of the wrapper class' "read_samples()" method which next fails when trying to do
```
        # add the static params and attributes
        addatrs = (list(self.static_params.items()) +
                   list(self[self.samples_group].attrs.items()))
        for (p, val) in addatrs:
            setattr(samples, p, val)
```
in base_hdf.py


I might take another look at this tomorrow, or go back to using python2 for the moment, but any input is welcome either way!

dbkeitel:
Closing this as a more global solution is clearly needed. (see #2944).

PR #2948, opened 2019-09-20
Setting initial conditions from checkpoint files

sum33it:
The initial condition for samplers at the start has been moved from pycbc_inference module to set_initial_condition function in BaseMCMC class. This will help in setting up initial condition for different samplers in their own way.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -134,21 +134,6 @@ with ctx:
     # set up output/checkpoint file
     sampler.setup_output(opts.output_file, force=opts.force)
 
-    # Figure out where to get the initial conditions from: a samples file,
-    # the checkpoint file, the prior, or an initial prior.
-    samples_file = opts.samples_file
-    # use the checkpoint file instead if resume from checkpoint
-    if not sampler.new_checkpoint:
-        samples_file = sampler.checkpoint_file
-    if samples_file is not None:
-        logging.info("Initial positions taken from last iteration in %s",
-                     samples_file)
-        init_prior = None
-    else:
-        # try to load an initial distribution from the config file
-        init_prior = inference.sampler.initial_dist_from_config(
-            cp, sampler.variable_params)
-
     sampler.set_initial_conditions(initial_distribution=init_prior,
                                    samples_file=samples_file)
 
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -134,8 +134,9 @@ with ctx:
     # Note: the pool is created at this point. This means that,
     # unless you enjoy angering your cluster admins,
     # NO SAMPLES FILE IO SHOULD BE DONE PRIOR TO THIS POINT!!!
-    sampler = inference.sampler.load_from_config(
-        cp, model, output_file=opts.output_file, nprocesses=opts.nprocesses, use_mpi=opts.use_mpi)
+    sampler = inference.sampler.load_from_config(cp, model,
+        output_file=opts.output_file, nprocesses=opts.nprocesses,
+        use_mpi=opts.use_mpi)
 
     # Run the sampler
     sampler.run()
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -91,6 +91,15 @@ fft.verify_fft_options(opts, parser)
 opt.verify_optimization_options(opts, parser)
 scheme.verify_processing_options(opts, parser)
 
+# check that the output file doesn't already exist
+if os.path.exists(opts.output_file):
+    if force:
+        pass
+        #os.remove(opts.output_file)
+    else:
+        raise OSError("output-file already exists; use force if you "
+                      "wish to overwrite it.")
+
 # set seed
 numpy.random.seed(opts.seed)
 logging.info("Using seed %i", opts.seed)
@@ -129,12 +138,7 @@ with ctx:
     # unless you enjoy angering your cluster admins,
     # NO SAMPLES FILE IO SHOULD BE DONE PRIOR TO THIS POINT!!!
     sampler = inference.sampler.load_from_config(
-        cp, model, nprocesses=opts.nprocesses, use_mpi=opts.use_mpi)
-
-    # set up output/checkpoint file
-    sampler.setup_output(opts.output_file, force=opts.force)
-
-    sampler.set_initial_conditions(cp)
+        cp, model, output_file=opts.output_file, nprocesses=opts.nprocesses, use_mpi=opts.use_mpi)
 
     # Run the sampler
     sampler.run()
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -92,13 +92,10 @@ opt.verify_optimization_options(opts, parser)
 scheme.verify_processing_options(opts, parser)
 
 # check that the output file doesn't already exist
-if os.path.exists(opts.output_file):
-    if force:
-        pass
-        #os.remove(opts.output_file)
-    else:
-        raise OSError("output-file already exists; use force if you "
-                      "wish to overwrite it.")
+if os.path.exists(opts.output_file) and not opts.force:
+    raise OSError("output-file already exists; use force if you "
+                  "wish to overwrite it.")
+
 
 # set seed
 numpy.random.seed(opts.seed)
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -134,8 +134,8 @@ with ctx:
     # Note: the pool is created at this point. This means that,
     # unless you enjoy angering your cluster admins,
     # NO SAMPLES FILE IO SHOULD BE DONE PRIOR TO THIS POINT!!!
-    sampler = inference.sampler.load_from_config(cp, model,
-        output_file=opts.output_file, nprocesses=opts.nprocesses,
+    sampler = inference.sampler.load_from_config(
+        cp, model, output_file=opts.output_file, nprocesses=opts.nprocesses,
         use_mpi=opts.use_mpi)
 
     # Run the sampler
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -134,8 +134,7 @@ with ctx:
     # set up output/checkpoint file
     sampler.setup_output(opts.output_file, force=opts.force)
 
-    sampler.set_initial_conditions(initial_distribution=init_prior,
-                                   samples_file=samples_file)
+    sampler.set_initial_conditions(cp)
 
     # Run the sampler
     sampler.run()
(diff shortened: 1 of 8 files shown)


sum33it:
@dfinstad Could you go through the changes I made in multinest sampler and try to run with/without a checkpoint file so that we can make sure its not broken in this PR? 
@cdcapano I have tested this PR with emcee_pt, Dynesty and CPnest samplers.

cdcapano:
@sum33it Can you rebase to master and fix the merge conflicts? My patch to fix the effective number of samples (#3122) is what caused the conflict. The problematic new lines are:
https://github.com/gwastro/pycbc/blob/master/pycbc/inference/sampler/base_mcmc.py#L469-L474
Those need to stay there now, so when you resolve the merge conflict, accept the changes on master. This means you'll also need to turn back on setting the `new_checkpoint` attribute in `setup_output` for MCMC samplers. Sorry, I forgot that that that was needed in #3122 .

cdcapano:
@sum33it should I review this now?

sum33it:
@cdcapano Can you start reviewing now? I have tested this PR with and without checkpoint files for all samplers modulo multinest.

pycbc/inference/sampler/multinest.py
```diff
@@ -396,4 +397,9 @@ def from_config(cls, cp, model, nprocesses=1, use_mpi=False):
                            v is not None}
         obj = cls(model, nlivepoints, constraints=constraints,
                   **optional_kwargs)
+        obj.setup_output(obj, output_file)
+        if not obj.new_checkpoint:
+            obj.resume_from_checkpoint(cp)
+        else:
+            init_prior = initial_dist_from_config(cp, obj.variable_params)
```
dfinstad: In these cases where the sampler doesn't have a `set_p0` method, is there a reason this is still making a call to `initial_dist_from_config`?

pycbc/inference/sampler/dynesty.py
```diff
@@ -138,11 +138,19 @@ def from_config(cls, cp, model, nprocesses=1, loglikelihood_function=None,
         obj = cls(model, nlive=nlive, dlogz=dlogz, nprocesses=nprocesses,
                   loglikelihood_function=loglikelihood_function,
                   use_mpi=use_mpi, **extra)
+        setup_output(obj, output_file)
+        if not obj.new_checkpoint:
+            obj.resume_from_checkpoint(cp)
+        else:
+            init_prior = initial_dist_from_config(cp, obj.variable_params)
```
dfinstad: Same comment as the multinest case, this call seems unnecessary.

pycbc/inference/sampler/cpnest.py
```diff
@@ -117,6 +118,12 @@ def from_config(cls, cp, model, nprocesses=1, use_mpi=False):
         obj = cls(model, nlive=nlive, maxmcmc=maxmcmc, nthreads=nthreads,
                   verbose=verbose,
                   loglikelihood_function=loglikelihood_function)
+
+        setup_output(obj, output_file)
+        if not obj.new_checkpoint:
+            obj.resume_from_checkpoint(cp)
+        else:
+            init_prior = initial_dist_from_config(cp, obj.variable_params)
```
dfinstad: Same as others, seems unnecessary.

pycbc/inference/sampler/base_mcmc.py
```diff
@@ -436,13 +436,27 @@ def set_p0(self, samples_file=None, prior=None):
         self._p0 = p0
         return self.p0
 
-    def set_initial_conditions(self, initial_distribution=None,
-                               samples_file=None):
+    def set_initial_conditions(self, cp):
```
cdcapano: Delete this function; you're now doing this in either `resume_from_checkpoint` or `from_config`.

pycbc/inference/sampler/multinest.py
```diff
@@ -396,4 +397,9 @@ def from_config(cls, cp, model, nprocesses=1, use_mpi=False):
                            v is not None}
         obj = cls(model, nlivepoints, constraints=constraints,
                   **optional_kwargs)
+        obj.setup_output(obj, output_file)
+        if not obj.new_checkpoint:
+            obj.resume_from_checkpoint(cp)
```
dfinstad: There doesn't seem to be a `resume_from_checkpoint` implementation in the multinest sampler like you've put in all the others. I think this will fail because the abstract method hasn't been overridden.

pycbc/inference/sampler/base_mcmc.py
```diff
@@ -464,14 +486,6 @@ def run(self):
...
         # target_eff_nsamples is not None, this is the effective number of
         # samples; otherwise, this is the total number of samples).
-        # _lastclear is the number of iterations that the file already
-        # contains (either due to sampler burn-in, or a previous checkpoint)
-        if self.new_checkpoint:
-            self._lastclear = 0
-        else:
-            with self.io(self.checkpoint_file, "r") as fp:
-                self._lastclear = fp.niterations
-                self.thin_interval = fp.thinned_by
```
cdcapano: Put this block of code back, as it's needed to make sure the thinning is done correctly.

pycbc/inference/sampler/emcee_pt.py
```diff
@@ -171,6 +172,14 @@ def from_config(cls, cp, model, nprocesses=1, use_mpi=False):
...
         # set prethin options
         obj.set_thin_interval_from_config(cp, section)
+        # Set up the output file
+        setup_output(obj, output_file)
+        if not obj.new_checkpoint:
+            obj.resume_from_checkpoint()
+        else:
+            init_prior = initial_dist_from_config(cp, obj.variable_params)
+            obj.set_p0(prior=init_prior)
+            obj._lastclear = 0
```
cdcapano: To please code climate, and to reduce copying code, create a function in `BaseMCMC` called something like `set_start_from_config` that does these three lines of code:
```
def set_start_from_config(self, cp):
    init_prior = initial_dist_from_config(cp, self.variable_params)
    self.set_p0(prior=init_prior)
```
(Don't do the `_lastclear` here, since you'll be putting that back into the beginning of the `run` function.)

Then call `self.set_start_from_config(cp)` here.

pycbc/inference/sampler/multinest.py
cdcapano: I don't think you need this, but not sure... @dfinstad might have a coment.

bin/inference/pycbc_inference
```diff
@@ -129,28 +135,7 @@ with ctx:
...
-                     samples_file)
-        init_prior = None
-    else:
-        # try to load an initial distribution from the config file
-        init_prior = inference.sampler.initial_dist_from_config(
-            cp, sampler.variable_params)
-
-    sampler.set_initial_conditions(initial_distribution=init_prior,
-                                   samples_file=samples_file)
+        cp, model, output_file=opts.output_file, nprocesses=opts.nprocesses, use_mpi=opts.use_mpi)
```
cdcapano: Is this PEP-8 compliant? The line looks longer than 79 characters.

bin/inference/pycbc_inference
```diff
@@ -128,29 +134,9 @@ with ctx:
...
-    else:
-        # try to load an initial distribution from the config file
-        init_prior = inference.sampler.initial_dist_from_config(
-            cp, sampler.variable_params)
-
-    sampler.set_initial_conditions(initial_distribution=init_prior,
-                                   samples_file=samples_file)
+    sampler = inference.sampler.load_from_config(cp, model,
+        output_file=opts.output_file, nprocesses=opts.nprocesses,
+        use_mpi=opts.use_mpi)
```
cdcapano: Still PEP8 issue... put `cp, model` on new line... it's fine to just indent on a line continuation, but if you do, you're not supposed to have anything after the `(` on the first line. Otherwise, if you put things on the same line as the `(` you have to align all subsequent things with that.

cdcapano:
@dfinstad Do you approve now?

pycbc/inference/sampler/multinest.py
```diff
@@ -396,4 +402,7 @@ def from_config(cls, cp, model, nprocesses=1, use_mpi=False):
                            v is not None}
         obj = cls(model, nlivepoints, constraints=constraints,
                   **optional_kwargs)
+        obj.setup_output(obj, output_file)
+        if not obj.new_checkpoint:
+            obj.resume_from_checkpoint()
```
dfinstad: Sorry to change my mind a bit, but after a little more thought it seems a little misleading to call this method here when the method just passes. I'd prefer these lines be removed if they're not doing anything. Same comment goes for the corresponding lines in the dynesty and cpnest samplers.

cdcapano: What if he sticks a logging warning message in their `resume_from_checkpoint` functions that say "checkpointing not supported yet"?

dfinstad: I'm not sure about cpnest and dynesty, but for multinest it actually does support checkpointing, it's just done through its own auxiliary files.

cdcapano: Oh, I didn't realize that about multinest. In that case, yeah, the `resume_from_checkpoint` call should be removed. Where are the files stored? It's a little unnerving that the run will recover from a checkpoint using files  that the executable is unaware of. For instance, is it possible to accidentally resume from the wrong file, or not resume because the files are getting written in a different place than multinest expects?

For dynesty, Sumit is going to be adding checkpointing to that next. Getting ready for that is what prompted this PR.

CPNest does something similar with saving its own files, but it only kinda sorta works, and definitely doesn't work in condor. Fixing that properly may require a fix to CPNest itself; since we haven't been using it much, that's low on the priority list. So for all intents and purposes, our implementation of CPNest doesn't support checkpointing.

dfinstad: Okay, then removing the lines here but leaving them in place for dynesty and cpnest sounds good to me.

I agree that the auxiliary file situation with multinest is a little cumbersome, but I did try to mitigate some risk by forcing the sampler to use the pycbc output file as a base for all the auxiliary files so they'll always be stored in the same location. It is still possible though to run into unexpected resume behavior in a couple (hopefully rare) edge cases. For instance, if the pycbc output files exist but the auxiliary files do not, the run will start from scratch. Conversely if the pycbc output files don't exist but the auxiliary files do, the run will resume.

In the future it would be good to integrate the files somehow, either by making `pycbc_inference` aware of the auxiliary files, or by writing all the necessary resume info to the pycbc files and having the auxiliary files generated on resume.

sum33it:
Hi @cdcapano, shall we merge it now?

pycbc/inference/sampler/multinest.py
```diff
@@ -201,6 +201,11 @@ def set_initial_conditions(self, initial_distribution=None,
         if samples_file is not None:
             self.set_state_from_file(samples_file)
 
+    def resume_from_ceckpoint(self):
```
cdcapano: `resume_from_ceckpoint` -> `resume_from_checkpoint`

sum33it:
@cdcapano Thanks for pointing this out. Just made this change in new commit.

PR #2950, opened 2019-09-20
Change CUDA chisq to use more accurate phase calculation

josh-willis:
This MR attempts to deal with one issue I have encountered when testing the CUDA implementation of the standard power-chi-squared test. Not uncommonly, there were larger than expected differences in chi-squared (a few to tens of percents) when comparing a CPU and GPU trigger, even though SNR and similar quantities were much closer.  There are several reasons for this, but one subset were traceable to differences in the exponential factor `exp(2*pi*freq_index*trig_index/N)`, especially when the index of the trigger is large.  

The problem is that the complex exponential is implemented via calls to the CUDA intrinsic `__sincos` function.  But that function becomes increasingly less accurate, particularly for argument reduction (the first stage in calculating sine and cosine: reduce the argument modulo `2 pi`). For a little more detail see [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#intrinsic-functions).

To compensate for this loss of accuracy, we in this case have a strategy that is not always available: we know how the argument to the trig functions is calculated, before we multiply it by `2 pi`. Conceptually, we would like to replace `__sincos(2*pi*j*k/N)` with `__sincos(2*pi* (j*k % N)/N)`; that is, calculate the remainder first using the two integer indices, then multiply and divide.  But that is also complicated, because `%` is a slow operator in CUDA, mapping to several machine instructions.

Therefore, following the suggestion [here](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#arithmetic-instructions) (scroll down to the section sub-titled "Integer Arithmetic"), I utilize the fact that when `N` is a power of two, `k % N = (k&(N-1))`. So this MR now has two paths for the chi-squared calculation, depending on whether or not the size of our time series is a power of two or not. If it is, the tricks above are used to calculate the complex exponential using `__sincos` but without sacrificing accuracy.  If not, then it falls back to using `sincos`, which should always be (reasonably) accurate.  I've tested this for the power of two case.

### DIFF
--- a/pycbc/vetoes/chisq_cuda.py
+++ b/pycbc/vetoes/chisq_cuda.py
@@ -232,7 +232,8 @@ _pchisq_cache_pow2 = {}
 def get_pchisq_fn_pow2(np, fuse_correlate=False):
     if np not in _pchisq_cache_pow2:
         nt = 256
-        mod = SourceModule(chisqkernel_pow2.render(NT=nt, NP=np, fuse=fuse_correlate))
+        mod = SourceModule(chisqkernel_pow2.render(NT=nt, NP=np,
+                                                   fuse=fuse_correlate))
         fn = mod.get_function("power_chisq_at_points_%s_pow2" % (np))
         if fuse_correlate:
             fn.prepare("PPPI" + "I" * np + "PPPI")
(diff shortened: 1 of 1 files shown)


josh-willis:
@ahnitz Poke on this.  I've fixed those of the original codeclimate complaints that I was willing to, but silencing a bogus warning raised two others, so I can revert that if you want.  But mostly I'm interested in the review of the actual change.

PR #2951, opened 2019-09-23
Add mchirp_area codes

veronica-villa:


tdent:
Can you try and fix the Travis issue by editing the docstring at the beginning of the file?

veronica-villa:
@tdent Sorry I thought I had already done it. I think it is fixed now.

tdent:
Actually I think the codeclimate complaints about 'constant name not in upper case' are due to it thinking that the plotting codes which have .py file names being modules rather than scripts. 

If we want these to go away we would have to change the file names to not have .py at the end. However this is tricky or impossible to do that in the middle of a pull request via the web browser as far as I can see so we should possibly postpone that.

pycbc/mchirp_area.py
```diff
@@ -0,0 +1,242 @@
...
+from scipy.integrate import quad
+
+
+def src_mass_from_z_det_mass(z, del_z, mdet, del_mdet):
+    """This function takes values of redshift, redshift uncertainty,
+    detector mass and its uncertainty and computes the source mass
+    and its uncertainty.
+    """
+    msrc = (mdet) / (1 + z)
+    del_msrc = msrc * ((del_mdet/mdet)**2 + (del_z/(1 + z))**2)**(0.5)
```
tdent: PEP8 issues not caught by codeclimate .. This should have more spacing - ideally spaces around each operator.  Also '1' and '2' should be '1.' and '2.' to force them to be floats. 
Also 'mdet' and '0.5' shouldn't be in brackets when they are on their own.

veronica-villa:
@tdent Equation spacing issue is  now fixed, as well as the renaming of the scripts.

pycbc/mchirp_area.py
```diff
@@ -0,0 +1,241 @@
+#!/usr/bin/env python
+#
+# Integration of the area laying in the different cbc regions
+# By A. Curiel Barroso
+# August 2019
+
+"""This script computes the area corresponding to different CBC on the m1 & m2
```
tdent: Since this isn't a script any more, better to say something like 'Functions to compute the area ..' 
Also, we don't need the `#!/usr/bin/env python` line at the beginning of the file as it's not an executable

pycbc/mchirp_area.py
```diff
@@ -0,0 +1,241 @@
...
+plane when given a central mchirp value and uncertainty.
+It also includes a function that calculates the source frame when given the
+detector frame mass and redshift.
+"""
+
+from pycbc.conversions import mass2_from_mchirp_mass1 as m2mcm1
+from scipy.integrate import quad
+
+
+def src_mass_from_z_det_mass(z, del_z, mdet, del_mdet):
```
tdent: Docstrings look fine, but you don't need to write 'this function' at the beginning of each one, it's clear that they are functions ;)

veronica-villa:
@tdent Travis is failing due to the rename of the scripts.. I don't know if it has a cache I can remove or something similar...
```
[Fri Oct  4 11:13:14 UTC 2019] running /home/travis/virtualenv/python2.7.15/bin/pycbc_mchirp_plots --help
    FAILED!
```

tdent:
What happens when you run the help command, i.e. pycbc_mchirp_plots --help ?

tdent:
See the second answer here : https://stackoverflow.com/questions/19425857/env-python-r-no-such-file-or-directory 
Looks like your text editor has not been doing the desired thing with linebreaks.

veronica-villa:
@tdent ok thanks, I will try to fix it!

tdent:
Ah, mass_area_plot now has a different Travis error : 
```---------------------------------------------------------
  File "/home/travis/virtualenv/python2.7.15/bin/pycbc_mass_area_plot", line 12, in <module>
    from mchirp_area import calc_areas
ImportError: No module named mchirp_area

File: bin/plotting/pycbc_mass_area_plot [hunk 4]
```diff
@@ -0,0 +1,109 @@
...
+# Integration of the area laying in the different cbc regions
+# By A. Curiel Barroso
+# August 2019
+
+"""This script computes the area corresponding to different CBC on
+the m1 & m2 plane when given a central mchirp value and uncertainty.
+"""
+
+import argparse
+from mchirp_area import calc_areas
```
tdent: These will then need to be changed since mchirp_area is a pycbc module - i.e. `from pycbc.mchirp_area import` etc

veronica-villa:
@tdent I think all Travis issues are fixed now!

tdent:
OK to merge, we will look at rebasing, splitting up functions etc. later.

PR #2952, opened 2019-09-24
few tweaks to multiifo workflow

tdent:


tdent:
Ah, hmm, we have an error :

> Traceback (most recent call last):
>   File "/home/tdent/pycbc-aug18/bin/pycbc_create_offline_search_workflow", line 674, in <module>
>     symlink_path(f, rdir.base)
>   File "/home/tdent/pycbc-aug18/bin/pycbc_create_offline_search_workflow", line 46, in symlink_path
>     os.symlink(f.storage_path, os.path.join(path, f.name))
> AttributeError: 'tuple' object has no attribute 'storage_path'

tdent:
I suspect this may be due to different treatment of the summary SNR-IFAR plots - now there is one for each of the coinc types, rather than just one plot for H1L1.  I am not sure how I would make it so that these 4 plots are grouped as 2*2 on the page, I naively used `list(layout.grouper(snrifar_summ, 2))` however that seems to produce too many layers of listed tuples for the file storage to digest.

GarethDaviesGW:
I think (not tested if this breaks elsewhere) you should be able to exchange the `snrifar_summ += closed_page` to `snrifar_summ += [snrifar_cb]` and get back what you want. This may become more complicated when `ifar_cb` is being made again though

tdent:
Have removed the attempt at making 2x2 snrifar plot layout, workflow generation now runs without an error so will merge when Travis is done

PR #2956, opened 2019-09-26
Add pruning in to fit_sngls_split_binned

GarethDaviesGW:
Basic pruning added to split binned histograms, at this point, simply takes the loudest trigger in each split, removes anything within specified window, and repeats this procedure as many times as specified

An 'outlier test' might be good to add, but I'm not sure of how best to do this
I have not set a 'maximum number to remove' flag as in the unsplit pruning procedure

### DIFF
--- a/bin/hdfcoinc/pycbc_fit_sngls_split_binned
+++ b/bin/hdfcoinc/pycbc_fit_sngls_split_binned
@@ -91,7 +91,7 @@ parser.add_argument("--stat-fit-threshold", type=float, required=True,
 parser.add_argument("--fit-function",
                     choices=["exponential", "rayleigh", "power"],
                     help="Functional form for the maximum likelihood fit")
-parser.add_argument("--prune-number", type=int, default=-1,
+parser.add_argument("--prune-number", type=int, default=0,
                     help="number of loudest events to remove from each split "
                          "[default do not prune]")
 parser.add_argument("--prune-window", type=float, default=0.1,
@@ -189,13 +189,6 @@ for i, lower_2, upper_2 in zip(range(args.split_two_nbins),
                                sp_two_bounds.lower(), sp_two_bounds.upper()):
     id_in_bin2[i] = np.intersect1d(np.argwhere(params[args.split_param_two] > lower_2),
                                    np.argwhere(params[args.split_param_two] <= upper_2))
-id_in_both = [[[]]*args.split_two_nbins]*args.split_one_nbins
-for x in range(args.split_one_nbins):
-    id_bin1 = id_in_bin1[x]
-    for y in range(args.split_two_nbins):
-        id_bin2 = id_in_bin2[y]
-        id_in_both[x][y] = np.intersect1d(id_bin1, id_bin2)
-id_in_both = np.array(id_in_both)
 
 logging.info('getting template boundaries from trigger file')
 boundaries = trigf[args.ifo + '/template_boundaries'][:]
@@ -226,11 +219,15 @@ logging.info('{} out of {} trigs removed after vetoing with {} from {}'.format(
 logging.info('Applying pruning around loudest triggers in each split up to a maximum of {}'.format(
                   args.prune_number))
 for x in range(args.split_one_nbins):
+    if not args.prune_number: break
+    id_bin1 = id_in_bin1[x]
     for y in range(args.split_two_nbins):
-        if len(id_in_both[x,y]) == 0: continue
+        id_bin2 = id_in_bin2[y]
+        id_in_both = np.intersect1d(id_bin1, id_bin2)
+        if len(id_in_both) == 0: continue
         vals_inbin = []
         time_inbin = []
-        for idx in id_in_both[x,y]:
+        for idx in id_in_both:
             where_idx_start = boundaries[idx]
             vals_inbin += list(stat[where_idx_start:where_idx_end[idx]])
             time_inbin += list(time[where_idx_start:where_idx_end[idx]])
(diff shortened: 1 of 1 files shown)


GarethDaviesGW:
The effects of the pruning can be seen in the output of this chunk without pruning:

image

and with pruning:
image

Though the bottom left split may show that it is being over-pruned (short duration line has a sharp downturn)

GarethDaviesGW:
Ideas on what an outlier test might entail would be welcome (particularly thinking of @ahnitz, @tdent) as this would hopefully protect from overpruning

GarethDaviesGW:
single pruning version of the above plot: 
image

Still has some overpruning, which should be sorted by an outlier test

tdent:
1 loudest trigger seems to be enough (which ifo in which data chunk?)
If the fits are not affected much I would not worry about 'over' pruning just now

GarethDaviesGW:
> 1 loudest trigger seems to be enough (which ifo in which data chunk?)
> If the fits are not affected much I would not worry about 'over' pruning just now

This is O2 A21 H1

GarethDaviesGW:
As a note - something has gone wrong somewhere and the eta split number is not being displayed properly, and all the sections are the same

GarethDaviesGW:
Updated plots as above but without repeated split errors
image

image

tdent:
Well, I'm sure you can think of something to quantify outlierness. Not highest priority to deal with though.

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -183,16 +195,68 @@ boundaries = trigf[args.ifo + '/template_boundaries'][:]
 max_boundary_id = np.argmax(boundaries)
 sorted_boundary_list = np.sort(boundaries)
 
-logging.info('getting stat values from trigger file')
+logging.info('Processing template boundaries')
+where_idx_end = np.zeros_like(boundaries)
+for idx, idx_start in enumerate(boundaries):
+    if idx == max_boundary_id:
+        where_idx_end[idx] = trigf[args.ifo+'/end_time'].size
```
tdent: spaces round +

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -183,16 +195,68 @@ boundaries = trigf[args.ifo + '/template_boundaries'][:]
...
+    else:
+        where_idx_end[idx] = sorted_boundary_list[
+            np.argmax(sorted_boundary_list == idx_start) + 1]
+
+logging.info('calculating single stat values from trigger file')
 stat = get_stat(args.sngl_stat, trigf[args.ifo])
 
 logging.info('applying DQ vetoes')
 time = trigf[args.ifo+'/end_time'][:]
```
tdent: spaces around + here too

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -183,16 +195,68 @@ boundaries = trigf[args.ifo + '/template_boundaries'][:]
...
+            max_statval = vals_inbin[max_val_arg]
+
+            logging.info('Prune {}: pruning triggers around time {}s'.format(
+                                     count_pruned, args.prune_window,
+                                     time[max_val_arg]))
+            remove = np.nonzero(abs(time_inbin[max_val_arg] - time)
+                                    < args.prune_window)[0]
+            remove_inbin = np.nonzero(abs(time_inbin[max_val_arg] - time_inbin)
+                                      < args.prune_window)[0]
+            logging.info('Removing {} triggers, {} in this split'
```
tdent: seems a bit over-verbose overall. can you combine this log message with the previous one? & for instance you don't have to tell the user the size of the prune window since it's already in the command line

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -183,16 +195,68 @@ boundaries = trigf[args.ifo + '/template_boundaries'][:]
...
+                         ''.format(remove.size,remove_inbin.size))
+            vals_inbin[remove_inbin] = np.zeros_like(remove_inbin)
```
tdent: comment as to why setting these to zero is the right thing to do or is OK?

GarethDaviesGW: Added, but earlier in the code where it is done for the vetoing (as well as a pointer to it here)

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -46,6 +46,8 @@ parser.add_argument('--output-file', type=str, required=True,
 parser.add_argument('--verbose', action='store_true')
 parser.add_argument('--bin-param', default='template_duration', type=str,
                     help="parameter for on-plot splitting. Default: template_duration")
+parser.add_argument('--bin-param-log', default=False, action='store_true',
```
tdent: btw, this was the previous behaviour : https://github.com/gwastro/pycbc/blob/v1.14.1/bin/hdfcoinc/pycbc_fit_sngls_binned#L99 
probably less confusing for users to have the same behaviour now.

tdent:
BTW, can you remind me why there is a max bin param option but not min bin param? (Ie why was the option put there?)
Not asking for a change particularly but wondering what the use case was.

GarethDaviesGW:
> BTW, can you remind me why there is a max bin param option but not min bin param? (Ie why was the option put there?)
> Not asking for a change particularly but wondering what the use case was.
This was just from wanting to see the variation in the shorter-duration times, e.g up to 4s or similar. I was using it for doing diagnostics at one point, and so it might make sense to keep it, and put in a minimum version as well, and then maybe add manual min-max options for the other parameters as well (I senr this as reply to email before merging but just had notification that it didn't get through)

PR #2960, opened 2019-10-03
Amaldi stat - geometric ave of 3 2-ifo foreground densities

tdent:


tdent:
This is provided mainly as a baseline for testing and comparison against Alex's updated multi-ifo histogram #2943 .  If that works as desired then this PR will not be merged to master.

PR #2961, opened 2019-10-04
add foreground ifar plot for multi-ifo pipeline

tdent:


ahnitz:
@tdent I don't have any major questions about this. Only to note that it might be useful for this to be able to support taking in multiple trigger files.

tdent:
multiple trigger files meaning for analyses of different disjoint periods of time which should be combined in the plot?  (overlapping times would be harder ..)

ahnitz:
@tdent yes, for disjoint times

tdent:
if you feel like testing this out on >1 statmap fill go ahead .. so far it's just my guesses at what ought to work

PR #2968, opened 2019-10-07
add injection_style checks to add statmap

GarethDaviesGW:
Hierarchical removal changes were not tested against injection style add_statmap - some keyes were not present in injection statmaps, so this fixes these

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_add_statmap
+++ b/bin/hdfcoinc/pycbc_multiifo_add_statmap
@@ -60,7 +60,9 @@ parser.add_argument('--output-file', help="name of output file")
 
 args = parser.parse_args()
 
-if args.max_hierarchical_removal and args.background_files:
+injection_style = args.background_files != None
+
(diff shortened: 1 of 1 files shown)


bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -113,11 +113,12 @@ logging.info('Copying foreground datasets')
 for k in files[0]['foreground']:
     if not k.startswith('fap') and k not in all_ifos:
         pycbc.io.combine_and_copy(f, files, 'foreground/' + k)
-
-logging.info('Copying background datasets')
-for k in files[0]['background']:
-    if k not in all_ifos:
-        pycbc.io.combine_and_copy(f, files, 'background/' + k)
+injection_style = args.background_files != None
```
tdent: it might be preferable to put this right after argument parsing (rather than in the middle of two parallel steps) since it's a flag with global effect. otherwise looks fine

PR #2975, opened 2019-10-21
More detailed description for GRB sky error func

a-r-williamson:
Also included a new argument to allow for different probability
coverage. Default is the same as old hardcoded value of 90%.

### DIFF
--- a/pycbc/workflow/grb_utils.py
+++ b/pycbc/workflow/grb_utils.py
@@ -282,36 +282,66 @@ def make_gating_node(workflow, datafind_files, outdir=None, tags=None):
     return condition_strain_nodes, condition_strain_outs
 
 
-def get_sky_grid_scale(
-        sky_error=0.0, Fermi=False, upscale=False, core_sigma=3.6,
-        core_frac=0.98, tail_sigma=29.6, containment=0.9, precision=1000):
+def fermi_core_tail_model(
+        sky_err, rad, core_frac=0.98, core_sigma=3.6, tail_sigma=29.6):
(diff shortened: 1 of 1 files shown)


a-r-williamson:
Tessa Carver is checking whether the Fermi-GBM systematic value is relevant for O3A GRBs. Will wait for confirmation or correction before merging.

pannarale:
@a-r-williamson, could you fix the last 2 codclimate issues, please?  That's: line too long, and blank line at EOF.

pannarale:
Hi @tcarver1234, your latest commits introduced some codeclimate issues (whitespaces and indentations mainly).  Please fix as many as you can.

Then there is a conflict to be resolved (tagging @a-r-williamson too).

a-r-williamson:
@tcarver1234 You removed the precision argument but the variable is still used in the function. Did you mean to change the function as well in some way?

a-r-williamson:
This function is not being called in master.
https://github.com/gwastro/pycbc/blob/3f6c49a2802d9f13cdac9f18582d2dd7537b0689/pycbc/workflow/matched_filter.py#L303-L319
It looks like `master` has thrown the `coh_PTF` interface away. Is this function still needed, and if so is this the correct place for it?

ahnitz:
@a-r-williamson What is the status of this? Is this PR still active?

pannarale (approved):
We should push this small change to master before it lags behind other development or is coded up again.

[We will be forced to revisit this whole file further down the line in any case, once when we start using it again, e.g., some parts will not work with Pegasus5.]

PR #2976, opened 2019-10-21
Vt ratio logx

GarethDaviesGW:
the plot_vt_ratio plot gets cramped at small mchirp when used with the standard values in the pycbc-config settings

This PR adds in an option to have log x axis so that these are more evenly spread across the plot


GarethDaviesGW:
Somethings funky with the way I've commited this. Will delete and retry

PR #2977, opened 2019-10-21
add log-x option to plot_vt_ratio

GarethDaviesGW:
standard mchirp options were making the vt ratio comparison plots bunched up on one side - add in ability to use log x axis

### DIFF
--- a/bin/plotting/pycbc_plot_vt_ratio
+++ b/bin/plotting/pycbc_plot_vt_ratio
@@ -106,7 +106,7 @@ for idv in idxs:
 
 if args.log_x and args.log_y:
     plt.loglog()
-elif args.log_x and args.log_y:
+elif args.log_x and not args.log_y:
     plt.semilogx()
 elif args.log_y and not args.log_x:
     plt.semilogy()
--- a/bin/plotting/pycbc_plot_vt_ratio
+++ b/bin/plotting/pycbc_plot_vt_ratio
@@ -23,7 +23,8 @@ parser.add_argument('--outfile', type=str, required=True,
 parser.add_argument('--ifars', type=float, required=True, nargs='+',
                     help='IFAR values to plot VT ratio for. Note that the plotted values '
                          'will be the closest values available from the VT files')
-
+parser.add_argument('--log-x', action='store_true', default=False,
+                    help='Use logarithmic x-axis')
 args = parser.parse_args()
 
 # Load in the two datasets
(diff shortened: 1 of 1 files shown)


bin/plotting/pycbc_plot_vt_ratio
```diff
@@ -54,9 +55,13 @@ x_param = r'$' + tuple(ffirst_set['data'].keys())[0].split('\\in')[0].strip('$')
...
-# offset different ifars by 1/20th of the distance between one parameter value and the next
-# for the last value use the previous difference
-xpos_add_dx = 0.05 * np.append(np.diff(xpos), np.diff(xpos)[-1])
+# offset different ifars by 1/20th of the mean distance parameter value diffs
+if args.log_x:
+    xpos_logdiffmean = np.diff(np.log(xpos)).mean()
+    xpos_add_dx = 0.05 * np.ones_like(xpos) * xpos_logdiffmean
+else:
+    xpos_diffmean = np.diff(xpos).mean()
+    xpos_add_dx = 0.05 * np.ones_like(xpos) * xpos_diffmean
```
tdent: I found another case which needs special treatment, namely if there's only one parameter bin so np.diff fails. 
I dealt with this via try / except IndexError and then arbitrarily setting `xpos_add_dx = 0.05`

bin/plotting/pycbc_plot_vt_ratio
```diff
@@ -80,14 +85,27 @@ for idv in idxs:
...
+    else:
+        xvals = xpos + xpos_add_dx * (colorcount -
+                                      float(len(args.ifars) - 1) / 2.)
+    ax_mi.errorbar(xvals, ys,
         yerr=[yerr_errlow, yerr_errhigh], fmt='o', markersize=7, linewidth=5,
-        label='IFAR = %d yr' % ffirst_set['xvals'][idv], capsize=5, capthick=2, mec='k',
-        color=colors[colorcount % 10])
-    plt.xticks(xpos, labels, rotation='horizontal')
+        label='IFAR = %d yr' % ffirst_set['xvals'][idv], capsize=5,
+        capthick=2, mec='k', color=colors[colorcount % 10])
```
tdent: I found this was a bug when trying to plot more than 6 IFAR values - since there are 6 colo(u)rs in the list, should be `colorcount % 6`

bin/plotting/pycbc_plot_vt_ratio
```diff
@@ -80,14 +85,27 @@ for idv in idxs:
...
     colorcount += 1
 
-# get the limit of the x axes, and draw a black line in order to highlight equal comparison
+if args.log_x:
+    plt.semilogx()
+    plt.xticks(xpos, labels, rotation='horizontal')
```
tdent: not sure why we need and else and why this xticks line turns up twice - just do it outside the 'if'.

PR #2980, opened 2019-10-23
Live trigger ifar calculation

GarethDaviesGW:
Current outline of idea:
-  Fit triggers split into bins by template duration
-  Use fits from previous N days (N currently unknown) to extrapolate IFAR

This PR contains:
Setting up a code which does trigger fitting
-  Splits the bank by template duration into (default 6, split in log) bins, these are each fit with an exponential model
-  The `live_time` is approximated by adding 8s for each file found with triggers generated. This means that this will always be an underestimate, and so IFAR is always overestimated. Therefore this may need investigation for some kind of correction
-  My thoughts for this were to run one per day, and then it can easily be averaged over multiple days/files in the way that fit_over_param does in the offline search. We can then set an amount of time to get the distribution coefficients from, either by specifying that the live_time attributes must add up to a certain amount, or by a set number of days. This may be done by another 
 
`pycbc/events/singles.py` is also being modified to deal with LiveSingles with a set IFAR or calculated IFAR. This is ongoing at point of first setting up the PR

This will then be used for calculating the IFAR for triggers above the loudest trigger in the previous days (note that it may be worth saving the trigger stat values and doing a `calculate_n_louder` and then the extrapolation is only done above the loudest trigger in full set)


titodalcanton:
@GarethDaviesGW since PyCBC Live is moving to Python 3, for this request please make sure to do all your development and testing in Python 3.

titodalcanton:
I also suggest rebasing this on master as we did recently change the code a bit.

GarethDaviesGW:
I thought I'd add an explanation of what the IFAR calculation is doing and what choices we have to make:

If there are any triggers which meet the newsnr threshold, duration threshold and chi-squared threshold, then these will have a false alarm rate calculated

This false alarm rate is based on a fit to some set of triggers, we will probably use the set of all O3 triggers so far. The fit is to these triggers is an exponential of the form:

rate(newsnr) = total count above fit threshold * alpha * exp(-alpha(newsnr-fit_threshold)). 

We calculate alpha with a max likelihood fit. The fit threshold shouldn't really matter, but I chose it to be 6 in the fits that I've done so far (this can be revisited if requested), the count is just a count.

The false alarm rate is then the amount of live time from which the triggers have been collected, divided by the count of triggers above this newsnr (i.e. integrate the above expression from newsnr to infinity)

So given alpha, we can choose which distribution to use in the calculation
One option is to simply use the fit values for all triggers in each detector. Which should be the most representative FAR.
Another option, which is called 'conservative' in the is to use the smallest alpha from any template duration bin and any detector, the largest count and the smallest live time. This will give the largest false alarm rate of the triggers and will mean that we are more confident with any alerts

So the things we need to decide are:
- Which distribution to use. I would probably lean towards the 'conservative' one
- newsnr threshold for FAR calculation
   - As the distribution has a 1:1 correspondence with newsnr, we can choose to set the newsnr threshold close to where the alert FAR threshold would be, and this will be more efficient. For the conservative fit this would be around 9.35, for the least conservative fits this would be around 8.9
- duration threshold
   - Looking at the live bank, and calculating remnant mass for each template, we get [this plot](https://ldas-jobs.ligo.caltech.edu/~gdavies/testoutput/live_bank_plots/duration_remnant_mass-LL_log.png), (note that if remnant mass=0 it is not plotted) and we see that if the template duration is less than around 7s, there is no remnant mass. So this duration cutoff should probably be around 7s
- chi-squared threshold
   - I don't know the best value for this, it is currently set as 4, but may be best reduced?
- the gracedb upload threshold - the same as that for coincs?

titodalcanton:
@GarethDaviesGW some more thoughts on your last points:
* I would pick a NewSNR threshold at 9 as a round number.
* A duration threshold of 7 seems a no brainer given your study of the remnant mass.
* I would choose an even more aggressive chi^2 threshold of 2 based on the past events, although it risks killing long signals with glitches in the middle of them.
* Using the same upload threshold for coincs seems a reasonable approach. We will in any case check the actual rate vs IFAR as part of the checks.

Given that you seem to have a working code, could you update the request and resolve the conflict with master?

bin/pycbc_live
```diff
@@ -309,27 +309,30 @@ class LiveEventManager(object):
...
-                end_time = int(single['foreground/%s/end_time' % ifo])
-                fname = 'single-%s-%s.xml.gz' % (ifo, end_time)
-                fname = os.path.join(self.path, fname)
-                logging.info('Single-detector candidate! Saving as %s', fname)
-                if args.enable_single_detector_upload:
-                    event.upload(fname, gracedb_server=args.gracedb_server,
-                                 testing=self.gracedb_testing)
-                else:
-                    event.save(fname)
+        if len(active) != 0:# == 1:
```
titodalcanton: I suggest replacing this line with
```python
if active:
```

titodalcanton: Actually, since the `if` statement only contains a for loop, you can just remove the `if` altogether.

bin/pycbc_live
```diff
@@ -309,27 +309,30 @@ class LiveEventManager(object):
...
+            for ifo in active:
+                single = sngl_estimator[ifo].check(results[ifo], data_reader[ifo])
+
+                if single is not None:
```
titodalcanton: Personally in cases like this I prefer to do
```python
if single is None:
    continue
```
and carry on with the rest of the code. Not sure what PEP8 says about that, and this was here before your change, so up to you.

bin/pycbc_live
```diff
@@ -505,6 +508,9 @@ parser.add_argument('--enable-profiling', type=int, metavar='RANK',
                          ' with given rank at the end of program execution')
 
 parser.add_argument('--enable-background-estimation', default=False, action='store_true')
+parser.add_argument('--conservative-sngl-ifar-est', default=True, action='store_true',
+                    help='Use conservative trigger fit values in estimation of '
+                         'single trigger ifars. Default=True')
```
titodalcanton: How does a `store_true` option work if it defaults to `True`? How do you set it to `False`?

pycbc/events/single.py
```diff
@@ -1,28 +1,51 @@
...
 from pycbc.types import MultiDetOptionAction
+from pycbc import conversions as conv
+from pycbc import bin_utils
 
-class LiveSingleFarThreshold(object):
+class LiveSingle(object):
     def __init__(self, ifo,
                  newsnr_threshold=10.0,
-                 reduced_chisq_threshold=5,
+                  reduced_chisq_threshold=5,
```
titodalcanton: Remove extra whitespace.

pycbc/events/single.py
```diff
@@ -1,28 +1,51 @@
...
     @staticmethod
     def insert_args(parser):
         parser.add_argument('--single-newsnr-threshold', nargs='+',
                             type=float, action=MultiDetOptionAction)
         parser.add_argument('--single-reduced-chisq-threshold', nargs='+',
                             type=float, action=MultiDetOptionAction)
-        parser.add_argument('--single-fixed-ifar', nargs='+',
+        parser.add_argument('--single-fixed-ifar', nargs='+', default=None,
                             type=float, action=MultiDetOptionAction)
+        parser.add_argument('--fit-definition-file',
```
titodalcanton: Is `--single-fit-file` any better? I prefer if the option itself contains the word "single".

pycbc/events/single.py
```diff
@@ -1,28 +1,51 @@
...
+                            help="File which contains definitons of fit "
+                                 "coefficients and counts for specific "
+                                 "single trigger IFAR fitting.")
+        if '--single-fixed-ifar' in sys.argv:
```
titodalcanton: I am not sure what the purpose of this code is, but I would not recommend peeking into `sys.argv` from a module. There should be a cleaner way to obtain the desired behavior.

pycbc/events/single.py
```diff
@@ -31,31 +54,65 @@ def from_cli(cls, args, ifo):
         return cls(
            ifo, newsnr_threshold=args.single_newsnr_threshold[ifo],
            reduced_chisq_threshold=args.single_reduced_chisq_threshold[ifo],
-           fixed_ifar=args.single_fixed_ifar[ifo],
            duration_threshold=args.single_duration_threshold[ifo],
+           fit_file=h5py.File(args.fit_definition_file),
```
titodalcanton: Could you instead open the file with a `with` statement, read whatever data is needed, and pass it to the `cls` constructor? I would like to minimize the number of open files during operation.

titodalcanton: Also, you probably want to open the file explicitly in read-only mode.

pycbc/events/single.py
```diff
@@ -31,31 +54,65 @@ def from_cli(cls, args, ifo):
...
         the current data.
         """
-        if len(triggers['snr']) == 0:
+        if len(trigs['snr']) == 0:
+            return None
+
+        # Apply cuts to trigs before clustering
+        valid_idx = (trigs['template_duration'] > self.duration_threshold) & \
+                    (trigs['chisq'] < self.reduced_chisq_threshold)
+        if not np.count_nonzero(valid_idx):
```
titodalcanton:
```python
if not np.any(valid_idx):
```

pycbc/events/single.py
```diff
@@ -31,31 +54,65 @@ def from_cli(cls, args, ifo):
...
             return None
+        cutdurchi_trigs = {k: trigs[k][valid_idx] for k in trigs}
+        nsnr_all = ranking.newsnr(cutdurchi_trigs['snr'],
+                                  cutdurchi_trigs['chisq'])
+        nsnr_idx = nsnr_all > self.newsnr_threshold
+        if not np.count_nonzero(nsnr_idx):
```
titodalcanton:
```python
if not np.any(nsnr_idx):
```

GarethDaviesGW:
Thanks for the comments Tito, I've tried to address them in the latest commits and will add specific comments where necessary above

bin/pycbc_live
GarethDaviesGW: Well that was a thinko and didn't make sense - fortunately I'm not using this option any more (forgot to remove it) so it doesn't matter

pycbc/events/single.py
GarethDaviesGW: Again I've worked around this by simply setting that if single-fixed-ifar is set, then it will override any other options. The sys.argv call was an attempt to have optional defaults depending on whether other options are set

pycbc/events/single.py
GarethDaviesGW: Yeah I had wanted to avoid having it open for too long, but there was an issue with having too many attributes for the class - I got around that by using dictionaries as attributes

bin/pycbc_live
```diff
@@ -309,27 +309,33 @@ class LiveEventManager(object):
...
+
+            end_time = int(single['foreground/%s/end_time' % ifo])
+            fname = 'single-%s-%s.xml.gz' % (ifo, end_time)
+            fname = os.path.join(self.path, fname)
+            logging.info('Single-detector candidate! Saving as %s',
+                         fname)
+            ifar = single['foreground/ifar']
+            # apply a trails factor of the number of active detectors
+            ifar /= len(active)
+            single['foreground/ifar'] = ifar
```
titodalcanton: I think the application of the trials factor (lines 329-332) should go before instantiating the `SingleCoincForGraceDB`. Otherwise GraceDB will show the uncorrected FAR, right?

Also trails → trials.

pycbc/events/single.py
```diff
@@ -1,61 +1,141 @@
...
+            "thresh": fit_thresh}
 
     @staticmethod
     def insert_args(parser):
         parser.add_argument('--single-newsnr-threshold', nargs='+',
                             type=float, action=MultiDetOptionAction)
         parser.add_argument('--single-reduced-chisq-threshold', nargs='+',
                             type=float, action=MultiDetOptionAction)
-        parser.add_argument('--single-fixed-ifar', nargs='+',
+        parser.add_argument('--single-fixed-ifar', nargs='+', default=None,
```
titodalcanton: Args implicitly default to `None`, so you can drop the explicit default setting.

pycbc/events/single.py
```diff
@@ -1,61 +1,141 @@
...
                             type=float, action=MultiDetOptionAction)
+        parser.add_argument('--single-fit-file',
+                            help="File which contains definitons of fit "
+                                 "coefficients and counts for specific "
+                                 "single trigger IFAR fitting.")
+        parser.add_argument('--sngl-ifar-est-dist', nargs='+',
+                            default='conservative',
```
titodalcanton: If this option only takes a limited set of choices, you can use the `choose=<list>` kwarg here.

pycbc/events/single.py
```diff
@@ -1,61 +1,141 @@
...
+            fit_coeffs = None
+        else:
+            with h5py.File(args.single_fit_file, 'r') as fit_file:
+                bin_edges = fit_file['bins_edges'][:]
+                fit_bins = bin_utils.IrregularBins(bin_edges)
+                dist_grp = fit_file[ifo + '/' + args.sngl_ifar_est_dist[ifo]]
+                live_time = fit_file[ifo].attrs['live_time']
+                fit_rates = dist_grp['counts'][:] / live_time
+                fit_coeffs = dist_grp['fit_coeff'][:]
+                fit_thresh = fit_file.attrs['fit_threshold']
```
titodalcanton: You could also avoid passing many arguments to the class constructor by having the constructor itself read the HDF5 file, I guess (i.e. only pass the file name to the constructor). But this looks good enough if codeclimate is happy.

pycbc/events/single.py
```diff
@@ -1,61 +1,141 @@
...
+
+            # 'cluster' by taking the maximal newsnr value over the trigger set
+            i = nsnr_all[nsnr_idx].argmax()
+
+            # This uses the pycbc live convention of chisq always meaning the
+            # reduced chisq.
+            rchisq = cutall_trigs['chisq'][i]
+            nsnr = ranking.newsnr(cutall_trigs['snr'][i], rchisq)
+            dur = cutall_trigs['template_duration'][i]
+            break
```
titodalcanton: Is the `while` loop appropriate here? On a quick glance, it looks like you will actually never loop around due to the final `break`.

bin/pycbc_live
GarethDaviesGW: Yes that's right, don't know why I put it afterwards. Sorted now

pycbc/events/single.py
GarethDaviesGW: Updated to do this so that fewer args are passed through the cli constructor

pycbc/events/single.py
GarethDaviesGW: The final break is on purpose. This was the way I thought of which would allow breaks so that if all triggers are filtered out, it would be able to break the loop without performing the next bits of code

The other option I can think of would be many nested if statements, or a for loop with just a single loop which can be continued past

GarethDaviesGW: Okay I've put this bit into a set of if statements which will prevent it from running if all triggers have been cut. This is better practice than the loop

GarethDaviesGW:
I think this is ready to review and no longer a work in progress now

titodalcanton:
Thanks @GarethDaviesGW. I actually have a more basic question: if we do get a coincident event, will your code *also* generate single-detector triggers? Or is there a mechanism that "vetoes" the single detector triggers in case of a multidetector trigger?

GarethDaviesGW:
> Thanks @GarethDaviesGW. I actually have a more basic question: if we do get a coincident event, will your code _also_ generate single-detector triggers? Or is there a mechanism that "vetoes" the single detector triggers in case of a multidetector trigger?

Good point - I'd just plugged it in where there was an existing singles-with-fixed-ifar search. I'll try and see how that part works and get back to you. When I've been running tests, I've been getting coincs without associated singles but that may be as they are not significant as singles but are as coincs

If there is a signal which is loud enough in single detector and coinc, fits the duration criteria, and so triggered both in coinc and single, would the mutiple events mechanism of gracedb not prevent multiple alerts in this case? I know that we should not rely on that, but I can see how it may actually be preferable to have single and coinc events sent to gracedb in this case

pycbc/events/single.py
```diff
@@ -31,31 +65,69 @@ def from_cli(cls, args, ifo):
...
-        dur = triggers['template_duration'][i]
-
-        if nsnr > self.newsnr_threshold and \
-                rchisq < self.reduced_chisq_threshold and \
-                dur > self.duration_threshold:
-            fake_coinc = {'foreground/%s/%s' % (self.ifo, k): triggers[k][i]
-                          for k in triggers}
+        trigsremain = True
+        if len(trigs['snr']) == 0:
+            trigsremain = False
```
titodalcanton: Hmm, this function still smells. What happens if this first `if` succeeds? The desired behavior is that the function returns `None`, right? With the current code, `trigsremain` will be set to `False`. Then every other `if` in the function will fail, and we will get to the final `return fake_coinc`. But `fake_coinc` will not be defined at that point.

Why not just returning `None` straight away here, instead of having `trigsremain` and the various `if` statements later on? I.e. every time you do a check that fails, just `return None` at that point. Exit the function as early as you can. You also save a lot of indentation.

titodalcanton: Nevermind, I was wrong in my previous comment. I will actually work. But this looks unnecessarily complex, so my second comment still stands :)

GarethDaviesGW: Yeah that was what I wanted to do, and what it originally did, but having quite a few return statements made codeclimate complain

pycbc/events/single.py
```diff
@@ -31,31 +65,69 @@ def from_cli(cls, args, ifo):
...
+
+            # This uses the pycbc live convention of chisq always meaning the
+            # reduced chisq.
+            rchisq = cutall_trigs['chisq'][i]
+            nsnr = ranking.newsnr(cutall_trigs['snr'][i], rchisq)
+            dur = cutall_trigs['template_duration'][i]
+
+        if trigsremain and nsnr > self.thresholds['newsnr'] and \
+                dur > self.thresholds['duration'] and \
+                rchisq < self.thresholds['reduced_chisq']:
```
titodalcanton: At this point the trigger already passed the newsnr, duration and rchisq thresholds right? Can you avoid checking the thresholds again?

pycbc/events/single.py
titodalcanton: Weird. It seems to me that having multiple returns would lead to a far more readable code in this case. But who am I to disagree with the codeclimate overlords...

pycbc/events/single.py
GarethDaviesGW: Thats right - I will remove the check here, but I have just realised that I should be doing an SNR cut in the first round of cuts so that newsnr is only calculated if SNR is above threshold

titodalcanton:
> If there is a signal which is loud enough in single detector and coinc, fits the duration criteria, and so triggered both in coinc and single, would the mutiple events mechanism of gracedb not prevent multiple alerts in this case? I know that we should not rely on that, but I can see how it may actually be preferable to have single and coinc events sent to gracedb in this case

Right, it should not create problems thanks to the superevent mechanism. I prefer the search to report whatever it thinks is the best trigger, and only one trigger. Additional uploads might slow things down, waste computing cycles and create confusion. But I also do see the benefit of knowing that a candidate would *also* be significant as a single. Again though, you could convey that information in the log of the single upload, without doing multiple uploads.

Maybe this is a question for the whole group.

pycbc/events/single.py
GarethDaviesGW: So this was fixed in the end by solving the below issue - I had been using 5 returns when the maximum is 4 - it should be more readable now

pycbc/events/single.py
```diff
@@ -1,61 +1,152 @@
...
-                dur > self.duration_threshold:
-            fake_coinc = {'foreground/%s/%s' % (self.ifo, k): triggers[k][i]
-                          for k in triggers}
-            fake_coinc['foreground/stat'] = nsnr
-            fake_coinc['foreground/ifar'] = self.fixed_ifar
-            fake_coinc['HWINJ'] = data_reader.near_hwinj()
-            return fake_coinc
-        return None
+        rchisq = cutall_trigs['chisq'][i]
+        nsnr = ranking.newsnr(cutall_trigs['snr'][i], rchisq)
```
titodalcanton: Looks like you could recycle the previous NewSNR calculation, and replace the above two lines with
```python
nsnr = nsnr_all[nsnr_idx][i]
```
Really just aesthetics though, up to you.

GarethDaviesGW:
> I left one last minor comment but I think this is ready to go into master, assuming you have done some minimal testing. Once it is merged, I will get it running on the replay data.

Great! - it might be best to move the current single-fits-file to the pycbc.live account - it is currently on CIT in `/home/gareth.davies/public_html/testoutput/live_sngls_fitting/cluster/O3a-TRIGGER-FITS.hdf`

I agree with your above suggestion of thresholds of newsnr 9, duration 7 and chisq 2 to begin with

Issue #2979, opened 2019-10-23
Fake data SEOBNRV4 injections sample frequency problems


GarethDaviesGW:
When generating fake coloured Gaussian noise data, some inspiral jobs were failing with SEOBNRV4 injections

Error as follows:
```
XLAL Error - XLALSimIMRSpinAlignedEOBWaveformAll: Ringdown frequency &gt; Nyquist frequency!
At present this situation is not supported.
XLAL Error - XLALSimIMRSpinAlignedEOBWaveformAll (LALSimIMRSpinAlignedEOB.c:980): Input domain error
XLAL Error - XLALSimInspiralChooseTDWaveform (LALSimInspiral.c:920): Internal function call failed: Input domain error
```

Slack discussion when this was brought up:

Thomas Dent
@ Gareth Davies just a note to make sure that we don’t forget, you said you had some ‘ringdown>Nyquist’ failures in SEOBNRv4 injections in a run on gaussian noise. when you get a chance, would be good to try and work out where these are coming from

Ian Harry
Ah, when running on Gaussian noise the noise is generated at the sampling frequency (4096Hz) not at the 16384Hz that data is normally sampled at ... There is then no downsampling. But SEOB in particular doesn't like being generated at 4096Hz, so this might cause failures

Maybe we need an option to do downsampling on Gaussian noise to avoid this??

Thomas Dent 
SEOBNRv3 worked, for some reason

Alexander Harvey Nitz
@ Ian Harry @ tdent If you want SEOBNRv4, you could just inject the ROM version. That shouldn't have problems at lower sample rates.

Ian Harry
v3 has that failsafe hack, where it will generate again at higher sample rate if it fails initially
(Adding v4 there is another potential method to getting v4 to generate)

Thomas Dent
Well, we aren’t really needing to inject SEOBNRv4 over the whole space from component mass 2 upwards
@ Ian Harry though, why are you stating 4096Hz - standard analysis sampling rate is 2048Hz

Alexander Harvey Nitz
@ tdent I think just a thinko, it's resampled first to the requested sample rate before injections are made. https://github.com/gwastro/pycbc/blob/master/pycbc/strain/strain.py#L381
pycbc/strain/strain.py:381
            strain = resample_to_delta_t(strain, 1.0/opt.sample_rate)
￼gwastro/pycbc | Added by GitHub

Thomas Dent
@ Alexander Harvey Nitz hang on, how does the colored_noise function know what sampling rate to use before resampling?

[...]

PR #2981, opened 2019-10-23
Changes to perform downsampling after injections in fake data

GarethDaviesGW:
Addresses problems in issue #2979 

Perform down-sampling after injections - this means that the injections can be made up to the ring-down frequency in as many cases as possible

Note that in the zeroNoise case, the frequency is now hard-coded to 16384 for injection application, and it is then down-sampled. I don't see a way to avoid this, and it is returned to the sample rate that it would have been at a later point so should not make a significant difference

### DIFF
--- a/pycbc/strain/strain.py
+++ b/pycbc/strain/strain.py
@@ -780,6 +780,7 @@ def insert_strain_option_group_multi_ifo(parser, gps_times=True):
 
     return data_reading_group_multi
 
+
 ensure_one_opt_groups = []
 ensure_one_opt_groups.append(['--frame-cache','--fake-strain',
                               '--fake-strain-from-file',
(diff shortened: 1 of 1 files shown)


tdent:
@spxiwh @ahnitz is there a reason why we have `method='ldas'` for resampling in the real data conditioning path but not in the fake data one? 
Ideally if the fake data path is OK without it we should keep that aspect the same so that this change does not make any difference to the actual output in gaussian noise (or zero noise!) data.

ahnitz:
@tdent It won't really matter for most purposes (I can't think of any atm) if you use 'ldas' or not. LDAS is an FIR filter, and the default is a backwards forwards butterworth filter. Within enough padding though, it shouldn't matter which is used.

tdent:
Will approve & merge after Travis etc. run if no further comments

PR #2984, opened 2019-10-25
Optional arguments in mchirp, q <-> mass1, mass2 transform

cdcapano:
This allows the name of the parameters used for mass1, mass2, mchirp, and q for the `MchirpQtoMass1Mass2` transform (and inverse) to be an option set in the configuration file. This is needed to sample in source masses instead of detector frame masses (in which case the prior parameter is something like `srcmass(1|2)` instead of `mass(1|2)`).

Default is to use the standard parameter names, so no changes need to be made to the standard config files.

### DIFF
--- a/pycbc/transforms.py
+++ b/pycbc/transforms.py
@@ -1530,7 +1530,7 @@ class Mass1Mass2ToMchirpQ(MchirpQToMass1Mass2):
         self.q_param = q_param
         self._inputs = [self.mass1_param, self.mass2_param]
         self._outputs = [self.mchirp_param, self.q_param]
-        super(MchirpQToMass1Mass2, self).__init__()
+        BaseTransform.__init__(self)
 
 
 class Mass1Mass2ToMchirpEta(MchirpEtaToMass1Mass2):
--- a/pycbc/transforms.py
+++ b/pycbc/transforms.py
@@ -373,10 +373,10 @@ class MchirpQToMass1Mass2(BaseTransform):
             of transformed values.
         """
         out = {}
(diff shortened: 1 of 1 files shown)


pycbc/transforms.py
```diff
@@ -1492,13 +1509,30 @@ class Mass1Mass2ToMchirpQ(MchirpQToMass1Mass2):
...
+            mchirp_param = parameters.mchirp
+        if q_param is None:
+            q_param = parameters.q
+        self.mass1_param = mass1_param
+        self.mass2_param = mass2_param
+        self.mchirp_param = mchirp_param
+        self.q_param = q_param
+        self._inputs = [self.mass1_param, self.mass2_param]
+        self._outputs = [self.mchirp_param, self.q_param]
+        super(MchirpQToMass1Mass2, self).__init__()
```
cmbiwer: I think codacy is right that this should be `` super(Mass1Mass2ToMchirpQ, self).__init__()``.

cdcapano: @cmbiwer Ok, I called the `BaseTransform` directly there.

cdcapano:
@cmbiwer Catching up on old PRs... I replaced the call to `super` with a direct call to the appropriate class, as requested.

cdcapano:
@cmbiwer poke

cdcapano:
Making @dfinstad a reviewer since @cmbiwer hasn't responded.

cdcapano:
Thanks @dfinstad... looks like I can't commit until @cmbiwer approves since he requested changes.

dfinstad:
Should be good to go now @cdcapano

PR #2986, opened 2019-10-25
Add interpolation for comoving volume

cdcapano:
The functions `distance_from_comoving_volume` and `redshift_from_comoving_volume` are relatively slow. A single call can take O(10ms). When using this functions in a PE run (as you would if you want to use a prior uniform in comoving volume), this can make the waveform transforms be slower than even waveform generation, making it the dominant cost in the analysis. This fixes this by adding an interpolator that is similar to what is used for the distance to redshift function. The first time either of these functions is called with a standard cosmology, the interpolator is setup. The initial setup take O(10s). However, all successive calls are very fast, O(10us).

Since the initial call is slow, using interpolation is optional (with default is set to True). So, if you're just calling the function a few times, you can set `interp=False` to turn it off.

Plots comparing the interpolations to no interpolation:
 * [redshift](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/prs/interp_comoving_vol/z_check.png)
 * [distance](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/prs/interp_comoving_vol/d_check.png)

### DIFF
--- a/pycbc/cosmology.py
+++ b/pycbc/cosmology.py
@@ -435,13 +435,23 @@ _v2zs = {_c: ComovingVolInterpolator('redshift', cosmology=_c)
          for _c in astropy.cosmology.parameters.available}
 
 
-def redshift_from_comoving_volume(vc, **kwargs):
+def redshift_from_comoving_volume(vc, interp=True, **kwargs):
     r"""Returns the redshift from the given comoving volume.
 
     Parameters
     ----------
     vc : float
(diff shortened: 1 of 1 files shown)


cdcapano:
Making @ahnitz reviewer since I haven't heard from @sum33it. This patch was used in the 2-OGC results to use the uniform in comoving volume prior.

cdcapano:
@sum33it or @ahnitz poke

PR #2987, opened 2019-10-28
PyCBC Live: fix str/bytes error related to PTA files on Python 3

### LINKED ISSUE #2983: PyCBC Live: phase-time-amp files broken on Python 3
PyCBC Live is currently broken on Python 3, due to this:
```
RuntimeError: We already have one file with stat attr = phasetd_newsnr. Can't provide more than one!
```
Modify/remove [this loop](https://github.com/gwastro/pycbc/blob/master/pycbc/events/coinc.py#L766) and pass all the files to line 775, along with the ifos that are requested. Also modify the 'stat' attribute name in each file.

titodalcanton:
Fixes #2983. Requires a change in the PTA histogram files such that their `stat` attribute lists the detectors to which they apply.

### DIFF
--- a/pycbc/events/coinc.py
+++ b/pycbc/events/coinc.py
@@ -762,17 +762,8 @@ class LiveCoincTimeslideBackgroundEstimator(object):
         self.num_templates = num_templates
         self.analysis_block = analysis_block
 
-        # Only pass a valid stat file for this ifo pair
-        for fname in stat_files:
-            f = h5py.File(fname, 'r')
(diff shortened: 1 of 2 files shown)


tdent:
can you if you have rebased to current master? looks like one verbose message was already removed relative to the changes being displayed.

titodalcanton:
@tdent this PR is based off the last commit currently on master (ce6c2eb5eca44e2a4f716ea086a25073cec2cac1).

tdent:
ah, I see Alex's commit added one which is now being removed.

PR #3000, opened 2019-11-06
Add OSG tools to docker container

josh-willis:
The primary change introduced in this PR is to add the OSG repo to the Docker container, and then using that to additionally install the `osg-wn-client` and `osg-ca-certs` packages into the container. This in particular pulls in globus tools, so that the container can invoke `gfal-copy` without needing it to be present on the worker node via CVMFS or some other source. Since the worker node environments are by default present in container execution (so various directories under `/cvmfs` may have been added to `PATH`, `LD_LIBRARY_PATH`, `PYTHONPATH` and others), this new addition is intended to be invoked by adding the following three lines to the condor submit file for any OSG job desiring to ensure that the containerized versions of these tools are used:
```
+InitializeModulesEnv = False
+SingularityCleanEnv = True
getenv = False
```
Doing this will restrict the environment inside the Singularity container to contain only those variables defined by the `environment = ` line of the submit files, as well as some other, white-listed variables needed behind the scenes (such as `_CONDOR_<VARNAME>` variables, `X509_USER_PROXY`, and similar. The full white-list is in [these lines](https://github.com/opensciencegrid/osg-flock/blob/99428e261e538ae60b950a91cacf9ee55e3ea19a/job-wrappers/user-job-wrapper.sh#L371-L403)). 

If the lines above are _**not**_  added to the submit file, then the host node environment will continue to be propagated as before into the container, so users used to doing that should be able to run as before.

In addition to that primary change, this PR incorporates a few other improvements to the Docker and Singularity containers:
 1. We were [adding](https://github.com/gwastro/pycbc/blob/31de6c737ab88153b463c91c28aff2942700e259/docker/.singularity.d/env/90-environment.sh#L9-L14) `/opt/intel/composer_xe_2015.0.090/mkl/lib/intel64` to `LD_LIBRARY_PATH`, but that directory doesn't exist, and we instead now install `mkl` via `pip`. However we did [need to ensure](https://github.com/josh-willis/pycbc/commit/b978bb7c0085c1ca1145ef3894d99c3726f7f36b) that the container has run `ldconfig`, since `pip` didn't know to do that.
 1. In fact, there is also no reason to add the other directory to `LD_LIBRARY_PATH`. So [this commit](https://github.com/josh-willis/pycbc/commit/c7cce81bab259ca3ee65868e7161b73b375164a6) removes it, and adds the appropriate file to `/etc/ld.so.conf.d`. 
 1. Finally, I've [fixed the permissions](https://github.com/josh-willis/pycbc/commit/e6d8b69f1e9505c64d6097e885f84af3fdc9664f) on the `/opt/pycbc`directory in the container, since Singularity doesn't allow us to `su` to the `pycbc` user, and hence we could never access that directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,11 +8,11 @@ ADD docker/etc/cvmfs/60-osg.conf /etc/cvmfs/60-osg.conf
 ADD docker/etc/cvmfs/config-osg.opensciencegrid.org.conf /etc/cvmfs/config-osg.opensciencegrid.org.conf
 
 # Set up extra repositories
-RUN rpm -ivh http://software.ligo.org/lscsoft/scientific/7/x86_64/production/l/lscsoft-production-config-1.3-1.el7.noarch.rpm && yum install -y https://centos7.iuscommunity.org/ius-release.rpm && yum install -y https://ecsft.cern.ch/dist/cvmfs/cvmfs-release/cvmfs-release-latest.noarch.rpm && yum install -y cvmfs cvmfs-config-default && yum clean all && yum makecache && \
+RUN rpm -ivh http://software.ligo.org/lscsoft/scientific/7/x86_64/production/l/lscsoft-production-config-1.3-1.el7.noarch.rpm && yum install -y https://centos7.iuscommunity.org/ius-release.rpm && yum install -y https://ecsft.cern.ch/dist/cvmfs/cvmfs-release/cvmfs-release-latest.noarch.rpm && yum install -y cvmfs cvmfs-config-default && yum -y install http://repo.opensciencegrid.org/osg/3.4/osg-3.4-el7-release-latest.rpm && yum clean all && yum makecache && \
     yum -y groupinstall "Compatibility Libraries" \
                         "Development Tools" \
                         "Scientific Support" && \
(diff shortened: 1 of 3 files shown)


duncan-brown:
@josh-willis are all the `_CONDOR*` variables passed through in addition to the ones whitelisted at https://github.com/opensciencegrid/osg-flock/blob/99428e261e538ae60b950a91cacf9ee55e3ea19a/job-wrappers/user-job-wrapper.sh#L371-L403 ?

duncan-brown:
@josh-willis this all looks straightforward to me. If the answer to my question above is affirmative, please go ahead and merge. If not, we should discuss what `_CONDOR` variables should be whitelisted.

couvares:
> are all the `_CONDOR*` variables passed through in addition to the ones whitelisted at [here](https://github.com/opensciencegrid/osg-flock/blob/99428e261e538ae60b950a91cacf9ee55e3ea19a/job-wrappers/user-job-wrapper.sh#L371-L403)?

I believe that's what [this last line in your snippet](https://github.com/opensciencegrid/osg-flock/blob/99428e261e538ae60b950a91cacf9ee55e3ea19a/job-wrappers/user-job-wrapper.sh#L403) (developed by @bbockelm) attempts to do, yes, but he should confirm.

duncan-brown:
Oh, one more request: please can you document if and when people should set
```
+InitializeModulesEnv = False
+SingularityCleanEnv = True
getenv = False
```
and also make any necessary updates to the documentation here:

https://github.com/gwastro/pycbc/blob/master/docs/workflow/pycbc_make_coinc_search_workflow.rst#running-on-the-open-science-grid

josh-willis:
@duncan-brown I'll work on the doc updates this afternoon. @couvares is correct that [this line](https://github.com/opensciencegrid/osg-flock/blob/99428e261e538ae60b950a91cacf9ee55e3ea19a/job-wrappers/user-job-wrapper.sh#L403) will add any environment variable whose name begins with `_CONDOR_` to the whitelist.

PR #3001, opened 2019-11-07
Obtain QNM params from other QNM

micamu:
Conversion functions that take the frequency and damping time of a particular QNM and return the frequency and damping time of a new desired QNM. If the provided QNM frequency and damping time correspond to a negative mass or a BH spin > 0.9996, it returns NAN


micamu:
Codeclimate complains about similar blocks of code, but I would like to leave it as separate functions. @cdcapano @JulianWesterweck any comments?

micamu:
I did now encounter the same bug that @JulianWesterweck had experienced and applied his fix

ahnitz:
@micamu The code looks properly formated and  pretty straightforward. @JulianWesterweck @cdcapano If there aren't any objects in the next day or so, I think this should merger. Have you take a look at this yet?

PR #3002, opened 2019-11-07
qscans not working for fake data

GarethDaviesGW:
This input argument is needed for creation of fake strain, but is not included when running `pycbc.strain.insert_strain_option_group(parser)`

The other place this strain creation/argument is used is in pycbc_inspiral, which uses the input argument in other places, hence why is is not in the stain module option group

This change does not affect normal running, but fixes for fake data

### DIFF
--- a/bin/plotting/pycbc_plot_qscan
+++ b/bin/plotting/pycbc_plot_qscan
@@ -60,6 +60,8 @@ parser.add_argument('--time-windows', required=True, type=t_window,
                          'be provided as start1,end1 start2,end2 ... '
                          'where start/end times are relative to --center-time '
                          'and both positive')
+parser.add_argument("--low-frequency-cutoff", type=float,
+                    help="The low frequency cutoff to use for filtering (Hz)")
 
 parser.add_argument('--qtransform-delta-t', default=0.001, type=float,
                     help='The time resolution to interpolate to (optional)')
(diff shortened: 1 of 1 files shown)


GarethDaviesGW:
runs as normal on an example from standard workflow

titodalcanton:
After this patch, what happens if one tries to do a Q scan with fake data and does *not* give a `--low-frequency-cutoff`? Does the CLI handler produce a helpful message, or do we get an exception downstream?

GarethDaviesGW:
> After this patch, what happens if one tries to do a Q scan with fake data and does _not_ give a `--low-frequency-cutoff`? Does the CLI handler produce a helpful message, or do we get an exception downstream?

Tried it, and it produces an exception downstream:

```Traceback (most recent call last):
  File "/home/gdavies/environments/pycbc-dev/bin/pycbc_plot_qscan", line 127, in <module>
    strain = pycbc.strain.from_cli(opts, pycbc.DYN_RANGE_FAC)
  File "/home/gdavies/environments/pycbc-dev/lib64/python2.7/site-packages/pycbc/strain/strain.py", line 352, in from_cli
    raise ValueError('Please provide low frequency cutoff to '
ValueError: Please provide low frequency cutoff to generate a fake strain```

tdent:
@titodalcanton the error message here seems to be the intended one, someone wrote it in the strain `from_cli` method to catch this specific problem.  It seems to be OK for this purpose.  There might be a couple of ways to make it 'nicer' in general - eg provide a generic strain error checking routine to be invoked directly after parsing args - but that would require more extensive changes.  Also since the `low-frequency-cutoff` option name is used by different executables with *different* meanings (filter cutoff, PSD calculation cutoff ..) we can't isolate it in the strain option group.  

In principle we could change the name of the option when used to produce fake strain to eg `fake-strain-low-frequency` and then put it into the strain option group. However that doesn't seem to be necessary immediately.  (It may be in the future if we continue doing fake strain analysis ..)

PR #3003, opened 2019-11-08
PyCBC Live: misc improvements

### LINKED ISSUE #2713: pycbc live depends on bank template start frequency > low_frequency_cutoff
Low frequency cutoff is not enforced in pycbc live. It relies on the template starting frequency being greater than the low frequency cutoff (which is used for the psd estimation).

titodalcanton:
* Make sure the low-frequency cutoff used to make the PSD is ok for all templates (fixes #2713).
* Improve followup plots in GraceDB (sort detectors alphabetically, show more PSD).
* Show GraceDB ping response when testing access.
* Shorten a very common log messange.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -562,16 +562,17 @@ pycbc.init_logging(args.verbose, format=log_format)
 ctx = scheme.from_cli(args)
 fft.from_cli(args)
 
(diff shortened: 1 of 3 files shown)


titodalcanton:
This is now tested.

PR #3009, opened 2019-11-18
max mchirp option: slim down code & None-proof

tdent:


tdent:
I figured there was no point doing `if self.mcm is not None` twice ..

PR #3011, opened 2019-11-19
pycbc_optimal_snr: support f_ref/f_final options

titodalcanton:


titodalcanton:
Yes, it may make sense to have an `add_injection_options` class method in InjectionSet (used both by `strain.py` and `pycbc_optimal_snr`). I will have another look at the code to see how other options would play with that.

titodalcanton:
It looks like a unified option-inserting function comes with a few issues:

* It changes `pycbc_optimal_snr`'s command line interface (replace `--input-file` with `--injection-file`). In principle ok, but I do not feel like doing it right now.
* It prevents making `--injection-file` a required option which, is the current behavior in `pycbc_optimal_snr`.

So for now I am merging this as is, and we can consider some refactoring later with a bigger scope (including sine-Gaussian injections etc).

PR #3021, opened 2019-11-21
StopIteration Fixed in qtransform

lbarosi:
From Python 3.7 StopIteration raises a Runtime Execution Error.
This broken qtransform.
Fix proposed in PEP479 implemented. (Change Rayse StopIteration() to retuirn)


ahnitz:
@lbarosi Thank you very much for your contribution!

PR #3025, opened 2019-11-25
page_coincinfo to work in multiifo general case

GarethDaviesGW:
Make page_coincinfo work in the general case

This is as full data foreground, full data background and injections all have different requirements and this seems the most general way to get them all to work, and for it to be obvious that the requested value does not exist

### DIFF
--- a/bin/minifollowups/pycbc_page_coincinfo
+++ b/bin/minifollowups/pycbc_page_coincinfo
@@ -72,46 +72,25 @@ else:
 
 # Make a table for the coincident information #################################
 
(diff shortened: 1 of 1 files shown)


GarethDaviesGW:
I agree that it would be better if the -1 was not shown - similar situation to the way the PSD variation part works. I will see if it looks simple enough to code this up, but will merge this for now

GarethDaviesGW:
Looks like this is working such that the values which are not there are now not printed

This has been tested and works on both multiifo-style and 2-ifo style statmap files

Re-requesting review as this has been a non-trivial change

PR #3027, opened 2019-11-26
make pillow optional

### LINKED ISSUE #3026: Does PyCBC actually need pillow?
PyCBC has a dependency on Pillow:

https://github.com/gwastro/pycbc/blob/60502a8a097e0692c9930cc9f3f0816e46eb6c4b/setup.py#L45

but I can't see that being used anywhere. Can this dependency be removed?

ahnitz:
This makes the Pillow library optional. Codes won't fail without it. It's still in requirements.txt, as it's useful for all the webpage metadata, but the base library won't require it if that's all you want.

Closes #3026.


duncanmmacleod:
@ahnitz, can you rebase this and merge?

ahnitz:
I've found an edge case where this causes a problem. This PR isn't super high priority, so I'm holding off for the moment.

PR #3032, opened 2019-11-27
further cleanup of __init__.py

ahnitz:
Depends on #2946 

* Removes variables in __init__.py only required for weave
* Remove use of functions that only set up weave


ahnitz:
@spxiwh We have avoid the 1.15.4 version of numpy to avoid a bug that affects astropy. I've moved the minimum version to 1.16.0 to make it simple.

PR #3036, opened 2019-11-28
Several fixes to pycbc_optimal_snr

titodalcanton:
Several little fixes to `pycbc_optimal_snr` prompted by the observation of a weird memory leak and subsequent digging into various scripts and functions.

* Add an option to choose the sorting of input injections. Sorting by time may help opening less PSD files, although it does not seem to matter much anyway.
* LALSimulation's `SimInspiralTaylorLength()` (used in the injection module) is much more complex than I thought. It does a numerical integration and seems to have a [memory leak](https://git.ligo.org/lscsoft/lalsuite/issues/224). Replace with a trivial call to `pycbc.conversions.tau0_from_mass1_mass2()`, which gives very close or systematically larger values.
* Avoid actually using multiprocessing when using a single core.
* Clean up code and fix an integer division.

### DIFF
--- a/bin/pycbc_optimal_snr
+++ b/bin/pycbc_optimal_snr
@@ -46,13 +46,12 @@ class TimeIndependentPSD(object):
 
 class TimeVaryingPSD(object):
     def __init__(self, file_name, length=None, delta_f=None, f_low=None):
-        f = h5py.File(file_name, 'r')
-        self.file_name = file_name
-        detector = tuple(f.keys())[0]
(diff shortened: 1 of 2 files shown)


titodalcanton:
This has now been checked by comparing optimal SNRs from one of the offline boxes before and after the change. Over several thousand SpinTaylorT2 BNS injections and SEOBNRv4 BBH injections, I observe a maximum absolute SNR difference of 1e-5.

PR #3037, opened 2019-11-28
Chirp vol

tdent:
If we do injections out to a fixed chirp distance to approximate the maximum sensitive distance for a given mass, the new calculation effectively estimates the fraction of the (mass-dependent) potential sensitive volume where we are actually able to find injections.


tdent:
use existing dchirp function -> prone to typing errors

PR #3044, opened 2019-12-04
Injections are optional, fix inj_summ var

stevereyes01:
The issue comes from the assignment of the `inj_summ` variable in an if statement for handling injection files. If there aren't any injections then the workflow will try to give the injection summary list to the summary list when `inj_summ` hasn't been created.

Patch is just to add an if-else statement. Not the most elegant solution to fixing `pycbc_create_offline_search_workflow`. But now the workflow generates and passes with this small patch.


stevereyes01:
I'm going to squash and merge it then.

PR #3050, opened 2019-12-06
Allow for mutiple wells in pycbc results pages and allow them to be collapsable

ahnitz:
To invoke multiple wells you need to use the 'unique' keyword argument with your layout command. Add  any string so it can make a new one.  To make collapseable, add the collapse=True keyword.

### DIFF
--- a/pycbc/results/layout.py
+++ b/pycbc/results/layout.py
@@ -18,19 +18,21 @@
 import os.path
 from six.moves import zip_longest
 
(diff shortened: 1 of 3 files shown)


ahnitz:
@cdcapano Did this work out for you? If so, we can probably merge to master.

ahnitz:
@cdcapano @cmbiwer Any thoughts?

PR #3052, opened 2019-12-09
Remove usetex from remaining plotting scripts

duncanmmacleod:
TexLive is extremely hard to build and distribute alongside other projects, and is not supported by Anaconda or Conda-forge to the level required for Matplotlib. This PR removes the remaining `rcParams['text.usetex'] = True` settings to remove the LaTeX dependency from PyCBC.

### DIFF
--- a/bin/hdfcoinc/pycbc_page_ifar
+++ b/bin/hdfcoinc/pycbc_page_ifar
@@ -40,9 +40,6 @@ def calculate_time_slide_duration(ifo1_segments, ifo2_segments, offset=0):
    return duration
 
 
-pylab.rc('text', usetex=True)
-pylab.rc('font', **{'family': 'serif', 'serif': ['Computer Modern']})
-
 # parser command line
 parser = argparse.ArgumentParser(usage='pycbc_page_ifar [--options]',
                     description='Plots a cumulative histogram of IFAR for'
--- a/bin/hdfcoinc/pycbc_page_snrifar
+++ b/bin/hdfcoinc/pycbc_page_snrifar
@@ -32,9 +32,6 @@ def _far_from_p(p, livetime, max_far):
 
 far_from_p = numpy.vectorize(_far_from_p)
 
-pylab.rc('font', **{'family': 'serif', 'serif': ['Computer Modern']})
-pylab.rc('text', usetex=True)
-
 parser = argparse.ArgumentParser()
 # General required options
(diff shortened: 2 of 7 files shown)


ahnitz:
@duncanmmacleod Ok, I see the point of this PR, but I'm not sure how I feel about the result. Do you know if there is there way to check for a working latex install?

duncanmmacleod:
Given that latex is only used in _some_ of the pycbc plots, I was hoping this would squeak through unnoticed. Anyways, gwpy does two things to test whether latex is present, first in the runtime library, the following is normally enough:

https://github.com/gwpy/gwpy/blob/7b90932b25291990d43d7d1fb9c3686d0665e107/gwpy/plot/tex.py#L37-L53

But, we also have a more robust check used in the test suite:

https://github.com/gwpy/gwpy/blob/7b90932b25291990d43d7d1fb9c3686d0665e107/gwpy/testing/fixtures.py#L53-L67

ahnitz:
@spxiwh What do you think?

duncanmmacleod:
@spxiwh, my suggestion is to remove the latex _requirement_ from the pipeline code, meaning that the standard pycbc engine runs without latex, making it significantly more portable. However, the idea of retaining the ability to opt-into using latex is a good one.

Regarding latex and conda, this is indeed frustrating, but is borne out of the fact that latex is not packaged 'properly' by the upstream developers for any system, but huge efforts are put in by debian/redhat teams to make it available on those platforms - we simply don't have the knowledge or resources to take this on (seemingly Anaconda doesn't have the desire, and conda-forge also doesn't have the resources). Because of the way it is implemented in `matplotlib`, all you have to do is make sure latex is available on the host outside of the conda environment, and it will be used by `matplotlib` inside - this is the best solution we have today for those requiring latex.

duncanmmacleod:
@ahnitz, @spxiwh, I have rebased this branch to resolve the conflicts. AFAICT latex is the only issue preventing pycbc analyses from running end-to-end from conda environments, so it would be great to get this resolved.

duncanmmacleod:
Ping @ahnitz, @spxiwh.

ahnitz:
@duncanmmacleod I'll punt this to @spxiwh to decide. I think it affects people in the LSC more (as you all are more likely to use some of the standard build environments). It's a non-issue either way for me personally, so I have no objections at this point.

PR #3060, opened 2019-12-17
Results page logo

GarethDaviesGW:
Adding in a logo to the orange.html template

This also moves the page title and subtitle to the centre of the page, in order to make room for the logo

There are a couple of things I've done that I would have preferred a 'proper' way to do it
1) I had to make a new .png logo so that it would show up - I'll let you decide what it should be so at the moment it is a placeholder
2) The logo is currently copied to the root of the folder and the symlinked to all subfolders, I would prefer a 'proper' way to do it - using the 'root' of the webpage (HTML-style) doesn't work
3) I don't know if the user even _should_ be allowed to choose their own logo - it may be best to point to the pycbc-logo repository, in which case this change can be made a _lot_ easier than it currently is

Also as this is aesthetic, it can be ignored if there is opposition to it

Example using the placeholder logo I added:
image


pycbc/results/templates/orange.html
```diff
@@ -100,11 +100,14 @@
 
         <!-- title and subtitle -->
         <div class="row" style="padding-top:10px">
-            <div class="col-md-6">
-                <p class="text-left"><b>{{analysis_title}}</b></p>
+            <div class="col-md-4">
```
cmbiwer: Although this works, it used to be the case that Bootstrap scaffolding default rows were defined in blocks of 12 (.e.g 6 + 6 as it was previously). https://getbootstrap.com/2.3.2/scaffolding.html

I would suggest altering this so that it equals 12 again.

bin/pycbc_make_html_page
```diff
@@ -214,6 +219,12 @@ for cwd in dirs:
     cwd.files.sort(key=lambda x: x.path)
 dirs_0.sort(key=lambda x: x.title())
 
+local_logo = opts.logo[:4] != "http"
+if local_logo:
+    logo_ext = opts.logo.split(".")[-1]
+    logo_loc = opts.output_path + '/logo.' + logo_ext
+    shutil.copy2(opts.logo, logo_loc)
```
cmbiwer: In the HTML template the substitution is only ``{logo}`` which seems appropriate. But why is there so much work being done to rename the image to ``logo + logo_ext``? Why not use ``os.path.basename`` and then copy ``shutil.copy2`` and keep the original file? It seems a lot of this code could be removed.

bin/pycbc_make_html_page
```diff
@@ -224,6 +235,18 @@ for cwd in dirs:
                'dir'               : cwd,
                'dot_dot_str'       : cwd.level() * '../',
                'plots_dir'         : opts.plots_dir}
+    if local_logo:
+        if not cwd.path == "":
+            print("CWD path: " + cwd.path)
```
cmbiwer: I would recommend removing this debug print statement.

bin/pycbc_make_html_page
```diff
@@ -224,6 +235,18 @@ for cwd in dirs:
...
+            cwd_logo_loc = opts.output_path + cwd.path + '/logo.' + logo_ext
+            try:
+                os.symlink(logo_loc, cwd_logo_loc)
```
cmbiwer: Why simlink? Why not link via the HTML path directly to the one image you copied already?

cmbiwer: Ah, wait nevermind. Thinking about this for a few seconds, I know why it doesn't work. Would have to think of an alternative.

pycbc/results/templates/orange.html
GarethDaviesGW: I've added in an empty div so that it still adds up to 12

bin/pycbc_make_html_page
GarethDaviesGW: I thought that it may be better to have logo being standardly named so that it would be more easily identified within the directory, but that doesn't make sense if allowing different extensions

using os.path.basename now as suggested

bin/pycbc_make_html_page
GarethDaviesGW: I used the dot_dot_str as you suggested in slack to allow the logo to be held only in the base directory

cmbiwer:
I've approved the changes, looks good now.

Though line 222 is not PEP-8. If PyCBC is still enforcing that strictly, then you need spaces around the `=` char. Comments help as well but I think its pretty straightforward.

ahnitz:
@GarethDaviesGW I would support just pointing to the pycbc logo repo and committing the png version there instead.

GarethDaviesGW:
For adding the png to the repo, see https://github.com/gwastro/pycbc-logo/pull/1, and I will update the default location to be there once that has been merged, as well as the PEP8 changes

PR #3068, opened 2020-01-03
enforce betas order required by emceept

ahnitz:
This is a convenience modification. I've found that if you get the betas order wrong, emcee pt will happily accept it but produce nonsense results. The order it requires is always descending, so this just makes it so that we always presort to it's desired order.


ahnitz:
@cdcapano Thoughts?

PR #3070, opened 2020-01-05
allow for negation of pycbc types

ahnitz:
This implements the negation operator so you can do 

x = -a

instead of the obtuse

x = -1 * a


josh-willis:
@ahnitz Perhaps I'm forgetting, but does this really need the `@_checkother`, since the only argument is `self`?

ahnitz:
@josh-willis You are right, that can be removed. I've updated the PR without it.

PR #3076, opened 2020-01-08
hierarchical iterations counter bug

GarethDaviesGW:
h_iterations counter is higher than number of hierarchical removals by 1 at the time that it is put into output files - this was not taken into account at the time of writing

(My fault for not properly testing previously - sorry!)


GarethDaviesGW:
Requesting more reviewers as has been a while (review should be easy - it's just moving an increment counter to after the break in a while statement!)

Issue #2962, opened 2019-10-04
Add global IFAR plot for multi-ifo pipeline


tdent:
While we can't assign IFARs to time slides in the multi-ifo pipeline globally (ie over all coinc types) we can show the foreground IFARs against the expected noise count. Eg this made from an O2 chunk 20 run - 
![image](https://user-images.githubusercontent.com/3316412/66209369-7c47e180-e6b7-11e9-85d9-cbbdd15f2e0a.png)
The code to produce the plot exists (PR #2961), it may need some plumbing to insert into the pipeline.

GarethDaviesGW:
Fixed by #3078

PR #3078, opened 2020-01-10
Global ifar plot

GarethDaviesGW:
Adding in global IFAR plot to summary pages

Tested for closed and open boxes, two ifo statmap and multiifo combined statmap

PR #3076 is required in order to produce correct results when hierarchical removal number is not given

Solves issue #2962

### DIFF
--- a/bin/plotting/pycbc_ifar_catalog
+++ b/bin/plotting/pycbc_ifar_catalog
@@ -232,6 +232,8 @@ if opts.open_box:
     # scale the plot around the foreground
     pylab.ylim(0.7, 0.05 * len(fore_cumnum))
     pylab.xlim(30 * min(fore_ifar), 3 * max_ifar)
+else:
+    pylab.ylim(0.7, max(expected_cumnum))
 pylab.grid()
 pylab.legend(loc='upper right', fontsize=13)
 pylab.ylabel('Cumulative Number', size='large')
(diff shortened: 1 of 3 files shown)


tdent:
do you have a link for the tests?

GarethDaviesGW:
> do you have a link for the tests?

See https://ldas-jobs.ligo.caltech.edu/~gdavies/testoutput/ifar_plot_alltypes/ - all the plots which start with multiifo or twoifo

tdent:
hmm, the closed plots show no useful data so should they be made at all?

the open plots cut off the contours at 10^-4yr, is there a specific reason for that?  are there options for controlling the min ifar plotted and/or the min ifar at which contours exist? 

in the multiifo open plots, what should be the difference between 'hrone' and 'hrnone'?

GarethDaviesGW:
> hmm, the closed plots show no useful data so should they be made at all?

Probably not, I put them in simply as the --open-box argument was vailable and the closed box page_ifar plot was made already - I will remove these from the workflow

> the open plots cut off the contours at 10^-4yr, is there a specific reason for that? are there options for controlling the min ifar plotted and/or the min ifar at which contours exist?

This is an option and would be set in the ini files, I chose 10^4 years ifar as the option that I thought looked prettiest

> in the multiifo open plots, what should be the difference between 'hrone' and 'hrnone'?

this test was to check that they _do_ look the same - hrnone means that no HR level was given as an argument, in which case it defaults to the number of HR removals as given by the attribute in the statmap file, which was 1 in this case

PR #3079, opened 2020-01-10
Silence annoying matplotlib messages

cdcapano:
For the past year, I've been getting a bunch of annoying messages from matplotlib about finding fonts whenever I run any of the inference plotting codes with verbose on. It turns out the issue is that the `pycbc.init_logging` function sets the logging level to `DEBUG` if you pass `verbose=True`. At some point, matplotlib must of added the finding font messages at the DEBUG level. The problem goes away if you set the logging level to `INFO`.

So... this patch sets the initial logging level in `pycbc.init_logging` to `INFO` if `verbose=True` rather than `DEBUG`. I added an option `level` keyword argument if you want to set it to `DEBUG` instead. This is more sane I think, since it will result in all of our `logging.log` messages to be printed without all the extra crud.

### DIFF
--- a/pycbc/__init__.py
+++ b/pycbc/__init__.py
@@ -46,12 +46,22 @@ except:
 __version__ = pycbc_version
 
 
-def init_logging(verbose=False, format='%(asctime)s %(message)s',
-                 level=None):
-    """ Common utility for setting up logging in PyCBC.
+def init_logging(verbose=False, format='%(asctime)s %(message)s'):
+    """Common utility for setting up logging in PyCBC.
 
     Installs a signal handler such that verbosity can be activated at
     run-time by sending a SIGUSR1 to the process.
+
+    Parameters
+    ----------
+    verbose : bool or int, optional
+        What level to set the verbosity level to. Accepts either a boolean
+        or an integer representing the level to set. If True/False will set to 
+        ``logging.INFO``/``logging.WARN``. For higher logging levels, pass
+        an integer representing the level to set (see the ``logging`` module
+        for details). Default is ``False`` (``logging.WARN``).
+    format : str, optional
+        The format to use for logging messages.
     """
     def sig_handler(signum, frame):
         logger = logging.getLogger()
@@ -66,14 +76,15 @@ def init_logging(verbose=False, format='%(asctime)s %(message)s',
 
     signal.signal(signal.SIGUSR1, sig_handler)
 
-    if level is None:
-        level = logging.INFO
-    if verbose:
-        initial_level = level
-    else:
+    if not verbose:
         initial_level = logging.WARN
+    elif int(verbose) == 1:
+        initial_level = logging.INFO
+    else:
+        initial_level = int(verbose)
     logging.getLogger().setLevel(initial_level)
     logging.basicConfig(format=format, level=initial_level)
+
 
 def makedir(path):
     """
--- a/pycbc/__init__.py
+++ b/pycbc/__init__.py
@@ -56,7 +56,7 @@ def init_logging(verbose=False, format='%(asctime)s %(message)s'):
     ----------
     verbose : bool or int, optional
         What level to set the verbosity level to. Accepts either a boolean
-        or an integer representing the level to set. If True/False will set to 
+        or an integer representing the level to set. If True/False will set to
         ``logging.INFO``/``logging.WARN``. For higher logging levels, pass
         an integer representing the level to set (see the ``logging`` module
         for details). Default is ``False`` (``logging.WARN``).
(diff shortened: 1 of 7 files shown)


ahnitz:
@cdcapano Two things to think about here. 
1) How does this patch interact with the toggle ability? It looks like you'll be back to debug if you use that. Does that make sense? 
2) Should another option really be added here? The main purpose is to support making to do logging from the executable perspective. Would the natural thing be to upgrade this function to support a numeric instead of a boolean? This way you can do things like -vvv or similar. To get different logging levels. 
@titodalcanton What do you think?

cdcapano:
@ahnitz 

> How does this patch interact with the toggle ability? It looks like you'll be back to debug if you use that. Does that make sense?

Won't the signaling now cause you to turn on INFO rather than DEBUG? If you want DEBUG, yeah, that will take some more work. Maybe make it so that SIGUSR1 gives INFO, while SIGUSR2 gives DEBUG?

> Should another option really be added here? The main purpose is to support making to do logging from the executable perspective. Would the natural thing be to upgrade this function to support a numeric instead of a boolean? This way you can do things like -vvv or similar. To get different logging levels.

Yeah, I agree that would be best, and more similar to what programs generally do. However, a boolean verbose option is used in a large number of our executables, so changing that will require a large patch. We could make it so that the first argument accepts a level, along with a boolean. If a boolean is passed for verbose option, then it turns on INFO (so, basically mapping `True/1` to `logging.INFO (= 20)`. Otherwise, it sets the level. Then all current executables would work with it as-is, with the option of adding more verbosity being there if you want to add it to a program.

pycbc/__init__.py
```diff
@@ -65,8 +66,10 @@ def sig_handler(signum, frame):
 
     signal.signal(signal.SIGUSR1, sig_handler)
 
+    if level is None:
```
titodalcanton: I suggest setting `level=logging.INFO` in the function arguments so you can remove this `if` statement.

titodalcanton:
@cdcapano I like your last proposal of supporting increased verbosity levels, but defaulting to `INFO` if only the traditional `--verbose` is given. Of course we will have to make each program aware of that, but that can take its time.

Another think to consider is whether we actually use `logging.info()` or `logging.debug()` throughout our codebase. I think we have a bit of a mix, and I would like to be sure that we do not accidentally disable some important logging messages by applying this PR. I.e. it is probably worth doing a `git grep logging.debug`.

ahnitz:
@titodalcanton @cdcapano I think this could be done in a backwards compatible way, if the scripts accepts True and 1 as the same thing (and zero and false). That way we can have 2 (double verbose) be debug, etc or whatever higher levels we want and old scripts don't need to move over until we feel like it.

titodalcanton:
By the way, it looks like @spxiwh already used [another solution](https://github.com/gwastro/pycbc/blob/master/bin/pycbc_optimize_snr#L137) to the matplotlib verbosity issue. Maybe we can get rid of that when this PR is merged.

cdcapano:
@titodalcanton @ahnitz Ok, I changed the patch so that you can set the debugging level with the verbose option, with `True` mapping to `logging.INFO`. I did not try to change the signal interupt though; if you do that, it will switch to `logging.DEBUG`. I figure if you're doing that, you probably want to debug anyway.

@titodalcanton I checked for usage of `logging.debug` as you suggested. Here is what I got:
```
bin/pycbc_optimal_snr:            logging.debug('Trying injection %s', inj.simulation_id)
bin/pycbc_optimal_snr:            logging.debug('Injection %s completed', inj.simulation_id)
bin/pygrb/pycbc_pygrb_pp_workflow:    logging.debug("Executable name: %s" % exec_name)
bin/pygrb/pycbc_pygrb_pp_workflow:    logging.debug("Executable name: %s" % exec_name)
pycbc/vetoes/bank_chisq.py:            logging.debug(errMsg)
pycbc/workflow/core.py:            logging.debug("Using %s executable "
pycbc/workflow/core.py:        logging.debug("%s executable will run as %s universe"
pycbc/workflow/core.py:    logging.debug(msg)
pycbc/workflow/core.py:    logging.debug("Call successful, or error checking disabled.")
pycbc/workflow/datafind.py:            logging.debug(msg)
pycbc/workflow/datafind.py:    logging.debug("Asking datafind server for frames.")
pycbc/workflow/datafind.py:    logging.debug("Frames returned")
```
This does mean with the new default for verbose these messages won't be printed. Do you want me to change these to `logging.info`?

ahnitz:
@cdcapano I would change those to info so they print by default (like now). We can make it more granular when those exes decide to implement setting a level.

cdcapano:
@ahnitz Ok, done. Once Travis passes, I'll merge.

PR #3080, opened 2020-01-10
Remove ArrayWithAligned

josh-willis:
The original motivation for sub-classing `ndarray` in this way was to allow checking that an array had sufficient alignment for SIMD vectorized operations without incurring the overhead of performing that calculation over and over.

Since we have moved to providing classes to implement our time-critical operations (e.g. FFT, correlation, peak-finding) we avoid this overhead simply by using the same arrays over and over, and only checking alignment at class instantiation. In the meantime, it is often inconvenient to have a subclass of ndarray, though arguably this is because other packages don't properly check class types via `isinstance`. However, we have no control over what other packages do, so I think for us it is better to eliminate this subclass.

I noted that at some point our SIMD vectorized version of correlation got decoupled from our class types (e.g. `CPUCorrelator`). That should be plugged back in at some point, at which point a check for alignment should be added, as that function will fail if its arguments are not aligned.

### DIFF
--- a/pycbc/fft/fftw.py
+++ b/pycbc/fft/fftw.py
@@ -320,14 +320,14 @@ def execute(plan, invec, outvec):
 
 def fft(invec, outvec, prec, itype, otype):
     theplan, destroy = plan(len(invec), invec.dtype, outvec.dtype, FFTW_FORWARD,
-                            get_measure_level(),(check_aligned(invec._data) and check_aligned(outvec._data)),
+                            get_measure_level(),(check_aligned(invec.data) and check_aligned(outvec.data)),
                    _scheme.mgr.state.num_threads, (invec.ptr == outvec.ptr))
     execute(theplan, invec, outvec)
     destroy(theplan)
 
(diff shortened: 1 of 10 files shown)


josh-willis:
@ahnitz I've fixed the codeclimate issues here that I'm willing to fix---I don't agree with always truncating lines to a certain length at the expense of readability.  

When you get a chance, please take a look, as I have some other pending changes I'd like to propose after this is merged (that's assuming there are no further travis issues after the rebase, of course).

Alternatively, if you'd rather not make this change now, let me know, and I will propose an alternative that keeps the `ArrayWithAligned` class but modifies the `.numpy()` method in `pycbc.types.Array` to always return an actual, real instance of `numpy.ndarray`.

josh-willis:
@ahnitz or @spxiwh (whom I just added).  I'd like to get this change merged, or, if people want to keep the ArrayWithAligned class, to make and approve an alternate PR that allows the `.numpy()` method to return an actual numpy array.  Some other PRs I want to make are blocked on this.  Thanks.

PR #3081, opened 2020-01-13
Improve PyCBC Live SNR followup

titodalcanton:
* Do not steal CPU cores from the main search when doing the SNR optimization
* SNR optimization timeout is now configurable from the command line
* Minor fixes


titodalcanton:
I bet. Maybe we should consider raising an error if PyCBC Live finds out that it is running in a too restricted platform.

PR #3082, opened 2020-01-14
add in followup data to live singles

GarethDaviesGW:
Add in ability to get followup data from other interferometers but without recalculating the IFAR of a single detector event

Example output: https://gracedb-playground.ligo.org/events/T244462

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -115,7 +115,7 @@ class LiveEventManager(object):
             raise RuntimeError("Not root process")
 
     def compute_followup_data(self, ifos, triggers, data_readers, bank,
-                              followup_ifos=None):
+                              followup_ifos=None, recalculate_ifar=False):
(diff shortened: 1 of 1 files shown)


bin/pycbc_live
```diff
@@ -314,8 +317,11 @@ class LiveEventManager(object):
 
             if single is None:
                 continue
+
+            other_ifos = [i for i in active if i is not ifo]
```
titodalcanton: Use set operations here, they are cool. Grep for `set(` for examples in `pycbc_live`.

GarethDaviesGW: I've looked and implementation of set operations in this case looks tricky, and degrades readability (and efficiency) of the code quite a bit, mainly as the ifos are not already available as a list at this point

I think that leaving it without set operations makes sense in this case

PR #3083, opened 2020-01-14
Pin versions of setuptools and pip in Dockerfile to workaround dropped python2 support

josh-willis:
Our travis tests are all failing the `PYCBC_CONTAINER=pycbc_docker DOCKER_IMG=pycbc/pycbc-el7` test because the release of `setuptools 45.0.0` dropped support for Python 2.  Thus we cannot simply upgrade pip and setuptools in the Dockerfile.  As a workaround, this PR simply pins both setuptools and pip to working versions (pip has not yet shown a problem, but I went ahead and included it in the pin).

We should of course revert this as soon as we are able to drop python2 support ourselves.


josh-willis:
@spxiwh I think that sounds reasonable, and I'll do that once merged unless someone else comments about  a better, alternate system for tracking such things.

PR #3087, opened 2020-01-18
Fixed the overlapping label problem

KoustavChandra:
Fixed the pycbc_plot_waveform overlapping subplots label problem. 
Also changed the caption as no insets are displayed

### DIFF
--- a/bin/plotting/pycbc_plot_waveform
+++ b/bin/plotting/pycbc_plot_waveform
@@ -21,12 +21,9 @@
 import sys, argparse, matplotlib, numpy, math
 matplotlib.use('Agg')
 import matplotlib.pyplot as plt
-from mpl_toolkits.axes_grid1.inset_locator import zoomed_inset_axes
-from mpl_toolkits.axes_grid1.inset_locator import mark_inset
 from pycbc import waveform, io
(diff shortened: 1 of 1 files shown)


GarethDaviesGW:
Do you have any examples of the fixed label issues?

The zoomed waveform plot should still happen in the cases where the waveform duration is above a set value so these imports should not be removed, make sure you test the longer waveforms for this case as well. 

See https://github.com/gwastro/pycbc/blob/29822cc643def18020ba327d5a0aafd17b20665b/bin/plotting/pycbc_plot_waveform#L147 for the code and an example output is https://ldas-jobs.ligo.caltech.edu/~gdavies/o3/runs/hl/c00/a8_initial/3._single_triggers/3.10_H1_loudest_all_time_duration_gt_4.0_sec/H1-PLOT_WAVEFORM_P1_0-1243161342-625572.png

It would be good to update the caption so it mentions the inset plot only if it was made

KoustavChandra:
Hi, Gareth I have addressed that issue.  The images can be accessed by following this [link](https://ligo.gravity.cf.ac.uk/~koustav.chandra/LSC/images/).

Issue #3088, opened 2020-01-18
PyCBC Live can still crash due to inconsistently sampled SNRs


titodalcanton:
An old problem shows up again:
```
2020-01-17 19:10:54,744 node742 Adding V1 to candidate, pvalue 0.8742921857304643, 3531 samples
2020-01-17 19:10:54,745 node742 Generating SPAtmplt, 312.0s, 1740, starting from 25.1 Hz
2020-01-17 19:10:55,062 node742 Adding L1 to candidate, pvalue 0.32945178594792623, 7642 samples
Traceback (most recent call last):
  File "/usr/lib64/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/pycbc.live/testing/O2_replay/pycbc_live_O2_replay_env_py3/lib/python3.6/site-packages/mpi4py/__main__.py", line 7, in <module>
    main()
  File "/home/pycbc.live/testing/O2_replay/pycbc_live_O2_replay_env_py3/lib/python3.6/site-packages/mpi4py/run.py", line 196, in main
    run_command_line(args)
  File "/home/pycbc.live/testing/O2_replay/pycbc_live_O2_replay_env_py3/lib/python3.6/site-packages/mpi4py/run.py", line 47, in run_command_line
    run_path(sys.argv[0], run_name='__main__')
  File "/usr/lib64/python3.6/runpy.py", line 263, in run_path
    pkg_name=pkg_name, script_name=fname)
  File "/usr/lib64/python3.6/runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "/usr/lib64/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/pycbc.live/testing/O2_replay/pycbc_live_O2_replay_env_py3/bin/pycbc_live", line 802, in <module>
    args.low_frequency_cutoff)
  File "/home/pycbc.live/testing/O2_replay/pycbc_live_O2_replay_env_py3/bin/pycbc_live", line 350, in check_singles
    channel_names=args.channel_name)
  File "/home/pycbc.live/testing/O2_replay/pycbc_live_O2_replay_env_py3/lib/python3.6/site-packages/pycbc/io/live.py", line 119, in __init__
    "delta_t for all ifos do not match"
[...]

PR #3089, opened 2020-01-20
Fix delta_t in singles followup data

GarethDaviesGW:
This applies the fix from https://github.com/gwastro/pycbc/blob/29822cc643def18020ba327d5a0aafd17b20665b/bin/pycbc_live#L222
To the singles followup data in
https://github.com/gwastro/pycbc/blob/29822cc643def18020ba327d5a0aafd17b20665b/bin/pycbc_live#L339

Testing will be started momentarily

#3088


GarethDaviesGW:
Tested now and works

PR #3090, opened 2020-01-20
allow to pass initialization args to dynesty

ahnitz:
This adds the ability to pass additional arguments from the config file to dynesty to help configure how it will run.


ahnitz:
@sum33it Any thoughts?

sum33it:
@ahnitz Looks good to me.

PR #3091, opened 2020-01-20
Inference data from config

cdcapano:
This moves the data options currently loaded on the command line in `pycbc_inference` to the configuration file instead. The data options are now loaded from the `[data]` section. This is only done by the `GaussianNoise` model and all models that inherit from it (like `MarginalizedPhase`).

For the most part, the options from the config file are a one-to-one mapping from what was previously set on the command line. One difference, however, is I've added a `trigger-time` option, and replaced the `gps-start/end-time` options with `analysis-start/end-time`. The analysis start/end times and the psd star/end times are now measured with respect to the `trigger-time` option. This way you can set the trigger time in one place, then have everything measured with respect to that. (This includes the tc prior; using our config file substitution, you can refer to the trigger time in the `[data]` section when setting the tc prior.) If you do not specify a trigger time, then the analysis and psd times will be with respect to 0, meaning you need to put in the full GPS time, as before.

This also adds the ability to automatically check for valid times, and to shift the psd times until a valid time can be found. If not valid time can be found for a particular detector, that detector is removed from the analysis. If no detectors are left, an error is raised. (I would've moved this ability to a separate PR, but it got mixed in with the other changes, and it would be hard to disentangle.)

This PR will break the workflows, and some documentation. I'll fix these in separate PRs.

Putting on hold for a bit until I can compile the other PRs.


ahnitz:
@cdcapano I think a bunch of the CC issues look worrying. Looking through the code, the overall change seems reasonable though. Once CC passes, I'll take a closer look through.

cdcapano:
@ahnitz Yeah, I have to address the code climate issues. Some of them look important. I've mostly been working on the docs, which has made this PR get quite long. I might try separating out the commits involving documentation, although it'll be difficult since the docs are dependent on the examples, which in turn are used by Travis.

Also, this now depends on #3101, so has to wait until that's approved anyway.

cdcapano:
@ahnitz Ok, I think this is ready for review. I've added a Travis test for the BBH injection example, which tests both `create_injection`, and `pycbc_inference` with simulated noise + injection. [Here is an example](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/prs/inference_data_config1/sample_docs/latest/html/inference.html) of what the new doc page will look like. I've added a table listing all of the data options, as this was a worry several people brought up about removing the options from parser help. I've also added a table listing all of the available model PSDs, to make the data options table shorter.

Since the PR is so long, below is a summary of why I made changes to each file (in the order they appear on the Files changed tab). The key changes are in `models/data_utils.py` and `models/gaussian_noise.py`.
 * `pycbc_inference`: removes all options and lines related to loading and conditioning data
 * `doc/_include/_dict_to_rst.py`: Added a function for formatting references to functions. Needed for the list of psd models that have been added to the docs page.
 * `docs/_include/inference_data_opts-table.py`: This creates the table of data options for the docs page. It does it by parsing the usual parser help message, converting it into a rST table.
 * `docs/_include/psd_models-table.py`: This creates the table of available model PSDs.
 * `docs/_static/theme_overrides.css`: Modifies how tables are presented on the docs page. I had to do this because the RTD theme flattens all table rows into a single line, regardless of what formatting you use. This made the data-options table hard to read, since you had to do a large amount of side scrolling. The modification causes the white space in the table to be respected (although the description column in the data options table is narrower than what the generated rST made it... I'm not sure why that is, but at least it's readable).
 * `docs/conf_std.py`: Adds the theme overrides to the  list of CSS files that are loaded when building the docs.
 * `docs/inference.rst`: Updates the inference docs page.
 * `examples/inference/bbh-injection/data.ini`: Adds a data config file to the BBH injection example.
 * `examples/inference/bbh-injection/inference.ini`: Removed. The configuration file in the examples is now split up between the prior, sampler, and data, to make it easier to document and to highlight the fact that you can provide multiple config files.
 * `examples/inference/bbh-injection/make_injection.sh`: Added a `--force` argument when running `create_injections`. Did this for the Travis test; I wasn't sure if it would leave behind an old injection file. If so, that would cause the test to fail without the `--force`.
 * `examples/inference/bbh-injection/run.sh`: Updated the BBH example run script to use the new config files and to remove the data options from the command line.
 * `examples/inference/bbh-injection/run_test.sh`: Added this script for Travis tests. Similar to the `run_test` used in the GW150914 example.
 * `examples/inference/gw150914/data.ini`: Added data config file for GW150914.
 * `examples/inference/gw150914/plot.sh`: Changed `tc` to `delta_tc` in the GW150914 example plot script, since `delta_tc` is now the varied parameter in the example run.
 * `examples/inference/gw150914/run.sh`: Updated the GW150914 example run script.
 * `examples/inference/gw150914/run_test.sh`: Updated the Travis GW150914 test. Note that I'm now getting the prior and sampler config file from a different directory.
 * `examples/inference/gw150914/inference.ini → examples/inference/priors/gw150914_like.ini`: Moved the GW150914 config file to a `priors` directory, and removed info about the sampler. The idea is that we can accumulate standard priors in this directory.
 * `examples/inference/samplers/emcee_pt-gw150914_like.ini`: Moves the emcee_pt settings to its own config file into a `samplers` directory. The idea is that we can accumulate standard sampler settings in this directory.
 * `pycbc/inference/io/base_hdf.py`: Lets the user now the injection file is being written.
 * `pycbc/inference/models/base_data.py`: Makes the injection file an attribute of `BaseDataModel`. Did this so the injection file would get written to the output when the model writes its metadata. (Previously, this was done in `pycbc_inference`).
 * `pycbc/inference/models/data_utils.py`: This contains the bulk of the key changes in this PR. All of the code related to loading the data has been moved to functions in this (new) module. The functions to check for valid times and shift the PSD times around are also here.
 * `pycbc/inference/models/gaussian_noise.py`: Loading the data has been moved to `GaussianNoise.from_config`, which calls the functions in `data_utils`. Loading the recalibration class and the gates are also moved into `from_config`. Substituting injection values for static params set to `FROM_INJECTION` (this was formerly in `pycbc_inference`) has also been moved here. To turn on checking for valid data times and automatically shifting PSD times, you need to set `check-for-valid-times =` and `shift-psd-times-to-valid =`, respectively, in the model section. You can also provide `err-on-missing-detectors =` in the section if you want an error to be raised if any of the detectors do not have valid time. Otherwise, the detector will automatically removed from the analysis. To accomodate this, I also had to allow low frequency cutoffs to be provided for detectors that may not actually get analyzed. Note that if you do not provide these arguments, valid times are not checked for (the current functionality). NaNs will be checked for though (see `data_utils.from_cli`), which will at least prevent the user from accidentally running on GWOSC data with missing segments.
 * `pycbc/inference/option_utils.py`: Removes functions related to parsing command-line data options.
 * `pycbc/inference/sampler/base.py`: Removes the injection file from the function used to create a new hdf file. Needed because `setup_output` is done by `pycbc_inference`, which no longer knows about injection files (since this is handled by the model). The injection is now written by the model when it writes its metadata.
 * `pycbc/strain/gate.py`: Just removes a spurious period in the help message.
 * `tools/pycbc_test_suite.sh`: Adds the test of the BBH injection example to the Travis tests.

I haven't mentioned anything about the automatic checking for valid times and shifting the PSD times in the documentation. I'll add something about them in another PR.  Since these are by default turned off, I don't think that documentation needs to be added to this.

PR #3097, opened 2020-01-21
Added pycbc_make_skymap

Fergus-Hayes:
Added a new executable `bin/pycbc_make_skymap` that quickly produces sky localization given a the time, masses and spins of a trigger by utilizing a single template matched filter on strain data from a number of detectors. The sky localization is produced from the resulting SNR time series using BAYESTAR and therefore the user requires `ligo.skymap`installed in their environment. The script outputs a .png file and a .fits file of the skymap as well as an .xml file containing the SNR timeseries.

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -1,0 +1,343 @@
+#!/usr/bin/env python
+
+"""Fast and simple sky localization for a specific compact binary merger event.
+Runs a single-template matched filter on strain data from a number of detectors
+and calls BAYESTAR to produce a sky localization from the resulting set of SNR
+time series."""
+
+import h5py, sys, pycbc, os, subprocess, argparse, json, tempfile, shutil
+import time as TIME
+import numpy as np
+import lal
+import logging
+from pycbc import filter
+from pycbc.io import live
+from pycbc.types import TimeSeries, FrequencySeries
+from glue.ligolw import utils as ligolw_utils
+from ligo.gracedb.rest import GraceDb
+
+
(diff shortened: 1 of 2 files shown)


titodalcanton:
Also look into `setup.py`, I think the script must be listed there, otherwise it will not be installed.

titodalcanton:
Ah nevermind, it looks like `setup.py` just installs every appropriate file under `/bin/` now

Fergus-Hayes:
@titodalcanton Although the tests seem to fail as `ligo.gracedb` is not installed. However I see `ligo-gracedb` is listed in `companion.txt`. Are these packages from `companion.txt` not installed from `setup.py`? If not should `setup.py` be altered to install the `ligo.gracedb` package?

Also will the tests pass given that `ligo.skymap` is required although not imported?

titodalcanton:
Please add the new script to the list of test exceptions [here](https://github.com/gwastro/pycbc/blob/master/tools/pycbc_test_suite.sh#L33). We have other scripts that require GraceDB, so the strategy for now is just to skip those tests.

titodalcanton:
I just had the same exact issue on another PR :) Another solution is to move the GraceDB import to where it is actually needed, so the current test will not actually try it.

micamu:
@Fergus-Hayes Could you fix the help message in line 305 before merging? It should request just one single time (which is what the code takes now). If you want the user to give a dictionary of trigger times you need to make a few changes to the main function

ahnitz:
@titodalcanton Any reason this hasn't been merged?

PR #3098, opened 2020-01-21
PyCBC Live: add more info to GraceDB log

titodalcanton:
Some improvements to comments on the logs of uploaded G events. Add explanation of which G events result from coincidences and which ones from single detectors. Link original and SNR-optimized G events with actual anchors and both ways.

Also includes some code cleanup.

Needs testing.


titodalcanton:
Example from replay analysis: https://gracedb-playground.ligo.org/events/G248201/view/

titodalcanton:
@spxiwh this is ready for review.

PR #3100, opened 2020-01-22
Implement relative likelihood model

dfinstad:
This implements a relative likelihood model according to the method introduced in https://arxiv.org/abs/1806.08792. The model uses a fiducial waveform whose parameters are known to be near the peak of the likelihood, and template waveform likelihoods are calculated relative to this fiducial waveform.

Currently the fiducial and all template waveforms are generated using the `SPAtmplt` approximant, and this patch makes necessary modifications to the `SPAtmplt` code to allow evaluation at specific frequencies which are provided as an input argument. Support for other approximants can be added in a future patch. 

Fiducial waveform parameters should be supplied in the `[model]` section of the inference config file, e.g.:
```
[model]
name = relative
low_frequency_cutoff = 20.
high_frequency_cutoff = 1000.
; fiducial waveform params
mass1 = 1.5059660588995867
mass2 = 1.2586878921474083
spin1z = 0.
spin2z = 0.
ra = 3.44615914
dec = -0.40808407
tc = 1187008882.43
```

Below is an example of results produced by the relative model analyzing GW170817:
gw170817

### DIFF
--- a/pycbc/inference/models/__init__.py
+++ b/pycbc/inference/models/__init__.py
@@ -26,6 +26,7 @@ from .gaussian_noise import GaussianNoise
 from .marginalized_gaussian_noise import (MarginalizedPhaseGaussianNoise,
                                           MarginalizedGaussianNoise)
 from .single_template import SingleTemplate
+from .relbin import Relative
 
 
 # Used to manage a model instance across multiple cores or MPI
@@ -183,5 +184,6 @@ models = {_cls.name: _cls for _cls in (
     GaussianNoise,
     MarginalizedPhaseGaussianNoise,
     MarginalizedGaussianNoise,
-    SingleTemplate
+    SingleTemplate,
+    Relative
 )}
--- a/pycbc/inference/models/__init__.py
+++ b/pycbc/inference/models/__init__.py
@@ -26,7 +26,7 @@ from .gaussian_noise import GaussianNoise
 from .marginalized_gaussian_noise import (MarginalizedPhaseGaussianNoise,
                                           MarginalizedGaussianNoise)
 from .single_template import SingleTemplate
-from .relbin import Relative
+from .relbin import RelativeSPA
 
 
 # Used to manage a model instance across multiple cores or MPI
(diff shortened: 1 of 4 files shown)


pycbc/inference/models/__init__.py
```diff
@@ -183,5 +184,6 @@ def read_from_config(cp, **kwargs):
     GaussianNoise,
     MarginalizedPhaseGaussianNoise,
     MarginalizedGaussianNoise,
-    SingleTemplate
+    SingleTemplate,
+    Relative
```
ahnitz: Might want to name this something like RelativeSPA or something like that just so its clear it's specific to one waveform model. Down the line, we'd like to have a more generic version, but this will likely remain.

pycbc/inference/models/relbin.py
```diff
@@ -0,0 +1,355 @@
...
+        # store fiducial waveform params
+        mass1 = float(mass1)
+        mass2 = float(mass2)
+        spin1z = float(spin1z)
+        spin2z = float(spin2z)
+        ra = float(ra)
+        dec = float(dec)
+        tc = float(tc)
+
+        # get detector-specific arrival times relative to end of data
```
ahnitz: One change you've made from my earlier implementation is that this goes back to including time in as a parameter which needs to be also modeled in the frequency shift. I see now why you wanted to provide a sky location as you need to determine the time of the peak likelihood in each detector. If this is far off, the approximation could get problematic. Is this right? How much is it a problem then if you don't have the right sky location or don't have the reference time to more than 50-100 ms ?

pycbc/inference/models/__init__.py
dfinstad: Okay good idea, I've renamed it as you've suggested.

pycbc/inference/models/relbin.py
dfinstad: In principle yes, setting fiducial parameters far from the peak will cause problems for the approximation, but in practice the method seems fairly robust to inaccuracies in sky location and coalescence time. For example, changing the above config settings for GW170817 to
```
ra = 2.0
dec = 0.0
tc = 1187008882.5
```
produces very similar posteriors. I don't have a more general answer yet for how far is "too far" but I'll look at mapping that out in the near future.

PR #3102, opened 2020-01-23
Remove redundant functions in workflow config

cdcapano:
As discussed in #3101, the `WorkflowConfigFile` class currently has two methods (`from_args` and `from_cli`) that do the same thing: parse command-line options. There is also a method (`add_config_opts_to_parser`) that is partial copy of `add_workflow_command_line_group`. This patch removes `from_args` in favor of `from_cli`. Although `from_args` is the older function, it isn't consistent with PEP8, and it does checks that are already done in the `__init__` function. This patch also gets rid of the `add_config_opts_to_parser`, which isn't being used by anything.

@spxiwh I know you had suggested renaming `from_cli` to `from_args`. However, grepping the repository, I found that `from_args` is only used in one place (in `workflow.core`), which I've updated. Calling it `from_cli` is more consistent with the rest of pycbc.

### DIFF
--- a/pycbc/workflow/configuration.py
+++ b/pycbc/workflow/configuration.py
@@ -443,63 +443,23 @@ class WorkflowConfigParser(glue.pipeline.DeepCopyableConfigParser):
             self.write(fp)
             fp.close()
 
-
     @classmethod
-    def from_args(cls, args):
+    def from_cli(cls, opts):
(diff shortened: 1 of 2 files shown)


cdcapano:
@spxiwh I added back the longer doc string and merged.

PR #3103, opened 2020-01-23
Write analysis and psd segments to inference file

cdcapano:
Following up #3091: this writes the analyzed detectors, analysis segment, and PSD segment to the inference output file's `attrs`. Currently, this information isn't being saved anywhere, making it difficult to know if the PSD times were shifted or not, and if so, to where. Also, without the the analyzed detectors in the attrs, it is impossible to know which detectors were used if a posterior file is created with the `data` group.

In the process, I had to break up the `data_from_config` function into a few different functions. That should make code climate more happy anyway.

I've also added more documentation to the `GaussianNoise.from_config` function.

### DIFF
--- a/pycbc/inference/models/data_utils.py
+++ b/pycbc/inference/models/data_utils.py
@@ -394,11 +394,11 @@ def data_from_cli(opts, check_for_valid_times=False,
     Returns
     -------
     strain_dict : dict
-        Dictionary of instruments -> `TimeSeries` strain.
-    stilde_dict : dict
-        Dictionary of instruments -> `FrequencySeries` strain.
-    psd_dict : dict
-        Dictionary of instruments -> `FrequencySeries` psds.
+        Dictionary of detectors -> time series strain.
+    psd_strain_dict : dict or None
+        If ``opts.psd_(start|end)_time`` were set, a dctionary of
+        detectors -> time series data to use for PSD estimation. Otherwise,
+        ``None``.
     """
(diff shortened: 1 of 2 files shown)


cdcapano:
Modified this a bit... now the time used for PSD estimation is added as an attribute to the PSDs before they are passed to `__init__`. The PSD segment is retrieved by the class from that attribute. (Before, the PSD segments were being passed as another init argument.)

Example of what the output looks like in a file (from the GW150914 example):
```
>>> from pycbc.inference import io
>>> fp = io.loadfile('inference.hdf', 'r')
>>> fp.attrs.items()
[(u'H1_analysis_segment', array([1.12625945e+09, 1.12625947e+09])),
 (u'H1_likelihood_low_freq', 20.0),
 (u'H1_psd_segment', array([1.12625921e+09, 1.12625972e+09])),
 (u'L1_analysis_segment', array([1.12625945e+09, 1.12625947e+09])),
 (u'L1_likelihood_low_freq', 20.0),
 (u'L1_psd_segment', array([1.12625921e+09, 1.12625972e+09])),
 (u'analyzed_detectors', array(['H1', 'L1'], dtype='|S2')),
 ...

>>> fp.attrs['H1_analysis_segment'] - int(fp.attrs['trigger_time'])
array([-10.,   6.])
>>> fp.attrs['H1_psd_segment'] - int(fp.attrs['trigger_time'])
array([-256.,  256.])

PR #3105, opened 2020-01-24
pycbc_page_injtable_multi_ifo to create a table of missed/found 

KoustavChandra:
Create a multi-ifo table of missed/found injection.


KoustavChandra:
You can find the corresponding table for missed and found injections for [this run](https://ldas-jobs.ligo.caltech.edu/~connor.mcisaac/o3/runs/hlv/c01/a16_initial/5._injections/5.05_NSBHSEOBNRV4_INJ/#A11) for the multi-ifo case [here](https://ligo.gravity.cf.ac.uk/~koustav.chandra/LSC/images/pycbc_page_injtable/)

GarethDaviesGW:
Do you have an updated version of the produced tables with the new changes?

KoustavChandra:
Hi,
I have fixed the changes. You can view the html for 2 detectors and 3 detectors here: [Link](https://ligo.gravity.cf.ac.uk/~koustav.chandra/LSC/images/pycbc_page_injtable/)

bin/hdfcoinc/pycbc_page_injtable
```diff
@@ -1,10 +1,12 @@
-#!/usr/bin/env python
-""" Make a table of found injection information
+#!/usr/bin/env python	
+""" Make a table of found injection information
```
GarethDaviesGW: remove whitespace

bin/hdfcoinc/pycbc_page_injtable
```diff
@@ -26,15 +28,41 @@ else:
...
+        tdiff = (found['time1'][:] - found['time2'][:]) * 1000
+        tdiff_str = '%s - %s time (ms)' % (f.attrs['detector_1'], f.attrs['detector_2'])
+        ids = {f.attrs['detector_1']: found['trigger_id1'][:],
+               f.attrs['detector_2']: found['trigger_id2'][:],}
+        found_cols = [tdiff, found['stat'], found['ifar'], found['ifar_exc']]
+        found_names = [tdiff_str, 'Ranking Stat.', 'Inc. IFAR (yrs)', 'Exc. IFAR']
+        found_formats =  ['##.##', '##.##', '##', '##']
+
+    else:
+        detectors = ['L1','H1','V1','K1','I1']
```
GarethDaviesGW: You should be able to use something like `f.attrs['ifos'].split(' ')` to get a list of the detectors used in this file

bin/hdfcoinc/pycbc_page_injtable
```diff
@@ -26,15 +28,41 @@ else:
...
+        detectors_used = []
+        found = f['found_after_vetoes']
+        for det in detectors:
+            if(det in keys):
+                detectors_used.append(det)
+        detectors_two_combinations= numpy.array(list(combinations(detectors_used,2)))
+        tdiff = []
+        tdiff_str = []
+        tdiff_format =[]
+        for i in range(len(detectors_two_combinations)):
```
GarethDaviesGW: as i is just a number used to reference detectors_two_combinations in this loop, you should be able to loop over detectors_two_combinations directly (I think)

bin/hdfcoinc/pycbc_page_injtable
```diff
@@ -45,34 +73,42 @@ else:
...
             
             found_names += ["%s SNR" % ifo, "CHISQ", "NewSNR"]
-            found_cols += [snr, chisq, newsnr]
+            found_cols += [numpy.round(snr,2), numpy.round(chisq,2), numpy.round(newsnr,2)]
             found_formats += ['##.##', '##.##', '##.##']
-          
-dec_dist = numpy.maximum(inj['eff_dist_h'][:][idx], inj['eff_dist_l'][:][idx])
+
+eff_dist = {'eff_dist_l':'Eff Dist(L)','eff_dist_h':'Eff Dist(H)','eff_dist_v':'Eff Dist(V)',
+            'eff_dist_k':'Eff Dist(K)','eff_dist_i':'Eff Dist(I)'}
```
GarethDaviesGW: again, try not to hard-code the IFOs. The involved IFOs will be in the attributes of the file

bin/hdfcoinc/pycbc_page_injtable
```diff
@@ -45,34 +73,42 @@ else:
...
+keys = inj.keys()
```
GarethDaviesGW: If you look at https://ligo.gravity.cf.ac.uk/~koustav.chandra/LSC/images/pycbc_page_injtable/two_detector_found.html, you can see that we are given a Virgo effective distance even though it isn't involved in the search. This will be solved by using the file attribute for ifo list

bin/hdfcoinc/pycbc_page_injtable
```diff
@@ -45,34 +73,42 @@ else:
             newsnr = pycbc.events.ranking.newsnr(snr, chisq)
             
             found_names += ["%s SNR" % ifo, "CHISQ", "NewSNR"]
-            found_cols += [snr, chisq, newsnr]
+            found_cols += [numpy.round(snr,2), numpy.round(chisq,2), numpy.round(newsnr,2)]
```
GarethDaviesGW: This list should also include the `if time < 0: output = 'n/a'` flag for snr, chisq and newsnr, as the table at the moment contains a lot of snr values where there wasn't a trigger which contributed to the found injection coincidence

bin/hdfcoinc/pycbc_page_injtable
```diff
@@ -81,4 +117,3 @@ kwds = { 'title' : title,
         'caption' : "A table of %s and their coincident statistic information." % title.lower(),
         'cmd' :' '.join(sys.argv), }
 pycbc.results.save_fig_with_metadata(str(html_table), args.output_file, **kwds)
```
GarethDaviesGW: The newline is generally wanted at the end of a file

bin/hdfcoinc/pycbc_page_injtable
```diff
@@ -45,34 +73,42 @@ else:
...
-           inj['spin1x'][:][idx], inj['spin1y'][:][idx], inj['spin1z'][:][idx],
-           inj['spin2x'][:][idx], inj['spin2y'][:][idx], inj['spin2z'][:][idx],
-           inj['distance'][:][idx],
-           inj['eff_dist_h'][:][idx], inj['eff_dist_l'][:][idx], 
-           ] + found_cols
+columns = [numpy.round(dec_chirp_dist,2), numpy.round(inj['end_time'][:][idx],2), 
+           numpy.round(m1,2), numpy.round(m2,2), numpy.round(mchirp,2), numpy.round(eta,2),
+           numpy.round(inj['spin1x'][:][idx],2), numpy.round(inj['spin1y'][:][idx],2), numpy.round(inj['spin1z'][:][idx],2),
+           numpy.round(inj['spin2x'][:][idx],2), numpy.round(inj['spin2y'][:][idx],2), numpy.round(inj['spin2z'][:][idx],2),
+           numpy.round(inj['distance'][:][idx],2)] + eff_distance + found_cols
```
GarethDaviesGW: is the rounding not done in the html_table already?

bin/hdfcoinc/pycbc_page_injtable
KoustavChandra: The two detector does content effective distance column corresponding to the Virgo detector and hence it displayed it.

bin/hdfcoinc/pycbc_page_injtable
KoustavChandra: The travel time can be negative. Right? Or do you mean. I should set snr, chisq and newsnr to be 'N/A' if travel time is 'N/A'

bin/hdfcoinc/pycbc_page_injtable
KoustavChandra: When you write 'N/A' as one of the entries then by default your matrix no longer remain a list of float but converts into a list of strings. The '##.##' can't deal with it. Checked.

bin/hdfcoinc/pycbc_page_injtable
KoustavChandra: This is just used to search for whether such an entry exists or not. I found even for two detector runs, there was eff distance entry for V1.

bin/hdfcoinc/pycbc_page_injtable
GarethDaviesGW: So this will be if the "H1/time" value is -1, not the travel time. This will be the same as checking if travel time is n/a

bin/hdfcoinc/pycbc_page_injtable
GarethDaviesGW: So this is why we should use the ifos attribute of the injfind file - if V1 wasn't involved in the search, then V1 won't be in that attribute, and so you won't be looking for it in the injection file, even if it is there

bin/hdfcoinc/pycbc_page_injtable
GarethDaviesGW: Ah thank you, makes sense

tdent:
@KoustavChandra The Travis build has failed on this. Do we know why? Can it be fixed in order to proceed with the MR?

GarethDaviesGW:
> @KoustavChandra The Travis build has failed on this. Do we know why? Can it be fixed in order to proceed with the MR?

The build has failed on other PRs as well - it looks like its a problem with building lalsuite, but I can't be certain

bin/hdfcoinc/pycbc_page_injtable
GarethDaviesGW: This is now reported as 'nan' rather than 'n/a', and gives nan for snr/schisq/newsnr where appropriate

GarethDaviesGW:
Results with latest version of the code are at https://ldas-jobs.ligo.caltech.edu/~gdavies/testoutput/injtable_3det/

tdent:
Is it a conscious choice to print 'nan' in fields where there is no meaningful value?  It might suggest that the code tried to calculate something but there was an underflow/overflow/division by zero etc.

tdent:
(if we have a choice I would prefer to leave the non-meaningful entries blank since it appears less messy.)

GarethDaviesGW:
I will put that into a future PR - I've just realised we didnt put it into the workflow so I will sort that as well

PR #3106, opened 2020-01-24
Allow multinest to work with mpi

dfinstad:
This patch is reviving #2836 which I'm not able to re-open for some reason. Additional changes made here address the comments by @cdcapano on that PR:

- `setup_output` has been made an abstract method of `BaseSampler` and its current functionality moved to a standalone convenience function in `base.py`
- `setup_output` methods have been added to `BaseMCMC`, `CPNestSampler`, and `DynestySampler` which call the new standalone function
- the `setup_output` method of `MultinestSampler` now manages the master process check and references the standalone function

### DIFF
--- a/pycbc/inference/sampler/base.py
+++ b/pycbc/inference/sampler/base.py
@@ -145,58 +145,10 @@ class BaseSampler(object):
         """Do any finalization to the samples file before exiting."""
         pass
 
+    @abstractmethod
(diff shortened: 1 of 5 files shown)


dfinstad:
Thanks @cdcapano, I've made the change you suggested. I think the new codeclimate issue is not relevant, so this should be ready to merge.

PR #3107, opened 2020-01-26
Add BaseGaussianNoise model to fix SingleTemplate and RelativeSPA

cdcapano:
Fixes the `SingleTemplate` and `RelativeSPA` models for the recent changes to how data is loaded. A `BaseGaussianNoise` class is introduced.

The `from_config` (which includes loading data) and everything related to setting up the inner product weight and calculation of the noise log likelihood is moved to `BaseGaussianNoise`. The `_loglikelihood` function is also implemented, as the difference between `_loglr` and `lognl`. A `_loglr` function is not implemented. The idea is that anything that assumes stationary Gaussian noise for the noise model of the (GW) data should inherit from this class, and implement there own `_loglr` function (since that is dependent on the signal model). `GaussianNoise`, `MarginalizedPhaseGaussianNoise`, `SingleTemplate`, and `RelativeSPA` now inherit from `BaseGaussianNoise`.

In the process, I've introduced some fixes to `SingleTemplate`. By virtue of inheriting from `BaseGaussianNoise`, it now has the correct `loglikelihood`. It can also accept different frequency cutoffs for different detectors.

For `RelativeSPA`, I changed the names of the fiducial parameters, adding `_ref` at the end of their names so as not to confuse them with the variable parameters. These are also now written to the `attrs` of the output file. I didn't change anything else though; the class will just raise a `ValueError` if you try to provide different frequency cutoffs. @dfinstad you can change that if you like in a future update.

I only did a quick test to make sure that these ran without error. I didn't check that they give sensible posteriors. @dfinstad @ahnitz It would be good if you guys added some example scripts for each of these (maybe using GW170817?), and some corresponding documentation on the doc page. Since both of these are runs you can do on a laptop, it'd be a real nice example to have for users.

Putting on hold for now because this depends on #2887.

### DIFF
--- a/pycbc/inference/models/relbin.py
+++ b/pycbc/inference/models/relbin.py
@@ -322,29 +322,6 @@ class RelativeSPA(BaseGaussianNoise):
             llr += (hd - 0.5 * hh)
         return float(llr)
 
-    def _loglikelihood(self):
-        r"""Computes the log likelihood of the parameters,
-
-        .. math::
-
-            \log p(d|\Theta, h) = \log \alpha -\frac{1}{2}\sum_i
-                \left<d_i - h_i(\Theta) | d_i - h_i(\Theta)\right>,
-
-        at the current parameter values :math:`\Theta`.
-
-        Returns
-        -------
-        float
-            The value of the log likelihood evaluated at the given point.
-        """
-        return self.loglr + self.lognl
-
-    def _lognl(self):
-        """Calculate the log of the noise likelihood. Currently not
-        implemented, so just returns zero.
-        """
-        return 0.
-
     def write_metadata(self, fp):
         """Adds writing the fiducial parameters and epsilon to file's attrs.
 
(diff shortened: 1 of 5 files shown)


cdcapano:
@ahnitz @dfinstad This is ready for review now.

pycbc/inference/models/relbin.py
```diff
@@ -338,22 +346,19 @@ def _lognl(self):
         return 0.
```
dfinstad: I think both these methods (`_loglikelihood` and `_lognl`) can be removed now. I had avoided implementing the full `_lognl` calculation when I wrote this class because it would've required replicating a lot of the data handling code in `GaussianNoise`, but now that you've moved it all to the new class which is inherited by `RelativeSPA` we should just be able to inherit the `_lognl` method from there. Similarly the `_loglikelihood` method here is just replicating what you now have in `BaseGaussianNoise`

cdcapano: Thanks for catching @dfinstad, I missed those two functions. I've removed them and updated.

PR #3108, opened 2020-01-27
Remove MarginalizedGaussianNoise model for now

cdcapano:
The class `MarginalizedGaussianNoise` has been broken for awhile. That class was supposed to implement the ability to optionally marginalize over distance, time, and/or phase. However, putting that all in one class was a mistake on my part (I told Charlie to do this), as it made the class way too complicated and difficult to figure out bugs. I think it would be better to implement separate classes for each possibility; e.g., having a `MarginalizedDistancePhase`, for distance and phase; `MarginalizedTimeDistancePhase`; etc. That will be a bit of work. For now, this patch just removes the broken class. When someone wants to tackle this, they can revert this patch then edit from there. 

Putting on hold as this depends on #3107.


cdcapano:
Rebased now that #3107 is merged.

PR #3116, opened 2020-01-29
Remove automatic transforms when loading results

cdcapano:
Currently, some of the plotting codes in the inference module try to apply transforms automatically to add some common cbc parameters if they do not exist in an input file, such as `mchirp` and `chi_eff`. The original intent was to save on having to write out functions on the command line when creating plots. However, this has led to a number of issues if you use anything but a standard CBC analysis. It's also impossible to know what transforms have been applied and what haven't, leading to confusion as to what should be called as a function and what should be called as an array. 

For example, the `--file-help` message in `pycbc_inference_plot_posterior` tells you that you can use functions in the conversions module on any of the parameters in your file. One of the functions is `chi_eff`. If you had `mass1`, `mass2`, `spin1z`, `spin2z` in your file, you might think you could do `--parameters chi_eff(mass1, mass2, spin1z, spin2z)`. But if the automatic transform was applied (and it's difficult to tell if it has or not), you'd get an error, since in that case `chi_eff` is an array that has silently been created, not a function (meaning you should have done`--parameters chi_eff`).

To clear this up, this patch removes the attempt at automatic transforms from the functions that load the results and all of the plotting executables. If a user gets tired retyping functions on the command line, they can use `pycbc_inference_extract_samples` to add the derived parameter to a file instead.

### DIFF
--- a/bin/inference/pycbc_inference_plot_inj_recovery
+++ b/bin/inference/pycbc_inference_plot_inj_recovery
@@ -13,7 +13,6 @@ import pycbc
 import pycbc.version
 from matplotlib import cm
 from pycbc import inject
-from pycbc import transforms
 from pycbc import __version__
 from pycbc.inference import (option_utils, io)
 
@@ -51,12 +50,6 @@ ax = fig.add_subplot(111)
 samples = [samples] if not isinstance(samples, list) else samples
 fp = [fp] if not isinstance(fp, list) else fp
 injs = io.injections_from_cli(opts)
-_, ts = transforms.get_common_cbc_transforms(opts.parameters,
-                                                 injs.fieldnames)
-
-# add parameters not included in injection file
-inj_parameters = transforms.apply_transforms(injs, ts)
-
 
 # if user wants a colorbar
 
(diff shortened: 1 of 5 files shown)


soumide1102:
@cdcapano So this means for example in a run that has the variable spin parameters in spherical coordinates, one would have to do `chi_eff_from_spherical(mass1, mass2, spin1_a, spin1_polar, spin2_a, spin2_polar):` ? Earlier, doing `chi_eff` I think worked, provided `spin1_azimuthal` and `spin2_azimuthal` were also present in the file.

cdcapano:
@soumide1102 Yes, that's right, you'd have to use `chi_eff_from_spherical(mass1, mass2, spin1_a, spin1_polar, spin2_a, spin2_polar)`. But, that's not terrible, and if you really don't want to retype it a bunch of times, you can just create a posterior file using `extract_samples` in which you add `chi_eff`. In other words, the inconvenience of having to type out that function is not worth the confusion and bugs that arise from the code silently trying to do it on its own.

PR #3125, opened 2020-01-31
Allow user to pass extra arguments to start/stop/status/debug scripts

duncanmmacleod:
This PR patches `pycbc_submit_dax` to add `$@` to the helper scripts that it writes, allowing users to add arbitrary extra (optional) arguments when calling, e.g. `./status`.


ahnitz:
@duncanmmacleod Seems like a very good idea. Thanks!

PR #3126, opened 2020-01-31
Allow FromFile distribution to use posterior files

cdcapano:
This adds a `datagroup` option to the `FromFile` distribution to specify which group in hdf file to get samples from. (Currently, it tries to get samples from the top level.) By setting this argument to `samples`, you can use posterior files produced by `pycbc_inference_extract_samples` for the `FromFile` distribution. This allows the posterior from one event to be used as a prior for another.

This also fixes a python 3 bug in the `Arbitrary` distribution that was discovered when testing this.


cdcapano:
@micamu poke

PR #3133, opened 2020-02-04
split_binned histogram plots dealing with low numbers of triggers

GarethDaviesGW:
As seen by @duncanmmacleod - if the number of triggers is low, then pycbc_fit_sngls_split_binned will fail

This is as there may be some templates with no triggers, in which case, we need to fall back to the original (slower) template duration calculation rather than reading it from the trigger fits file

The code now works in this case, but is slow, so the work in progress is to work out the trigger-to-template number ratio where we would expect this to happen, so that we don't do both procedures if we're likely to need this exception

### DIFF
--- a/bin/hdfcoinc/pycbc_fit_sngls_split_binned
+++ b/bin/hdfcoinc/pycbc_fit_sngls_split_binned
@@ -128,21 +128,29 @@ usedparams = [args.split_param_two, args.split_param_one, args.bin_param]
 extparams = [par for par in ['mchirp', 'eta', 'chi_eff', 'template_duration']
              if par in usedparams]
 
+
+
 for ex_p in extparams:
     logging.info('calculating {}'.format(ex_p))
     manual_duration_calc = False
     if ex_p == 'template_duration':
-        logging.info('atempting to use duration from trigger file')
-        params[ex_p] = np.array([trigf[args.ifo + '/template_duration'][ref][0]
-                                 for ref in trigf[args.ifo + '/template_duration_template'][:]])
-        if any(params[ex_p] == 0):
-            # Some of the templates didn't get any triggers.
-            # This will produce zeros in template duration, so recalculate.
-            logging.warn('Some templates contained no triggers. This plot '
+        logging.info('attempting to use duration from trigger file')
+        durs_list = []
+        for ref in trigf[args.ifo + '/template_duration_template'][:]:
+            temp_ref_dur = trigf[args.ifo + '/template_duration'][ref][0]
+            if temp_ref_dur:
+                durs_list += [temp_ref_dur]
+            else:
+                logging.warning('Some templates contained no triggers. This plot '
                          'is only sensible for large statistics. Calculating '
                          'template duration manually.')
-            manual_duration_calc = True
+                manual_duration_calc = True
+                break
+        if not manual_duration_calc:
+            params[ex_p] = np.array(durs_list)
+
     if ex_p != 'template_duration' or manual_duration_calc:
+        logging.info("Calculating duration from template parameters")
         params[ex_p] = trigs.get_param(ex_p, args, params['mass1'],
                                        params['mass2'], params['spin1z'],
                                        params['spin2z'])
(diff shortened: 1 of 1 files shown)


GarethDaviesGW:
A calculation for whether this would be needed is non-trivial. So I would say that this is sufficient (it's a rare use case anyway)

tdent:
what does 'manual' mean here? 
obviously the code doesn't do anything 'manually' ..

GarethDaviesGW:
'manual' in this case means that we are not using the dereferencing that was designed to speed up the code, rather calculating the duration from template parameters - which requires f-lower and min-duration to be in the args

GarethDaviesGW:
poke on either/both reviewers here

GarethDaviesGW:
Last commit was used to kick travis - no changes

tdent:
I'm extremely busy with 190521 paper tasks and will be all the rest of today.  Duncan requested this change anyway.

One thing I haven't emphasized yet on this is that a code should _fail_ rather than _silently do something that makes a plot but it's a thing you didn't ask for_.

tdent:
Can we go back a bit - is the problem just that *some* templates have no triggers? If so, is there some way to still produce the histogram omitting those templates? I don't think that would be an incorrect outcome.

GarethDaviesGW:
That's correct, and should be a fairly easy fix I think - I'll take a look

tdent:
was this ready for review?

GarethDaviesGW:
yes this is, sorry I forgot to put a comment on to say so

GarethDaviesGW:
There are example plots for the situation where there are many empty templates in this directory:
https://ldas-jobs.ligo.caltech.edu/~gdavies/testoutput/fit_sngls_split_binned_low_numbers/

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -129,6 +129,7 @@ for ex_p in extparams:
         params[ex_p] = np.array([trigf[args.ifo + '/template_duration'][ref][0]
                                  for ref in trigf[args.ifo + '/template_duration_template'][:]])
     else:
+        logging.info("Calculating " + ex_p + " from template parameters")
```
tdent, line 132: Can we remove the logging statement on line 126 since now there is a statement for the parameter regardless of what it is?  (also, if it's template duration it isn't actually calculated.)

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -153,16 +154,25 @@ if args.max_bin_param:
     pbin_upper_lim = float(args.max_bin_param)
 else:
     pbin_upper_lim = params[args.bin_param].max()
-bb_input = (params[args.bin_param].min(), pbin_upper_lim, args.num_bins)
+
+
+if args.bin_param == 'template_duration' and params[args.bin_param].min() == 0:
```
tdent, line 159: should it be clear why having min duration of 0 indicates that some templates are empty?  (at least I don't immediately see that)

bin/hdfcoinc/pycbc_fit_sngls_split_binned
```diff
@@ -153,16 +154,25 @@ if args.max_bin_param:
...
+    logging.warn('WARNING: Some templates do not contain triggers')
+    pbin_lower_lim = params[args.bin_param][params[args.bin_param] > 0].min()
```
tdent, line 161: comment this line to tell us what it does

bin/hdfcoinc/pycbc_fit_sngls_split_binned
GarethDaviesGW: Comment added, hopefully it is clear

bin/hdfcoinc/pycbc_fit_sngls_split_binned
GarethDaviesGW: Ditto as above: Comment added, hopefully it is clear

bin/hdfcoinc/pycbc_fit_sngls_split_binned
tdent: not really, i don't see what is meant .. 'dereferencing' is something you do with pointers in C afaik, and where above? 
if you mean a previous line of code, probably better to explain it there (ie say what gives rise to the zero value)

PR #3134, opened 2020-02-04
Add utility scripts to stage out a failed workflow and copy the output map

connor-mcisaac:
### pycbc_stageout_failed_workflow

Given a workflow directory look at the main map file and stage-out any files that exist with non-zero size.
This is useful if one type of injection fails (as was the case recently with SpinTaylorT2) and many of the completed inspiral jobs are therefore never staged out.

### pycbc_copy_output_map

Copy the output map of a workflow to a new file, with options to copy subsets of files (eg. inspirals) and to check that the files exist.

### DIFF
--- a/bin/pycbc_copy_output_map
+++ b/bin/pycbc_copy_output_map
@@ -74,6 +74,10 @@ do
 	echo "A line contained ${#ARRAY[@]} space seperated entries."
 	echo "A map file should contain 3 per line"
 	exit 1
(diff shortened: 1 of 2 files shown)


ahnitz:
@connor-mcisaac Oh, wow, those scripts sound really handy! I've had to manually do this in the past, which can be a pain.

connor-mcisaac:
Great, adding in the ability to deal with sub-workflows would definitely be good.

PR #3139, opened 2020-02-05
add very basic support for ultranest sampler

ahnitz:
This adds minimal support for ultranest (https://github.com/JohannesBuchner/UltraNest). 

Does not include support for checkpointing, configuring ultranest, or really much except running the sampler in default mode. 

It works similarly to dynesty, so adding support is very similar. 

Why ultranest? 

Well, see the docs https://johannesbuchner.github.io/UltraNest/. It seems to include a lot of the latest advancements in sampling (or at least more recent than many other packages) and can target final error uncertainties or desired number of independent samples (not actually supported in this patch). 

It's also by someone who has developed a version of this sampler and written papers on it for years and is the author of pymultinest as well. This seems to be his most actively developed sampler and given past history it would seem he is likely to support this for a while.

### DIFF
--- a/pycbc/inference/io/ultranest.py
+++ b/pycbc/inference/io/ultranest.py
@@ -1,0 +1,31 @@
+# Copyright (C) 2019 Collin Capano, Sumit Kumar, Alex Nitz
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 3 of the License, or (at your
+# self.option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+
+#
+# =============================================================================
+#
+#                                   Preamble
+#
+# =============================================================================
+#
+"""Provides IO for the ultranest sampler.
+"""
+from .base_nested_sampler import BaseNestedSamplerFile
+
+class UltranestFile(BaseNestedSamplerFile):
+    """Class to handle file IO for the ``dynesty`` sampler."""
+
+    name = 'ultranest_file'
--- a/pycbc/inference/io/ultranest.py
+++ b/pycbc/inference/io/ultranest.py
@@ -27,6 +27,6 @@ from .base_nested_sampler import BaseNestedSamplerFile
 
 
 class UltranestFile(BaseNestedSamplerFile):
-    """Class to handle file IO for the ``dynesty`` sampler."""
+    """Class to handle file IO for the ``ultranest`` sampler."""
 
     name = 'ultranest_file'
--- a/pycbc/inference/io/ultranest.py
+++ b/pycbc/inference/io/ultranest.py
@@ -25,6 +25,7 @@
 """
 from .base_nested_sampler import BaseNestedSamplerFile
 
+
 class UltranestFile(BaseNestedSamplerFile):
     """Class to handle file IO for the ``dynesty`` sampler."""
 
--- a/pycbc/inference/sampler/base_cube.py
+++ b/pycbc/inference/sampler/base_cube.py
@@ -78,7 +78,7 @@ class CubeModel(object):
         """
         returns log likelihood function
         """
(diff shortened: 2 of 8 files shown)


examples/inference/analytic-normal2d/normal2d.ini
```diff
@@ -2,9 +2,7 @@
 name = test_normal
 
 [sampler]
-name = emcee
-nwalkers = 5000
-niterations = 100
+name = ultranest
```
cdcapano: I think you accidentally committed this. Please don't change this example script; better to add a new one if you want to show one using ultranest.

pycbc/inference/io/ultranest.py
```diff
@@ -0,0 +1,31 @@
...
+#                                   Preamble
+#
+# =============================================================================
+#
+"""Provides IO for the ultranest sampler.
+"""
+from .base_nested_sampler import BaseNestedSamplerFile
+
+class UltranestFile(BaseNestedSamplerFile):
+    """Class to handle file IO for the ``dynesty`` sampler."""
```
cdcapano: For the ``ultranest`` sampler.

cdcapano:
@ahnitz Are you going to finish this patch?

ahnitz:
@cdcapano I think I've now addressed your concerns. Can you check over again?

ahnitz:
I think the remaining CC issues don't need to be resolved.

pycbc/inference/sampler/base_cube.py
```diff
@@ -0,0 +1,100 @@
...
+
+def call_global_logprior(cube):
+    return models._global_instance.prior_transform(cube)
+
+
+def setup_calls(model, nprocesses=1,
+                loglikelihood_function=None, copy_prior=False):
+    """ Configure calls for MPI support
+    """
+    model_call = CubeModel(model, loglikelihood_function, copy_prior=copy_prior)
```
cdcapano: Line is over 79 characters.

pycbc/inference/sampler/base_cube.py
```diff
@@ -0,0 +1,100 @@
...
+    else:
+        prior_call = model_call.prior_transform
+        log_likelihood_call = model_call.log_likelihood
+
+    return log_likelihood_call, prior_call
+
+
+class CubeModel(object):
+    """
+    Class for making PyCBC Inference 'model class'
```
cdcapano: Should start on the same line as the ```"""`, with blank line between this and `Parameters`.

pycbc/inference/sampler/ultranest.py
```diff
@@ -0,0 +1,184 @@
...
+                'max_iters': int,
+                'max_ncalls': int,
+                'max_num_improvement_loops': int,
+                'min_num_live_points': int,
+                'cluster_num_live_points:': int}
+        for opt_name in opts:
+            if cp.has_option('sampler', opt_name):
+                value = cp.get('sampler', opt_name)
+                skeys[opt_name] = opts[opt_name](value)
+        return cls(model, **skeys)
```
cdcapano: You forgot to call `setup_output`! I think you may have started this before Sumit's patch was merged. The `from_config` should take an output file keyword argument. After initializing the cls, you then run `setup_output` using the output file. See the Dynesty sampler for an example.

pycbc/inference/sampler/ultranest.py
```diff
@@ -0,0 +1,184 @@
...
+        Parameters
+        ----------
+        sampler : sampler instance
+            Sampler
+        output_file : str
+            Name of the output file.
+        force : bool, optional
+            If the output file already exists, overwrite it.
+        """
+        setup_output(self, output_file)
```
cdcapano: Since you're just calling `setup_output` from `sampler.base`, just remove this class method and call it directly in the `from_config`.

pycbc/inference/sampler/ultranest.py
```diff
@@ -0,0 +1,184 @@
...
+#
+# =============================================================================
+#
+#                                   Samplers
+#
+# =============================================================================
+#
+
+class UltranestSampler(BaseSampler):
+    """This class is used to construct an Dynesty sampler from the dynesty
```
cdcapano: [D|d]ynesty -> [U|u]tranest.

ahnitz:
@cdcapano Ok, this should now be addressed.

cdcapano:
@ahnitz Github seems broken... I had a few more comments, but couldn't add them in the review. So, here they are:
 * Line 87 of `sampler/ultranest.py`: drop the `**kwds` from the `from_config` function since it's not being used.
 * Line 30 of `io/ultranest.py`: still has `dynesty` in the doc string.
 * Line 81 of `sampler/base_cube.py`: code climate is right, you can simplify this to `params = dict(zip(self.model.sampling_params, cube))`
Please feel free to merge once you address those.

PR #3141, opened 2020-02-07
Add ability to turn off redshifting masses in TOV transforms

cdcapano:
This adds an option to the `LambdaFromTOVFile` transform to turn off redshifting masses. This is needed if you want to use this transform with a prior that is defined in the source frame (since the masses are already redshifted). The default is to redshift, as before. To turn it off, you add `do-not-redshift-mass =` to the transform's section in the config file.

This also will check if the requested mass is larger than the maximum mass supported by the equation of state. If so, it will set lambda to 0, assuming that it is a black hole.

### DIFF
--- a/pycbc/transforms.py
+++ b/pycbc/transforms.py
@@ -1015,14 +1015,15 @@ class CartesianSpinToChiP(BaseTransform):
 class LambdaFromTOVFile(BaseTransform):
     """Transforms mass values corresponding to Lambda values for a given EOS
     interpolating from the mass-Lambda data for that EOS read in from an
-    external ASCII file. The interpolation of the mass-Lambda data is a
-    one-dimensional piecewise linear interpolation. The mass values to be
-    transformed are assumed to be detector frame masses, so a distance should
-    be provided along with the mass for transformation to the source frame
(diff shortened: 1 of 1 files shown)


cdcapano:
@soumide1102 poke

cdcapano:
@soumide1102 pokity poke

soumide1102 (dismissed):
@cdcapano Sorry, I was traveling and missed this.

One request---It might be nice to add something to the docstring for `LambdaFromTOVFile` to describe this as well. Otherwise looks fine to me. 

Looks like Travis is failing while processing dependencies etc, not because of anything related to the changes in this PR. Code Climate's complaints can be ignored.

ahnitz:
@cdcapano @soumide1102 What's the status of this?

cdcapano:
@ahnitz I added the doc strings as @soumide1102 requested, and rebased to current master. I'm not sure Soumi is checking PyCBC much anymore, so I've added @ahnitz as a reviewer.

ahnitz (approved):

cdcapano:
Hmm... looks like I can't merge until @soumide1102 signs off.

PR #3140, opened 2020-02-05
Update inference workflow

cdcapano:
This updates `pycbc_make_inference_workflow` for the recent changes to `pycbc_inference` in which data is loaded from the inference config file. It also adds a number of improvements to the workflow including:
 * simplifies the options needed to be provided to `pycbc_make_inference_workflow`;
 * allows for multiple events to be analyzed at once by providing `[event]` sections in the workflow config file;
 * adds creation of posterior files from the samples files;
 * allows for multiple runs on the same event to be analyzed at once (with the results combined into a single posterior file);
 * adds jobs to create `.fits` files and sky maps (python 3 only);
 * overall cleanup of the results page;
 * fixes a number of issues with the workflow (such as how plots are named);
 * makes sampler-specific diagnositc plots (currently only `plot_samples` and `inference_rate`) optional, so the workflow can be used with both mcmc and nested samplers (to turn these off, you just remove them from the list of executables in the workflow config file);
 * adds another workflow generator, `pycbc_make_inference_plots_workflow` that just makes a posterior file and plots (i.e., everything the main workflow does except for running `pycbc_inference`).

The biggest change from a user perspective is how you specify what to analyze. Previously, you provided an `inference-config-file` on the command line to give to `pycbc_inference`; this was separate from the workflow config file. Now, you only provide a config file(s) for the workflow via the `--config-file` option. Inside of that, you provide `[event-{NAME}]` sections to specify what events to analyze. In these sections you specify the config files to give to `pycbc_inference`, using `config-files`, `config-overrides`, and `config-deletes` (same as what you would do on the command line). You can also provide an `nruns` option to have multiple inferences of `pycbc_inference` run on the same event with different starting seeds. The output from all of the instances are combined into a single posterior file for an event, with all posterior plots generated from that posterior file. You can also provide a `label` for more human-readable titles on the results page. Example:
```
[event-gw150914]
label = GW150914+09:50:45UTC
config-files = /work/cdcapano/WWW/scratch/prs/update_inference_workflow/inference_workflow_test/inference_config_files/sampler.ini
	/work/cdcapano/WWW/scratch/prs/update_inference_workflow/inference_workflow_test/inference_config_files/gw150914_like.ini
	/work/cdcapano/WWW/scratch/prs/update_inference_workflow/inference_workflow_test/inference_config_files/o1_data.ini
config-overrides = data:trigger-time:1126259462.43
nruns = 3
```
To analyze multiple events, you just provide more sections.

[...]

PR #2596, opened 2019-03-26
Implement condor checkpointing

duncan-brown:
This patch allows the user to specify a signal in the workflow ini file which triggers the sampler to exit after it has written a checkpoint. If the right set of ini options are specified, then the condor will restart the job, taking care of the transfer of checkpoint files.

To enable this using `SIGUSR2`, add the following pegasus profiles for the executable:
```
[pegasus_profile-inference]
pegasus|gridstart = NoGridStart
condor|+CheckpointExitBySignal = True
condor|+CheckpointExitSignal = 12
condor|+SuccessCheckpointExitBySignal = True
condor|+SuccessCheckpointExitSignal = 12
condor|+WantFTOnCheckpoint = True
condor|+CheckpointSig = 12
condor|periodic_remove = (HoldReasonCode is 13)
```
and the following option for the sampler:
```
[sampler]
checkpoint-signal = USR2
```

PR #3144, opened 2020-02-09
Restore condor checkpointing in inference

cdcapano:
In #3140, I inadvertently removed the condor checkpointing for inference jobs that was implemented in #2596. When setting up the inference nodes, I used `core.Executable`. I didn't realize there was a `PycbcInferenceExecutable` class in `jobsetup.py` that had extra things added to its `create_node` function to alert condor about the checkpoint file. This switches back to using that class in `jobsetup.py`, and updates that class to remove data-related things that are no longer set on the command line.


cdcapano:
@ahnitz I can't get Travis to complete successfully, and I've restarted it multiple times. The error looks related to virtualenv, but I haven't been able to find anything about it via Google. Any ideas?

PR #3145, opened 2020-02-12
numpy hstack FutureWarning fix

GarethDaviesGW:
Numpy hstack is giving a FutureWarning that generators will not be allowed as input - this changes the input to hstack from a generator to a list


GarethDaviesGW:
I used git grep to check if hstack is used elsewhere, it is also used in `bin/pycbc_multi_inspiral` and `pycbc/population/scale_injections.py` on things that are already list/np array so are safe

PR #3147, opened 2020-02-12
Separate the ordered ifo list from the number of detectors in tags

GarethDaviesGW:
So that changes can be made easily to config for all combinations within a certain number of detectors, separate the ordered ifo list from the number of detectors when creating tags


GarethDaviesGW:
This doesn't affect any currently used config files in gwastro/pycbc-config (or any in LIGO gitlab for that matter), but if we would like to implement different (for e.g.) parallelization factors in 2det vs 3det situations, then this will make the config file much more manageable

PR #3148, opened 2020-02-12
Percentile-percentile plot upgrades

cdcapano:
The executable for making PP plots has the default injection group set to an older value. This fixes that. It also makes it better at plotting multiple parameters at once. The default style is now `seaborn-colorblind` to make colorblind-friendly plots. It also adds a line style iterator. When plotting multiple parameters, there is now a unique combination of color/line style for up to 18 combinations; more if a different `mpl-style` is used. The aspect ratio of figures is now square to make it easier to compare to y = x line, and doing a KS test is optional. Finally, I renamed the executable from `pycbc_inference_plot_inj_intervals` to `pycbc_inference_plot_pp`.

Example [here](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/prs/fix_pp_plot/test2.png).

Putting on hold as this depends on #3146.


cdcapano:
@ahnitz This is ready for review now.

PR #3149, opened 2020-02-12
Extend source classification to all kinds of GraceDB uploads

veronica-villa:
Currently the source classification in PyCBC Live is only performed for triggers resulting from a two-detector coincidence. This PR extends it for the triggers resulting from the SNR maximization followup (`pycbc_optimize_snr`) as well as single-detector triggers (`check_singles()` in `pycbc_live`).

### DIFF
--- a/bin/pycbc_optimize_snr
+++ b/bin/pycbc_optimize_snr
@@ -20,7 +20,7 @@ from pycbc.conversions import (
 )
 from scipy.optimize import differential_evolution
 from pycbc.io import live
-from pycbc.io.hdf import recursively_load_hdf5_group_to_dict
+from pycbc.io.hdf import load_hdf5_to_dict
 from pycbc.detector import Detector
 from pycbc.psd import interpolate
 
@@ -317,12 +317,13 @@ channel_names = {}
 for ifo in fp['channel_names'].keys():
     channel_names[ifo] = fp['channel_names'][ifo][()]
 
+mc_area_args = load_hdf5_to_dict(fp, 'mc_area_args/')
+
 kwargs = {'psds': {ifo: followup_data[ifo]['psd'] for ifo in ifos},
           'low_frequency_cutoff': flow,
           'followup_data': followup_data,
           'channel_names': channel_names,
-          'mc_area_args': recursively_load_hdf5_group_to_dict(fp,
-                                                              'mc_area_args/')}
+          'mc_area_args': mc_area_args}
 
 doc = live.SingleCoincForGraceDB(ifos, coinc_results,
                                  upload_snr_series=True, **kwargs)
--- a/bin/pycbc_optimize_snr
+++ b/bin/pycbc_optimize_snr
@@ -20,6 +20,7 @@ from pycbc.conversions import (
 )
 from scipy.optimize import differential_evolution
 from pycbc.io import live
+from pycbc.io.hdf import recursively_load_hdf5_group_to_dict
 from pycbc.detector import Detector
 from pycbc.psd import interpolate
 
@@ -319,7 +320,9 @@ for ifo in fp['channel_names'].keys():
 kwargs = {'psds': {ifo: followup_data[ifo]['psd'] for ifo in ifos},
           'low_frequency_cutoff': flow,
           'followup_data': followup_data,
-          'channel_names': channel_names}
+          'channel_names': channel_names,
+          'mc_area_args': recursively_load_hdf5_group_to_dict(fp,
+                                                              'mc_area_args/')}
 
 doc = live.SingleCoincForGraceDB(ifos, coinc_results,
                                  upload_snr_series=True, **kwargs)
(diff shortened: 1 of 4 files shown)


veronica-villa:
An example of a trigger from SNR optimization can be seen in: https://gracedb-playground.ligo.org/events/T256619/view/

bin/pycbc_optimize_snr
```diff
@@ -319,7 +320,9 @@ for ifo in fp['channel_names'].keys():
 kwargs = {'psds': {ifo: followup_data[ifo]['psd'] for ifo in ifos},
           'low_frequency_cutoff': flow,
           'followup_data': followup_data,
-          'channel_names': channel_names}
+          'channel_names': channel_names,
+          'mc_area_args': recursively_load_hdf5_group_to_dict(fp,
```
titodalcanton: Just a minor code readability point: here I would do
```python
mc_area_args = recursively_load_hdf5_group_to_dict(fp, ...)
kwargs = {'psd': ...
          'mc_area_args': mc_area_args}
```

pycbc/io/hdf.py
```diff
@@ -960,13 +960,53 @@ def recursively_save_dict_contents_to_group(h5file, path, dic):
...
-        if isinstance(item, (np.ndarray, np.int64, np.float64, str, bytes, tuple, list)):
+        if isinstance(item, (np.ndarray, np.int64, np.float64, str, int, float,
+                             bytes, tuple, list)):
             h5file[path + str(key)] = item
         elif isinstance(item, dict):
             recursively_save_dict_contents_to_group(h5file, path + key + '/', item)
         else:
             raise ValueError('Cannot save %s type' % type(item))
 
+def load_hdf5_to_dict(filename):
```
titodalcanton: Do you actually use this function? If not, please remove.

pycbc/io/hdf.py
```diff
@@ -960,13 +960,53 @@ def recursively_save_dict_contents_to_group(h5file, path, dic):
...
+    filename:
+	hdf5 file to be loaded as a dictionary
+
+    Returns
+    -------
+    dictionary with hdf5 file content
+
+    """
+    with h5py.File(filename, 'r') as h5file:
+        recursively_load_hdf5_group_to_dict(h5file, '/')
```
titodalcanton: Looks like a `return` is missing here (but see previous comment).

pycbc/io/hdf.py
```diff
@@ -960,13 +960,53 @@ def recursively_save_dict_contents_to_group(h5file, path, dic):
...
+
+def recursively_load_hdf5_group_to_dict(h5file, path):
```
titodalcanton: Can this function be called simply `load_hdf5_to_dict()`?

pycbc/mchirp_area.py
```diff
@@ -320,7 +320,7 @@ def calc_probabilities(mchirp, snr, eff_distance, src_args):
     # equal to the maximum mass, the probability for BBH is 100%
     mc_max = mass_limits['max_m1'] / (2 ** 0.2)
     if trig_mc_det['central'] > mc_max * (1 + z['central']):
-        if mass_gap is not False:
+        if mass_gap:
```
titodalcanton: 👍

titodalcanton:
Thanks @veronica-villa, this looks great. Please see the comments I left and the code climate issues.

pycbc/io/hdf.py
veronica-villa: I wrote it this way to mimic the two already existing functions `recursively_save_dict_contents_to_group` and `save_dict_to_hdf5`,  but it's simpler as you say

pycbc/io/hdf.py
```diff
@@ -960,13 +960,38 @@ def recursively_save_dict_contents_to_group(h5file, path, dic):
...
+    Returns
+    -------
+    dic:
+        dictionary with hdf5 file group content
+    """
+    dic = {}
+    for key, item in h5file[path].items():
+        if isinstance(item, h5py._hl.dataset.Dataset):
+            dic[key] = item[()]
+        elif isinstance(item, h5py._hl.group.Group):
```
titodalcanton: This is raising some more Codeclimate issues. Is it possible to just use `h5py.Group` and `h5py.Dataset` here? Or is there a specific reason to use those in `_hl.dataset`?

veronica-villa: I just used the output I got from doing `type(h5file[path])` for groups and datasets but I checked and `h5py.Group` and `h5py.Dataset` are valid types too, so thanks for solving this

titodalcanton: Ok nice, but make sure it works! They may be slightly different derived classes.

veronica-villa: I've checked it with ipython and it seems to work correctly

titodalcanton:
The Travis failure is unrelated to this, so merging.

Issue #3160, opened 2020-02-23
is there any typo?


yanyuechuixue:
https://github.com/gwastro/pycbc/blob/8161365fe6e5e86f0ac591dc62651251f2ed4b54/pycbc/inference/sampler/multinest.py#L405

maybe it should be 
``` python
obj.setup_output(output_file)
```

cdcapano:
Yes, that's a bug, thanks for catching. I've filed a pull request to fix it.

cdcapano:
Fixed by #3162.

PR #3162, opened 2020-02-24
Fix setup_output calls

cdcapano:
As caught in Issue #3160, PR #2948 introduced a bug into setting up the multinest sampler. This fixes that. Also removes unused `setup_output` class methods in other samplers, and removes the `force` keyword argument in the `setup_output` function in `pycbc/inference/sampler/base.py`, since that isn't used by anything anymore.

### DIFF
--- a/pycbc/inference/sampler/base.py
+++ b/pycbc/inference/sampler/base.py
@@ -155,15 +155,12 @@ class BaseSampler(object):
 #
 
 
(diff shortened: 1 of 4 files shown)


ahnitz:
@cdcapano before merging, please update the commit message to explicitly acknowledge @yanyuechuixue finding this bug.

PR #3166, opened 2020-02-28
install companion for travis

ahnitz:


ahnitz:
Also install companion libs so that optional parts show up in documentation and can be used in testing.

PR #3169, opened 2020-03-03
make the ValueError message more accurate by including dynesty

yi-fan-wang:
This is a minor modification. Both Dynesty and Ultranest do not support sampling transforms, therefore I modify the ValueError message to make it more accurate.


ahnitz:
@yi-fan-wang Thanks!

PR #3170, opened 2020-03-04
population run keywords into dynesty

ahnitz:


ahnitz:
This passes all the keyword arguments that Dynesty's run command uses.

PR #3179, opened 2020-03-11
Optionally import epsie file

### LINKED ISSUE #3178: epsie should not be hard-required
Espsie imports still leak in when running inference code. @cdcapano Can you sanitize these so epsie is not a hard requirement.

cdcapano:
Makes it so that `EpsieFile` will only be imported into the `inference.io` namespace if epsie is installed. This should fix #3178 .


cdcapano:
@ahnitz I can't get travis to succeed. Seems to be an unrelated error. I've tried restarting the build, but no luck.

PR #3181, opened 2020-03-12
Hack to remove setuptools upgrade until it is fixed

GarethDaviesGW:
Temporary fix to travis failure issues


GarethDaviesGW:
(Source of the problem found by @KoustavChandra)

PR #3186, opened 2020-03-13
Add in Gracedb search functionality to results page

GarethDaviesGW:
Add in a link to results pages which links to a search on GraceDB for superevents 3s around a single or coinc time in snglinfo and coincinfo

Note that it is possible to include this as a gracedb api search within python, but requires grid proxy credentials, so is not useful when using --no-grid with pycbc_submit_dax. I couldn't get it to work within the workflow either (standalone worked fine, within workflow didn't work) - I can add in this as an attempt to search but if the link fails then it falls back to the search, but thought this was cleaner for now

### DIFF
--- a/bin/minifollowups/pycbc_page_coincinfo
+++ b/bin/minifollowups/pycbc_page_coincinfo
@@ -21,6 +21,7 @@ import matplotlib; matplotlib.use('Agg')
 import numpy, lal, datetime
 import pycbc.version, pycbc.results, pycbc.pnutils
 from pycbc.events import ranking
+from pycbc.results import followup
 
 parser = argparse.ArgumentParser()
 parser.add_argument('--version', action='version',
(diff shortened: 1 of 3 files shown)


GarethDaviesGW:
(added tito as reviewer as it was discussion with him for the idea)

bin/minifollowups/pycbc_page_coincinfo
```diff
@@ -46,6 +47,9 @@ trig_input.add_argument('--trigger-id', type=int,
 parser.add_argument('--include-summary-page-link', action='store_true',
     help="If given, will include a link to the DQ summary page on the "
          "single detector trigger tables.")
+parser.add_argument('--include-gracedb-link', action='store_true',
+    help="If given, will provide a link to search GraceDB for events "
+         "within a 3s window around the coincidence time.")
```
titodalcanton: I see the two options `--include-summary-page-link` and `--include-gracedb-link` are used by two scripts. Have you considered using a `add_followup_options()` function to the `followup` module?

PR #3190, opened 2020-03-19
Start to add in codes for inclusion of offline singles

GarethDaviesGW:
Starting to add offline singles functionality:

This first commit will almost definitely cause codeclimate to complain due to repeated code between this and coinc_findtrigs and coinc_statmap. I'll sort this when I can

The initial IFAR calculation method is bad (count number louder in foreground), and will never be larger than the chunk size. So hopefully will get an improved version in  - this is just to get the code into a shared space

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_sngls_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_sngls_findtrigs
@@ -114,19 +114,21 @@ else:
 data = {'stat': [], 'decimation_factor': [], 'timeslide_id': [],
         'template_id': [], '%s/time' % ifo : [], '%s/trigger_id' % ifo: []}
 
+tid_counter = 0
 for tnum in template_ids:
-    logging.info('Obtaining trigs for template %i ..' % (tnum))
+    if not tid_counter % 500:
+        logging.info('Obtaining trigs for template %i out of %i',
+                     tid_counter, len(template_ids))
+    tid_counter += 1
     sngl = reader
     tids_full = sngl.set_template(tnum)
     times_full = sngl['end_time']
     sds_full = rank_method.single(sngl)
-    logging.info('%s' % len(tids_full))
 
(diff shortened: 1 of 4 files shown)


bin/hdfcoinc/pycbc_multiifo_sngls_findtrigs
```diff
@@ -0,0 +1,292 @@
...
+parser.add_argument("--ranking-statistic", choices=stat.statistic_dict.keys(),
+                    default='newsnr',
+                    help="The single-detector ranking statistic to calculate")
+parser.add_argument("--statistic-keywords", nargs='*',
+                    default=[],
+                    help="Provide additional key-word arguments to be sent to "
+                         "the statistic class when it is initialized. Should "
+                         "be given in format --statistic-keywords "
+                         "KWARG1:VALUE1 KWARG2:VALUE2 KWARG3:VALUE3 ...")
+parser.add_argument("--loudest-keep-values", type=str, nargs='*',
```
ahnitz: This is unused no?

GarethDaviesGW: yes thats right - I thought I'd managed to remove all unused arguments must have missed that one

ahnitz:
@GarethDaviesGW Can you look at some of the CC issues? Some of them look like they should be addressed.

GarethDaviesGW:
> @GarethDaviesGW Can you look at some of the CC issues? Some of them look like they should be addressed.

CC is saying that it is fixed - though most of these probably should have been caught before as it was just moving the code! - hopefully these fixes are acceptable

GarethDaviesGW:
I'm not sure what the status is here for codebeat - it's saying that it has found some issues but hasn't decided whether it has failed or passed

GarethDaviesGW:
well it's let me merge anyway

PR #3198, opened 2020-03-30
Prior histogram normalisation

GarethDaviesGW:
Update prior histogram creation to be normalised to integrate to 1 rather than so that the largest count is 1


GarethDaviesGW:
Wrong Dtphase version - ignore this PR

PR #3205, opened 2020-04-07
Made the pycbc_page_found_missed colourblind friendly.

KoustavChandra:
Made the found-missed plot colourblind friendly.

### DIFF
--- a/bin/hdfcoinc/pycbc_page_foundmissed
+++ b/bin/hdfcoinc/pycbc_page_foundmissed
@@ -6,6 +6,10 @@ import matplotlib; matplotlib.use('Agg')
 import matplotlib.pyplot as plot
 import pycbc.results.followup, pycbc.pnutils, pycbc.results, pycbc.version
 import pycbc.pnutils
+import seaborn as sns
+
+sns.set_style("whitegrid")
+sns.set_palette('colorblind')
 
 parser = argparse.ArgumentParser(description=__doc__)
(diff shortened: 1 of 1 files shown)


KoustavChandra:
We need to update requirements.txt since it doesn't contain seaborn. In the section: # requirements for most basic library use

GarethDaviesGW:
> We need to update requirements.txt since it doesn't contain seaborn. In the section: # requirements for most basic library use

Do you have any example outputs that we could show here?

You should be able to update requirements.txt and commit that and it should then pass Travis checks

ahnitz:
@KoustavChandra Is there a way to do this without pulling in the seaborn dependency or does matplotlib not have any good colorblind-friendly colormaps?

KoustavChandra:
@ahnitz I believe there is. The simplest solution as per me is not to use green and red in the same plot since the most common type of colourblindness is to separate green and red colour. So changing the colourmap to something else (from Viridis) to cividis will be enough.

GarethDaviesGW:
cividis is based on viridis but optimised for colour deficiency and doesn't include green (which I guess was the issue with the red corsses) - maybe try that out?

KoustavChandra:
@GarethDaviesGW you can find an example of the plot [here](https://ligo.gravity.cf.ac.uk/~koustav.chandra/LSC/images/colorblind/test.png)

PR #3207, opened 2020-04-07
Change unused newsnr_sgchisq_psdvar function

tdent:
New stat scales both SNR and chisq by psd_var

Also change 'bchisq' where it appears and is actually reduced-bruce-chisq to 'brchisq' to remind us it is 'reduced'

Also put functions in more logical/chronological order


tdent:
@ahnitz if you have no objections / comments I'll merge soon.

ahnitz:
no objection

PR #3210, opened 2020-04-10
add flexibility to UltraNest sampler

JohannesBuchner:
This exposes a few more options of UltraNest allowing resuming and plotting (log_dir) and efficient sampling of high-dimensional problems.

### DIFF
--- a/pycbc/inference/sampler/ultranest.py
+++ b/pycbc/inference/sampler/ultranest.py
@@ -52,11 +52,16 @@ class UltranestSampler(BaseSampler):
     ----------
     model : model
         A model from ``pycbc.inference.models``.
+    log_dir : str
+        Folder where files should be stored for resuming (optional).
+    stepsampling : bool
+        If false, uses rejection sampling. If true, uses
+        hit-and-run sampler, which scales better with dimensionality.
     """
     name = "ultranest"
     _io = UltranestFile
 
-    def __init__(self, model, **kwargs):
+    def __init__(self, model, log_dir=None, stepsampling=False, **kwargs):
         super(UltranestSampler, self).__init__(model)
 
         import ultranest
@@ -65,7 +70,12 @@ class UltranestSampler(BaseSampler):
         self._sampler = ultranest.ReactiveNestedSampler(
             list(self.model.variable_params),
             log_likelihood_call,
-            prior_call)
+            prior_call, log_dir=log_dir, resume=True)
+        
+        if stepsampling:
+            import ultranest.stepsampler
+            self._sampler.stepsampler = ultranest.stepsampler.RegionBallSliceSampler(
+                nsteps=100, adaptive_nsteps='move-distance', region_filter=True)
 
         self.nlive = 0
         self.ndim = len(self.model.variable_params)
@@ -74,6 +84,9 @@ class UltranestSampler(BaseSampler):
 
     def run(self):
         self.result = self._sampler.run(**self.kwargs)
+        self._sampler.print_results()
+        self._sampler.plot()
+
 
     @property
     def io(self):


pycbc/inference/sampler/ultranest.py
```diff
@@ -74,6 +84,9 @@ def __init__(self, model, **kwargs):
 
     def run(self):
         self.result = self._sampler.run(**self.kwargs)
+        self._sampler.print_results()
+        self._sampler.plot()
```
cdcapano, line 88: What does `print_results` and `plot` do?

pycbc/inference/sampler/ultranest.py
```diff
@@ -52,11 +52,16 @@ class UltranestSampler(BaseSampler):
...
+        Folder where files should be stored for resuming (optional).
+    stepsampling : bool
+        If false, uses rejection sampling. If true, uses
+        hit-and-run sampler, which scales better with dimensionality.
     """
     name = "ultranest"
     _io = UltranestFile
 
-    def __init__(self, model, **kwargs):
+    def __init__(self, model, log_dir=None, stepsampling=False, **kwargs):
```
cdcapano, line 64: Thanks for adding support for stepsampling. In order to set this when running, you'll need to add it to the [list of known options](https://github.com/gwastro/pycbc/blob/6679ae985f26ebaf1a4fbf4d4f34a0bff7257fb5/pycbc/inference/sampler/ultranest.py#L105-L118) in the `from_config` function, so that it can be turned on/off using the configuration file.

ahnitz:
@JohannesBuchner Thank you for this PR! We are happy for any help. 

The reason this is currently failing is because of the plotting option added. Could you amend the patch so that the built in ultranest plotting is not enabled by default? 

Keep in mind that we usually run our inference in a headless mode, so there isn't necessarily going to be the ability to do plotting like this.

```
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/pycbc/inference/sampler/ultranest.py", line 88, in run
    self._sampler.plot()
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/ultranest/integrator.py", line 2222, in plot
    self.plot_corner()
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/ultranest/integrator.py", line 2242, in plot_corner
    cornerplot(self.results, logger=self.logger if self.log else None)
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/ultranest/plot.py", line 62, in cornerplot
    labels=paramnames, show_titles=True)
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/corner/corner.py", line 207, in corner
    fig, axes = pl.subplots(K, K, figsize=(dim, dim))
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/matplotlib/pyplot.py", line 1184, in subplots
    fig = figure(**fig_kw)
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/matplotlib/pyplot.py", line 533, in figure
    **kwargs)
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/matplotlib/backend_bases.py", line 161, in new_figure_manager
    return cls.new_figure_manager_given_figure(num, fig)
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/matplotlib/backends/_backend_tk.py", line 1046, in new_figure_manager_given_figure
    window = Tk.Tk(className="matplotlib")
  File "/opt/python/2.7.15/lib/python2.7/lib-tk/Tkinter.py", line 1819, in __init__
    self.tk = _tkinter.create(screenName, baseName, className, interactive, wantobjects, useTk, sync, use)
_tkinter.TclError: no display name and no $DISPLAY environment variable

ahnitz:
> About the log files: what type of files are being saved for checkpointing? I'd like to avoid saving extra files if possible (we're currently doing this with multinest, but it's something I'd like to undo in the future). Our preferred modus operandi is to limit all file I/O to the `.checkpoint` and `.bkup` (which is just a copy of the former) hdf files. Are you just using pickle for saving the log files in Ultanest? If so, would it be possible to use the [dump_state]

@cdcapano Looking at the ultranest code, I think the important state is being saved as some data in an hdf5 file. From our side, I think we'd just want a way to specify the file pointer so that we can control the overall file setup in our usual way.

cdcapano:
@ahnitz Ah, ok, that makes it even easier then. @JohannesBuchner can you confirm?

JohannesBuchner:
* Yes, all the important information is in the hdf5 file `log_dir + '/results/points.hdf5'`.

* re: matplotlib: You can solve this by specifying a working matplotlib backend on the headless server, e.g. put "backend: Agg" into file matplotlibrc placed in the current directory or $HOME/.matplotlib/matplotlibrc.

* plot() places plots in log_dir + '/plots/'
* print_results() prints to stdout a summary of the uncertainties for each parameter and the logZ estimate.

JohannesBuchner:
I am interested in the correctness and performance of samplers, so would be interested in systematic comparisons you make, and suggestions for test problems. There may be faster samplers than UltraNest, but it tends to give correct results in a wider range of problems.

I am currently using [the pycbc-tutorial problem](https://github.com/gwastro/pycbc-tutorials/blob/master/tutorial/inference_1_ModelsAndPEByHand.ipynb), slightly modified to constrain simulateously the distance, location, chirp mass & mass ratio of GW170817. Is that a representative challenging problem?

ahnitz:
@JohannesBuchner Yes, if you are sampling over those parameters simultaneously, that's a reasonably good example problem, though I would also include inclination and polarization if you weren't already.

cdcapano:
Superseded by #3332

ahnitz:
@JohannesBuchner We are merging your patch with just a small change to make the plotting optional (with a flag) in #3332. We'll remember not to squash to preserve your patch contribution. 

@JohannesBuchner Also, I'm quite curious if you had a chance to try out any GW problems yet. Might be interesting what you've learned about tuning the ultranest sampler (or maybe it's good with default settings for our problems).

JohannesBuchner:
> @JohannesBuchner We are merging your patch with just a small change to make the plotting optional (with a flag) in #3332. We'll remember not to squash to preserve your patch contribution.
Sounds good. 
> @JohannesBuchner Also, I'm quite curious if you had a chance to try out any GW problems yet. 
> Might be interesting what you've learned about tuning the ultranest sampler (or maybe it's good with default settings for our problems).

In my test (as described above), I now included inclination, but polarization is unconstrained, so I excluded it again.

It works well. On some problems, especially with many parameters or poor data you it may seem stuck, and you probably want to turn on `stepsampling=True`.

PR #3214, opened 2020-04-14
JointDistribution: introduce loop for renormalization due to constraints, and improve documentation

dbkeitel:
@cdcapano @cmbiwer trying to follow the [example](https://pycbc.org/pycbc/latest/html/pycbc.distributions.html#module-pycbc.distributions.joint) for a JointDistribution with constraints, I ran into two errors, one trivial and one not:

1.  import needs to be `Uniform` not `uniform`
1. Since the example constraint function `mtotal_lt_30` is not vectorized, but evaluated internally over an array of amples for the pdf renormalization, it throws an
```
ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

At first I thought the right fix would be to vectorize the example function, e.g.
```
def mtotal_lt_30(params):
    return np.less(params["mass1"] + params["mass2"],30)
```
but as far as I know, the resulting JointDistribution object won't be able to be called over arrays anyway, so that seems to be a red herring. (That said, vectorizing the whole module would be an amazing overall improvement!)

Instead, this PR introduces an explicit loop over the test samples, along with additional documentation fixes.

Commit message:
```
fix renormalization due to constraints in JointDistribution, and doc

* since the resulting distribution cannot be evaluated on arrays,
  the constraint functions need not / cannot be vectorized either
  (and the given example isn't),
  so internally need to loop over the random samples
  used for pdf renormalization
* also document n_test_samples kwarg
* and minor doc fixes
```

### DIFF
--- a/pycbc/distributions/joint.py
+++ b/pycbc/distributions/joint.py
@@ -127,6 +127,10 @@ class JointDistribution(object):
             # set new scaling factor for prior to be
             # the fraction of acceptances in random sampling of entire space
             self._pdf_scale = result.sum() / float(n_test_samples)
+            if self._pdf_scale == 0.0:
+                raise ValueError("None of the random draws for pdf "
+                    "renormalization satisfied the constraints. "
+                    " You can try increasing the 'n_test_samples' keyword.")
 
         else:
             self._pdf_scale = 1.0
--- a/pycbc/distributions/joint.py
+++ b/pycbc/distributions/joint.py
@@ -36,11 +36,13 @@ class JointDistribution(object):
         (retrieved from the distributions' params attribute) must be the same
         as the set of variable_args provided.
     \*\*kwargs :
-        Valid keyword arguments include `constraints`. `constraints` is a list
-        functions that accept a dict of parameters with the parameter name as
-        the key. If the constraint is satisfied the function should return
-        True, if the constraint is violated, then the function should return
-        False.
+        Valid keyword arguments include:
+        `constraints` : a list of functions that accept a dict of parameters
+        with the parameter name as the key. If the constraint is satisfied the
+        function should return True, if the constraint is violated, then the
+        function should return False.
+        `n_test_samples` : number of random draws used to fix pdf normalization
+        factor after applying constraints.
 
     Attributes
     ----------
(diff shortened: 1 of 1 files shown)


dbkeitel:
For the `Too many local variables (17/15)` codeclimate failure, how about moving the whole `if self._constraints` block to a `_renormalize_pdf_scale` method?

dbkeitel:
I've added a 2nd commit to catch a problem that's otherwise tricky to debug down the line:
```
JointDistribution: throw error if self._pdf_scale == 0.0
    
    * this means that none of the random draws
      for pdf renormalization satisfied the constraints
    * can happen for self-contradictory constraints,
      or just for harsh constraints and too low n_test_samples
    * would lead to an unusable distribution
```

pycbc/distributions/joint.py
```diff
@@ -117,12 +119,17 @@ def __init__(self, variable_args, *distributions, **kwargs):
...
-            for constraint in self._constraints:
-                result = constraint(samples) & result
+            for n in numpy.arange(n_test_samples):
+                sample = {param: samples[param][n] for param in distparams}
+                for constraint in self._constraints:
+                    result[n] = constraint(sample) & result[n]
 
             # set new scaling factor for prior to be
             # the fraction of acceptances in random sampling of entire space
             self._pdf_scale = sum(result) / float(n_test_samples)
```
cdcapano, line 129: Doing `result.sum()` here will be slightly faster.

pycbc/distributions/joint.py
```diff
@@ -117,12 +119,17 @@ def __init__(self, variable_args, *distributions, **kwargs):
...
+            if self._pdf_scale == 0.0:
+                raise ValueError("None of the random draws for pdf "
+                    "renormalization satisfied the constraints.")
```
cdcapano, line 132: Thanks for catching this! Could you add a message to the error message telling the user to try to increase the number of test points by setting `n_test_samples` to something large than its current value (giving the current value)? That way they know what to do if they run into this error.

ahnitz:
@cdcapano To add another use case for vectorizing the package, there are samplers which can do vectorized sampling (such as Ultranest which we support). We have some models where the likelihood is fast enough it would be worth speeding up the prior evaluation as well. Vectorization would the least painful first step.

dbkeitel:
Thanks @cdcapano , both minor changes done.

Regarding overall vectorization, if the project I'm currently experimenting with (and hence where I stumbled over these fixes) works out, I may be interested in helping out at a later point. `JointDistribution` is really quite a powerful class!

cdcapano:
Excellent, thanks @dbkeitel. Yes, if you want to help vectorize this at some point, that would be great!

PR #3216, opened 2020-04-15
changes to PTA histogram normalisation

GarethDaviesGW:
Renormalise PTA histogram to a PDF when reading

Take account of expected noise in PTA histogram in ExpFitSGFgBgNormNewStatistic

(also renaming some statistics)

### DIFF
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -278,6 +278,7 @@ class PhaseTDNewStatistic(NewSNRStatistic):
         self.ref_snr = 5.0
         self.relsense = {}
         self.swidth = self.pwidth = self.twidth = None
+        self.srbmin = self.srbmax = None
         self.max_penalty = None
         self.pdtype = []
         self.weights = {}
@@ -1194,10 +1195,10 @@ class ExpFitSGFgBgNormNewStatistic(PhaseTDNewStatistic,
         # noise coincs for n_ifos > 2
         noise_twindow = coinc_rate.multiifo_noise_coincident_area(
                             self.hist_ifos, kwargs['time_addition'])
-        hist_volume = noise_twindow * (2 * numpy.pi * (self.srbmax - self.srbmin)
-                                       * self.swidth) ** (n_ifos - 1)
+        hist_vol = noise_twindow * (2 * numpy.pi * (self.srbmax - self.srbmin)
+                                    * self.swidth) ** (n_ifos - 1)
 
-        logr_n = - numpy.log(hist_volume)
+        logr_n = - numpy.log(hist_vol)
 
         # Combine to get final statistic
         loglr = logr_s - logr_n + network_logvol - ln_noise_rate
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -1182,11 +1182,14 @@ class ExpFitSGFgBgNormNewStatistic(PhaseTDNewStatistic,
         benchmark_logvol = s[0][1]['benchmark_logvol']
         network_logvol -= benchmark_logvol
 
-        # Use prior histogram to get log signal rate
+        # Use prior histogram to get Bayes factor for signal vs noise given
+        # time, phase and SNR differences between IFOs
+        # First get signal term logr_s
         stat = {ifo: st for ifo, st in s}
-        logr_shist = self.logsignalrate_multiifo(stat,
+        logr_s = self.logsignalrate_multiifo(stat,
                                              slide * step, to_shift)
 
+        # Calculate noise term of phase-time-amplitude Bayes factor, logr_n
         noise_twindow = coinc_rate.multiifo_noise_coincident_area(
                             self.hist_ifos, kwargs['time_addition'])
         # Calculate the volume of the histogram which could be filled by
@@ -1205,10 +1208,13 @@ class ExpFitSGFgBgNormNewStatistic(PhaseTDNewStatistic,
 
         # One over this volume will be the PDF assuming uniform noise
         # distributions in each dimension.
-        logr_nhist = - numpy.log(hist_vol)
+        logr_n = - numpy.log(hist_vol)
+
+        # Combine the signal and noise terms of the PTA Bayes factor
+        pta_bf = logr_s - logr_n
 
         # Combine to get final statistic
-        loglr = logr_shist - logr_nhist + network_logvol - ln_noise_rate
+        loglr = network_logvol - ln_noise_rate + pta_bf
         # cut off underflowing and very small values
         loglr[loglr < -30.] = -30.
         return loglr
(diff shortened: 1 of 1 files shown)


GarethDaviesGW:
Note: Codebeat issues are ones which were there before, but are showing up as new due to statistic renaming

pycbc/events/stat.py
```diff
@@ -1174,26 +1182,41 @@ def coinc_multiifo(self, s, slide, step, to_shift,
         benchmark_logvol = s[0][1]['benchmark_logvol']
         network_logvol -= benchmark_logvol
 
+        # Use prior histogram to get log signal rate
         stat = {ifo: st for ifo, st in s}
         logr_s = self.logsignalrate_multiifo(stat,
                                              slide * step, to_shift)
 
-        loglr = logr_s + network_logvol - ln_noise_rate
+        # Calculate noise rate in each bin given uniform phase time diffs
```
tdent, line 1190: I'm not seeing why we want to calculate noise rate in each bin if the signal rate term is now a proper density, i.e. rate per volume not rate per bin

pycbc/events/stat.py
```diff
@@ -1174,26 +1182,41 @@ def coinc_multiifo(self, s, slide, step, to_shift,
...
+        # and SNR ratio.
+        # Calculate bin & histogram volumes
```
tdent, line 1192: comment seems out of place - the code it's commenting only defines the n_ifos variable, which doesn't require explanation

pycbc/events/stat.py
```diff
@@ -1174,26 +1182,41 @@ def coinc_multiifo(self, s, slide, step, to_shift,
...
+        # noise coincs for n_ifos > 2
+        noise_twindow = coinc_rate.multiifo_noise_coincident_area(
+                            self.hist_ifos, kwargs['time_addition'])
+        hist_vol = noise_twindow * (2 * numpy.pi * (self.srbmax - self.srbmin)
+                                    * self.swidth) ** (n_ifos - 1)
+
+        logr_n = - numpy.log(hist_vol)
+
+        # Combine to get final statistic
+        loglr = logr_s - logr_n + network_logvol - ln_noise_rate
```
tdent, line 1204: this seems to have two noise rate terms, logr_n and ln_noise rate, I don't immediately see how they are supposed to fit together

pycbc/events/stat.py
```diff
@@ -1174,26 +1182,41 @@ def coinc_multiifo(self, s, slide, step, to_shift,
...
-        loglr = logr_s + network_logvol - ln_noise_rate
+        # Calculate noise rate in each bin given uniform phase time diffs
+        # and SNR ratio.
+        # Calculate bin & histogram volumes
+        n_ifos = len(self.hist_ifos)
+        # Effective histogram volume, as not all bins will be occupied by
+        # noise coincs for n_ifos > 2
+        noise_twindow = coinc_rate.multiifo_noise_coincident_area(
+                            self.hist_ifos, kwargs['time_addition'])
+        hist_vol = noise_twindow * (2 * numpy.pi * (self.srbmax - self.srbmin)
```
tdent, line 1198: what calculation is being done here needs to be explained a bit more, if we aren't referring to an equation : what is 2 pi doing?  what is the srbmax/srbmin thing for?  why is it relevant that not all bins are occupied?  (as we don't use bins to estimate density of noise coincs)

tdent:
@GarethDaviesGW the idea here is to rename the statistic that is being used for the multi-ifo paper and also to 'fix' the relative normalization? 
I think this should be OK given that we probably don't want to use the 'wrongly/randomly normalized' previous incarnation of the statistic.  If we did want to compare the new one to the previous one, the procedure would be to go back to an older version that contained the previous calculation .. well, or use 'a run I made earlier'.

pycbc/events/stat.py
```diff
@@ -1174,26 +1182,41 @@ def coinc_multiifo(self, s, slide, step, to_shift,
...
+
+        # Combine to get final statistic
+        loglr = logr_s - logr_n + network_logvol - ln_noise_rate
         # cut off underflowing and very small values
         loglr[loglr < -30.] = -30.
         return loglr
 
 
-class TwoOGCStatistic(ExpFitSGFgBgRateNewStatistic):
+class ExpFitSGPSDFgBgNormDensityStatistic(ExpFitSGFgBgNormNewStatistic):
```
tdent, line 1210: I don't think we need `Density` in the name here

pycbc/events/stat.py
```diff
@@ -1174,26 +1182,41 @@ def coinc_multiifo(self, s, slide, step, to_shift,
...
     def __init__(self, files=None, ifos=None, **kwargs):
-        ExpFitSGFgBgRateNewStatistic.__init__(self, files=files, ifos=ifos,
+        ExpFitSGFgBgNormNewStatistic.__init__(self, files=files, ifos=ifos,
                                               **kwargs)
         self.get_newsnr = ranking.get_newsnr_sgveto_psdvar_scaled
 
 
-class TwoOGCBBHStatistic(ExpFitSGFgBgRateNewStatistic):
+class ExpFitSGPSDFgBgNormDensityBBHStatistic(ExpFitSGFgBgNormNewStatistic):
```
tdent, line 1217: ditto here, can remove `Density`

GarethDaviesGW:
I've tried to improve commenting and explanation for each of the steps of the new calculation

pycbc/events/stat.py
```diff
@@ -1174,26 +1182,54 @@ def coinc_multiifo(self, s, slide, step, to_shift,
...
         stat = {ifo: st for ifo, st in s}
         logr_s = self.logsignalrate_multiifo(stat,
                                              slide * step, to_shift)
 
-        loglr = logr_s + network_logvol - ln_noise_rate
+        # Calculate noise term of phase-time-amplitude Bayes factor, logr_n
+        noise_twindow = coinc_rate.multiifo_noise_coincident_area(
+                            self.hist_ifos, kwargs['time_addition'])
+        # Calculate the volume of the histogram which could be filled by
+        # noise coincidences. Not all bins may be occupied due to
```
tdent, line 1196: i'd suggest not talking about 'the histogram' here - a histogram is just a clunky way of calculating/storing a density over some space. what is being calculated is the volume of phase/time/amplitude space. 

I'll edit via the web interface since other points look fine

tdent:
@GarethDaviesGW @ahnitz I've rejigged a bunch of comments, resolved (hopefully correctly) the conflict from Ian's 2-dethist improvement, and also dealt with #3222 , again hopefully correctly. Can you take a look and see if it all makes sense?

GarethDaviesGW:
self.hist usage looks correct to me - it isn't used as a dict where this is applied so use as a flag makes sense

PR #3217, opened 2020-04-15
PTA histogram - wrap phase, remove extraneous snr ratio bins

GarethDaviesGW:
Some bins were outside of the ranges of phase difference and SNR ratio due to smoothing

This PR makes the phases wrap back into the 0-2pi range when smoothing and removes any bins above/below the SNR ratio allowed range

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_dtphase
+++ b/bin/hdfcoinc/pycbc_multiifo_dtphase
@@ -6,10 +6,10 @@ doing a simple monte-carlo
 Output is the relative amplitude, time, and phase as compared to a reference
 IFO. A separate calculation is done for each IFO as a possible reference.
 The data is stored as two vectors : one vector gives the discrete integer bin
-corresponding to a particular location in 3*(Nifo-1)-dimensional 
+corresponding to a particular location in 3*(Nifo-1)-dimensional
 amplitude/time/phase space, the other gives the weight assigned to that bin.
 To get the signal rate this weight should be scaled by the local sensitivity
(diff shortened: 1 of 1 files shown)


tdent:
is there a plot of the effect of the wrapping / cutting on the phase/time histogram or (say) the snr/time histogram ?

GarethDaviesGW:
>is there a plot of the effect of the wrapping / cutting on the phase/time histogram or (say) the snr/time histogram ?
>(just 2-ifo examples would be fine)

There's a preliminary plot at https://ldas-jobs.ligo-wa.caltech.edu/~gdavies/testoutput/prior_hist_wrap/phasewrapping.png (I'll sort labels and get the plot central in time another day)

GarethDaviesGW:
> is there a plot of the effect of the wrapping / cutting on the phase/time histogram or (say) the snr/time histogram ?
> (just 2-ifo examples would be fine)

H-L plots of the histogram for phase diff vs time diff and snr ratio vs time diff are below:
image

image

GarethDaviesGW:
I think that the weight threshold is different between the two files. I'll see if I can re-run with comparable files

GarethDaviesGW:
With the updated threshold the occupied bins in the allowed region remains the same:
image

image

tdent:
Looks good to me. Which threshold is going to be used in production?

GarethDaviesGW:
The lower threshold (i.e. more bins filled) is the one that has been used in production previously, so I would say to continue with that

PR #3218, opened 2020-04-16
Add colorbar labels to foundmissed plots. General tidy-up of the foundmissed script

GarethDaviesGW:
Adding colorbar tick labels to foundmissed plots - this is currently hard-coded to give 5 labels

I have also done some tidying of the code having seen that quite a bit has been repeated in various if statements. The colorblind-friendly colour scheme of #3205 also hadn't been applied to the plots which did not have --gradient-far set

Examples:
Gradient FAR - Found on top
image

Gradient FAR - Missed on top
image

Set FAR - Found on Top
image

Set FAR - Missed on Top
image

### DIFF
--- a/bin/hdfcoinc/pycbc_page_foundmissed
+++ b/bin/hdfcoinc/pycbc_page_foundmissed
@@ -1,11 +1,12 @@
 #!/usr/bin/env python
 """ Plot found and missed injections.
 """
-import h5py, numpy, logging, os.path, argparse, sys, matplotlib
+import h5py, numpy, logging, os.path, argparse, sys
 import matplotlib; matplotlib.use('Agg')
 import matplotlib.pyplot as plot
 import pycbc.results.followup, pycbc.pnutils, pycbc.results, pycbc.version
 import pycbc.pnutils
+from pycbc import init_logging
 
 parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument('--injection-file',
@@ -17,8 +18,9 @@ parser.add_argument('--distance-type', default='decisive_optimal_snr',
                     choices=['decisive_distance', 'dec_chirp_distance',
                              'chirp_distance', 'comb_optimal_snr',
                              'decisive_optimal_snr', 'redshift'],
-                    help="Variable related to injected distance. Decisive distance and dec"
-                         " chirp distance only available for 2-ifo search")
+                    help="Variable related to injected distance. Decisive "
+                         "distance and dec chirp distance only available for "
+                         "2-ifo search")
 parser.add_argument('--plot-all-distance', action='store_true', default=False,
                     help="Plot all values of distance or SNR. If not given, "
                          "the plot will be truncated below 1")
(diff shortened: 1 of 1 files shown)


bin/hdfcoinc/pycbc_page_foundmissed
```diff
@@ -243,4 +253,4 @@ if ('.html' in args.output_file):
 pycbc.results.save_fig_with_metadata\
```
titodalcanton, line 253: Could you also PEP-ify this call? I would prepare a `title` variable before the call so you can just pass `title=title` and semplify the arg list.

bin/hdfcoinc/pycbc_page_foundmissed
```diff
@@ -201,6 +210,7 @@ ax = plot.gca()
 plot.xlabel(labels[args.axis_type])
 plot.ylabel(labels[args.distance_type])
 plot.grid()
+plot.tight_layout()
```
titodalcanton, line 213: I am debated on the `tight_layout()` here. It will definitely tidy up a *particular* plot, but it will also make the plot change slightly every time you make it with different input data. This slightly spoils the "flipping between two plots" feature that many of us love. Up to you though, especially if it is a mess to get a tidy plot without `tight_layout()`.

titodalcanton:
image

Modulo the comments above, and that you did a before/after comparison.

bin/hdfcoinc/pycbc_page_foundmissed
GarethDaviesGW: Good point, I'll remove this call - I hadn't realised it would ruin the flipping-between plots thing, which is probably one of the things this is used for most

bin/hdfcoinc/pycbc_page_foundmissed
GarethDaviesGW: Done (I hadn't actually scrolled that far in the tidy up!)

GarethDaviesGW:
For completeness here is a "before" plot - it is, however before all the changes including the colour scheme change that Koustav did
image

bin/hdfcoinc/pycbc_page_foundmissed
```diff
@@ -145,48 +146,56 @@ fig = plot.figure()
...
     caption = (fig_title + ": Red x's are missed injections. "
-               "Circles are found injections. The color indicates the value of "
-               "the false alarm rate." )
-    plot.subplots_adjust(right=0.99)
+               "Circles are found injections. The color indicates the value "
+               "of the false alarm rate." )
+
+points = plot.scatter(fvals, fdval, c=color, linewidth=0, s=16, norm=norm,
+                      marker='o', label='found', zorder=zfound,
+                      cmap=args.colormap)
```
KoustavChandra, line 187: Hi @GarethDaviesGW I don't see any arguments with colormap. Initially I recall I have put an option 

parser.add_argument('--colormap',default='cividis_r',
                   help="Type of colormap to be used for the plots.")

Such that anyone can switch the colourmap as they please. I wanted non-colourblind people to enjoy the spectrum of colours.

GarethDaviesGW: Do you know if cividis without the _r reversal is okay? - I think that will look a bit better, but want to keep it suitable for as many as possible

KoustavChandra: The markers were difficult to decipher without reversal.

GarethDaviesGW: If I've understood the comment correctly, the colormap argument is still there - the file will be collapsed on the github display so that it doesn't have too much of the file which is unchanged by this PR

KoustavChandra: One more thing args.colormap need parser.add_argument('--colormap') I don't see it here. Did I miss something?

KoustavChandra: Ouch. I missed it. Approved

GarethDaviesGW: It's still there on line 25: https://github.com/gwastro/pycbc/pull/3218/files#diff-c831f075d5a9aa6f90d34b98f9aef674R25

PR #3224, opened 2020-04-20
Generalize relative model to all frequency domain approximants

dfinstad:
This patch generalizes the relative likelihood model to be able to use any available frequency domain approximant, by switching the waveform generation from `spa_tmplt` to `get_fd_waveform_sequence`. This has been tested and confirmed to work for `TaylorF2`, `IMRPhenomD`, and `SEOBNRv4_ROM`.

To enable the use of different approximants, the fiducial waveform parameters required by the model are no longer provided explicitly to the `__init__` method. Instead these parameters are processed by `extra_args_from_config` into a single kwarg `fiducial_parameters`, using the same `{PARAM}_ref = {VALUE}` convention in the `[model]` section of the config file.

Other minor changes include:
 - restructure the loglr calculation to more closely match that done by `MarginalizedPhaseGuassianNoise`
 - optimize the binning algorithm to prevent edge cases where bins could be repeated or incorrectly ordered
 - change the linear approximation to anchor each bin at the left edge instead of the center, which ensures this model reduces to the non-relative likelihood in the limit of large number of bins

### DIFF
--- a/docs/_include/inference_example_relative.sh
+++ b/docs/_include/inference_example_relative.sh
@@ -1,0 +1,3 @@
+sh ../../examples/inference/relative/get.sh
+sh ../../examples/inference/relative/run.sh
+sh ../../examples/inference/relative/plot.sh
--- a/docs/inference/examples/relative.rst
+++ b/docs/inference/examples/relative.rst
@@ -1,0 +1,54 @@
+-----------------------------------
+Using the relative model
+-----------------------------------
+
+The relative model is useful for when you know the parameters of your signal
+(say the masses, spins, etc of a merger) that are near the peak of the
+likelihood. In this case, we don't need to calculate the likelihood at the full
+frequency resolution in order to accurately sample the neighborhood of the
+maximum likelihood. This can greatly speed up the calculation of the likelihood.
+To use this model you provide the near-peak parameters as fixed arguments as in
+the configuration file below.
+
+This example demonstrates using the ``relative`` model with the
+``emcee_pt`` sampler. First, we create the following configuration file:
+
+.. literalinclude:: ../../../examples/inference/relative/relative.ini
+   :language: ini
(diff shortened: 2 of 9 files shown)


ahnitz:
@dfinstad Can I ask at this point that you also add an example to the pycbc inference documetation?  Something similar to the single template example (which also runs directly in the documentation / unittests), but maybe add some constrained sampling over chirp mass say? Something which can complete in ~1-2 minutes. 

A page equivalent to this would be great. https://pycbc.org/pycbc/latest/html/inference/examples/single.html

dfinstad:
@ahnitz Sure thing, the documentation page and example should be added now.

ahnitz:
@dfinstad Do you have an example of what that looks like by chance?

dfinstad:
@ahnitz the plot produced by the example is below

image

ahnitz:
@dfinstad Was there any performance loss from dropping the SPA tmplt support? Or is there a bottleneck elsewhere atm? In past testing the sequence interface is not that fast compared to SPAtmplt.

ahnitz:
@dfinstad Small problem, there's no link to the new docs. You can see the page here though https://pycbc.org/pycbc/latest/html/inference/examples/relative.html. Can you add a link like we have for the other examples? Look at the inference rst files to see how.

dfinstad:
@ahnitz yeah I just noticed that myself, I'm testing the fix now and will file a PR once I've seen it works correctly.

Regarding the performance question, I actually didn't see any significant performance loss in switching to the sequence interface. A standalone timing test I put together showed comparable speed:
```
spatmplt average runtime: 0.0003283
get_fd_waveform_sequence average runtime (TaylorF2): 0.0004348
```
and full inference runs seem to finish in about the same amount of time I saw before the switch.

PR #3229, opened 2020-04-22
Updates to snrratehist

GarethDaviesGW:
snrratehist had some issues which come up when using with the new statistic
- X label was hard-coded to something incorrect (now more generic)
- Default values of x-min and bin-size were sensible for sum-of-SNRs ranking statistic, but not for others:
  - x-min default changes from 8.0 to np.floor of the peak of the exclusive background histogram
  - bin-size default changes from 0.1 to 1/100th of the statistic range
- Fill the background at the end (this was just personal preference for aesthetics)
- Clarify the involved IFOs in the title and caption

Example output (default values):
image

Previous output (default values):
image

### DIFF
--- a/bin/hdfcoinc/pycbc_page_snrratehist
+++ b/bin/hdfcoinc/pycbc_page_snrratehist
@@ -22,8 +22,8 @@ parser.add_argument('--version', action='version', version=pycbc.version.git_ver
 parser.add_argument('--trigger-file')
 parser.add_argument('--verbose', action='count')
 parser.add_argument('--output-file')
-parser.add_argument('--bin-size', type=float, default=0.1)
-parser.add_argument('--x-min', type=float, default=8.0)
+parser.add_argument('--bin-size', type=float)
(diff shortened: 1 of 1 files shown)


tdent:
Either capitalize 'statistic' or de-capitalize 'Ranking', also remove the comma

Prefer 'bin size' without capitals

If the 'grey area' at the end isn't labelled as meaning something then I vote against it (i.e. people will ask 'why is it grey' and then we'll have to tell them 'because gareth likes grey')

tdent:
actually the legend gives grey as meaning background uncorrelated with foreground .. so it might confuse people.

GarethDaviesGW:
image

Updated plot based on Tom's comments

tdent:
Looks OK to me..

tdent:
@ahnitz @stevereyes01 Ping to review this please.

stevereyes01:
@tdent @GarethDaviesGW You probably don't want my review. No longer in academia :-). Code looks fine.

PR #3230, opened 2020-04-22
ifar_catalog bug

GarethDaviesGW:
If --truncate-threshold is not set, then tries to take log10 of None --> error

Move the 'or' statement so that this doesn't fail


tdent:
Seems OK. Is the code used for anything in the pipeline atm?

GarethDaviesGW:
The code is used to calculate the ifar plots given the combined statmap (given no bg ifars)

It wasn't being called without truncate-threshold, but this showed up when I accidentally used an old version of config files

PR #3232, opened 2020-04-23
Add option to plot waveform track on top of qscan

micamu:
This PR adds the option to plot a waveform time-freq track on top of an omega scan in pycbc_plot_qscan. To avoid having to give too many new arguments, it takes the masses in the same format as the time_window (i.e. mass1,mass2). I wanted to do the same with the spins, but it fails if the first spin is negative (if someone knows how to fix that, I'll be happy to change it).

P.S: I have an example plot in my laptop but I don't see any way to attach it here. I can send it on slack if someone wants to see it.

### DIFF
--- a/bin/plotting/pycbc_plot_qscan
+++ b/bin/plotting/pycbc_plot_qscan
@@ -43,7 +43,7 @@ def t_window(s):
         start, end = map(float, s.split(','))
         return [start, end]
     except:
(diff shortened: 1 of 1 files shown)


micamu:
1126106138_H1
Here is an example plot!

titodalcanton:
Thanks Miriam. Same request as Ian's.

micamu:
Done. Waiting for checks to complete, otherwise it should be ready to merge

PR #3240, opened 2020-04-28
multiifo_dtphase time size

GarethDaviesGW:
dynamically choose the datatype depending on number of IFOs

Error if max possible tdiff is less than (approx) Earth diameter travel time


bin/hdfcoinc/pycbc_multiifo_dtphase
```diff
@@ -125,6 +125,16 @@ numpy.random.seed(args.seed)
 size = args.batch_size
 chunks = int(args.sample_size / size) + 1
 
+n_ifos = len(args.ifos)
+bin_dtype = numpy.int8 if n_ifos > 2 else numpy.int16
```
ahnitz, line 129: I think it would be better if the dtype where just an option, or there was a flag like --store-as-int8, and otherwise it was int32. Probably safe in the future if this is explicit, to make sure some thought is made before choosing yes/no on the flag.

PR #3241, opened 2020-04-28
get sizes for in-memory weights from the keys

ahnitz:


pycbc/events/stat.py
```diff
@@ -376,17 +375,19 @@ def get_hist(self, ifos=None):
                 # "weights" table a O(N) rather than O(NlogN) operation. It
                 # sacrifices RAM to do this, so is a good tradeoff for 2
                 # detectors, but not for 3!
-                pb_iinfo = numpy.iinfo(self.param_bin[ifo]['c0'].dtype)
-                self.pb_int_size = pb_iinfo.max - pb_iinfo.min + 1
+                # Size should be same for each reference ifo!
+                self.c0_size = 2 * (abs(self.param_bin[ifo]['c0']).max() + 1)
+                self.c1_size = 2 * (abs(self.param_bin[ifo]['c1']).max() + 1)
+                self.c2_size = 2 * (abs(self.param_bin[ifo]['c2']).max() + 1)
```
GarethDaviesGW, line 381: you need to take the min() value from this as param bin can take negative values (for time at least)

GarethDaviesGW: well the 2* thing will work, but be twice the size that's needed for e.g. phase

ahnitz: @GarethDaviesGW That's why I took the absolute value.

ahnitz: @GarethDaviesGW That's true, but it shouldn't be any more memory than we use now and preserves the logic elsewhere in the code. If I change the zero point then we have to keep track of it.

GarethDaviesGW: I hadn't seen the 2* abs thing before the first comment (the max should always be abs anyway I think)

Issue #3243, opened 2020-04-30
time series add_into alignment error


MarlinSchaefer:
When using the add_into function of a TimeSeries with a TimeSeries where the epoch is not exactly the same as the one of the original TimeSeries, a ValueError 'different epoch' is raised. This happens even if the epoch is manually set to floating point precision to be: original epoch + integer * original delta_t. The reason seems to be an internal check, where the GPS times are compared rather than the floating point representations. I've [attached a script](https://github.com/gwastro/pycbc/files/4558350/add_into_bug.txt) showcasing both manual alignment and random starting times.

ahnitz:
This should be fixed by #3244

PR #3244, opened 2020-04-30
make add into more robust

ahnitz:
This should address #3243. 

This makes the check on start time check for reasonable floating point differences and ensures that rounding differences don't cause a function failure.

### DIFF
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -849,7 +849,7 @@ class TimeSeries(Array):
             raise ValueError('Sample rate must be the same')
 
         # Other is disjoint
-        if ((other.start_time > self.end_time) or
+        if ((other.start_time >= self.end_time) or
            (self.start_time > other.end_time)):
             return self.copy()
 
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -178,7 +178,7 @@ class TimeSeries(Array):
 
         start_idx = (start - self.start_time) * self.sample_rate
         end_idx = (end - self.start_time) * self.sample_rate
-        
+
         if mode == 'floor':
             start_idx = int(start_idx)
             end_idx = int(end_idx)
@@ -870,10 +870,10 @@ class TimeSeries(Array):
         ts = self.copy()
         start = max(other.start_time, self.start_time)
         end = min(other.end_time, self.end_time)
-        
+
         opart = other.time_slice(start, end, mode='nearest')
         part = ts.time_slice(start, end, mode='nearest')
-        part[:] += opart[:] 
+        part[:] += opart[:]
         return ts
 
     @_nocomplex
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -53,7 +53,8 @@ class TimeSeries(Array):
     sample_rate
     """
 
-    def __init__(self, initial_array, delta_t=None, epoch=None, dtype=None, copy=True):
+    def __init__(self, initial_array, delta_t=None,
+                 epoch=None, dtype=None, copy=True):
         if len(initial_array) < 1:
             raise ValueError('initial_array must contain at least one sample.')
         if delta_t is None:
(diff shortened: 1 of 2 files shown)


pycbc/types/timeseries.py
```diff
@@ -84,6 +83,13 @@ def __init__(self, initial_array, delta_t=None, epoch="", dtype=None, copy=True)
         self._delta_t = delta_t
         self._epoch = epoch
 
+    def epoch_close(self, other):
+        """ Check if the epoch is close enough to allow operations """
+        dt = abs(float(self.start_time) - float(other.start_time))
+        if dt > 1e-7:
+            return False
+        return True
```
titodalcanton, line 91: I suggest condensing this into a one-liner:
```python
return dt <= 1e-7
```

pycbc/types/timeseries.py
```diff
@@ -84,6 +83,13 @@ def __init__(self, initial_array, delta_t=None, epoch="", dtype=None, copy=True)
         self._delta_t = delta_t
         self._epoch = epoch
 
+    def epoch_close(self, other):
+        """ Check if the epoch is close enough to allow operations """
+        dt = abs(float(self.start_time) - float(other.start_time))
```
titodalcanton, line 88: Considering cases where `start_time` is not a float (e.g. `LIGOTimeGPS`), is it better to subtract first and convert to float later? I suspect it will not make a big difference given the tolerance on `dt`, but I am wondering if you thought about it.

pycbc/types/timeseries.py
```diff
@@ -101,14 +107,12 @@ def _typecheck(self, other):
         if isinstance(other, TimeSeries):
             if not self.sample_rate_close(other):
                 raise ValueError('different delta_t')
-            if self._epoch != other._epoch:
+            if not self.epoch_close(other):
                 raise ValueError('different epoch')
```
titodalcanton, line 111: While you are at it, I suggest
```python
raise ValueError('different epoch, {} vs {}'.format(self.start_time, other.start_time))
```

pycbc/types/timeseries.py
```diff
@@ -169,8 +173,18 @@ def time_slice(self, start, end):
...
+        end_idx = (end - self.start_time) * self.sample_rate
+
+        if mode == 'floor':
+            start_idx = int(start_idx)
+            end_idx = int(end_idx)
+        elif mode == 'nearest':
+            start_idx = int(round(start_idx))
+            end_idx = int(round(end_idx))
+        else:
+            raise ValueError("Invalid mode", mode)
```
titodalcanton, line 186:
```python
raise ValueError('Invalid mode "{}"'.format(mode))
```

pycbc/types/timeseries.py
```diff
@@ -844,16 +855,19 @@ def add_into(self, other):
 
         other = other.copy()
         dt = float((other.start_time - self.start_time) * self.sample_rate)
+
         if not dt.is_integer():
-            diff = (dt - _numpy.floor(dt))
+            diff = (dt - _numpy.floor(dt)) * self.delta_t
```
titodalcanton, line 860: This seems to change the meaning of `diff` from samples to seconds, is it intended? Was the previous code incorrect?

titodalcanton:
Apart from the comments above, this looks ok to me. Please rebase on master in order to include the new PyCBC Live test.

pycbc/types/timeseries.py
ahnitz: @titodalcanton Yes, it does. Cyclic_time_shift takes sense, I think this was a bug. I'm not sure how I missed this before.

pycbc/types/timeseries.py
ahnitz: I haven't, I'll make the minor change just in case. Longer term, we should think about if we can abandon LIGOTimeGPS altogether. I've found some cases where it is kind of a pain. Not a high priority though.

pycbc/types/timeseries.py
ahnitz: Ok, suggestion taken.

ahnitz: Also changed the different detla_t message while I was at it.

PR #3255, opened 2020-05-06
Add basic example/test of PyCBC Live

titodalcanton:
First step toward a more serious automatic test of PyCBC Live on Travis. Add an example script that creates hundreds of seconds of simulated strain data and runs a PyCBC Live analysis on it. The analysis is configured similarly to what used in O3, but it is simplified in such a way that it does not require large auxiliary files (namely the phase-time-amplitude histograms and the single-detector trigger fits) and it uses a baby template bank with several thousand templates only. The simulated strain only consists of Gaussian noise and no injections. There are also no sanity checks of the results. The script completes in a few minutes on CIT machines.

Opening this to gauge if we want to immediately add other features, add this as a Travis test right away, both, or neither.


titodalcanton:
Requires #3254.

ahnitz:
I think we should find a different place for the template bank file. I'd like to keep out more binary files from the main pycbc repo (where we care about the history). Maybe pycbc-config would work for that? 

Also, yes, I think it should be added immediately to the test suite. Not needed for this PR, but it would also be good to expose this as part of a documentation page for pycbc live (at least quickstart guide of sorts)

examples/live/run.sh
```diff
@@ -0,0 +1,103 @@
...
+        --frame-duration 32
+}
+
+simulate_strain H1 aLIGOMidLowSensitivityP1200087 1234
+simulate_strain L1 aLIGOMidLowSensitivityP1200087 2345
+simulate_strain V1 AdVEarlyLowSensitivityP1200087 3456
+
+echo -e "\\n\\n>> [`date`] Running PyCBC Live"
+
+mpirun \
```
ahnitz, line 35: I think this is for the intel mpirun api. We'll need to see what Travis supports in this regard. Most likely openmpi  or mpich

titodalcanton: It is actually for Open MPI. We are already installing some MPI version on Travis in order to run the `--help` test, but yes, we will see if this example works with that.

titodalcanton:
Nice, so far it seems to pass Travis. Addressing the bank comment now.

PR #3262, opened 2020-05-10
updated inverse cdf interfaces so we can support correlated parameters

ahnitz:


ahnitz:
@cdcapano I've taken off the work in progress label. I'm happy if you can take a look at this PR now.

ahnitz:
@cdcapano Yes, it's a helper function effectively. I've renamed it and change the interface of cdfinv to use dicts as requested.

PR #3264, opened 2020-05-11
PyCBC Live: fix empty versions in uploads

titodalcanton:
PyCBC Live tries to report the git hash when uploading candidates,
but that is an empty string when using actual releases, so the
version info is lost in production setups. This patch improves
that and also logs the location of the code.


titodalcanton:
Example: https://gracedb-playground.ligo.org/events/T305655/view/

PR #3267, opened 2020-05-11
allow external distributions if you can provide a logpdf and cdfinv function

ahnitz:
This adds a new distribution that gets its critical functions from an external module which you provide. The interface would be

[prior-x+y]
module = mymod
logpdf = function1
cdfinv = function2

Where you can import mymod.function1 and mymod.function2


ahnitz:
requires #3262

ahnitz:
@cdcapano I think this should address your comments.

ahnitz:
@cdcapano Poke

PR #3272, opened 2020-05-13
Add missing cast to float

lpsinger:
`round()` is not defined for instances of `lal.LIGOTimeGPS`.


ahnitz:
@lpsinger This looks like the right solution. I'm somewhat confused that LIGOTimeGPS behaves differently in python2, so noting here in case a similar issue is noticed in the future (should hopefully have stopped using python2 environments soon anyway).

PR #3274, opened 2020-05-14
redo indexing in 'add into' method

ahnitz:
@lpsinger Pointed out the current indexing will produce off by one errors. This patch redoes the indexing to avoid the time_slice method.

### DIFF
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -837,8 +837,8 @@ class TimeSeries(Array):
         fft(tmp, f)
         return f
 
-    def insert(self, other):
-        """Return copy of self with other inserted into it.
+    def inject(self, other):
+        """Return copy of self with other injected into it.
 
         The other vector will be resized and time shifted with sub-sample
         precision before adding. This assumes that one can assume zeros
@@ -888,7 +888,8 @@ class TimeSeries(Array):
         ts = self.copy()
         ts[left:right] += other[oleft:oright]
         return ts
-    add_into = insert # maintain backwards compatibility for now
+
+    add_into = inject  # maintain backwards compatibility for now
 
     @_nocomplex
     def cyclic_time_shift(self, dt):
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -837,7 +837,7 @@ class TimeSeries(Array):
         fft(tmp, f)
         return f
 
-    def add_into(self, other):
+    def insert(self, other):
         """Return copy of self with other inserted into it.
 
         The other vector will be resized and time shifted with sub-sample
@@ -888,6 +888,7 @@ class TimeSeries(Array):
         ts = self.copy()
         ts[left:right] += other[oleft:oright]
         return ts
+    add_into = insert # maintain backwards compatibility for now
 
     @_nocomplex
     def cyclic_time_shift(self, dt):
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -838,14 +838,14 @@ class TimeSeries(Array):
         return f
 
     def add_into(self, other):
-        """Return the sum of the two time series accounting for the time stamp.
+        """Add another time series into this one where they overlap.
 
         The other vector will be resized and time shifted with sub-sample
         precision before adding. This assumes that one can assume zeros
         outside of the original vector range.
         """
         # only handle equal sample rate for now.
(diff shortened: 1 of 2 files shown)


pycbc/types/timeseries.py
```diff
@@ -838,14 +838,14 @@ def to_frequencyseries(self, delta_f=None):
         return f
 
     def add_into(self, other):
-        """Return the sum of the two time series accounting for the time stamp.
+        """Add another time series into this one where they overlap.
 
         The other vector will be resized and time shifted with sub-sample
```
lpsinger, line 843: The docstring should make it clear that this method returns a new object and does not modify self.

Also, the name of the method is counterintuitive: you would think that `a.add_into(b)` would mean that `a` is time shifted and added to `b`, but the method does the opposite.

ahnitz:
> This still isn't right. In pseudocode, here's what I would do:
> 
> 1. Find the integer and fractional parts of the sample offset between the two arrays.
> 2. Apply a Fourier shift by the fractional part.
> 3. Slice by the integer part and add.

Please clarify. This is exactly how I would describe what the code is already doing. It finds the fractional offset, shifts the other vector, and then slices both. Which part do you think differs?

pycbc/types/timeseries.py
lpsinger: The method name `insert` runs the risk of being confused with the `insert` method on builtin Python container types, which behaves very differently. How about `add_from`?

pycbc/types/timeseries.py
```diff
@@ -856,19 +856,39 @@ def add_into(self, other):
...
+            # fd shift to the right
             other = other.cyclic_time_shift(diff)
 
-        ts = self.copy()
-        start = max(other.start_time, self.start_time)
-        end = min(other.end_time, self.end_time)
+        # get indices of other with respect to self
+        # this is already an integer to floating point precission
+        left = float(other.start_time - self.start_time) * self.sample_rate
+        left = int(round(left))
```
lpsinger, line 872: Should be floor here, right?

lpsinger:
> > This still isn't right. In pseudocode, here's what I would do:
> > 
> > 1. Find the integer and fractional parts of the sample offset between the two arrays.
> > 2. Apply a Fourier shift by the fractional part.
> > 3. Slice by the integer part and add.
> 
> Please clarify. This is exactly how I would describe what the code is already doing. It finds the fractional offset, shifts the other vector, and then slices both. Which part do you think differs?

I see. Yes, now it does. The problem, I think, is just the `round()`.

I suggest adding unit tests.

pycbc/types/timeseries.py
ahnitz: @lpsinger No here no. The time stamps should already be aligned so they'll differ by exactly an integer up to floating point precision. If you do floor, then you can have an off-by-one issue if the floating point number just slightly drifts below the correct integer. (i.e. 1 - 1e7)

pycbc/types/timeseries.py
```diff
@@ -856,19 +856,39 @@ def add_into(self, other):
         other = other.copy()
         dt = float((other.start_time - self.start_time) * self.sample_rate)
 
+        # This coaligns other to the time stepping of self
         if not dt.is_integer():
             diff = (dt - _numpy.floor(dt)) * self.delta_t
```
lpsinger, line 861: Might be cleaner to use `divmod` here so that you get the integer and fractional parts at the same time.

pycbc/types/timeseries.py
lpsinger: I see.

pycbc/types/timeseries.py
ahnitz: Changed to 'inject' as this is a synonym and accurate about what it is doing.

pycbc/types/timeseries.py
ahnitz: I'd prefer to keep as is.

ahnitz:
@lpsinger @spxiwh I've added a simple unittest for this and renamed to 'inject'. It's still aliased for backwards compatibility.

ahnitz:
@spxiwh @lpsinger Poke

lpsinger:
I'm sorry, I'm swamped as usual. I'll do my best to do a review later today.

ahnitz:
@lpsinger I'll merge this for now as it is clearly better. We can re-open a new PR if this doesn't meet your needs.

PR #3281, opened 2020-05-18
Pycbc geom nonspin python3

SumeetKul:
Removed python 2 has_key() dict syntax to make the pycbc_geom_nonspinbank module compatible with python3.


ahnitz:
I'm assigning this to @spxiwh as this is primarily his code.

SumeetKul:
Thanks! Looks ready to merge.

PR #3286, opened 2020-05-20
MultiIFO workflow singles development

GarethDaviesGW:
Various changes for development of the multiifo singles search (this has been edited from when the PR was first put up)

Complete rework of `pycbc_multiifo_sngls_findtrigs`:
- not to use ReadByTemplate and behaves more efficiently
- Add ability to cut on SNR and reduced chisq before dealing with triggers (add other cut options?)
- Remove IFAR calculation
- Basically, this code now simply calculates the ranking statistic of triggers which meet the cuts above

New executable, `pycbc_multiifo_sngls_pastro`:
- Used after findtrigs statistic calculation
- Takes in triggers from full_data sngls_statmap files and calculate p_astro according to either callister or truncated_shelf methods
  - It is possible to give multiple files here in order to analyse multiple chunks at the same time
  - takes in single-detector statistic of injection triggers and calculates the signal distribution in order to allow the callister/truncated_shelf calculation
  - also possible to give multiple files here
  - requires hdfinjfind file to also be given in order to calculate weightings for the injection triggers
- Outputs "foreground" and "background":
  - foreground is all triggers from the sngls_statmap files
  - background contains only triggers in coincident time which are not in foreground veto time. So the background triggers are also entirely within the foreground
  - The N loudest triggers can be removed from the background using --remove-n-loudest (after cutting down to coinc time and removing foreground ttriggers). This prevents any signals from coinc time which were not seen as coincs from affecting the background estimate
    - This is something under discussion, as it removes triggers from signals in non-coinc time so may cause a bias

Change to pycbc_page_snglinfo:
- Allow a pastro file to be given in order to report a p_astro on the results pages for single triggers (not when using n_loudest, mostly because this was taking a long time to test)
- Change headings from ("Stat")  to the full name (e.g. newsnr_sgveto), and the newly calculated single-detector statistic is called "Single-detector statistic". These needed some more clarity in order to distinguish them from one another

Changes in stat.py:
 - Provide a default for single_multiifo, to use the single-detector statistic as used in the generation of the coinc statistic
 - Add in SNR ** - 4 rescaling to single-detector statistic
 - allow more than one fits file to be given when only one ifo is under consideration (means that all fits files can be given but then not necessarily used)

I envision that this will be implemented by performing an initial pastro calculation during a standard offline run, which gives a bit more info in results pages. This analysis would then be re-run using a longer period of time (e.g. a full analysis run) in order to get better background estimates and a larger number of expected signals

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_sngls_findtrigs
+++ b/bin/hdfcoinc/pycbc_multiifo_sngls_findtrigs
@@ -1,20 +1,25 @@
 #!/usr/bin/env python
-import h5py, argparse, logging, numpy, numpy.random
+import argparse, logging, numpy as np
 from ligo.segments import infinity
 from pycbc.events import veto, coinc, stat
 import pycbc.conversions as conv
 import pycbc.version
-from numpy.random import seed, shuffle
-from pycbc.io.hdf import ReadByTemplate
+from pycbc import io
+from pycbc.events import trigger_fits as trfits
+
+ifar_model_options = ['n_louder', 'exp_fit', 'gaussian_kde']
 
 parser = argparse.ArgumentParser()
-parser.add_argument("--verbose", action="count")
-parser.add_argument("--version", action="version", version=pycbc.version.git_verbose_msg)
-parser.add_argument("--veto-files", nargs='*', action='append', default=[],
+parser.add_argument("--verbose", action='count')
+parser.add_argument("--version", action='version',
+                    version=pycbc.version.git_verbose_msg)
+parser.add_argument("--veto-file", default=None,
                     help="Optional veto file. Triggers within veto segments "
-                         "contained in the file are ignored")
-parser.add_argument("--segment-name", nargs='*', action='append', default=[],
-                    help="Optional, name of veto segment in veto file")
+                         "contained in the file are ignored. Required if "
+                         "--segment-name given.")
+parser.add_argument("--segment-name", default=None,
+                    help="Optional, name of veto segment in veto file. "
+                         "Required if --veto-file given.")
 parser.add_argument("--trigger-file",type=str,
                     help="File containing single-detector triggers")
 parser.add_argument("--template-bank", required=True,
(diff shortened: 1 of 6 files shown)


GarethDaviesGW:
This is ready for review now I think

GarethDaviesGW:
so I never actually requested a review before.... :facepalm:

GarethDaviesGW:
Example plot output from L1 in O2 chunk 20:
image

This uses a single chunk of data for the background estimate, and so due to the short amount of data only has ~1.25 expected signals, leading to a maximum p_astro of just over 0.5. The trigger with statistic just below 30 is the GW170809 trigger

bin/hdfcoinc/pycbc_multiifo_sngls_findtrigs
```diff
@@ -1,20 +1,24 @@
 #!/usr/bin/env python
```
spxiwh, line 1: A general comment: This is definitely going to conflict with #3345. My take is the easiest approach is to merge that first, and then rebase this on top of that.

bin/hdfcoinc/pycbc_multiifo_sngls_findtrigs
```diff
@@ -31,59 +35,126 @@ parser.add_argument("--statistic-keywords", nargs='*',
...
+parser.add_argument('--inj-filter-file',
+                    help="If analysing injections, use a HDF_INJFIND file to "
+                         "select the trigger ids of found injections")
+parser.add_argument('--inj-filter-ifar', default=0, type=float,
+                    help="If analysing injections, set a minimum IFAR for "
+                         "deciding if an injection is found. Default=0")
 parser.add_argument("--output-file",
                     help="File to store the candidate triggers")
-parser.add_argument("--batch-singles", default=5000, type=int,
-                    help="Number of single triggers to process at once")
```
spxiwh, line 46: Why is this being removed? It was added to manage memory problems with a single-template search (e.g. lensing)

bin/hdfcoinc/pycbc_multiifo_sngls_findtrigs
```diff
@@ -31,59 +35,126 @@ parser.add_argument("--statistic-keywords", nargs='*',
                          "the statistic class when it is initialized. Should "
                          "be given in format --statistic-keywords "
                          "KWARG1:VALUE1 KWARG2:VALUE2 KWARG3:VALUE3 ...")
-parser.add_argument("--template-fraction-range", default="0/1",
-                    help="Optional, analyze only part of template bank. Format"
-                         " PART/NUM_PARTS")
```
spxiwh, line 36: Wait. We definitely need these 3 options no? (Well maybe we don't need the randomize one, but it does help balance load)

bin/hdfcoinc/pycbc_multiifo_sngls_findtrigs
```diff
@@ -31,59 +35,126 @@ parser.add_argument("--statistic-keywords", nargs='*',
...
+                         'Default=do not cluster')
+parser.add_argument('--reduced-chisq-cut', type=float,
+                    help='Only consider triggers below given reduced '
+                         'chisquared.')
+parser.add_argument('--inj-filter-file',
+                    help="If analysing injections, use a HDF_INJFIND file to "
+                         "select the trigger ids of found injections")
+parser.add_argument('--inj-filter-ifar', default=0, type=float,
+                    help="If analysing injections, set a minimum IFAR for "
+                         "deciding if an injection is found. Default=0")
```
spxiwh, line 52: This code runs independent of injection information currently. I can't really parse what this file is used for? As the output of this code is needed to produce the HDF_INJFIND, this seems to introduce a cyclical dependency

spxiwh:
@GarethDaviesGW I'm getting bogged down by the significant changes to the coincidence code here. Can you explain why a patch to do singles p-astro is making such a significance change to this code? I'm also worried that necessary functionality seems to be being removed from that code.

spxiwh:
Ah, wait, that's not the coinc findtrigs executable, is it ... I think I am failing at reviewing this.

GarethDaviesGW:
I've removed the injfind-based cuts in the statistic calculation stage, as discussed on slack. I was actually performing these anyway in the pastro calculation stage as well, so this was fairly inefficient

GarethDaviesGW:
Outline in the opening comment is updated, and should be accurate now

pycbc/events/stat.py
```diff
@@ -878,6 +876,9 @@ def single(self, trigs):
         stat = thresh - (logr_n / self.alpharef)
         return numpy.array(stat, ndmin=1, dtype=numpy.float32)
 
+    def single_multiifo(self, s):
+        return s[1]['snglstat']
```
GarethDaviesGW, line 880: Note that I should probably use the self.single_increasing part of https://github.com/gwastro/pycbc/pull/3345 here

GarethDaviesGW:
CC issues are the same as in #3345, where there is repeated code from earlier commits

GarethDaviesGW:
I'll rebase then merge and when Alex/Tom are less busy (and I have finished moving) we can look over and discuss any required changes

PR #3288, opened 2020-05-21
Add support for dynesty's dynamic nested sampler

prayush:
This change enables the inference module to use the dynamic nested sampler provided (and actively developed) by JS in [`dynesty`](https://github.com/joshspeagle/dynesty).

### DIFF
--- a/pycbc/inference/sampler/dynesty.py
+++ b/pycbc/inference/sampler/dynesty.py
@@ -86,10 +86,18 @@ class DynestySampler(BaseSampler):
         self.names = model.sampling_params
         self.ndim = len(model.sampling_params)
         self.checkpoint_file = None
-        self._sampler = dynesty.NestedSampler(log_likelihood_call,
-                                              prior_call, self.ndim,
-                                              nlive=self.nlive,
-                                              pool=pool, **kwargs)
+        if self.nlive < 0:
+            # Interpret a negative input value for the number of live points
+            # (which is clearly an invalid input in all senses)
+            # as the desire to dynamically determine that number
+            self._sampler = dynesty.DynamicNestedSampler(log_likelihood_call,
+                                                         prior_call, self.ndim,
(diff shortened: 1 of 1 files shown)


pycbc/inference/sampler/dynesty.py
```diff
@@ -86,10 +86,18 @@ def __init__(self, model, nlive, nprocesses=1,
         self.names = model.sampling_params
         self.ndim = len(model.sampling_params)
         self.checkpoint_file = None
-        self._sampler = dynesty.NestedSampler(log_likelihood_call,
-                                              prior_call, self.ndim,
-                                              nlive=self.nlive,
-                                              pool=pool, **kwargs)
+        if self.nlive < 0:
```
ahnitz, line 89: @prayush Maybe we can just check if nlive is set or not? That might be more obvious to a user.

ahnitz: @prayush I think this is a straightforward request, but means a couple more lines to change i..e 

1) The __ini__ signature (nlive should not be required -> nlive = None
2) The from_config method should default to None for nlive if the option doesn't exist. 
3) the check here should just see if nlive is set. 

Make sense?

prayush: @ahnitz - I thought about this choice for a while, i.e. whether to have `nlive` uninitialized or to have it initialized to impossible values to indicate the use of dynamic nested sampling. My reason for doing the latter is that it *always* makes clear to the user (say, a new user) what nested sampler exactly they are using since there is no mistaking that a negative value of nlive must have special meaning. An *uninitialized* will make many a user think they are using some reasonable default that the developers have set for `nlive`. So I would argue for keeping the current mechanism. I don't feel too strongly about this though, so I'm happy to be dissuaded!

ahnitz:
@sum33it Did you have any further comments?

PR #3291, opened 2020-05-29
avoid hash for pycbc live

ahnitz:
This may address #3263 for the case of pycbc live. I'm still worried about offline though, so I'll think about that too.

### DIFF
--- a/pycbc/waveform/bank.py
+++ b/pycbc/waveform/bank.py
@@ -490,10 +490,10 @@ class LiveFilterBank(TemplateBank):
         super(LiveFilterBank, self).__init__(filename, approximant=approximant,
                 parameters=parameters, **kwds)
         self.ensure_standard_filter_columns(low_frequency_cutoff=low_frequency_cutoff)
-        self.hash_lookup = {}
+        self.param_lookup = {}
         for i, p in enumerate(self.table):
-            hash_value =  hash((p.mass1, p.mass2, p.spin1z, p.spin2z))
-            self.hash_lookup[hash_value] = i
+            key =  (p.mass1, p.mass2, p.spin1z, p.spin2z)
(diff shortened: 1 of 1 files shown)


ahnitz:
Note: The reverse lookup needs to be possible as the template bank is *sliced*. This means the index for __getitem__ won't match the template_id from the original bank. However, we want htilde.id to be the *original* template id, even if the bank was sliced.

titodalcanton:
This seems a very reasonable solution. I will try it out and report.

pycbc/waveform/bank.py
```diff
@@ -490,10 +490,10 @@ def __init__(self, filename, sample_rate, minimum_buffer,
         super(LiveFilterBank, self).__init__(filename, approximant=approximant,
                 parameters=parameters, **kwds)
         self.ensure_standard_filter_columns(low_frequency_cutoff=low_frequency_cutoff)
-        self.hash_lookup = {}
+        self.param_lookup = {}
         for i, p in enumerate(self.table):
-            hash_value =  hash((p.mass1, p.mass2, p.spin1z, p.spin2z))
-            self.hash_lookup[hash_value] = i
+            key =  (p.mass1, p.mass2, p.spin1z, p.spin2z)
+            self.param_lookup[key] = i
```
titodalcanton, line 496: Just for good practice against crazy banks, could you add this:
```
assert(key not in self.param_lookup)
```

ahnitz: good idea

titodalcanton:
After this change, I can no longer reproduce issue #3263 with the test script. Please address my comment above and this will be ready for approval.

ahnitz:
@titodalcanton Ok, done

PR #3293, opened 2020-06-02
Use get_newsnr_sgveto_psdvar where appropriate

GarethDaviesGW:
Use get_newsnr_sgveto_psdvar in exp_fit_sg_fgbg_norm_psdvar and exp_fit_sg_fgbg_norm_psdvar_bbh statistics

Note that 2ogc and 2ogcbbh named statistics still use get_newsnr_sgveto_psdvar_scaled and get_newsnr_sgveto_psdvar_scaled_threshold respectively


GarethDaviesGW:
codeclimate is complaining about some similar code - can't really change this because it does similar things and we want it to do similar things

PR #3294, opened 2020-06-02
Add injections to PyCBC Live example

maxtrevor:
Added injections into the live example. Not sure what other features may be wanted.


examples/live/check_results.py
```diff
@@ -100,4 +100,9 @@ def close(a, b):
     log.error('No triggers found in %s', ', '.join(missing))
     fail = True
 
+if fail:
+    print('Test Failed')
```
titodalcanton, line 104: Please use `log.error()` instead of `print()` here.

examples/live/check_results.py
```diff
@@ -100,4 +100,9 @@ def close(a, b):
...
+else:
+    print('Test Passed')
```
titodalcanton, line 106: Please use `log.info()` instead of `print()` here.

examples/live/run.sh
```diff
@@ -34,6 +34,29 @@ pycbc_hdf5_splitbank \
...
+  --approximant SEOBNRv4 \
+  --mass1 25.0 \
+  --mass2 25.0 \
+  --inclination 0.0 \
+  --taper TAPER_START \
+  --waveform-low-frequency-cutoff 10 \
+  --geocentric-end-time 1272790440 \
+  --instruments H1 L1 V1 \
+  --low-frequency-cutoff 10 \
+  --sample-rate H1:16384 L1:16384 \
```
titodalcanton, line 53: Is a `V1:16384` also needed?

examples/live/run.sh
```diff
@@ -34,6 +34,29 @@ pycbc_hdf5_splitbank \
...
+  --gps-end-time 1272790500 \
+  --gps-start-time 1272790000 \
+  --psd-model H1:aLIGOMidLowSensitivityP1200087 L1:aLIGOMidLowSensitivityP1200087 V1:AdVEarlyLowSensitivityP1200087
+  
+gunzip hwinjcbc_1272790426.xml.gz
```
titodalcanton, line 58: I think PyCBC's injection code can read gzipped XML, so this should not be needed.

examples/live/run.sh
```diff
@@ -47,7 +70,8 @@ function simulate_strain { # detector PSD_model random_seed
         --sample-rate 16384 \
         --low-frequency-cutoff 10 \
         --channel-name $1:SIMULATED_STRAIN \
-        --frame-duration 32
+        --frame-duration 32 \
+        --injection-file hwinjcbc_1272790426.xml
```
titodalcanton, line 74: I would prefer not having to hardcode the XML file name, but I see `pycbc_generate_hwinj` is not super flexible in controlling the name, so I think this is ok for now.

titodalcanton:
One currently missing thing which we need is a check that the injection was found. This can be done in different ways with more or less precision: one could go from a basic check that a `coinc-*.xml.gz` file is present, to a full inspection of the content of the file. I would start with a basic but not completely trivial check:

1. Find all `coinc-*.xml.gz` files.
2. If there are no such files, then the injection was missed and the check should fail.
3. If there are "too many" (say > 10) then something else is probably wrong and the check should also fail.
4. For each such file, read its `sngl_inspiral` table and get the approximate merger time, the network SNR and the template parameters, and check that they are consistent with the injected values.

One slightly tricky aspect is that the injected masses and spins you chose do not necessarily match any of the templates in the bank used by the test, and the bank is definitely not effectual. The injection is loud enough that it will probably still be found, but it becomes harder to predict which template will find it. For this reason I would inject masses and spins that correspond exactly to one of the templates, making it easier to predict which template will trigger. This is a bit more complicated though: you have to read the bank and extract the masses and spins from one of the templates.

titodalcanton:
Another comment is that it would be nice to have one high-mass (GW150914-like) and one low-mass (GW170817-like) injection, but this may be a bit tricky to do with the limited template bank we are using. Maybe for a new PR...

maxtrevor:
> One currently missing thing which we need is a check that the injection was found. This can be done in different ways with more or less precision: one could go from a basic check that a `coinc-*.xml.gz` file is present, to a full inspection of the content of the file. I would start with a basic but not completely trivial check:
> 
> 1. Find all `coinc-*.xml.gz` files.
> 2. If there are no such files, then the injection was missed and the check should fail.
> 3. If there are "too many" (say > 10) then something else is probably wrong and the check should also fail.
> 4. For each such file, read its `sngl_inspiral` table and get the approximate merger time, the network SNR and the template parameters, and check that they are consistent with the injected values.
> 
> One slightly tricky aspect is that the injected masses and spins you chose do not necessarily match any of the templates in the bank used by the test, and the bank is definitely not effectual. The injection is loud enough that it will probably still be found, but it becomes harder to predict which template will find it. For this reason I would inject masses and spins that correspond exactly to one of the templates, making it easier to predict which template will trigger. This is a bit more complicated though: you have to read the bank and extract the masses and spins from one of the templates.

Ok I made the other changes you requested, I will start working on this now. The bank definitely did pick up this injection with a very different set of masses, ~ 72 and 6 instead of both being 25. I will start working on this stuff now

maxtrevor:
Made the requested changes

maxtrevor:
Updated the structure of the code. check_results is now a method inside result_checker.py, and takes the snr, time, masses and spins of the injection as arguments to test if results are consistent with the injection.

maxtrevor:
I see that this apparently failed a travis check? No clue why

titodalcanton:
Thanks for the changes @maxtrevor. I am not sure what the Travis failure is about, but seems orthogonal to your change. Hopefully it will clear itself at your next commit, if not we will followup with others. Now a couple comments on your changes.

First, please let's not rename `check_results.py`, as that will erase it previous git history and make the diff much longer than it has to be. It is fine to just add the new functionality to `check_results.py`.

The approach of turning `check_results.py` into a callable function which takes the injection parameters is ok for a single injection, however what will happen when we will have two? I propose the following algorithm instead (this would go into `check_results.py`):
* Read the list of injections from the XML file created by `pycbc_generate_hwinj`.
* For each injection, check if there is a `coinc-*.xml.gz` file within, say, a second of the injection time (you can pick any detector here, or the geocentric time, as it is just an approximate check).
* If there is no file, the injection was missed, so set the `fail` flag, print an error, and continue.
* If there is a file, read it and check the template parameters against the injection parameters.
One unfortunate drawback of this approach is that you will not be able to check the injection SNR precisely, because `pycbc_generate_hwinj` does not make it available in the `sim_inspiral` table (I think). You could use `pycbc_optimal_snr` to calculate the expected SNR, but instead I suggest just an even more basic check: simply verify that the SNRs in all detectors are positive, and that the network SNR is greater than 10. We will then make sure to always inject signals louder than that.

maxtrevor:
The name change is reverted, I didn't like having a function named check_results inside a file of the same name but I will live with it.

I previously had something similar to your proposal for how to deal with multiple injections. I was caught by the snr issue, and also wasn't sure how to deal with multiple injections that way. Looping over the injections makes sense and then checking if there a coinc trigger at a nearby time.

Instead I was thinking of making this callable method with the injection parameters and calling it once for each injection . It could also do the check for a coinc trigger with the parameters it is passed. 

Two questions though: is pycbc_generate_hwinj able to produce multiple injections in a single xml file? And do we plan on adding multiple injections to one time series, or using a seperate time series for each injection?

titodalcanton:
> The name change is reverted, I didn't like having a function named check_results inside a file of the same name but I will live with it.

That is ok and actually quite common for simple modules (e.g. `from signal import signal`, `from tqdm import tqdm`).

> Instead I was thinking of making this callable method with the injection parameters and calling it once for each injection . It could also do the check for a coinc trigger with the parameters it is passed.

A problem with that though is that, for every injection, it would also repeat the checks of the single-detector triggers, which is not super slow, but unnecessary. I also find it a bit weird to have this functionality split between Bash and Python when you can do everything in Python instead, but this is more my own personal taste :)

> Two questions though: is pycbc_generate_hwinj able to produce multiple injections in a single xml file? And do we plan on adding multiple injections to one time series, or using a seperate time series for each injection?

No, I think `pycbc_generate_hwinj` is meant to generate a single injection. It is indeed not really great for this particular application, but I thought it was simple enough for a start. Eventually I would replace it with either a call to `pycbc_generate_hwinj` (which I know is cool, but never used myself) or with a custom little Python script that draws masses and spins from the template bank, so you do not have to hardcode those parameters in `run.sh`. If you feel like making any of these changes right away, please do!

Multiple injections should all go into the same XML file, if that is what your second question was asking. `pycbc_condition_strain` will create a single time series with all injections specified by the XML file.

maxtrevor:
@titodalcanton I made the mistake of editing my reply while you were replying to it

maxtrevor:
Also you mentioned pycbc_generate_hwinj as an alternative to itself? If I'm understanding you correctly, pycbc_condition_strain can take one xml file with multiple injections, and produce strain data with all of those injections in one strain time series. My question then is how do I produce that one xml file containing multiple injections?

maxtrevor:
I'm going to finish getting a working build for a single injection with some of your suggestions. I'd like to pull that to the master branch once it's done, then start a new pull request for dealing with multiple injections.

titodalcanton:
> Also you mentioned pycbc_generate_hwinj as an alternative to itself?

Sorry, that was a mistake, I meant to say `pycbc_generate_hwinj` could be replaced with `pycbc_create_injections`.

> If I'm understanding you correctly, pycbc_condition_strain can take one xml file with multiple injections, and produce strain data with all of those injections in one strain time series.

Correct, that is how injections typically work in the search.

> My question then is how do I produce that one xml file containing multiple injections?

I know three basic options: `lalapps_inspinj` (still used by PyCBC offline analyses, legacy code, not very flexible nor recommended), `pycbc_create_injections` (modern code which will create an HDF injection file, not XML, but that should be fine) or your own custom code.

As I said earlier, we do not have to worry about multiple injections right now, but whatever we put in place here should allow us to do so in a future PR without completely rewriting the test.

examples/live/check_results.py
```diff
@@ -1,103 +1,206 @@
-#!/usr/bin/env python
-
```
titodalcanton, line 2: Please keep these lines.

examples/live/run.sh
```diff
@@ -122,10 +164,8 @@ python -m mpi4py `which pycbc_live` \
 --ifar-double-followup-threshold 0.0001 \
 --ifar-upload-threshold 0.0001 \
 --round-start-time 4 \
---size-override 20 \
 --start-time $gps_start_time \
 --end-time $gps_end_time
 
 echo -e "\\n\\n>> [`date`] Checking results"
-
-./check_results.py
```
titodalcanton, line 131: Please keep this as is.

examples/live/check_results_old.py
```diff
@@ -0,0 +1,179 @@
+import sys
```
titodalcanton, line 1: No need to keep old copies of files around, git's history does that for us.

maxtrevor:
Alright I have a working version with a single injection. Would like to get this pulled to the master before I start working on adding more injections. I chose to include the template bank in the files because not having to generate it every time reduces runtime substantially. Hope everything is acceptable.

examples/live/check_results.py
```diff
@@ -6,98 +6,202 @@
...
+                                    trigfp, detector)
+                              single_fail = True
+
+      # check that triggers were produced in all detectors
+      if detectors_with_trigs != tested_detectors:
+            missing = sorted(tested_detectors - detectors_with_trigs)
+            log.error('No triggers found in %s', ', '.join(missing))
+            single_fail = True
+      
+      if single_fail: log.info('Single Trigger Test Failed')
```
titodalcanton, line 105: Suggest
```python
if single_fail:
    log.error('Single trigger test failed')
```

examples/live/check_results.py
```diff
@@ -6,98 +6,202 @@
...
+      coinc_trig_paths = sorted(glob.glob('output/coinc*.xml.gz'))
+      l=len(coinc_trig_paths)
+      if l==0:
+            log.error('No coincident triggers detected')
+            coinc_fail = True
+      elif l>=10: 
+            log.error('Too many coincident triggers detected')
+            coinc_fail = True
+      else: 
+            log.info(str(l)+' coincident trigger(s) detected')
```
titodalcanton, line 120: The logging module lets you use format strings like this:
```python
log.info('%d coincident trigger(s) detected', l)
```
This is the recommended usage.

examples/live/check_results.py
```diff
@@ -6,98 +6,202 @@
...
+            mass2=sngl_inspiral_table.get_column('mass2')[0]    
+            spin1z=sngl_inspiral_table.get_column('spin1z')[0] 
+            spin2z=sngl_inspiral_table.get_column('spin2z')[0] 
+            log.info('IFO SNRs: '+str(snr_list))
+            log.info('Network SNR: '+str(network_snr))
+            log.info('IFO End Times: '+str(end_time)) 
+            log.info('Mass 1: '+str(mass1))
+            log.info('Mass 2: '+str(mass2))
+            log.info('Spin1z: '+str(spin1z))
+            log.info('Spin2z: '+str(spin2z))
```
titodalcanton, line 151: I think it is not necessary to print all these info, but up to you. If you *do* keep these lines, please use the string format as suggested above (`%f` for float variables).

examples/live/check_results.py
```diff
@@ -6,98 +6,202 @@
...
+                  coinc_fail = True
+                  log.error('Trigger spin1z does not match injection spin1z')
+            if not close(spin2z, inj_spin2z, 5e-7):
+                  coinc_fail = True
+                  log.error('Trigger spin2z does not match injection spin1z')   
+            if not close(network_snr, inj_snr, 0.1):
+                  coinc_fail = True
+                  log.error('Network SNR test failed')
+            
+      if coinc_fail: log.info('Coincident Trigger Test Failed')
```
titodalcanton, line 174: Similar comment as above,
```python
if coinc_fail:
    log.error('Coincident trigger test failed')
```

examples/live/check_results.py
```diff
@@ -6,98 +6,202 @@
...
+      log.error('No inj_xml found when checking results')
+      fail=True
+else:
+      path_to_inj_xml = inj_xmls[0]
+      log.info('injection xml found')
+      inj_doc = ligolw_utils.load_filename(
+            path_to_inj_xml, False, contenthandler=LIGOLWContentHandler)
+      inj_table = lsctables.SnglInspiralTable.get_table(inj_doc)
+   
+      tested_detectors = get_tested_detectors(inj_table)
```
titodalcanton, line 196: I see the reasoning behind this, but inferring the list of detectors from the injections is not going to be obvious. I suggest keeping the original definition of `tested_detectors`. The point is that check_results.py has to be consistent with whatever set of detectors we decided to simulate, regardless of the injections, and I think it is ok to hardcode this both in `check_results.py` and `run.sh`.

examples/live/generate_injections.sh
```diff
@@ -0,0 +1,29 @@
+#major work in progress
```
titodalcanton, line 1: I would not split this off to an extra file, I think having it in `run.sh` is fine. Up to you.

I also suggest removing the "major work in progress" comment.

examples/live/run.sh
```diff
@@ -10,49 +10,91 @@ export HDF5_USE_FILE_LOCKING="FALSE"
...
+
+
+if [[ "$(echo ./hwinjcbc*.xml.gz)" != "./hwinjcbc*.xml.gz" ]]
+then
+    inj_names='ls ./hwinjcbc*.xml.gz'
+    n_inj=${#inj_names[@]}
+    if [[ $n_inj = 1  ]]
+    then echo -e "\\n\\n>> [`date`] Pre-existing Inj Xml Found"
+    elif [[ $n_inj > 1 ]]
+    then echo -e "\\n\\n>> [`date`] Too many injection xmls, only first will be used"
```
titodalcanton, line 54: What motivates this check? We are never going to generate more than one injection file.

titodalcanton:
Thanks @maxtrevor, I think this is getting close to approval. The check for missing/found injections is still not quite the algorithm I suggested in the comments above, but let's go ahead and we will fix it in another PR. Please address my remaining comments above, and remove the template bank file from the PR. I also originally committed it, but it was decided not to. I think we generally frown upon binary files in the repo.

examples/live/check_results.py
maxtrevor: This was prep for the possibility of not always using HLV (I've been talking with Peter about doing a quick investigation into the effect of filtering out noise around 60 Hz on the SNR we pick up in H and L). My idea is that at some point in the future, the code will first check what detectors the injection is specified for, then only simulate strain for those detectors.

examples/live/generate_injections.sh
maxtrevor: This is also prep for the future. I think for doing multiple injections the generate_injections file will wind up getting much more complicated. It may wind up getting switch to a python file instead of a .sh . Just wanted to have this file ready to minimize the amount of changes that need to be done in run.sh for switching to multiple injections.

examples/live/run.sh
maxtrevor: The code is currently set up that it can take an injection xml as input instead of generating it's own by checking for a pre-existing xml injection file. I just wanted to make it clear that if for some reason there are two xml injection files sitting in the directory only one will be used.

maxtrevor:
> Thanks @maxtrevor, I think this is getting close to approval. The check for missing/found injections is still not quite the algorithm I suggested in the comments above, but let's go ahead and we will fix it in another PR. Please address my remaining comments above, and remove the template bank file from the PR. I also originally committed it, but it was decided not to. I think we generally frown upon binary files in the repo.

I will delete the template bank. The check for missing/found injections is different because I couldn't figure out the syntax for looping over multiple injections without actually having produced an file with multiple injections but I plan to switching to what you suggested once we are working with multiple injections.

examples/live/check_results.py
maxtrevor: This is prep for a possible future where the injection doesn't exactly match the template and we want to check which template is picking up the injection. I can however add an option to not print these out unless its wanted.

maxtrevor:
Think I've made all the required changes. A couple of things I responded to your comments to explain why I did things a certain way. Basically I'm planning to expand this test for a lot more functionality in the future.

examples/live/check_results.py
titodalcanton: Well, let's not mix other investigations into this, this script is strictly to provide a rapid sanity check of PyCBC Live and a simple example. I think hardcoding to HLV is ok for the foreseeable future of this script. It would of course be good to test other detector combinations in a more thorough test, but I doubt we want to run such a long test on Travis at every commit.

examples/live/generate_injections.sh
titodalcanton: I predict it will get *less* complicated once we use `pycbc_create_injections`, but ok, we can leave this as is.

examples/live/check_results.py
maxtrevor: Ok I will remove this from here. I'm probably going to wind up building a more complicated test system at some point but ya it probably won't be for travis.

maxtrevor:
Ok maybe now its ready.

examples/live/generate_injections.sh
maxtrevor: The problem is from what I've seen, pycbc_create_injections does NOT have a way to exactly specify multiple injections. It makes randomized injections from within a specified parameter space. So if we want to continue having the injections exactly match templates in the bank, we will not be able to use pycbc_create_injections

examples/live/run.sh
maxtrevor: Went ahead and made the suggested changes, checking for too many injection files will be a feature for a larger non-travis test in the future.

maxtrevor:
Ok I think its finally really done.

PR #3295, opened 2020-06-03
Prepare for release

GarethDaviesGW:


GarethDaviesGW:
will do - I've got a PR there as well which I was seeing if that would be approved before tagging config - I'll see how quick travis completes!

PR #3296, opened 2020-06-03
Expose GraceDB search tag as a PyCBC Live option

titodalcanton:
Needed for public early-warning test.


GarethDaviesGW:
I think the codebeat issue can be ignored (it's annoying that it doesn't just fail, but always says pending even if I has found issues)

titodalcanton:
This has now been tested using the `examples/live` (thanks to bits of @maxtrevor's PR) and it works as expected.

PR #3298, opened 2020-06-03
Fixed start frequency of FilterBankSkyMax

KoustavChandra:
Added statements to enable FilterBankSkyMax code to use low-frequency cutoff of the bank itself.

### DIFF
--- a/pycbc/waveform/bank.py
+++ b/pycbc/waveform/bank.py
@@ -829,11 +829,9 @@ class FilterBankSkyMax(TemplateBank):
 
         # Find the start frequency, if variable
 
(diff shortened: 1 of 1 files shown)


GarethDaviesGW:
To prevent the duplicated code warning you're seeing - it looks like the two bits of code are just safety wrappers of the function find_variable_start_frequency:
https://github.com/gwastro/pycbc/blob/master/pycbc/waveform/bank.py#L777

These are the only uses of this function in pycbc, so I'd suggest to move the checks to inside the function, this will need an extra variable adding to the function

PR #3300, opened 2020-06-03
add precognition to pycbc live

ahnitz:
This implements the following features to enable pre-merger detection with pycbc live

1) A new template called 'PreTaylorF2', where the 'f_final' argument is taken as the end of the waveform and the waveform is time-shifted from this point so the end is at zero.

2) Tracking of the inccurred time offset, which is propagated to the name of the coinc file and the coinc end time. I've left the other times alone as that might be better for understanding the event (i.e. the SNR is X up to that time). 

Only testing I've done so far is on 170817, which this produces trigger for as expected ~ 10 seconds before merger (which is exactly what I build a quick test bank for, so I wouldn't expect any other time).

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -242,8 +242,9 @@ class LiveEventManager(object):
             live_ifos = [ifo for ifo in fud if 'snr_series' in fud[ifo]]
 
             time_offset = 0
-            if 'foreground/time_offset' in coinc_results:
-                time_offset = coinc_results['foreground/time_offset']
+            rtoff = 'foreground/{}/time_offset'.format(coinc_ifos[0])
+            if rtoff in coinc_results:
+                time_offset = coinc_results[rtoff]
             event = SingleCoincForGraceDB(live_ifos, coinc_results, bank=bank,
                                           psds=psds, followup_data=fud,
                                           low_frequency_cutoff=f_low,
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -779,6 +779,7 @@ with ctx:
     while data_end() < args.end_time:
         t1 = time()
         logging.info('%s: Analyzing from %s', evnt.rank, data_end())
+
         results = {}
         evnt.live_detectors = set()
 
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -241,20 +241,15 @@ class LiveEventManager(object):
 
             live_ifos = [ifo for ifo in fud if 'snr_series' in fud[ifo]]
 
-            time_offset = 0
-            rtoff = 'foreground/{}/time_offset'.format(coinc_ifos[0])
-            if rtoff in coinc_results:
-                time_offset = coinc_results[rtoff]
             event = SingleCoincForGraceDB(live_ifos, coinc_results, bank=bank,
                                           psds=psds, followup_data=fud,
                                           low_frequency_cutoff=f_low,
                                           channel_names=args.channel_name,
                                           mc_area_args=self.mc_area_args,
-                                          time_offset=time_offset)
+                                          )
 
-            end_time = int(coinc_results['foreground/%s/end_time'
-                                         % coinc_ifos[0]])
-            fname = 'coinc-{}-{}.xml.gz'.format(end_time + time_offset, pycbc.random_string(6))
+            fname = 'coinc-{}-{}.xml.gz'.format(event.merger_time,
+                                                pycbc.random_string(6))
             fname = os.path.join(self.path, fname)
             logging.info('Coincident candidate! Saving as %s', fname)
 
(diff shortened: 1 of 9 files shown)


bin/pycbc_live
```diff
@@ -241,15 +241,20 @@ class LiveEventManager(object):
...
+            rtoff = 'foreground/{}/time_offset'.format(coinc_ifos[0])
+            if rtoff in coinc_results:
+                time_offset = coinc_results[rtoff]
             event = SingleCoincForGraceDB(live_ifos, coinc_results, bank=bank,
                                           psds=psds, followup_data=fud,
                                           low_frequency_cutoff=f_low,
                                           channel_names=args.channel_name,
-                                          mc_area_args=self.mc_area_args)
+                                          mc_area_args=self.mc_area_args,
+                                          time_offset=time_offset)
```
titodalcanton, line 253: Given that the above change is repeated twice (for coincs and singles) would it make sense to move all of that into the constructor of `SingleCoincForGraceDB`?

bin/pycbc_live
```diff
@@ -241,15 +241,20 @@ class LiveEventManager(object):
...
 
             end_time = int(coinc_results['foreground/%s/end_time'
                                          % coinc_ifos[0]])
-            fname = 'coinc-{}-{}.xml.gz'.format(end_time, pycbc.random_string(6))
+            fname = 'coinc-{}-{}.xml.gz'.format(end_time + time_offset, pycbc.random_string(6))
```
titodalcanton, line 257: Ah right, we also need the time offset down here. However, I propose giving `SingleCoincForGraceDB` a property called `merger_time_estimate`. This would allow us to replace the above *two* lines with this single line:
```python
fname = 'coinc-{:.0f}-{}.xml.gz'.format(event.merger_time_estimate,
                                        pycbc.random_string(6))
```

pycbc/events/single.py
```diff
@@ -153,7 +153,7 @@ def check(self, trigs, data_reader):
 
     def calculate_ifar(self, newsnr, duration):
         if self.fit_info['fixed_ifar']:
-            return self.fit_info['fixed_ifar']
+            return self.fit_info['fixed_ifar'][self.ifo]
```
titodalcanton, line 156: @GarethDaviesGW should probably comment on these changes to `pycbc/events/single.py`.

pycbc/waveform/bank.py
```diff
@@ -398,6 +398,10 @@ def end_frequency(self, index):
         """ Return the end frequency of the waveform at the given index value
         """
         from pycbc.waveform.waveform import props
+        try:
+            return self.table[index].f_final
+        except:
+            pass
```
titodalcanton, line 404: I suggest using a `if hasattr` instead of catching the exception, or being more specific about which exception should be ignored.

pycbc/waveform/bank.py
```diff
@@ -598,6 +605,9 @@ def get_template(self, index, min_buffer=None):
         htilde.approximant = approximant
         htilde.end_frequency = f_end
 
+        if time_offset:
```
titodalcanton, line 608: ...`is not None`?

pycbc/events/single.py
ahnitz: @GarethDaviesGW This change is needed, as the code is currently non-functional for setting a fixed ifar. I think when the if statement at the top of the code was added, it was forgotten that this is dict that is keyed by the ifo.

pycbc/waveform/bank.py
ahnitz: Makes sense, change made.

pycbc/waveform/bank.py
ahnitz: I don't that's necessary.

pycbc/waveform/premerger.py
```diff
@@ -0,0 +1,33 @@
...
+    from pycbc.waveform import get_fd_waveform
+    from pycbc.waveform.spa_tmplt import spa_length_in_time
+
+    p.pop('approximant')
+    hp, hc = get_fd_waveform(approximant="TaylorF2", **p)
+
+    removed = spa_length_in_time(mass1=p['mass1'],
+                                 mass2=p['mass2'],
+                                 f_lower=p['f_final'],
+                                 phase_order=-1)
```
titodalcanton, line 17: Probably not an issue for the current application, but do we have a function similar to `spa_length_in_time()` that includes PN spin effects in PyCBC?

bin/pycbc_live
ahnitz: Ok, moving to a property of SingleCoincForGracedb and consolidating.

pycbc/waveform/premerger.py
ahnitz: Hmm, I don't think so.

pycbc/events/single.py
GarethDaviesGW: Yes, sorry I must have missed that before. I think when moving code into the new structure I missed that it was set in from_cli before. I'm happy for it to be here instead

titodalcanton:
@ahnitz I would like to wait until #3294 is merged so we use it test that the standard analysis works, sounds good?

ahnitz:
@titodalcanton Yes, that's fine.

titodalcanton:
@ahnitz please rebase on master to trigger the new test.

ahnitz:
@titodalcanton Ok, rebased, so the new test should run in Travis.

pycbc/strain/strain.py
```diff
@@ -1374,7 +1374,7 @@ def __init__(self, frame_src, channel_name, start_time,
         self.psd = None
         self.psds = {}
 
-        strain_len = int(sample_rate * self.raw_buffer.delta_t * len(self.raw_buffer))
+        strain_len = int(max_buffer * self.sample_rate)
```
titodalcanton, line 1377: Could you explain this change and the one above?

ahnitz: The change above means that the raw strain (i.e. what directly comes out of the frame files) is only stored for the most recent 32 seconds instead of the entire length of what we store the *conditioned* strain for. We never actually need this buffer to go back to (or at least we don't at the moment as we only use the last chunks worth). The long buffer which was at 16834 took a *long* time to do the roll operation, which became a significant cost when you use shorter chunk sizes. 

The change highlighted here is to make sure that the *conditioned* strain is still the correct length as we do need that to be the full buffer worth, however it is at a much lower sample rate now and so isn't such a cost anyway to manipulate.

PR #3307, opened 2020-06-09
Add epsie proposals

cdcapano:
This adds support for more jump proposals in epsie, which were added in version 0.5. Specifically, this adds support for:
 * `BoundedNormal`: uses a truncated normal distribution for jump proposals. Useful for posteriors that peak at or near a prior boundary (like mass ratio).
 * `Angular`: for a parameters that are cyclic on [0, 2pi), like phase. Works by using a truncated normal distribution with bounds at +/- pi at every point, which are then wrapped around to produce the jump proposal.
 * `Discrete`: produces discretized jump proposals for integer parameters. A discretized version of the normal distribution is used for jumps.
 * `BoundedDiscrete`: produces discretized jump proposals with bounds. Basically, a discrete form of `BoundedNormal`.

For each one of these, an adaptive version is also included.

This also correctly fixes the issue that PR #3306 temporarily fixed, by bumping the minimum epsie version to 0.5.

Some tests on the 2D normal analytic model: [link](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/prs/add_epsie_proposals/).


cdcapano:
@ahnitz I checked that an epsie dependence hasn't slipped back by uninstalling epsie, then running the test inference scripts in tools. Didn't have any trouble, so merging.

PR #3308, opened 2020-06-09
Undo renormalize loop in JointDistribution

cdcapano:
The changes in PR #3214 made initializing the `JointDistribution` class incredibly slow if any constraints are used.  A renormalization constant is estimated by evaluating the constraints over 1 million points when the distribution is initialized. Previously, this was done by applying the constraints to a vector of samples. In #3214, that was changed to looping over each point, evaluating the constraints one point at a time. This has made initializing the joint distribution go from taking less than a second, to several minutes, if not hours, if any constraints are used. This makes using constraints in PE jobs or when creating injections virtually impossible.

This patch fixes that by going back to evaluating the constraints on the vector of samples (see lines 122-124 of `distributions/joint.py` in this patch). The joint distribution now initializes in under a second with constraints included.

The basic issue that @dbkeitel  pointed out in #3214  was that many of the distributions are not vectorized. This is still an issue that should be fixed. However, the custom constraint class was vectorized as long as the function passed to it as the constraint arg was. The example constraint function in the `JointDistribution` docs is not, which led to the `ValueError` that @dbkeitel described in #3214. The fix for the example function is that it should just return `params['mass1'] + params['mass2'] < 30.` rather than `True if params['mass1'] + params['mass2'] < 30. else False`. I've updated the doc string to reflect this.

The other constraint classes in `distributions/constraints.py` are not vectorized, which would lead to a `ValueError` if they are passed to the `JointDistribution` with the changes in this patch. However, those classes really aren't used for anything. They also have awkward initialization syntax (you need to pass them a `constraint_arg` since they inherit from the `Constraint`, but that isn't used for anything) as they had been written prior to when the custom constraint class was introduced. The custom constraint class is really the only thing that's used, so I've removed the other classes. If someone wants to reintroduce more complicated constraints that would require a class, they can do so in the future.

### DIFF
--- a/pycbc/distributions/constraints.py
+++ b/pycbc/distributions/constraints.py
@@ -28,37 +28,15 @@ class Constraint(object):
     def __init__(self, constraint_arg, transforms=None, **kwargs):
         self.constraint_arg = constraint_arg
         self.transforms = transforms
-        self._scratch = None
         for kwarg in kwargs.keys():
             setattr(self, kwarg, kwargs[kwarg])
-
-    def _getscratch(self, params):
-        """Gets FieldArray scratch space.
-
-        FieldArray creation can be slow. This tries to speed up successive
-        calls by creating a scratch FieldArray space based on the given
-        parameters.
-        """
-        if self._scratch is None or (set(params.keys()) != 
-                                     set(self._scratch.fieldnames)):
-            self._scratch = record.FieldArray.from_kwargs(**params)
-        else:
-            for p in params:
-                try:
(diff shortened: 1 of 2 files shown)


ahnitz:
@cdcapano And if the classes (or something similar) were re-added, it could be done with vectorization from the start.

dbkeitel:
Sorry @cdcapano for not coming up with the easy fix to the custom/example functions and thanks for generalizing/improving this feature again! I think it should be easy to adapt my use case accordingly, but will let you know if I run into any hurdles again.

PR #3309, opened 2020-06-11
Update pegasus template for 'local' site 

josh-willis:
There will shortly be changes to some LDG HTCondor configurations where jobs will no longer match to the local pool (even if the grid site is 'nogrid' and `+IS_GLIDEIN` is `False`) unless they also set `+flock_local = True` in their submit files.  This PR makes the necessary change. It should be safe to use even on sites that do not have that requirement, as they will ignore the setting of that optional condor variable.

@spxiwh @GarethDaviesGW This change will likely be needed during at least some of the upcoming offline runs, so in relatively short order it should be incorporated into a release.  But it can wait on any other changes that might be needed if there will be another point release before the production runs start.  Otherwise, we will probably need a point release just for this.

### DIFF
--- a/pycbc/workflow/pegasus_files/local-site-template.xml
+++ b/pycbc/workflow/pegasus_files/local-site-template.xml
@@ -13,3 +13,4 @@
     <profile namespace="condor" key="when_to_transfer_output">ON_EXIT_OR_EVICT</profile>
     <profile namespace="condor" key="+DESIRED_Sites">&quot;nogrid&quot;</profile>
     <profile namespace="condor" key="+IS_GLIDEIN">&quot;False&quot;</profile>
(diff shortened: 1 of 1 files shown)


josh-willis:
@spxiwh so to clarify, my understanding is that this will be required on nodes that also submit to OSG, and not necessarily on nodes that only submit to the local condor pool.  I believe the rationale is to ensure that users are intending that their jobs run locally in the vanilla condor universe, and without Glidein pilots and the environment they setup. Supposedly the plan had been that this should have been in place for some time, and there is now an effort to standardize the condor configuration on the different nodes throughout LSC/Virgo/KAGRA that will submit to OSG.

I'm going to go ahead and merge this, because I have no idea why the codebeat analysis of a one-line change has been pending for over twelve hours.

PR #3311, opened 2020-06-14
Allow different burn-in iteration and ACL for independent MCMC chains

cdcapano:
This separates ensemble MCMC (in which chains share information) methods from generic MCMC methods (in which chains are independent of each other) in the io, burn in, and sampler module. The primary purpose of this is to allow epsie (or any other future MCMC sampler) to use different burn in iterations and ACLs for different chains, resulting in a different number of effective samples being collected from different chains. This also lays ground work for implementing a transdimensional MCMC in the future.

This turned out to be way more work than I anticipated when I started it. I didn't realize how much ensemble MCMC-specific things were ingrained in the base MCMC classes. Hence the large patch. In terms of actual differences a user would notice, the output for emcee(_pt) files are largely unchanged. The only differences are burn-in information is now stored as datasets in the output file's `sampler_info` group as opposed to the `sampler_info`'s `attrs` (this is something I've been wanting to do for awhile), and the autocorrelation time is stored (again, as datasets rather than attrs) rather than autocorrleation length. I switched to the former because ACTs are independent of the thinning interval used in the file. However, both ACTs and ACLs can now be easily accessed with file attributes.

I've done several tests to check everything is working. I ran on the 1D normal likelihood with various levels of burn-in and checkpointing turned on. I did this for [epsie](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/prs/epsie_separate_acls/analytic_test/), [emcee_pt](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/prs/epsie_separate_acls/emcee_pt-analytic_test/), and [emcee](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/prs/epsie_separate_acls/emcee-analytic_test/) (click on the links to see the results). I also ran a workflow on GW150914 using `emcee_pt`, along with various settings for epsie. Results are [here](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/scratch/prs/epsie_separate_acls/py37_gw150914_8perchain/). Importantly, the `emcee_pt` results are what we expect.

Here is a summary of the changes that are made in each file:

`burn_in.py`:
 * Formerly there was a single BurnIn class for single-temperature samplers, and another one that inherited from it for parallel-tempered samplers. Now there are two separate classes for single-temperatures (and corresponding versions for parallel tempered), an `MCMCBurnInTests` for MCMCs with independent chains (epsie) and an `EnsembleMCMCBurnInTests` for ensemble MCMCs (emcee, emcee_pt). Both of these inherit from a `BaseBurnInTests` which implements the common functions. The difference between the MCMC and Ensemble classes is the former applies tests separately to each chain, returning an array of burn-in iterations and test results that can be different for each chain. The ensemble class, on the other hand, applies burn in tests to all chains, returning only a single test value and iteration for the entire ensemble (as it currently does). In other words, the ensemble test are only burned in if all the chains are burned in, whereas the independent ones may burn in separately.
 * The separate Ensemble and MCMC classes also allow different tests to be implemented for each. For example, the ensemble class has the ks-test (compares the distribution of walkers to what it was at the halfway point), but the MCMC class does not. I plan to take use this to add a Gelman-Rubin test to the MCMC class in the future, which would not be implemented for the Ensemble class.
 * The way the burn in classes store data is more simple. Previously, all data was stored to `burn_in_data` which was a dictionary keyed by tests that pointed to dictionaries. The sub-dictionaries had to have `burn_in_iteration` and `is_burned_in` in them, but could also have other test-specific things. Now, `burn_in_iteration` and `is_burned_in` for each test are separate attributes (called `test_(burn_in_iteration|is_burned_in)`) which are simple dictionaries that store test names -> result. Any extra information a test returns is stored as `test_aux_info`.
 * Writing burn-in results has been moved from a method of `BaseMCMC` io classes to methods of the burn in classes. I did this because the different burn-in classes may write different amount of information.

`io/base_hdf.py`:
 * thin start/interval/end are no longer stored to the file's attrs. Instead, these are attributes of the class. The base version just return 0, 1, None, respectively, but can be overridden by inheriting classes (which is done by the MCMC classes; see below). This was done to allow the thin attributes to be arrays, and also reduces the amount of bookkeeping that needs to be done in the file. For example, previously the burn-in index had to be copied to the thin-start attr in the file (and both needed to be updated whenver a new burn-in test was applied, or file thinning was done). Now, the class just handle that on the fly.
 * The `get_slice` method was simplified. It no longer tries to silently pull from the file's thin start/interval/end attributes. Instead, these need to be supplied to the function.
 * A `write_data` method is added to reduce code redundancy. This takes care of figuring out if a dataset already exists in a file when writing, overwriting the data if it does, and creating the dataset (along with the group) if it does not.

`io/base_mcmc.py`:
 * The `MCMCMetadataIO` class is now split into three classes: a `CommonMCMCMetadatIO`, `MCMCMetadataIO`, and `EnsembleMCMCMetadataIO`. `Common` implements all the methods common to both independent-chain MCMCs and ensemble MCMCs. `MCMCMetadataIO` implements methods specific to independent chain MCMCs; ditto for `Ensemble`.
 * Both `MCMCMetadataIO`  and `EnsembleMetadataIO` override the base file class's `thin_(start|interval|end)`. In both cases, the `thin_start` attribute returns the burn-in index if a burn-in test was done (0 otherwise), and `thin_interval` attribute returns the ACL if it was calculated (1 otherwise). This is what I meant about thinning settings handled on the fly above, removing these values being stored twice in the files. The difference between `Ensemble` and `MCMC` is the former returns integers for these values, whereas `MCMC` returns arrays giving the values for each chain. 
 * `CommonMCMCMetadataIO` now has `act`, `raw_acts`, `acl`, and `raw_acls` attributes. Setting the `*act` attributes writes the autocorrelation times to the file. The `raw_*` are the times per parameter and temperature.  The `act` is the maximum over the parameters and temperatures. For ensemble MCMCs, this is a single value for all the ensemble, whereas for independent-chain MCMCs this is an array for each chain. However, the file class doesn't need to worry about the distinction, since the results are stored as a dataset in either case. The `*acl` are derived on the fly from stored ACTs by dividing the ACTs by the file's `thinned_by` attribute.
 * `CommonMCMCMetadataIO` has both `nwalkers` and `nchains` attributes. These are just aliases of each other. Ensemble samplers call their chains "walkers" whereas a collection of MCMC chains are just "chains." Since we've coded up "walkers" in a bunch of places, I decided to just alias the two.
 * The `SingleTempMCMCIO` class has been removed. Formerly, this implemented the `write_samples` and `read_raw_samples` functions. Now, `write_samples` and `read_raw_samples` are implemented as stand-alone functions, which file classes explicitly call, rather than inheriting from a class. This was done because the `write_samples` functions are common between ensemble and independent MCMC samplers, but the `read_raw_samples` are different. Simplest to allow each sampler's file class pick which functions to use on their own.
 * The `SingleTempMCMCIO`'s `read_raw_samples` is now `ensemble_read_raw_samples`, as this is for single-temperature Ensemble MCMCs (emcee). An equivalent function for single-temperature independent-chain MCMCs is not implemented, since epsie is parallel-tempered. If a single-temperature single-chain MCMC is added in the future, it would be straight forward to add a `read_raw_samples` function for it.
 * Some of the things that were done inside of `read_raw_samples` -- getting the indices of iterations to load, getting the walker indices -- are now broken out to separate private functions, to reduce code redundancy. 
 * An `nsamples_per_chain` function is added to compute the number of samples you'll get back from an array given a thin start, thin interval, and number of iterations. The function is vectorized, so can be used with a single chain, or an array of chains.
 
`io/base_multitemper.py`:
 * `MultiTemperedMCMCMetadataIO` has been renamed`CommonMultiTemperedMCMCMetadataIO`, and inherits from `CommonMCMCMetadataIO`. This adds temps to the common MCMC metadata.
 * As was done in `base_mcmc`, `MultiTemperedMCMCIO` has been removed. That class's `write_samples ` and `read_raw_samples` for parallel-tempered MCMCs are now the stand-alone `write_samples` and `ensemble_read_raw_samples`. The former does not have the `ensemble_` prefix because it works with both types. The latter has the `ensemble_` because it is meant for ensemble parallel-tempered samplers.
 * A `read_raw_samples` stand-alone function is added for parallel-tempered MCMCs with independent chains. This is the function that epsie uses for reading samples now. It can handle chains having different burn-in iterations and ACLs. If an unflattened array is requested, it will use NaNs as fill values for chains that return fewer samples.
 * Getting the temperature indices to load is moved to a stand-alone private function, to reduce code redundancy between the two read functions.

`io/emcee.py`:
 * `EmceeFile` now inherits from `EnsembleMCMCMetadataIO`, `CommonMCMCMetadataIO`, and `BaseSamplerFile`.
 * The `write_samples` and `read_raw_samples` functions are now defined here, calling `base_mcmc.write_samples` and `base_mcmc.ensemble_read_raw_samples`, respectively.

`io/emcee_pt.py`:
 * `EmceePTFile` now inherits from `EnsembleMCMCMetadataIO`, `CommonMultiTemperedMetadataIO`, and `BaseSamplerFile`.
 * The `write_samples` and `read_raw_samples` functions are now defined here, calling `base_multitemper.write_samples` and `base_multitemper.ensemble_read_raw_samples`, respectively.

`io/epsie.py`:
 * `EpsieFile` now inherits from `MCMCMetadataIO`, `CommonMultiTemperedMetadataIO`, and `BaseSamplerFile`.
 * The `write_samples` and `read_raw_samples` functions are now defined here, calling `base_multitemper.write_samples` and `base_multitemper.read_raw_samples`, respectively.

`sampler/base_mcmc.py`:
 * Ensemble-MCMC specific attributes have been split off from the `BaseMCMC` class and into a support class called `EnsembleSupport`.
 * The `nwalkers` attribute in `BaseMCMC` has been renamed `nchains`. `EnsembleSupport` implements a `nwalkers` attribute, which is an alias of `nchains`.
 * The `acl` attribute is now an `abstractmethod` in `BaseMCMC`. The purpose of this is to convert the `raw_acls` into an ACL. `EnsembleSupport` implements this method for ensemble samplers, which is just the max over all ACLs.
 * `effective_nsamples` has been made an abstract method of `BaseMCMC` since the way this is calculated depends on if it is an ensemble (in which you have the same number of effective samples from all chains) or an independent set of chains (in which the samples from each chain can be different). The former `effective_nsamples` method has been moved to `EnsembleSupport` (although it now calls `io/base_mcmc/nsamples_in_chain` to reduce redundancy).
 * The `MCMCAutocorrSupport` class has been removed. Instead, `compute_acl` and `compute_acf` are moved to stand-alone functions. Since these were specific to Ensemble samplers, the new functions have the `ensemble_` prefix.

`sampler/base_multitemper.py`:
 * The `MultiTemperedMCMCAutocorrSupport` class has been removed. As was done in `base_mcmc`, its `compute_acl` and `compute_acf` methods have been moved to stand-alone functions, with the `ensemble_` prefix.
 * Stand-alone `compute_acl` and `compute_acf` functions are added for independent-chain MCMCs. These differ from the ensemble versions in that they do not average their chains together before computing the ACL. Instead, an ACL is calculated separately for each chain, and an array of ACLs is returned.
 * An `acl_from_raw_acls` function is added to get an ACL from a dictionary of raw ACLs for independent chains. This works by taking the maximum over the parameters and temperatures for each chain (but does not maximize over chains). An equivalent is not needed for `ensemble` samplers, since the simple taking the max over everything that is implemented in `EnsembleSupport.acl` works for both single-temperature and parallel-tempered ensemble MCMCs.

`sampler/emcee.py`:
 * The `EmceeEnsembleSampler` sampler now inherits from `EnsembleSupport`, `BaseMCMC`, and `BaseSampler`. Its `burn_in_class` is set to `EnsembleMCMCBurnInTests`.
 * Its `compute_acl` and `compute_acf` functions are implemented here, which call `sampler/base_mcmc.ensemble_compute_ac(l|f)`.

`sampler/emcee_pt.py`:
 * `EmceePTSampler` now inherits from `MultiTemperedSupport`, `EnsembleSupport`, `BaseMCMC`, and `BaseSampler`. Its `burn_in_class` is set to `EnsembleMultiTemperedMCMCBurnInTests`.
 * Its `compute_acl` and `compute_acf` functions are implemented here, which call `sampler/base_multitemper.ensemble_compute_ac(l|f)`.

`sampler/epsie.py`:
 * `EpsieSampler` now inherits from `MultiTemperedSupport`, `BaseMCMC`, and `BaseSampler`. Its `burn_in_class` is `MCMCBurnInTests` (same name as before, though different class; see changes to `burn_in.py`, above).
 * Its `compute_acl` and `compute_acf` function call `sampler/base_multitemper.compute_ac(l|f)`.
 * The `acl` property (which, because this does not use `EnsembleSupport` is inherited as an `abstractmethod` from `BaseMCMC`). Calls `base_multitemper.acl_from_acls`.
 * Implements `effective_nsamples` (again, inherited as an abstract method) which computes the total number of effective samples from all the chains.

`bin/inference/pycbc_inference_plot_acl`:
 * Update accessing the ACL from the file for emcee and emcee_pt for the new syntax. This executable will not work with epsie due to these changes (nor will plot samples or plot acf), but I'll address that in another patch.

### DIFF
--- a/bin/inference/pycbc_inference_plot_acl
+++ b/bin/inference/pycbc_inference_plot_acl
@@ -78,14 +78,11 @@ for param_name in parameters:
     logging.info("Plotting autocorrelation times")
     plt.hist(acls, opts.bins, histtype="step", label=labels[param_name])
 
-# get the file's acl
-fpacl = fp.thinned_by * fp[fp.sampler_group].attrs['acl']
-
 plt.xlabel("Autocorrelation time")
 plt.ylabel(r'Number of walkers')
 
-# plot autocorrelation length saved in hdf file
-plt.axvline(fpacl, linestyle='--')
+# plot autocorrelation time saved in hdf file
+plt.axvline(fp.act, linestyle='--')
 plt.legend()
 
 # save figure with meta-data
--- a/pycbc/inference/burn_in.py
+++ b/pycbc/inference/burn_in.py
@@ -29,6 +29,8 @@ have burned in.
 from __future__ import division
 
 import numpy
+from six import add_metaclass
+from abc import ABCMeta, abstractmethod
 from scipy.stats import ks_2samp
 
 from pycbc.io.record import get_vars_from_arg
@@ -149,6 +151,91 @@ def posterior_step(logposts, dim):
     return idx
 
 
+def nacl(nsamples, acls, nacls=5):
+    """Burn in based on ACL.
+
+    This applies the following test to determine burn in:
+
+    1. The first half of the chain is ignored.
+
+    2. An ACL is calculated from the second half.
+
+    3. If ``nacls`` times the ACL is < the length of the chain / 2,
+       the chain is considered to be burned in at the half-way point.
+
+    Parameters
+    ----------
+    nsamples : int
+        The number of samples of in the chain(s).
+    acls : dict
+        Dictionary of parameter -> ACL(s). The ACLs for each parameter may
+        be an integer or an array of integers (for multiple chains).
+    nacls : int, optional
+        The number of ACLs the chain(s) must have gone past the halfway point
+        in order to be considered burned in. Default is 5.
+
+    Returns
+    -------
+    dict
+        Dictionary of parameter -> boolean(s) indicating if the chain(s) pass
+        the test. If an array of values was provided for the acls, the values
+        will be arrays of booleans.
+    """
+    kstart = int(nsamples / 2.)
+    return {param: (nacls * acl) < kstart for (param, acl) in acls.items()}
+
+
+def evaluate_tests(burn_in_test, data):
+    """Evaluates burn in data from multiple tests.
+
+    The iteration to use for burn-in depends on the logic in the burn-in
+    test string. For example, if the test was 'max_posterior | nacl' and
+    max_posterior burned-in at iteration 5000 while nacl burned in at
+    iteration 6000, we'd want to use 5000 as the burn-in iteration.
+    However, if the test was 'max_posterior & nacl', we'd want to use
+    6000 as the burn-in iteration. This function handles all cases by
+    doing the following: first, take the collection of burn in iterations
+    from all the burn in tests that were applied.  Next, cycle over the
+    iterations in increasing order, checking which tests have burned in
+    by that point. Then evaluate the burn-in string at that point to see
+    if it passes, and if so, what the iteration is. The first point that
+    the test passes is used as the burn-in iteration.
+
+    Parameters
+    ----------
+    burn_in_test : str
+        The test to apply; e.g., ``'max_posterior & nacl'``.
+    data : dict of dict
+        Dictionary of test name -> dictionary that has format
+        ``{'is_burned_in': bool, 'burn_in_iteration': int}``, giving the
+        burn-in status and iteration of each test, respectively. Data for all
+        tests used by ``burn_in_test`` must be included.
+
+    Returns
+    -------
+    is_burned_in : bool
+        Whether or not the data passes all burn in tests.
+    burn_in_iteration :
+        The iteration at which all the tests pass. If the tests did not all
+        pass (``is_burned_in`` is false), then returns
+        :py:data:`NOT_BURNED_IN_ITER`.
+    """
+    burn_in_iters = numpy.unique([data[t]['burn_in_iteration']
+                                  for t in data])
+    burn_in_iters.sort()
+    for ii in burn_in_iters:
+        test_results = {t: (data[t]['is_burned_in'] &
+                            0 <= data[t]['burn_in_iteration'] <= ii)
+                        for t in data}
+        is_burned_in = eval(burn_in_test, {"__builtins__": None},
+                            test_results)
+        if is_burned_in:
+            break
+    if not is_burned_in:
+        ii = NOT_BURNED_IN_ITER
+    return is_burned_in, ii
+
+
 #
 # =============================================================================
 #
@@ -158,11 +245,11 @@ def posterior_step(logposts, dim):
 #
 
 
-class MCMCBurnInTests(object):
-    """Provides methods for estimating burn-in of an ensemble MCMC."""
-
+@add_metaclass(ABCMeta)
+class BaseBurnInTests(object):
+    """Base class for burn in tests."""
     available_tests = ('halfchain', 'min_iterations', 'max_posterior',
-                       'posterior_step', 'nacl', 'ks_test',
+                       'posterior_step', 'nacl',
                        )
 
     def __init__(self, sampler, burn_in_test, **kwargs):
@@ -177,8 +264,6 @@ class MCMCBurnInTests(object):
         # Arguments specific to each test...
         # for nacl:
         self._nacls = int(kwargs.pop('nacls', 5))
-        # for kstest:
-        self._ksthreshold = float(kwargs.pop('ks_threshold', 0.9))
         # for max_posterior and posterior_step
         self._ndim = int(kwargs.pop('ndim', len(sampler.variable_params)))
         # for min iterations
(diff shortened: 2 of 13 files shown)


ahnitz:
@cdcapano Can you rebase from master?

cdcapano:
@ahnitz Rebased

Issue #3302, opened 2020-06-05
Hash handling in offline templates


GarethDaviesGW:
Template hashes can clash in the python3 version of the hash() function

@titodalcanton noticed this issue in PyCBC Live #3263 and it was solved in Live by #3291 - we need to ensure that it is fixed in offline before being able to run in python 3.

Possible options are to follow the currently used live example and compare (possible float comparison issues) or use hashlib module  (_may_ still clash but v. unlikely, I think the library is more stable than the in-built function)

ahnitz:
One can't follow the live model, as it didn't actually need to use hashes to check for common templates between different files later on. Each process still had access to the whole bank (and hence what the ids meant) so it could avoid the hashes entirely. 

I think the options are
1) hashlib (as you suggest, might be an easiest extension)
2) get rid of hashes altogether, and put an explicit id column in the banks (or at least sub-banks). This would require a few more places to change though, but would get rid of the hash issue entirely.

ahnitz:
I believe @GarethDaviesGW has now addressed this.

PR #3314, opened 2020-06-16
Offline hash handling

GarethDaviesGW:
Makes a function which detects whether using python 2 or python 3 and decides whether to use native hash (unsafe in python 3) or hashlib module (utilises int.to_bytes which is unavailable in python2)

Solves issue #3302

Note that I haven't been able to test using bank2hdf, as I haven't been able to get that to work even with the original function! However this works with Tito's test script

### DIFF
--- a/pycbc/waveform/bank.py
+++ b/pycbc/waveform/bank.py
@@ -38,6 +38,7 @@ import pycbc.waveform.compress
 from pycbc import DYN_RANGE_FAC
 from pycbc.types import FrequencySeries, zeros
 import pycbc.io
+import six, hashlib
 
 def sigma_cached(self, psd):
     """ Cache sigma calculate for use in tandem with the FilterBank class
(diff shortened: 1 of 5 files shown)


titodalcanton:
As done in PyCBC Live, I recommend an explicit check for a hash collision using `assert` or `raise` as appropriate.

pycbc/waveform/bank.py
```diff
@@ -178,6 +179,26 @@ def parse_approximant_arg(approximant_arg, warray):
     """
     return warray.parse_boolargs(boolargs_from_apprxstr(approximant_arg))[0]
 
+def list_to_hash(list_to_be_hashed):
```
ahnitz, line 182: To avoid confusion, I don't think you should call this 'list_XXX'. Hashing a list doesn't actually work. It really does have to be a tuple.

GarethDaviesGW:
I managed to get bank2hdf working - which caused an issue as the 64-byte integer being made wouldn't fit into a native numpy datatype. I've updated which hash algorithm is being used to make the hash.

Note that though a 8-byte (64 bit) hash _looks_ unsafe, it's actually slightly better than the in-built hash which uses 62 bits

I've changed the name as suggest by Alex, and added to raise an error as suggested by Tito

ahnitz:
@GarethDaviesGW I don't really see any problems with this patch. Have you checked that nothing crops up with a small workflow in python 3?

GarethDaviesGW:
I havent done a small-workflow test yet
I'll try to set one up

GarethDaviesGW:
Note that I haven't forgotten this, just taking a while to get it through the workflow on CIT

GarethDaviesGW:
I've had to add a couple of fixes for using python3 in multiifo_statmap_inj and cluster_coincs_multiifo, and one bugfix in pycbc_ifar_catalog for the case where there are no hierarchical removals

I've added these here just because they came up when running the test workflow for the hash handling change

GarethDaviesGW:
(note I am re-requesting a review for the other things I had to add)

ahnitz:
@GarethDaviesGW The last changes seem very straightforward, so I am happy to approve.

PR #3316, opened 2020-06-16
add nested sampler base class and checkpoint dynesty

prayush:
This is currently for code sharing. See commit messages for detail.


ahnitz:
@prayush I'm closing this in favor of #3354. There are some subtle issues with checkpointing dynesty that has to be done in a different approach than this patch. For one, you can't actually interrupt the code at arbitrary times and checkpoint, as you can end up in an invalid internal state which will result in incorrect output. At the moment, we can only checkpoint the fixed live point sampler. There are some issues with how the DynamicSampler is implemented (addressed in https://github.com/joshspeagle/dynesty/pull/188) that make it difficult to meaningfully checkpoint. (we could force a checkpoint but it will cause the code to run in a very non-optimal way, so it isn't worth it).

PR #3317, opened 2020-06-17
LISA_detector

SSingh087:
file for LISA detector having methods for

- getting orbit of the detector
- Coordinate Transformation between ICRS frame and GCRS frame
- Distance from the LISA detector to a detector on earth at a given time
- Distance of the LISA detector from any location


pycbc/detector.py
```diff
@@ -322,3 +304,96 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
+        self.location = np.array([x,y,z])
+
+        return self.location
+
+    def plot_orbit(self):
+        from mpl_toolkits.mplot3d import Axes3D
+        import matplotlib.pyplot as plt
+        dec_center = []
+        for i in range(2000):
+          dec_center.append(self.get_pos(self.reference_time + i).mean(axis = 1))
```
SSingh087, line 350: tried passing GPS time in loop to continuously obtain positions

pycbc/detector.py
```diff
@@ -322,3 +304,96 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
+        ax.set_xlabel('X axis (AU)')
+        ax.set_ylabel('Y axis (AU)')
+        ax.set_zlabel('Z axis (AU)')
+
+    def time_delay_from_location(self, other_location, right_ascension,
+                                 declination, t_gps):
+        dec_loc = self.get_pos(t_gps)
+        """signal = coordinates.SkyCoord(ra = right_ascension, dec = declination,
+                                          unit = u.rad, frame = 'gcrs').transform_to('icrs')"""
+
```
SSingh087, line 371: Assuming the user is passing ICRS frame values(s).

pycbc/detector.py
```diff
@@ -322,3 +304,96 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
+        dx = np.array([other_location[0] - self.location[0],
+                       other_location[1] - self.location[1],
+                       other_location[2] - self.location[2]])
+
+        """ra_angle = self.gmst_estimate(t_gps) - right_ascension"""
```
SSingh087, line 376: Need this equation when in ICRS frame ?

pycbc/detector.py
```diff
@@ -322,3 +304,96 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
 def effective_distance(distance, inclination, f_plus, f_cross):
     return distance / np.sqrt( ( 1 + np.cos( inclination )**2 )**2 / 4 * f_plus**2 + np.cos( inclination )**2 * f_cross**2 )
 
+    """ LISA class  """
+
+class LISA(object):
+    def __init__(self, kappa, _lambda_, reference_time = 1126259462.0):
+        self.reference_time = reference_time
+        self.kappa = kappa
```
ahnitz, line 312: Why are you taking these in as arguments? (You also set them to zero in some methods)

pycbc/detector.py
```diff
@@ -322,3 +304,96 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
 def effective_distance(distance, inclination, f_plus, f_cross):
     return distance / np.sqrt( ( 1 + np.cos( inclination )**2 )**2 / 4 * f_plus**2 + np.cos( inclination )**2 * f_cross**2 )
 
+    """ LISA class  """
+
+class LISA(object):
+    def __init__(self, kappa, _lambda_, reference_time = 1126259462.0):
```
ahnitz, line 310: You don't need a reference time here. the purpose was a bit different for the ground based detector.

pycbc/detector.py
```diff
@@ -322,3 +304,96 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
+        self.reference_time = reference_time
+        self.kappa = kappa
+        self._lambda_ = _lambda_
+
+    def get_pos(self, t_gps):
+        if t_gps is None:
```
ahnitz, line 316: the gps time should be allowed to be 'none'.

pycbc/detector.py
ahnitz: @SSingh087 I don't think we need a method just for the plot. It would be better to have a stand-a-lone script of yours which calls into the pycbc to get the LISADetector object to then make the plot.

pycbc/detector.py
ahnitz: It would be good to not assume this. The best way to do that would be to add a method for both the regular detector class *and* this one that is somthing like "icrs_position". You can then have this method just call that on both self and other to get their positions and then calculate the distance.  Similarly, this method could be identical to the one for the regular detector class, so it works either way (i.e. you can ask the distance from lisa -> ligo or ligo -> lisa.

pycbc/detector.py
```diff
@@ -322,3 +304,96 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
+            t_gps = Time(val = self.reference_time, format = 'gps',
+                         scale = 'utc').to_datetime(timezone = None)
+        elif isinstance(t_gps, np.ScalarType):
+            t_gps = Time(val = t_gps, format = 'gps',
+                         scale = 'utc').to_datetime(timezone = None)
+
+        t_gps = np.sum(np.array([t_gps.year - 2034, t_gps.month/12,
```
ahnitz, line 323: use the astropy Time class directly to convert to gps time (it will be more generic).

pycbc/detector.py
```diff
@@ -322,3 +304,96 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
+
+        """ra_angle = self.gmst_estimate(t_gps) - right_ascension"""
+        cosd = cos(declination)
+        e0 = cosd * cos(right_ascension)
+        e1 = cosd * -sin(right_ascension)
+        e2 = sin(declination)
+        ehat = np.array([e0, e1, e2])
+        return dx.dot(ehat) / constants.c.value
+
+    def time_delay_from_detector(self, other_detector, right_ascension,
```
ahnitz, line 384: It would be good to not assume this. The best way to do that would be to add a method for both the regular detector class and this one that is somthing like "icrs_position". You can then have this method just call that on both self and other to get their positions and then calculate the distance. Similarly, this method could be identical to the one for the regular detector class, so it works either way (i.e. you can ask the distance from lisa -> ligo or ligo -> lisa.

pycbc/detector.py
```diff
@@ -322,3 +304,65 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
+class LISA(object):
+    def __init__(self):
+        None
+
+    def get_pos(self, t_gps):
+        t_gps = Time(val=t_gps, format='gps', scale='utc').jyear
+
+        n = np.array(range(1, 4))
+        kappa, _lambda_ = 0, 0
+        alpha = 2. * np.pi * t_gps/1 + kappa
```
ahnitz, line 332: This doesn't seem to be a gps time anymore, some perhaps use a better label. Also, is there not a fiducial start time for this that is a specific year? Should that not be accounted for in here.

pycbc/detector.py
```diff
@@ -322,3 +304,65 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
+        x = a*cos(alpha) + prod - (1 + sin(alpha)**2)*cos(beta_n)
+        y = a*sin(alpha) + prod - (1 + cos(alpha)**2)*sin(beta_n)
+        z = -np.sqrt(3)*a*e*cos(alpha - beta_n)
+        self.location = np.array([x, y, z])
+
+        return self.location
+
+    def time_delay_from_location(self, other_location, right_ascension,
+                                 declination, t_gps):
+        dx = dist_lisa_gbd(other_location, t_gps)
```
ahnitz, line 347: Instead of creating a function that exists on its own and needs switch statements, a more class based approached would be to have both the LISA class and the Detector class have a .get_icrs_pos method. Then you just call both those methods and use the output to calculate the distance between them.

PR #3319, opened 2020-06-18
add plugin handling for waveforms

ahnitz:
* This allows pycbc to properly find external packages which *advertise* a plugin for pycbc without relying on environment variables.
* An example plugin package is provided in https://github.com/gwastro/example-waveform-plugin

If the plugin is simply installed, pycbc will automatically pick it up and its waveform models will be available from PyCBC. I think this is the behavior we would like for external packages. The hook is the standard Python mechanism using the 'entrypoint' system of 'pkg_resources', which was explicitly designed for this purpose. 

* (this will eventually deprecated the older undocumented interface, but I'm not removing in this PR yet). "

It would be good if you all could sign off on the generic interface and merge this first. In the next PR, I'll add documentation to main pages on how to use this both from the package level and how to add a waveform within a single script (by calling pycbc.waveform.add_custom_waveform). That will be followed up by the corresponding tutorial example (which we want for the workshop next week).

### DIFF
--- a/docs/waveform_plugin.rst
+++ b/docs/waveform_plugin.rst
@@ -1,0 +1,91 @@
+.. _waveform_plugin:
+
+------------------------------------------------------
+Making new waveform approximants available to PyCBC
+------------------------------------------------------
+
+=================================================
+Adding a custom waveform model within a script
+=================================================
+
+By example, the following script shows how to write a waveform model
+in the form required for PyCBC. We can also make this new waveform directly
+accessible by using the `pycbc.waveform.plugin.add_custom_waveform` function.
+If you are developing in a notebook or self-contained script, this may be
+what you want to do. However, if you want to make your waveform available
+to pycbc-based executables such as PyCBC Inference, also read the next
+section. 
+
+Ther are two kinds of models you can make. In the example below, we
+make a time-domain model. You can also make a freuqency-domain model. The only
+difference is that your function should return pycbc FrequencySeries and
+the required sample step option is 'delta_f' instead of 'delta_t'. 
+
+Each waveform generation function must take only keyword arguments, and
+should be able to take an arbitrary number of them. You may add new parameters
+as you like. These will be automatically useable by PyCBC Inference and
+other pycbc codes.
+
+Each waveform model must have an associate 'approximant' name, which identifies
+the model and distinguishes it from any other. If the name has already been
+used, you may need to select a different name. 
+
+.. plot:: ../examples/waveform/add_waveform.py
+   :include-source:
+
+=================================================
+Creating a plugin for PyCBC
+=================================================
+
+To make a waveform model universally available to PyCBC so it can be called
+from PyCBC Inference, or the pycbc-based searched codes, you can create
+a plugin package which advertises your model. PyCBC will automatically
+detects your package and makes your waveform model available for use.
+
+The steps are:
+
+ * Create a waveform model just like as in the above example
+ * Create a python package for your module
+ * In your packages setup.py advertise that it contains a PyCBC compatible
+   waveform model in it's 'entry_points' option. 
+  
+The setup.py would like the following, the key addition being the 'entry_points'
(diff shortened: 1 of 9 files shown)


ahnitz:
@cdcapano @spxiwh I've updated this PR with docs. Please see https://www.atlas.aei.uni-hannover.de/work/ahnitz/projects/prs/_gh-pages/latest/html/waveform_plugin.html for an example.

docs/waveform_plugin.rst
```diff
@@ -0,0 +1,91 @@
...
+from PyCBC Inference, or the pycbc-based searched codes, you can create
+a plugin package which advertises your model. PyCBC will automatically
+detects your package and makes your waveform model available for use.
+
+The steps are:
+
+ * Create a waveform model just like as in the above example
+ * Create a python package for your module
+ * In your packages setup.py advertise that it contains a PyCBC compatible
+   waveform model in it's 'entry_points' option.
```
cdcapano, line 50: 'entry_points' -> `entry_points`

docs/waveform_plugin.rst
```diff
@@ -0,0 +1,91 @@
...
+
+The steps are:
+
+ * Create a waveform model just like as in the above example
+ * Create a python package for your module
+ * In your packages setup.py advertise that it contains a PyCBC compatible
+   waveform model in it's 'entry_points' option. 
+  
+The setup.py would like the following, the key addition being the 'entry_points'
+parameter passed to setup.py.
```
cdcapano, line 53: "The setup.py would like the following," -> "Your `setup.py` should look like the following,"
Also, 'entry_points' -> `entry_points`

docs/waveform_plugin.rst
```diff
@@ -0,0 +1,91 @@
...
+            'Intended Audience :: Science/Research',
+            'Natural Language :: English',
+            'Topic :: Scientific/Engineering',
+            'Topic :: Scientific/Engineering :: Astronomy',
+            'Topic :: Scientific/Engineering :: Physics',
+            'License :: OSI Approved :: GNU General Public License v3 (GPLv3)',
+        ],
+    )
+
+The format is "capability":"approximant_name = module_path:function_name".
```
cdcapano, line 84: Change to, "The format for the `entry_points` parameter is `"capability":"approximant_name = module_path:function_name"`." to make it clear you're referring to the `entry_points` (and use backticks for the format).

docs/waveform_plugin.rst
```diff
@@ -0,0 +1,91 @@
...
+The module path may include dots if the module is within a package or sub-package. The
+valid 'capbility' is 'pycbc.waveform.td' and 'pycbc.waveform.fd' for time and frequency
+domain waveform models,  respectively.
```
cdcapano, line 87: Use backticks instead of single quotes.

docs/waveform_plugin.rst
```diff
@@ -0,0 +1,91 @@
...
+the required sample step option is 'delta_f' instead of 'delta_t'. 
+
+Each waveform generation function must take only keyword arguments, and
+should be able to take an arbitrary number of them. You may add new parameters
+as you like. These will be automatically useable by PyCBC Inference and
+other pycbc codes.
+
+Each waveform model must have an associate 'approximant' name, which identifies
+the model and distinguishes it from any other. If the name has already been
+used, you may need to select a different name.
```
cdcapano, line 31: Change "may need" to "should"? Otherwise, you should address what happens if you use a name that's already used. (I'm guessing it would override it, since it looks like the custom approximants are being added to the cpu dicts last, but maybe not. Might be best to just to tell people not to do that.) Also, it would be good to point people to the docs [here](http://pycbc.org/pycbc/latest/html/waveform.html#what-waveforms-can-i-generate) so they know how to see what approximant names are currently used.

ahnitz: It raises an error by default.

PR #3320, opened 2020-06-19
Live combine fits missing bins bug

GarethDaviesGW:
One of the variables being looped over was in a different dimension than the others (fit file date rather than bin)

I hadn't noticed this as the number of bins had always been less than the number of days being analysed

Also prevents the same colour being used twice in plots given a large number of bins


titodalcanton:
@GarethDaviesGW can you confirm this only affects plotting?

GarethDaviesGW:
yes, I reran the O3a trigger fits combination code and all datasets are the same. The only difference is the addition of the `live_times` dataset under `ifo/daily_fits`

PR #3322, opened 2020-06-19
PyCBC Live Test w/ Multiple HDF injections

maxtrevor:


titodalcanton:
@maxtrevor it appears this is causing the Travis test to fail with this error:
```
>> [Fri Jun 19 19:26:19 UTC 2020] Generating injection
  File "./generate_injections.py", line 16, in <module>
    samples = record.FieldArray(2, dtype=[('approximant', h5py.string_dtype(encoding='utf-8')), \
AttributeError: module 'h5py' has no attribute 'string_dtype'
    FAILED!
```
Are you maybe relying on a fairly new feature of h5py?

examples/live/check_results.py
```diff
@@ -1,208 +1,236 @@
...
-      coinc_trig_paths = sorted(glob.glob('output/coinc*.xml.gz'))
-      l=len(coinc_trig_paths)
-      if l==0:
+    coinc_fail = False    
+    # gather coincident triggers
+    coinc_trig_paths = sorted(glob.glob('output/coinc*.xml.gz'))
+    n_coincs=len(coinc_trig_paths)
+    if n_coincs==0:
+        if verbose:
             log.error('No coincident triggers detected')
```
titodalcanton, line 114: Errors should be displayed regardless of any verbose flag.

examples/live/check_results.py
```diff
@@ -1,208 +1,236 @@
...
-            coinc_fail = True
-      elif l>=10: 
+        coinc_fail = True
+    elif n_coincs>=10: 
+        if verbose:
             log.error('Too many coincident triggers detected')
```
titodalcanton, line 118: Errors should be displayed regardless of any verbose flag.

examples/live/check_results.py
```diff
@@ -1,208 +1,236 @@
...
+        inj_spin1z = inj['spin1z'][:]
+        inj_spin2z = inj['spin2z'][:]
+        inj_time = inj['tc'][:]
+         
+    n_injs = len(inj_mass1)
+    if n_injs > n_coincs :
+        log.error('More injections than coincident triggers')
+        coinc_fail = True
+    if n_injs < n_coincs :
+        log.error('Fewer injections than coincident triggers')
```
titodalcanton, line 135: This second check should be removed, since we expect noise to produce extra coincidences. We should at least have as many coincs as the number of injections, but we may well have more (especially if we let the test run over a lot of data or a very large bank).

examples/live/check_results.py
```diff
@@ -1,208 +1,236 @@
...
-            xmldoc = ligolw_utils.load_filename(
+    #create field array to store properties of triggers
+    trig_props = record.FieldArray(n_coincs, dtype=[('mass1', float), \
+                ('mass2', float), ('spin1z', float), ('spin2z', float), \
+                ('tc', float), ('net_snr', float)])
+      
+    #store properties of coincident triggers
+    x=-1
+    for ctrigfp in coinc_trig_paths:     
+        x+=1
```
titodalcanton, line 146: If you are using `x` as an index for each coinc, there is a neat one-line way to achieve this:
```python
for x, ctrigfp in enumerate(coinc_trig_paths):
```

examples/live/check_results.py
```diff
@@ -1,208 +1,236 @@
...
+        trig_props['mass2'][x]=sngl_inspiral_table.get_column('mass2')[0]    
+        trig_props['spin1z'][x]=sngl_inspiral_table.get_column('spin1z')[0] 
+        trig_props['spin2z'][x]=sngl_inspiral_table.get_column('spin2z')[0] 
+        
+        
+        snr_list=sngl_inspiral_table.get_column('snr')
+        network_snr_squared=0
+        for snr in snr_list:
+            network_snr_squared+=snr**2
+        network_snr=np.sqrt(network_snr_squared)
```
titodalcanton, line 164: Another Python trick to shorten your code (and make it faster, though it will not be an issue here) is list comprehension:
```python
network_snr = sum([snr ** 2 for snr in snr_list]) ** 0.5
```

titodalcanton: and if `snr_list` is actually a Numpy array (I forget what `get_column()` returns) you can even do
```python
network_snr = sum(snr_list ** 2) ** 0.5
```

examples/live/check_results.py
```diff
@@ -1,208 +1,236 @@
...
+        if not has_match:
+            coinc_fail = True
+            log.error('Injection %i has no match',i)
+            
+    if coinc_fail:
+        log.error('Coincident Trigger Test Failed')
+    return coinc_fail
+
+      
+verbose = False
```
titodalcanton, line 209: Carrying around this `verbose` flag is not necessary, actually. The same functionality is achieved by setting the `level` kwarg of `log.basicConfig()`, i.e. if you se that to `log.ERROR` it will only show messages issued via `log.error()` and ignore `log.info()`.

examples/live/check_results.py
```diff
@@ -1,208 +1,236 @@
...
 log.basicConfig(level=log.INFO, format='%(asctime)s %(message)s')
 sim_gps_start = 1272790000
-sim_gps_end =  1272790500
+sim_gps_end =  1272790300
```
titodalcanton, line 212: Did you intend to make the simulation shorter, or was this just a local test? I would prefer to keep it to 500 s.

examples/live/check_results.py
```diff
@@ -1,208 +1,236 @@
...
 else:
-      path_to_inj_xml = inj_xmls[0]
-      log.info('injection xml found')
-      inj_doc = ligolw_utils.load_filename(
-            path_to_inj_xml, False, contenthandler=LIGOLWContentHandler)
-      inj_table = lsctables.SnglInspiralTable.get_table(inj_doc)
-   
-      tested_detectors = {'H1', 'L1', 'V1'}
+    if verbose:
+         log.info('injection hdf found')
```
titodalcanton, line 222: I do not think this block is necessary. If `test_inj.hdf` is missing, the code that tries to read it will just fail with a pretty obvious exception.

examples/live/check_results.py
```diff
@@ -1,208 +1,236 @@
...
+        log.error('Coincident Trigger Test Failed')
+    return coinc_fail
+
+      
+verbose = False
 log.basicConfig(level=log.INFO, format='%(asctime)s %(message)s')
 sim_gps_start = 1272790000
-sim_gps_end =  1272790500
+sim_gps_end =  1272790300
 sim_f_lower = 18
```
titodalcanton, line 213: Looks like this parameter has been moved to `generate_injections.py`, so this can be removed.

examples/live/generate_injections.py
```diff
@@ -0,0 +1,30 @@
...
+import os, sys
+from pycbc.io import record
+import h5py
+from pycbc.inject import InjectionSet
+
+if os.path.exists('./test_inj.hdf'):
+    raise OSError("output-file already exists")
+
+# 2 injections 
+static_params = { 'f_lower': 18.0, 'f_ref': 20.0, \
```
titodalcanton, line 12: I suggest setting `f_ref` equal to `f_lower` as not all approximants like them different.

examples/live/generate_injections.py
```diff
@@ -0,0 +1,30 @@
...
+               'taper': 'start', 'ra': 45.0, 'dec': 45.0, 'inclination': 0.0, \
+               'coa_phase': 0.0, 'polarization': 0.0}
+
+samples = record.FieldArray(2, dtype=[('approximant', h5py.string_dtype(encoding='utf-8')), \
+                                      ('mass1', float), ('mass2', float), \
+                                      ('spin1z', float), ('spin2z', float), \
+                                      ('tc', float), ('distance', float)])
+
+samples['approximant'] = [u'SEOBNRv4', u'SpinTaylorT4']
```
titodalcanton, line 21: Since the following lines contain a bunch of unexplained "magic values", in cases like this you should add a comment saying where they come from. In this case, they are chosen to match two templates in the bank.

maxtrevor:
> @maxtrevor it appears this is causing the Travis test to fail with this error:
> 
> ```
> >> [Fri Jun 19 19:26:19 UTC 2020] Generating injection
> Traceback (most recent call last):
>   File "./generate_injections.py", line 16, in <module>
>     samples = record.FieldArray(2, dtype=[('approximant', h5py.string_dtype(encoding='utf-8')), \
> AttributeError: module 'h5py' has no attribute 'string_dtype'
>     FAILED!
> ```
> 
> Are you maybe relying on a fairly new feature of h5py?

I don't know if it is new or not but this was the hardest part of the test to get working. It runs successfully on CIT. Initially I naively setting the dtype to str but that was throwing an error from h5py that it didn't know how to handle native python strings.  I can't say I understand all of it but I got it working this way after reading through http://docs.h5py.org/en/stable/strings.html

examples/live/check_results.py
maxtrevor: Ya not totally sure what I was thinking here. Will fix

examples/live/check_results.py
maxtrevor: Based on your other comment, should this even produce an error at all? If we "run over a lot of data or a very large bank" we could get 10 triggers?

examples/live/check_results.py
maxtrevor: Ya I will remove it. This was largely here for my own testing so I could have some extra messages to see what was going on, but I removed all those extra messages anyways

examples/live/check_results.py
maxtrevor: This was a local change to make it run slightly faster on a day when CIT was super slow, forgot to revert it

examples/live/check_results.py
maxtrevor: sim f_lower is used by your check_single_results algorithm so it is needed here still unless we want to pass the info to here in another way

maxtrevor:
Addressed your comments except I have no clue how to deal with that dtype issue. It runs fine on CIT, so maybe the easiest fix is to update h5py wherever Travis is running?

titodalcanton:
I suspect #3395 will take a bit to be resolved. In the mean time, could you try using the `SEOBNRv4_opt` approximant for both injections? That is a fast implementation of SEOBNRv4 which, for this purpose, should be entirely equivalent. I do not remember how low in mass it allows you to go, but give it a try. If it works for both injections, we can proceed with this PR.

maxtrevor:
Updated with a workaround to the issue I encountered in https://github.com/gwastro/pycbc/issues/3395 . generate_injections.py now generates two separate hdf injection files. This required some updates to run.sh and and check_results.py for compatibility with multiple hdf files.

examples/live/check_results.py
```diff
@@ -6,203 +6,230 @@
...
+                if not close(inj_mass2[i], trig_props['mass2'][j], 5e-7):
+                    match = False
+                    log.error("Injection %i mass2 does not match simultaneous trigger's mass2",i)
+                if not close(inj_spin1z[i], trig_props['spin1z'][j], 5e-7):
+                    match = False
+                    log.error("Injection %i spin1z does not match simultaneous trigger's spin1z",i)
+                if not close(inj_spin2z[i], trig_props['spin2z'][j], 5e-7):
+                    match = False
+                    log.error("Injection %i spin2z does not match simultaneous trigger's spin2z",i)
+                if not close(15.0, trig_props['net_snr'][j], 0.05):
```
titodalcanton, line 194: I suggest making the tolerance 1 here, as we do not generally expect the SNR to be stable to a better precision than that.

examples/live/check_results.py
```diff
@@ -6,203 +6,230 @@
...
-      tested_detectors = {'H1', 'L1', 'V1'}
+
+       
+tested_detectors = {'H1', 'L1', 'V1'}
    
-      single_fail = check_single_results(tested_detectors)
-      coinc_fail = check_coinc_results(inj_table)
-      fail = single_fail or coinc_fail
+single_fail = check_single_results(tested_detectors)
+coinc_fail = check_coinc_results('./test_inj.hdf')
```
titodalcanton, line 226: Looks like `check_coinc_results()` does not need any argument now, right? If so, please update.

examples/live/check_results.py
```diff
@@ -6,203 +6,230 @@
...
+        log.info('Mass 2: %f', trig_props['mass2'][x])
+        log.info('Spin1z: %f', trig_props['spin1z'][x])
+        log.info('Spin2z: %f', trig_props['spin2z'][x])
+     
+
+    # check if injections match trigger params
+    for i in range(n_injs):
+        has_match = False
+        for j in range(n_coincs):
+            if close(inj_time[i],trig_props['tc'][j],1.0):
```
titodalcanton, line 180: I think the match logic got a bit overcomplicated and could be simplified as follows. Start with `has_match = False` as you did. Then for each coinc, check if the coinc matches *all* relevant parameters (tc, mass1, mass2, spin1z, spin2z, network SNR). If everything matches, set `has_match` to true and break out of the `j` loop. Otherwise keep looping.

However, this will not work if your want to have a log message that says exactly which parameter did *not* match. I am not sure this is necessary for the Travis test, because if the test fails we would probably run a more verbose local check anyway. But if you absolutely want this log message, you can keep the code as is.

examples/live/generate_injections.py
```diff
@@ -0,0 +1,58 @@
...
+
+if os.path.exists('./test_inj1.hdf'):
+    raise OSError("output-file 1 already exists")
+    
+if os.path.exists('./test_inj2.hdf'):
+    raise OSError("output-file 2 already exists")
+
+    
+# injection 1    
+static_params = { 'f_lower': 18.0, 'f_ref': 18.0, 'approximant': 'SEOBNRv4', \
```
titodalcanton, line 15: Please remove the backslashes, as they are not required when you have a list or dict.

titodalcanton:
@maxtrevor have a look at the last few comments, after those I will approve this.

maxtrevor:
Should be ready to go now

PR #3325, opened 2020-06-21
Add spherical to cartesian transform

cdcapano:
Adds a generic `SphericalToCartesian` transform and its inverse.

Currently, we have a class that does this specifically for `spin1` called `SphericalSpin1ToCartesianSpin1` (ditto for spin2). Those two classes are tied to parameters with names "spin1{X}". This makes that generic, so you can use a single transform for any spherical <-> cartesian transforms. This means you would now do:
```
[waveform_transforms-spin1x+spin1y+spin1z]
name = spherical_to_cartesian
x = spin1x
y = spin1y
z = spin1z
radial = spin1_a
azimuthal = spin1_azimuthal
polar = spin1_polar
```
And same for spin2.

I've left the `SphericalSpin(1|2)ToCartesianSpin(1|2)` classes in for now, but have added deprecation warnings to them saying they will be removed in a future update.

This still leaves in automatic transforms in the waveform models (meaning that you don't actually need to include the waveform transforms section for the spins). I will remove that in another patch.

### DIFF
--- a/examples/inference/priors/gw150914_like.ini
+++ b/examples/inference/priors/gw150914_like.ini
@@ -71,6 +71,28 @@ name = uniform_solidangle
 polar-angle = spin2_polar
 azimuthal-angle = spin2_azimuthal
 
+; The waveform generator expects spins to be in cartesian coordinates, with
+; names spin(1|2)(x|y|z). We therefore need to provide a waveform transform
+; that converts the spherical coordinates that we have defined the spin prior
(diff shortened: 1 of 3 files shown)


ahnitz:
@cdcapano Error in travis looks real for this one.

cdcapano:
Yes, Travis did catch an issue. I've fixed it. As soon as the py37 issue is fixed, I'll rebase.

PR #3327, opened 2020-06-22
Remove automatic transforms from waveform generator

cdcapano:
This removes the automatic transforms that happen in the CBC waveform generators in pycbc inference. Those transforms tried to automatically figure out what transforms were needed to go from some parameters (like spherical spin coordinates and mchirp, q) to parameters understood by the waveform generators. While this meant not having to include a waveform transforms section for spherical spin parameters, the silent and difficult to follow nature of these transforms makes it easy to make mistakes. It also goes against the general philosophy we have of making analysis settings explicit in the configuration file.

Just removing the automatic transforms can be dangerous for spin parameters, however. We typically specify the prior in spherical coordinates. Removing the automatic transforms will cause an analysis that does not have the spherical -> Cartesian waveform transform in the config file to silently set all spins to zero, since lalsimulation expects Cartesian spin coordinates. Since we have not previously required people to include a waveform transform to convert the spherical coordinates to Cartesian, removing the automatic transforms could silently break many people's analyses if they update their code.

To protect against this, I've added a check for spin parameters in the base model. If any parameters are found in the variable or static params that start with `spin(1|2)` but there are no coordinates named `spin(1|2)(x|y|z)` in variable and static params, nor any in the output of the waveform transforms, an error will be raised. This test is applied separately for component 1 and component 2. The error message is:
```
ValueError: Spin parameters spin1_polar, spin1_azimuthal, spin1_a found in variable/static params for component 1, but no Cartesian spin parameters (spin1x, spin1y, spin1z) found in either the variable/static params or the waveform transform outputs. Most waveform generators only recognize Cartesian spin parameters; without them, all spins are set to zero. If you are using spherical spin coordinates, add the following waveform_transform to your config file:

[waveform_transforms-spin1x+spin1y+spin1z]
name = spherical_to_cartesian
x = spin1x
y = spin1y
z = spin1z
radial = spin1_a
azimuthal = spin1_azimuthal
polar = spin1_polar

Here, spin1_a, spin1_azimuthal, and spin1_polar are the names of your radial, azimuthal, and polar coordinates, respectively. If you intentionally did not include Cartesian spin parameters, (e.g., you are using a custom waveform or model) add

[no_err_on_missing_cartesian_spins]

to your config file as an empty section and rerun. This check will not be performed in that case.
```
As indicated in the error message, you can skip the check by adding an empty section titled `[no_err_on_missing_cartesian_spins]` to the config file.

The error will only be raised if no Cartesian spin coordinates present. If there is at least one, then it will not be triggered. I did that because you may be doing an aligned spin search, in which you only have the z components.

Depends on #3325, so putting on hold until that is merged.

### DIFF
--- a/docs/inference.rst
+++ b/docs/inference.rst
@@ -376,12 +376,6 @@ quadrupole moment of a neutron star ``dquad_mon1`` to its tidal deformation
    waveform generator understands. This is why in the above example we are
    able to use a custom transform without needing to provide a Jacobian.
 
-Some common transforms are pre-defined in the code. These are: the mass
-parameters ``mass1`` and ``mass2`` can be substituted with ``mchirp`` and
-``eta`` or ``mchirp`` and ``q``.  The component spin parameters ``spin1x``,
-``spin1y``, and ``spin1z`` can be substituted for polar coordinates
-``spin1_a``, ``spin1_azimuthal``, and ``spin1_polar`` (ditto for ``spin2``).
-
 ^^^^^^^^^^^^^^^^^^^^^^
 Calibration parameters
 ^^^^^^^^^^^^^^^^^^^^^^
--- a/pycbc/inference/models/base.py
+++ b/pycbc/inference/models/base.py
@@ -753,9 +753,18 @@ class BaseModel(object):
         # get any waveform transforms
         if any(cp.get_subsections('waveform_transforms')):
             logging.info("Loading waveform transforms")
-            args['waveform_transforms'] = \
-                transforms.read_transforms_from_config(
-                    cp, 'waveform_transforms')
+            waveform_transforms = transforms.read_transforms_from_config(
+                cp, 'waveform_transforms')
+            args['waveform_transforms'] = waveform_transforms
+        else:
+            waveform_transforms = []
+        # safety check for spins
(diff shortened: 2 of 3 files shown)


ahnitz:
@cdcapano I'm a bit wary of making this change all at once. Can you just submit a PR with the doc changes, so that people are encouraged to the do the right thing, and we can make sure to give people more heads up for the actual removal. (maybe put a deprecation warning in the docs change PR?). I.E. I'm not sure I'm comfortable for this to merge before we do a release tomorrow.

cdcapano:
@ahnitz Yeah, I can change the check to just issue a deprecation warning for now. Will address it in the morning.

ahnitz:
@cdcapano Is this still needed? It forget if it's already been done and all the docs haven't been using this for a while. Can we either apply (maye update PR) or close if no longer needed?

ahnitz (approved):

ahnitz:
@cdcapano Poke, is this still needed?

cdcapano:
@ahnitz Oh, yes, this is still needed. Sorry, I missed that you had approved this last August. This is so old though, I should probably rebase to master and re-push.

PR #3328, opened 2020-06-22
Avoid numpy 1.19

cdcapano:
ignore


ahnitz:
@cdcapano Pin it also in the setup.py and pyproject.toml and see

ahnitz:
@cdcapano Ok, this seems to pass now.

PR #3330, opened 2020-06-22
WIP: Dynesty checkpointing: With Hdf files as checkpoint files

sum33it:
To make the checkpoint filesystem uniform across samplers, I am starting this pull request. The sampler state is dumped to existing checkpoint pdf files using 'hickle' at regular intervals. Added two functions for this purpose in `pycbc.inference.io.dynesty`.

### DIFF
--- a/pycbc/inference/io/__init__.py
+++ b/pycbc/inference/io/__init__.py
@@ -29,7 +29,7 @@ import logging
 import h5py as _h5py
 from pycbc.io.record import (FieldArray, _numpy_function_lib)
 from pycbc import waveform as _waveform
-from pycbc.io.hdf import (dump_state,load_state)
+from pycbc.io.hdf import (dump_state, load_state)
 
 from pycbc.inference.option_utils import (ParseLabelArg, ParseParametersArg)
 from .emcee import EmceeFile
@@ -169,8 +169,8 @@ def check_integrity(filename):
     with loadfile(filename, 'r') as fp:
         if fp.name == 'dynesty_file':
             try:
-                load_state(fp,path='sampler_info/saved_state')
-            except:
+                load_state(fp, path='sampler_info/saved_state')
+            except IOError:
                 raise IOError("Could not recover pickled state")
         else:
             # check that all datasets in samples have the same shape
@@ -242,7 +242,7 @@ def validate_checkpoint_files(checkpoint_file, backup_file, sampler_name):
     if backup_valid:
         with loadfile(backup_file, 'r') as fp:
             backup_valid = fp.validate()
-    if sampler_name == 'dynesty' or sampler_name == 'cpnest':
+    if sampler_name in ['dynesty', 'cpnest']:
         pass
     else:
         # This check is not required by nested samplers
--- a/pycbc/inference/io/__init__.py
+++ b/pycbc/inference/io/__init__.py
@@ -29,6 +29,7 @@ import logging
 import h5py as _h5py
 from pycbc.io.record import (FieldArray, _numpy_function_lib)
 from pycbc import waveform as _waveform
+from pycbc.io.hdf import (dump_state,load_state)
 
 from pycbc.inference.option_utils import (ParseLabelArg, ParseParametersArg)
 from .emcee import EmceeFile
@@ -166,17 +167,23 @@ def check_integrity(filename):
     # if the file is corrupted such that it cannot be opened, the next line
     # will raise an IOError
     with loadfile(filename, 'r') as fp:
-        # check that all datasets in samples have the same shape
-        parameters = list(fp[fp.samples_group].keys())
-        # but only do the check if parameters have been written
-        if len(parameters) > 0:
-            group = fp.samples_group + '/{}'
-            # use the first parameter as a reference shape
-            ref_shape = fp[group.format(parameters[0])].shape
-            if not all(fp[group.format(param)].shape == ref_shape
-                       for param in parameters):
-                raise IOError("not all datasets in the samples group have the "
-                              "same shape")
+        if fp.name == 'dynesty_file':
+            try:
+                load_state(fp,path='sampler_info/saved_state')
+            except:
+                raise IOError("Could not recover dumped state")
+        else:
+            # check that all datasets in samples have the same shape
+            parameters = list(fp[fp.samples_group].keys())
+            # but only do the check if parameters have been written
+            if len(parameters) > 0:
+                group = fp.samples_group + '/{}'
+                # use the first parameter as a reference shape
+                ref_shape = fp[group.format(parameters[0])].shape
+                if not all(fp[group.format(param)].shape == ref_shape
+                           for param in parameters):
+                    raise IOError("not all datasets in the samples group have "
+                                  "the same shape")
             # check that we can read the first/last sample
             firstidx = tuple([0]*len(ref_shape))
             lastidx = tuple([-1]*len(ref_shape))
@@ -185,7 +192,7 @@ def check_integrity(filename):
                 _ = fp[group.format(param)][lastidx]
 
 
-def validate_checkpoint_files(checkpoint_file, backup_file):
+def validate_checkpoint_files(checkpoint_file, backup_file, sampler_name):
     """Checks if the given checkpoint and/or backup files are valid.
 
     The checkpoint file is considered valid if:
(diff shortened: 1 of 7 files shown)


pycbc/inference/io/dynesty.py
```diff
@@ -24,8 +24,20 @@
 """Provides IO for the dynesty sampler.
 """
 from .base_nested_sampler import BaseNestedSamplerFile
+import hickle
```
ahnitz, line 27: Why are you using this and not the function Collin wrote?

sum33it: Thanks for pointing. I fixed this issue. Also tested with a run with GW150914 and an analytical model. It seems to work now.

pycbc/inference/sampler/dynesty.py
```diff
@@ -175,10 +231,23 @@ def finalize(self):
...
 
+    def getstate(self):
+        """Strips unserializable attributes of sampler
+        """
+        self_dict = self._sampler.__dict__.copy()
+        # Remove multiprocessing / mpi pool-related/ rstate attributes
+        # as they cannot be pickled
+        del self_dict['pool']
+        del self_dict['M']
+        del self_dict['rstate']
```
ahnitz, line 246: This won't preserve the random seed for when you start back up from a resume. I think wee need to (1) set the random state explicitly. It might pickle on its own, which would be nice, but if it doens't, you'll have to explicitly preserve it and restore it on resume.

pycbc/inference/sampler/dynesty.py
```diff
@@ -158,10 +200,24 @@ def from_config(cls, cp, model, output_file=None, nprocesses=1,
...
+        for fn in [self.checkpoint_file]:
+            with loadfile(fn, 'a') as fp:
+                fp.write_pickled_data_into_checkpoint_file(pickle_obj)
 
     def resume_from_checkpoint(self):
-        pass
+        for fn in [self.checkpoint_file]:
+            with loadfile(fn, 'r') as fp:
+                self._sampler = fp.read_pickled_data_from_checkpoint_file()
+                self._sampler.rstate = numpy.random
```
ahnitz, line 215: See paired comment, this won't restore the random state to how it left off when checkpointed.

sum33it: @ahnitz Now I have changed it and using functions written by @cdcapano: `write_random_state` and `read_random_state` to keep track of random state and restore it from previous checkpoint.

pycbc/inference/sampler/dynesty.py
```diff
@@ -158,10 +200,35 @@ def from_config(cls, cp, model, output_file=None, nprocesses=1,
...
+            with loadfile(fn, 'a') as fp:
+                fp.write_pickled_data_into_checkpoint_file(pickle_obj)
+                fp.write_random_state()
 
     def resume_from_checkpoint(self):
-        pass
+        for fn in [self.checkpoint_file]:
+            with loadfile(fn, 'r') as fp:
+                self._sampler = fp.read_pickled_data_from_checkpoint_file()
+            self._sampler.rstate = numpy.random
```
ahnitz, line 216: Why is this line here?

pycbc/inference/sampler/dynesty.py
```diff
@@ -158,10 +200,35 @@ def from_config(cls, cp, model, output_file=None, nprocesses=1,
...
+            self._sampler.M = pool.map
+            self._sampler.pool = pool
+
+    def set_state_from_file(self, filename):
+        """Sets the state of the sampler back to the instance saved in a file.
+        """
+        with self.io(filename, 'r') as fp:
+            rstate = fp.read_random_state()
+        # set the numpy random state
+        numpy.random.set_state(rstate)
```
ahnitz, line 230: Since dynesty has an explicitly rstate, why don't you just set that when you instatiate the sampler first? That way we don't have to rely on the state of the global numpy.

ahnitz: Oh, is this a function that collin already had in place?

pycbc/inference/sampler/dynesty.py
ahnitz: Nvm, I think I see what you are doing.

pycbc/inference/io/__init__.py
```diff
@@ -185,7 +192,7 @@ def check_integrity(filename):
                 _ = fp[group.format(param)][lastidx]
 
 
-def validate_checkpoint_files(checkpoint_file, backup_file):
+def validate_checkpoint_files(checkpoint_file, backup_file, sampler_name):
```
cdcapano, line 195: Don't make this dependent on the sampler name. That will be a major pain to keep up to date. Just have it be a keyword argument, like `check_nsamples=True`, then below replace the:
```
if sampler_name in ['dynesty', 'cpnest']:
    pass
else:
```
with:
```
if check_nsamples:
```
By making this a kwarg with default of True, you don't have to change any of the other codes, and just have to pass in `check_nsamples=False` into the call to validate_checkpoint_files for dynesty.

pycbc/inference/sampler/dynesty.py
```diff
@@ -98,9 +112,35 @@ def __init__(self, model, nlive, nprocesses=1,
...
     def run(self):
-        self._sampler.run_nested(**self.run_kwds)
+        diff_niter = 1
+        if self.run_with_checkpoint is True:
+            if self.checkpoint_file:
+                try:
+                    self.resume_from_checkpoint()
+                    self.niter = self._sampler.results.niter
+                    logging.info('Successfully read from the checkpoint file')
+                except KeyError:
```
ahnitz, line 126: Is a KeyError the only possible error?

pycbc/inference/sampler/dynesty.py
```diff
@@ -80,12 +84,22 @@ def __init__(self, model, nlive, nprocesses=1,
...
+        self.maxcall = maxcall
+        self.checkpoint_time_interval = checkpoint_time_interval
         self.run_kwds = {} if run_kwds is None else run_kwds
         self.nlive = nlive
         self.names = model.sampling_params
         self.ndim = len(model.sampling_params)
         self.checkpoint_file = None
+        self.__getstate__ = self.getstate
+        self.niter = niter
+        if ('maxcall' or 'maxiter') in self.run_kwds:
```
ahnitz, line 98: I'm confused by this. So there will be no checkpointing by default, if you don't set maxcall in run_kwds? It seems self.maxcall isn't actually used anyway. Why the inconsistency?

ahnitz: Shouldn't checkpointing just be enabled if checkpoint_time_interval is set?

pycbc/inference/sampler/base.py
```diff
@@ -175,7 +175,7 @@ def setup_output(sampler, output_file):
     # check if we have a good checkpoint and/or backup file
     logging.info("Looking for checkpoint file")
     checkpoint_valid = validate_checkpoint_files(checkpoint_file,
-                                                 backup_file)
+                                                 backup_file, sampler.name)
```
cdcapano, line 178: Continuing my suggestion above, add a `check_nsamples` keyword argument to `setup_output` (with default True), which you then pass to `validate_checkpoint_files`.

pycbc/inference/sampler/base_mcmc.py
```diff
@@ -620,7 +620,7 @@ def checkpoint(self):
         # check validity
         logging.info("Validating checkpoint and backup files")
         checkpoint_valid = validate_checkpoint_files(
-            self.checkpoint_file, self.backup_file)
+            self.checkpoint_file, self.backup_file, self.name)
```
cdcapano, line 623: With the keyword argument change, you can remove this change from base mcmc.

pycbc/inference/sampler/cpnest.py
```diff
@@ -139,7 +139,7 @@ def finalize(self):
             self.write_results(fn)
         logging.info("Validating checkpoint and backup files")
         checkpoint_valid = validate_checkpoint_files(
-            self.checkpoint_file, self.backup_file)
+            self.checkpoint_file, self.backup_file, self.name)
```
cdcapano, line 142: Here, pass in `check_nsamples=False`.

pycbc/inference/sampler/dynesty.py
```diff
@@ -175,10 +242,23 @@ def finalize(self):
             self.write_results(fn)
         logging.info("Validating checkpoint and backup files")
         checkpoint_valid = validate_checkpoint_files(
-            self.checkpoint_file, self.backup_file)
+            self.checkpoint_file, self.backup_file, self.name)
```
cdcapano, line 245: And here pass in `check_nsamples=False`

pycbc/inference/sampler/multinest.py
```diff
@@ -327,7 +327,7 @@ def checkpoint(self):
                 f_p.write_niterations(self.niterations)
         logging.info("Validating checkpoint and backup files")
         checkpoint_valid = validate_checkpoint_files(
-            self.checkpoint_file, self.backup_file)
+            self.checkpoint_file, self.backup_file, self.name)
```
cdcapano, line 330: Here pass `check_nsamples=False`

pycbc/inference/sampler/dynesty.py
sum33it: At least in the testing it was the only error I encountered in various possible scenarios.

pycbc/inference/io/__init__.py
sum33it: Ok. Thanks for the suggestion. Let me do this.

pycbc/inference/sampler/dynesty.py
ahnitz: If the key is there but is corrupt, this won't be the error. This is one of the examples, where a blanket except would be correct.

ahnitz:
Closing in favor of #3354

PR #3335, opened 2020-06-23
Update format_lmns in ringdown module.

JulianWesterweck:
The inference injection workflow currently does not work with ringdown injections with multiple modes because the different length of the lmn modes array causes a dimension mismatch. 
This updates the ringdown module to read a string containing the list as produced by the ConfigParser of the workflow when given as a string in the config file.


ahnitz:
@micamu Any reason you are holding off on merging this? (I think you have permission to)

ahnitz:
@cdcapano @micamu @JulianWesterweck Any reason this hasn't been merged already?

Issue #2955, opened 2019-09-26
hdfinjfind can't run on valid statmap file produced by multi-ifo workflow


ahnitz:
hdfinjfind should be able to run on any statmap file, but it appears to be broken except if combined_statmap has been called first. 
```
Traceback (most recent call last):
  File "/work/ahnitz/projects/2-ogc/statmap_runs/env/bin/pycbc_coinc_hdfinjfind", line 4, in <module>
    __import__('pkg_resources').run_script('PyCBC===0b9dcf', 'pycbc_coinc_hdfinjfind')
  File "/work/ahnitz/projects/2-ogc/statmap_runs/env/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/work/ahnitz/projects/2-ogc/statmap_runs/env/local/lib/python2.7/site-packages/pkg_resources/__init__.py", line 1462, in run_script
    exec(code, namespace, namespace)
  File "/work/ahnitz/projects/2-ogc/statmap_runs/env/lib/python2.7/site-packages/PyCBC-0b9dcf-py2.7-linux-x86_64.egg/EGG-INFO/scripts/pycbc_coinc_hdfinjfind", line 251, in <module>
    fo[key].attrs['pivot'] = f[key].attrs['pivot']
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "/work/ahnitz/projects/2-ogc/statmap_runs/env/local/lib/python2.7/site-packages/h5py/_hl/group.py", line 262, in __getitem__
    oid = h5o.open(self.id, self._e(name), lapl=self._lapl)
  File "h5py/_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py/_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py/h5o.pyx", line 190, in h5py.h5o.open
KeyError: "Unable to open object (object 'H1L1' doesn't exist)"
```

tdent:
Is this urgent? If so should assign someone.

ahnitz:
@tdent This is not the most urgent, as it works for the 2-ifo workflow and the multi-ifo *if* you've run the statmap file through the add_statmap code. So there is a working path. It's just overly constrictive at the moment.

PR #3343, opened 2020-06-26
get hdfinjfind to work on multiifo_statmap files

GarethDaviesGW:
As requested in #2955, this checks whether the attribute-hanging group exists (in the case of combine_statmap) or not (multiifo_statmap) and gets the relevant attributes according to this


GarethDaviesGW:
Poke @ahnitz

PR #3347, opened 2020-07-01
Write checkpoint history

cdcapano:
This makes it so that some history is preserved in the inference checkpoint file on each checkpoint. Right now, only the MCMC samplers make use of this, but nested samplers could pick up the functionality later. This can be used for comparing performance between samplers, and to diagnose issues (like growing ACTs).

On each checkpoint, `update_checkpoint_history` is called. This is a method of `BaseSamplerFile`. It writes the current time and the time since the last checkpoint or last startup (which ever happened last). It then calls `_update_sampler_history`. This doesn't do anything in `BaseSamplerFile`, but can be overridden to write-sampler specific things. This is overriden by `CommonMCMCMetadataIO` to write the number of iterations, effective number of samples, burn-in iteration, and autocorrelation time. In the future, nested samplers could do something similar (e.g., writing the current dlogz). All information is written to `samlper_info/checkpoint_history/`.

To make this work, the ability to append data has been added to `BaseInferenceFile.write_data`. I also had to add the time that pycbc inference starts up, so that the time since the last checkpoint could be accurately computed. This is stored as `run_start_time` in the file's `attrs`. For completeness, I also added a `run_end_time`, which is written by `pycbc_inference` just before exiting.

I tested using the single template example in the docs, on emcee pt and epsie.

In another PR, I'll add an executable to plot the checkpoint history.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -158,6 +158,10 @@ if not opts.save_backup:
     logging.info("Deleting backup file")
     os.remove(sampler.backup_file)
 
+# write the end time
+with sampler.io(opts.output_file, 'a') as fp:
+    fp.write_run_end_time()
+
 if condor_ckpt:
    # create an empty checkpoint file
    open(sampler.checkpoint_file, 'a').close()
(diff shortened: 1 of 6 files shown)


cdcapano:
@ahnitz Can you take a look?

PR #3350, opened 2020-07-03
Add duration support for Phenom HM waveforms

cdcapano:
Adds a function `imrphenomhm_length_in_time` to `waveform.py` that estimates the duration of Phenom waveforms that include sub-dominant harmonics. The estimate is calculated by multiplying the PhenomD estimate by max m/2, where max m is the largest m in the requested modes. If no modes are provided, it uses max m = 4, since (according to their papers), the highest m mode is the (4,4). This also updates the dictionary of filter length functions to have point to this function, so that `get_waveform_filter_length_in_time` will use it.

This also means that the calling `get_td_waveform` will now default to calling our internal `get_td_waveform_from_fd`, which avoids the lalsimulation issue of truncating higher modes too soon.

### DIFF
--- a/pycbc/waveform/waveform.py
+++ b/pycbc/waveform/waveform.py
@@ -801,6 +801,18 @@ def imrphenomd_length_in_time(**kwds):
     """
     return get_imr_length("IMRPhenomD", **kwds)
 
+def imrphenomhm_length_in_time(**kwargs):
+    """Estimates the duration of IMRPhenom waveforms that include higher modes.
+    """
+    if 'mode_array' in kwargs and kwargs['mode_array'] is not None:
+        maxm = max(m for _, m in kwargs['mode_array'])
+    else:
+        # the highest m for all of these is 4 (from the 4,4 mode)
+        maxm = 4
+    # we'll use the PhenomD length, then scale it by maxm / 2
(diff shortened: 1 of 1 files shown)


pycbc/waveform/waveform.py
```diff
@@ -801,6 +801,18 @@ def imrphenomd_length_in_time(**kwds):
...
+    """Estimates the duration of IMRPhenom waveforms that include higher modes.
+    """
+    if 'mode_array' in kwargs and kwargs['mode_array'] is not None:
+        maxm = max(m for _, m in kwargs['mode_array'])
+    else:
+        # the highest m for all of these is 4 (from the 4,4 mode)
+        maxm = 4
+    # we'll use the PhenomD length, then scale it by maxm / 2
+    dur = get_imr_length("IMRPhenomD", **kwargs)
+    return dur * maxm/2.
```
ahnitz, line 814: I don't think this is the right equation. You should modidfy get_imr_length to give you the duration from f_lower / (maxm/2)). The duration doesn't just scale.

cdcapano: Good point. I had tested a few points using the m/2 thing, and it seemed to work. You're right that scaling the frequency makes more sense. See the update.

PR #3351, opened 2020-07-03
Write n-loudest in the background

bhooshan-gadre:
To store only n-loudest while writing live background. This can reduce storage while saving the background.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -562,6 +562,9 @@ parser.add_argument('--store-psd', action='store_true')
 parser.add_argument('--output-background', type=str, nargs='+',
                     help='Takes a period in seconds and a file path and dumps '
                          'the coinc backgrounds to that path with that period')
+parser.add_argument('output-background-n-loudest', type=int, default=0,
+                    help="If given an integer (assumed positive), it stores loudest n triggers"
+                    "(not sorted) for each of the coinc background")
 
 parser.add_argument('--newsnr-threshold', type=float, default=0)
 parser.add_argument('--max-batch-size', type=int, default=2**27)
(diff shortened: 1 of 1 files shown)


bin/pycbc_live
```diff
@@ -872,8 +875,14 @@ with ctx:
                 bg_fn = os.path.join(args.output_background[1], bg_fn)
                 with h5py.File(bg_fn, 'w') as bgf:
                     for bg_ifos, bg_data, bg_time in bg_dists:
-                        ds = bgf.create_dataset(','.join(sorted(bg_ifos)),
-                                                data=bg_data, compression='gzip')
+                        if args.output_background_n_loudest:
+                            n_loudest = args.output_background_n_loudest
+                            assert (n_loudest > 0), "We can only store positive int loudest triggers."
+                            ds = bgf.create_dataset(','.join(sorted(bg_ifos)),
+                                                    data=-numpy.partition(-bg_data, n_loudest), compression='gzip')
```
ahnitz, line 882: @bhooshan-gadre This doesn't look like it should work based on what the partition function does. I.E. it returns and array with the same length as its input. See https://numpy.org/doc/stable/reference/generated/numpy.partition.html

bhooshan-gadre: Write. Modified accordingly. Forgot last part choosing n

bhooshan-gadre:
Hi. Pinging you back for merging this.

ahnitz:
@bhooshan-gadre Before we can do a final review, make sure the unittest are passing. At the moment, your PR is causing pycbc live to break.

bhooshan-gadre:
> @bhooshan-gadre Before we can do a final review, make sure the unittest are passing. At the moment, your PR is causing pycbc live to break.

Taken care of. I think everything is OK now.

PR #3352, opened 2020-07-03
Vectorize joint distribution

cdcapano:
This further speeds up the joint distribution by allowing it work on an array of values. Since many (all?) of the distributions in the distributions module are not vectorized, you still can't do vectorized prior evaluation for most practical purposes, but this is a step in that direction.

My main reason for doing this was to speed up the rvs function when you have many constraints. Currently, this is quite slow, because it draws one point at a time, evaluating each as they go. In this patch, the rvs will draw a batch of points, then use the now vectorized `contains` function to evaluate them all at once. It will then keep the points that pass the constraints, and continue until it's acquired the requested size. On each loop it will also increase the size of the draws based on the previous loop's acceptance (up to a maximum of a million draws). In my test, the constraints resulted in only ~1 in 10000 draws being kept. With that, it would take potentially hours to draw 1000 points. With this patch, it takes ~minute.

### DIFF
--- a/pycbc/distributions/joint.py
+++ b/pycbc/distributions/joint.py
@@ -225,9 +225,10 @@ class JointDistribution(object):
         isin = self.contains(parray)
         if not isin.any():
             if return_atomic:
-                return -numpy.inf
+                out = -numpy.inf
             else:
-                return numpy.full(parrray.shape, -numpy.inf)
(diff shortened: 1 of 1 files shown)


ahnitz:
@cdcapano Try rebasing from master

PR #3355, opened 2020-07-06
use radians consistently for setting angular bounds

ahnitz:
The angular distributions produce values in radians, but the bounds are in fractions of PI. We should consistently use radians so that min/max parameters can be given in the same units. This makes it a lot easier to do bounded sky location priors.

### DIFF
--- a/pycbc/distributions/angular.py
+++ b/pycbc/distributions/angular.py
@@ -82,8 +82,7 @@ class UniformAngle(uniform.Uniform):
                 bnds._max = bnds._max.__class__(bnds._max)
             else:
                 # create a Bounds instance from the given tuple
-                bnds = boundaries.Bounds(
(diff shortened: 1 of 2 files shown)


ahnitz:
The problem is that ra / Dec are in radians.bIf you set them with any other distribution or manually it must be in radians. It's inconsistent to have it use a different set of units for this one distribution. In my case, I know what the sky location mar roughly but can't just input that here because the units have changed between sections in the same config file without changing the name of the parameters.

ahnitz:
@cdcapano I see your point about setting the bounds at the maximum boundaries, but the typical case, is either the entire space (which is done by default anyway) or you have a sky location in mind with some error region, in which case it is most natural to set that in radians.

ahnitz:
@cdcapano Please re-review, I've updated the docs, and included the one-line fix for the issue that @a-r-williamson reported.

PR #3357, opened 2020-07-08
Fix pruning for single templates

connor-mcisaac:
When using a single template in the offline workflow you get a ZeroDivision error in the fit singles code.
This is because the minimum and maximum of the pruning parameters are equal, and in the method "which_bin" it divides by "float(maxpar - minpar)".
This PR shifts these limits slightly to avoid errors.

### DIFF
--- a/bin/hdfcoinc/pycbc_fit_sngls_binned
+++ b/bin/hdfcoinc/pycbc_fit_sngls_binned
@@ -207,8 +207,8 @@ if args.prune_param:
     prpars = get_pars(args, 'prune',
                       templatef['mass1'][:], templatef['mass2'][:],
                       templatef['spin1z'][:], templatef['spin2z'][:])
-    minprpar = min(prpars)
-    maxprpar = max(prpars)
+    minprpar = min(prpars) - 0.001 * abs(min(prpars))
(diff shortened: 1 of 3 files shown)


tdent:
I'm not sure this is fixing in the right place - could it not be more basically addressed by putting in a try-except here https://github.com/gwastro/pycbc/blob/master/pycbc/events/trigger_fits.py#L244 
with a comment like "corner case : if binning only one value an identical min/max may be obtained, avoid division by 0"
basically, we should only 'hack' the min/max values if and when necessary.

connor-mcisaac:
Yeah, could put in an if statement "if minpar == maxpar: frac = 0" to avoid try-excepts.

tdent:
or even frac = 0.5 (as it would be when shifting min and max by equal and opposite amounts) .. anyway, yes, I think the `if` approach is more transparent and maintainable

tdent:
Looks OK. It may be worth checking that we don't expect the function to work with a numpy array input - I am guessing we don't.

tdent:
(i.e. I guess it would be nice to have this function work for 'par' being a numpy array, but I assume that is currently optional)

connor-mcisaac:
It doesn't seem to be intended to take an array at the moment as "float(par - minpar)" would through an error for an array.

PR #3364, opened 2020-07-09
pass boundary types to dynesty

ahnitz:
This passes the pycbc boundary types (where applicable) to dynest. Only the two-sided reflective and the cyclic boundaries are supported as those are the only ones Dynesty has. So, notable, there is no one-sided reflective boundary. 

Example of how to write the boundary in the config file. Note that if you specify the boundary type, you also need to explicitly
set the min/max of the parameter as well. I am doing so here for inclination. The units are currently radians per PI, but that will change to PI in #3355, so be careful until that patch is applied. 

```
#btype-min-distance = reflected
#btype-max-distance = reflected


[prior-inclination]	[prior-inclination]
name = sin_angle	name = sin_angle
cyclic-inclination =
min-inclination = 0
max-inclination = 3.141592653
```

### DIFF
--- a/examples/inference/single/single.ini
+++ b/examples/inference/single/single.ini
@@ -21,10 +21,11 @@ strain-high-pass = 15
 sample-rate = 2048
 
 [sampler]
-name = emcee_pt
-ntemps = 4
-nwalkers = 100
-niterations = 300
+name = dynesty
+nlive = 100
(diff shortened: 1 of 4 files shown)


ahnitz:
For sanity, let's say this depends on #3355

PR #3368, opened 2020-07-10
Changes in arm length value

SSingh087:
Correction in values of time
Correction in the calculation of position

### DIFF
--- a/pycbc/detector.py
+++ b/pycbc/detector.py
@@ -341,17 +341,17 @@ class LISA(object):
                Returns the position of all 3 sattelites with each row
                correspoding to a single axis.
         """
-        ref_time = 2034 - Time(val=ref_time, format='gps', scale='utc').jyear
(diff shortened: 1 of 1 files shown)


pycbc/detector.py
```diff
@@ -341,17 +341,17 @@ def get_pos(self, ref_time):
...
                correspoding to a single axis.
         """
-        ref_time = 2034 - Time(val=ref_time, format='gps', scale='utc').jyear
+        ref_time = Time(val=ref_time, format='gps', scale='utc').jyear - 2034
         n = np.array(range(1, 4))
         kappa, _lambda_ = 0, 0
         alpha = 2. * np.pi * ref_time/1 + kappa
         beta_n = (n - 1) + (2. * np.pi/3) + _lambda_
-        a, L = 1., .1
+        a, L = 1., 0.0334229
```
ahnitz, line 349: If possible, can you add a reference for some of these numbers? @SSingh087

PR #3369, opened 2020-07-10
update support for ultranest

ahnitz:
* output logl along with samples for our standard plotting
* add support for cyclic parameters with ultranest

### DIFF
--- a/pycbc/inference/sampler/ultranest.py
+++ b/pycbc/inference/sampler/ultranest.py
@@ -168,7 +168,7 @@ class UltranestSampler(BaseSampler):
 
         # we'll do the resampling ourselves so we can pick up
         # additional parameters
-        try: # Remove me on next ultranest release
+        try:  # Remove me on next ultranest release
             wsamples = self.result['weighted_samples']['v']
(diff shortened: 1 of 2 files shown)


pycbc/inference/sampler/ultranest.py
```diff
@@ -22,16 +22,18 @@
...
 """
 
 
 from __future__ import absolute_import
 
 import logging
+import numpy
 
 from pycbc.inference.io.ultranest import UltranestFile
+from pycbc.io.hdf import dump_state
```
cdcapano, line 36: I see you're adding this import, but the checkpoint function still doesn't do anything. Are you planning on adding checkpointing in another PR?

ahnitz: @cdcapano This was an obvious change. Even before checkpointing I need to figure out how to use it's MPI mode in our code. Otherwise, there won't be many problems fast enough to bother with.

ahnitz: @cdcapano To be clear, I'm just using dump_state here to preserve the ultranest output in its original form if needed.

PR #3370, opened 2020-07-10
docs shouldn't run separately

ahnitz:


ahnitz:
Running the docs with & allowed travis to think things were done and would try to deploy the docs before they were built. This caused the docs to get removed.

PR #3373, opened 2020-07-12
simplify running pycbc inference with mpi

ahnitz:
This simplifies running under MPI with pycbc inference by removing the requirement for the --use-mpi and nprocesses options. You can no simply do

```
mpirun -np 5 pycbc_inference ....
```

With no extra options. It infers internally how many processes to use based on the mpi launch size. If you provide --use-mpi it will enforce that mpi is actually available and throw an error if not. Otherwise --use-mpi and --nproceses are optional in this case. A warning is raised if you provide them. In the future the --use-mpi option can then be removed entirely.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -27,7 +27,7 @@ import numpy
 
 import pycbc
 from pycbc import (distributions, transforms, fft,
-                   opt, scheme)
+                   opt, scheme, pool)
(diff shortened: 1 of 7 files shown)


ahnitz:
@cdcapano I don't think mpi4py builds without an existing mpi installation and the build becomes machine specific. If  you want that, it should probably be in its own file, separate from the others.

PR #3376, opened 2020-07-13
Dynesty: add nested samples and write to backup file

sum33it:
This PR addresses two issues:
i) Added an option to write nested samples in file.
ii) Fix the issue of 'not writing into backup file'.

### DIFF
--- a/pycbc/inference/io/dynesty.py
+++ b/pycbc/inference/io/dynesty.py
@@ -31,10 +31,11 @@ from pycbc.io.hdf import (dump_state, load_state)
 from .base_nested_sampler import BaseNestedSamplerFile
 from .posterior import write_samples_to_file, read_raw_samples_from_file
 
-@staticmethod
 class CommonNestedMetadataIO(object):
     """Provides functions for reading/writing dynesty metadata to file.
     """
+
+    @staticmethod
     def extra_args_parser(parser=None, skip_args=None, **kwargs):
         """Create a parser to parse sampler-specific arguments for loading
         samples.
@@ -138,12 +139,6 @@ class DynestyFile(CommonNestedMetadataIO, BaseNestedSamplerFile):
         bool :
             Whether or not the file is valid as a checkpoint file.
         """
-        #try:
-        #    group = '{}/{}'.format(self.samples_group, self.variable_params[0])
-        #    checkpoint_valid = self[group].size != 0
-        #except KeyError:
-        #    checkpoint_valid = False
-        #return checkpoint_valid
         try:
             if 'sampler_info/saved_state' in self:
                 load_state(self, path='sampler_info/saved_state')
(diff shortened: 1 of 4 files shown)


ahnitz:
@sum33it Please address the CC issues.

cdcapano:
@sum33it @ahnitz Is there any reason not to store the nested samples in the `samples` group? That would have symmetry with the MCMC samplers, since the samples written there for them are not really the posterior either. If you store the `logwt`, you could then load the posterior from the file on the file, or, using the extra args parser, tell it to just load the nested samples. So, under this model, the nested sampler doesn't even try to write the posterior at the end, it just writes all the nested samples. You then use extract samples to generate the posterior file, just like we do with MCMCs.

ahnitz:
@sum33it What did you think of @cdcapano's suggestion?

ahnitz:
@sum33it Poke

pycbc/inference/io/dynesty.py
```diff
@@ -25,12 +25,96 @@
...
+                             "a parser is provided.")
+        if skip_args is None:
+            skip_args = []
+        actions = []
+
+        if 'raw_samples' not in skip_args:
+            act = parser.add_argument(
+                "--raw-samples", type=numpy.array, default=False,
+                help="Raw samples are the unweighted samples obtained from "
+                     "dynesty sampler."
```
cdcapano, line 73: "dynesty sampler" -> "nested sampler"

pycbc/inference/io/dynesty.py
```diff
@@ -25,12 +25,96 @@
...
+        elif kwargs:
+            raise ValueError("No other keyword arguments should be provded if "
+                             "a parser is provided.")
+        if skip_args is None:
+            skip_args = []
+        actions = []
+
+        if 'raw_samples' not in skip_args:
+            act = parser.add_argument(
+                "--raw-samples", type=numpy.array, default=False,
```
cdcapano, line 71: `action='store_true'`

cdcapano: And remove `type=numpy.array`

pycbc/inference/io/dynesty.py
sum33it: But this particular function deals specifically with dynesty sampler.

sum33it:
@cdcapano  I guess it is in good shape. I have done the required tests and it works fine:
i) Writing raw samples to inference output file along with log-weights.
ii) `pycbc_inference_extract_sanples` and `pycbc_inference_plot_posterior' executables tested.

ahnitz:
@cdcapano Is there any reason this cannot now be merged?

pycbc/inference/io/dynesty.py
cdcapano: Ok, fair enough

cdcapano (approved):
Looks good!

PR #3379, opened 2020-07-14
Update docker for singularity

josh-willis:
This PR is to convert our Docker images so that when `cvmfs-singularity-sync` switches over to using Singularity directly to convert docker images, we can update our docker images to work with the new conversion.

I'm opening the PR now to start the Travis test, and see if there are any problems there with the new docker configuration. Even if that is fine, I do not want to merge this until we have confirmation of the switch in CVMFS singularity sync, so that we do not have non-functional singularity images published.

But of course comments and review will also be welcome.


josh-willis:
The change has been made upstream on the `cvmfs-singularity-sync` repo, so I am merging this now.

PR #3381, opened 2020-07-16
Silence warnings from child nodes in MPI

cdcapano:
Stops child nodes from printing logging warnings to stdout, so you don't get the same warning repeated a bunch of times when running `pycbc_inference` with MPI.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -82,8 +82,6 @@ opts = parser.parse_args()
 # setup log
 # If we're running in MPI mode, only allow the parent to print
 use_mpi, size, rank = pycbc.pool.use_mpi(opts.use_mpi, log=False)
-if use_mpi:
-    opts.verbose &= rank == 0
 pycbc.init_logging(opts.verbose)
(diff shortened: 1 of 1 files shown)


bin/inference/pycbc_inference
```diff
@@ -84,6 +84,8 @@ use_mpi, size, rank = pycbc.pool.use_mpi(opts.use_mpi, log=False)
 if use_mpi:
     opts.verbose &= rank == 0
 pycbc.init_logging(opts.verbose)
+if use_mpi and rank != 0:
```
ahnitz, line 87: I think this makes lines 84-85 superfluous now. As the docs seem to indicate disabling in this way is cumulative, i.e. that level and less critical ones are all disabled.

ahnitz:
@cdcapano Can you address the comments so this can be merged?

bin/inference/pycbc_inference
cdcapano: @ahnitz Sorry for forgetting about this. I've removed the previous `if` block.

ahnitz (approved):

PR #3382, opened 2020-07-17
add_statmap to output combined segments of all types of coincidence

GarethDaviesGW:
This is wanted to be able to output times which are in any type of coincidence

This was previously doable by adding the segments from each type of coinc separately, but this makes things a bit easier


GarethDaviesGW:
(Note that this change will make the inputs to the singles pastro file much easier to organise/review)

PR #3385, opened 2020-07-19
open up nested sampling to complicated / arbitrary distributions with fixed sample distribution

ahnitz:
### Concept

This adds a new distribution "FixedSamples" which allows for arbitrary and complicated distributions to be used within nested sampling in an approximate way. It is likely suitable for parameter estimation in many cases. Depends on #3315 

The idea is to draw a large number of points from any standard prior (even the arbitrary one) (say 10^6-10^9). These points can be transformed as one likes (using anything available for waveform transforms, including custom). The inverse CDF function is then done numerically by picking one of these points, so it can be used within nested sampling. This requires that a sufficient number of points is drawn from the prior for the parameters in question to adequately represent the space. 

### Example 
Example configuration file. In this example instead of sampling in ra and dec, we use the hanford time relative to geocentric, and the time delay between Hanford and Livingston. However, the prior is still defined in terms of ra and dec to give us an isotropic prior over the sky. 

```
[model]
name = single_template

#; This model precalculates the SNR time series at a fixed rate.
#; If you need a higher time resolution, this may be increased
sample_rate = 32768
low-frequency-cutoff = 30.0

[data]
instruments = H1 L1 V1
analysis-start-time = 1187008482
analysis-end-time = 1187008892
psd-estimation = median
psd-segment-length = 16
psd-segment-stride = 8
psd-inverse-length = 16
pad-data = 8
channel-name = H1:LOSC-STRAIN L1:LOSC-STRAIN V1:LOSC-STRAIN
frame-files = H1:H-H1_LOSC_CLN_4_V1-1187007040-2048.gwf L1:L-L1_LOSC_CLN_4_V1-1187007040-2048.gwf V1:V-V1_LOSC_CLN_4_V1-1187007040-2048.gwf
strain-high-pass = 15
sample-rate = 2048

[sampler]
name = ultranest
#nlive = 200

[sampler-burn_in]
burn-in-test = min_iterations
min-iterations = 100

[variable_params]
; waveform parameters that will vary in MCMC
tc =
distance =
inclination =
dh =
dhl =

[static_params]
; waveform parameters that will not change in MCMC
approximant = TaylorF2
f_lower = 30
mass1 = 1.3757
mass2 = 1.3757
polarization = 0

[prior-tc]
; coalescence time prior
name = uniform
min-tc = 1187008882.4
max-tc = 1187008882.5

[prior-distance]
#; following gives a uniform in volume
name = uniform_radius
min-distance = 10
max-distance = 60

[prior-inclination]
name = sin_angle

[prior-dh+dhl]
name = fixed_samples
subname = mysky        
sample-size = 1e6

[mysky_sample-ra+dec]
name = uniform_sky

[mysky_transform-dh+dhl]
name = custom
inputs = ra, dec
dh = det_tc('H1', ra, dec, 1187008882.0, relative=1)
dhl = det_tc('L1', ra, dec, 1187008882.0, relative=1) - det_tc('H1', ra, dec, 1187008882.0, relative=1)

[waveform_transforms-ra+dec]
name = mysky
```
single

### Possible Future Improvements
(will not be included in this PR and looking for help if this is interesting to people)
 * No PDF function is provided, so it cannot be used outside of nested sampling. It may be possible to use this is other cases however where the prior is flat after transformation (such as effectively is the case for nested sampling), but the hooks are not there at the moment to enable that kind of mode for mcmc samplers. 
* Cannot at the moment return the prior-space parameters. It would be nice to have these in the output file as they may be the parameters you want to measure anyway. One would have to instantiate the distribution directly with the same random state to recover these by hand.
* Support for more than 2 dimensions. At the moment, we only support 1 or 2 dimensions. The current algorithm for providing a an invese cdf would not work well for higher dimensions and the code isn't set up to let you try at the moment.
* Increased accuracy for invcdf. We could had some minimal interpolation scheme to provide an increased accuracy result, however, that might preclude transformations that are not smooth in the original space, but are in the transformed space. At the moment this works fine as we only need smoothness in the transformed space, but interpolation may require smoothness in both simultaneously.

### DIFF
--- a/pycbc/inference/sampler/base_cube.py
+++ b/pycbc/inference/sampler/base_cube.py
@@ -75,8 +75,8 @@ class CubeModel(object):
         """
         params = dict(zip(self.model.sampling_params, cube))
         self.model.update(**params)
-        #if self.model.logprior == -numpy.inf:
-        #    return -numpy.inf
+        if self.model.logprior == -numpy.inf:
+            return -numpy.inf
         return getattr(self.model, self.loglikelihood_function)
 
     def prior_transform(self, cube):
--- a/pycbc/inference/sampler/base_cube.py
+++ b/pycbc/inference/sampler/base_cube.py
@@ -75,8 +75,8 @@ class CubeModel(object):
         """
         params = dict(zip(self.model.sampling_params, cube))
         self.model.update(**params)
-        if self.model.logprior == -numpy.inf:
-            return -numpy.inf
+        #if self.model.logprior == -numpy.inf:
+        #    return -numpy.inf
         return getattr(self.model, self.loglikelihood_function)
 
     def prior_transform(self, cube):
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -663,7 +663,7 @@ def distance_from_chirp_distance_mchirp(chirp_distance, mchirp, ref_mass=1.4):
 
 
 _detector_cache = {}
-def det_tc(detector_name, ra, dec, tc, ref_frame='geocentric'):
+def det_tc(detector_name, ra, dec, tc, ref_frame='geocentric', relative=False):
     """Returns the coalescence time of a signal in the given detector.
 
     Parameters
@@ -685,16 +685,20 @@ def det_tc(detector_name, ra, dec, tc, ref_frame='geocentric'):
     float :
         The GPS time of the coalescence in detector `detector_name`.
     """
+    ref_time = tc
+    if relative:
+        tc = 0
+
     if ref_frame == detector_name:
         return tc
     if detector_name not in _detector_cache:
         _detector_cache[detector_name] = Detector(detector_name)
     detector = _detector_cache[detector_name]
     if ref_frame == 'geocentric':
-        return tc + detector.time_delay_from_earth_center(ra, dec, tc)
+        return tc + detector.time_delay_from_earth_center(ra, dec, ref_time)
     else:
         other = Detector(ref_frame)
-        return tc + detector.time_delay_from_detector(other, ra, dec, tc)
+        return tc + detector.time_delay_from_detector(other, ra, dec, ref_time)
 
 def optimal_orientation_from_detector(detector_name, tc):
     """ Low-level function to be called from _optimal_dec_from_detector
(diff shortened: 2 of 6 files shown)


ahnitz:
On thing I forgot to mention. The distribution automatically provides access to the backward transformation if needed so that the waveform parameters can be provided. See example for useage.

pycbc/inference/sampler/base_cube.py
```diff
@@ -75,8 +75,8 @@ def log_likelihood(self, cube):
         """
         params = dict(zip(self.model.sampling_params, cube))
         self.model.update(**params)
-        if self.model.logprior == -numpy.inf:
-            return -numpy.inf
+        #if self.model.logprior == -numpy.inf:
```
ahnitz, line 78: @cdcapano This change is a question for you. I'm not really sure why this is here, and it isn't possible to make this work when you can't provide a logpdf function. For nested sampling, I would have though this would not be possible. Is this here to handle constraints or something like that?

ahnitz: @cdcapano Can you clarify why this is done?

cdcapano: @ahnitz Yes, this is to handle constraints. Since constraints can be arbitrary, you can't account for them when reshaping the hypercube. The only thing you can do is just treat points outside of the constraints as having 0 likelihood, which is what this line is enforcing. I don't follow, why is this an issue for this PR?

ahnitz: @cdcapano As I said originally, there is no log prior function to provide. It will remain non implemented.

ahnitz: @cdcapano To clarify more, we don't know the jacobian for a generic transformation which could be applied, so there is no logprior function at all for this class. The prior is only determined in the original space by sampling .

ahnitz: @cdcapano Actually, I guess a correct thing to do would be to return a constant, right? It won't be used anyway and is technically correct (up to a constant which I won't calculate) for the sample parameters.

ahnitz:
@cdcapano It now returns a dummy logp, can this be merged now?

cdcapano (approved):

pycbc/inference/sampler/base_cube.py
cdcapano: Yeah, that looks fine.

PR #3389, opened 2020-07-21
Reset _saved in roll-function.

MarlinSchaefer:
Resetting the memory of the Array after the roll is applied, rather then before.

### DIFF
--- a/pycbc/types/array.py
+++ b/pycbc/types/array.py
@@ -825,7 +825,6 @@ class Array(object):
     def roll(self, shift):
         """shift vector
         """
-        self._saved = LimitedSizeDict(size_limit=2**5)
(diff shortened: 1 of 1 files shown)


ahnitz:
@MarlinSchaefer We might have made this mistake elsewhere. Do you see any other instances like this?

ahnitz:
@MarlinSchaefer Travis errors don't appear related to your patch, but instead it looks like some LIGO links are down.

MarlinSchaefer:
@ahnitz Can/Should we just restart the Travis runs in that case?

ahnitz:
@MarlinSchaefer Please rebase your branch from the current master. It should pass travis then.

PR #3391, opened 2020-07-23
Fixes to live combine trigger fits

GarethDaviesGW:
Allow date format to not be YYYY_MM_DD, but (e.g.) YYYY_M_D
fix issue which would break things if no triggers are found in the duration bin

Note that this does not affect the validity of reviews - it is mostly due to dealing with the O2 triggers

### DIFF
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -58,9 +58,9 @@ for f in files:
     if 'days_since_epoch' in fits_f.attrs:
         analysis_dates += [fits_f.attrs['days_since_epoch']]
     elif 'analysis_date' in fits_f.attrs:
-        analysis_year = int(fits_f.attrs['analysis_date'][:4])
-        analysis_month = int(fits_f.attrs['analysis_date'][5:7])
-        analysis_day = int(fits_f.attrs['analysis_date'][8:10])
+        analysis_year = int(fits_f.attrs['analysis_date'].split('_')[0])
+        analysis_month = int(fits_f.attrs['analysis_date'].split('_')[1])
+        analysis_day = int(fits_f.attrs['analysis_date'].split('_')[2])
         n_since_epoch = (datetime.date(analysis_year,
                                        analysis_month,
                                        analysis_day)
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -68,15 +68,14 @@ for f in files:
                          - datetime.date(2000, 1, 1)).days
         analysis_dates += [n_since_epoch]
     else:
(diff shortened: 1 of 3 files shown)


GarethDaviesGW:
(And no rush on this - obviously paper takes precedence)

bin/live/pycbc_live_combine_single_fits
```diff
@@ -58,30 +58,33 @@ for f in files:
...
-        re_string = '([12]\d{3}_(0[1-9]|1[0-2])_(0[1-9]|[12]\d|3[01]))'
+        re_string = '([12]\d{3}_(0?[1-9]|1[0-2])_(0?[1-9]|[12]\d|3[01]))'
         m = re.search(re_string, f)
-        analysis_year = int(m.group(0)[:4])
-        analysis_month = int(m.group(0)[5:7])
-        analysis_day = int(m.group(0)[8:10])
+        date_array = m.group(0).split('_')
+        analysis_year = int(date_array[0])
+        analysis_month = int(date_array[1])
+        analysis_day = int(date_array[2])
```
titodalcanton, line 79: I am a bit confused by the parentheses in the RE and by the split on underscores. Can you not just do this:
```python
re_string = '([12]\d{3})_(0?[1-9]|1[0-2])_(0?[1-9]|[12]\d|3[01])'
m = re.search(re_string, f)
analysis_year = int(m.group(1))
analysis_month = int(m.group(2))
analysis_day = int(m.group(3))
```

titodalcanton:
I cannot speak for the functionality (I assume you tested it), but modulo the comment above this looks good.

bin/live/pycbc_live_combine_single_fits
GarethDaviesGW: that is something I didn't know you could do. That makes more sense

titodalcanton: Makes more sense now, but if my parsing of the RE is correct, should it not be
```
([12]\d{3})_(0?[1-9]|1[0-2])_(0?[1-9]|[12]\d|3[01])
```
instead of
```
([12]\d{3}_(0?[1-9]|1[0-2])_(0?[1-9]|[12]\d|3[01]))
```
(a subtle difference in parentheses)?

ahnitz:
@GarethDaviesGW @titodalcanton What's the status of this?

titodalcanton:
I think I still have an open question to @GarethDaviesGW about the correctness of the regexp.

GarethDaviesGW:
That's right - I had just forgotten about this. I'll implement Tito's
suggestion and then merge

GarethDaviesGW:
(well I'll merge after approval)

GarethDaviesGW:
oops - that was the wrong button

titodalcanton (approved):
The change looks good now, but I am mystified by the Travis failure...

PR #3393, opened 2020-07-24
update pycbc_submit_dax to deal with cache files behind authorisation

GarethDaviesGW:
Change to pycbc_submit_dax to use resolve_url to allow cache file download from web pages behind authentication

Replaces the curl method, as if this is attempted with a password-protected file, then it will return the html of the login page and not fail.

### DIFF
--- a/bin/pycbc_submit_dax
+++ b/bin/pycbc_submit_dax
@@ -74,9 +74,18 @@ while true ; do
                  cat ${FILE_URL} >> _reuse.cache
                  FOUND_URL=0
               else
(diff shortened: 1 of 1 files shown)


GarethDaviesGW:
I should also check with @ahnitz about this as well, particularly with respect to ensuring that the removal of the curl command is the correct thing to do, and that it will not screw things up when getting files from atlas/non-LIGO locations

ahnitz:
@GarethDaviesGW I don't think this is a problem.

PR #3401, opened 2020-07-30
guess channel name from version

ahnitz:
Allows the catalog interface to read whatever version frame that gwosc presents (assuming the channel name convention holds true). For the longer term, I've asked Jonah if the channel name can be put in the metadata. 

Enables this to work, since one frame is V1 and the other is V2

```
from pycbc.catalog.catalog import list_catalogs
from pycbc.catalog import Merger

sources = ["GW190412", "GW190814"]
ifos = ['H1', 'L1', 'V1']

for source in sources:
    m = Merger(source, source='O3_Discovery_Papers')
    data = {ifo: m.strain(ifo) for ifo in ifos}
    psd = {ifo: data[ifo].psd(2) for ifo in ifos}
```

### DIFF
--- a/pycbc/catalog/__init__.py
+++ b/pycbc/catalog/__init__.py
@@ -25,6 +25,8 @@
 """ This package provides information about LIGO/Virgo detections of
 compact binary mergers
 """
+
+import os
 import numpy
 from . import catalog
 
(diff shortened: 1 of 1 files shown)


ahnitz:
@spxiwh Should be good to go now

ahnitz:
@spxiwh Poke

PR #3406, opened 2020-08-03
polarization angle marginalization in single_template

shobhitrphy:


pycbc/inference/models/single_template_p.py
```diff
@@ -0,0 +1,123 @@
...
+        vloglr += abs(shloglr) + hhloglr 
+        #putting all vloglr values in exp(vloglr)
+        temp=numpy.exp(vloglr)
+        #summing over the values in the array exp(vloglr)
+        loglm=(numpy.sum(temp))
+        
+        #To handle the case where loglm=0; assigned a very small value to loglm that is not equal to 0
+        #taking the log of loglm to obtain new loglikelihood ratio marginalized over pol
+        
+        if loglm!=0:
```
ahnitz, line 120: I suspect you should not need this switch here if you implement the above. The reason you were probably getting identically zero is due to numerical error.

pycbc/inference/models/single_template_p.py
```diff
@@ -0,0 +1,123 @@
...
+            sh = self.sh[ifo].at_time(p['tc'] + dt) * htf
+            shloglr += sh
+            hhloglr += self.hh[ifo] * abs(htf) ** 2.0
+
+        vloglr = numpy.log(scipy.special.i0e(abs(shloglr)))
+        
+        #An array of vloglr for each polarization value
+        vloglr += abs(shloglr) + hhloglr 
+        #putting all vloglr values in exp(vloglr)
+        temp=numpy.exp(vloglr)
```
ahnitz, line 113: Line 113 and 115 should be combined into a single step with https://docs.scipy.org/doc/scipy-0.19.0/reference/generated/scipy.misc.logsumexp.html. It has much better accuracy than taking an exponential directly then summing.

pycbc/inference/models/single_template_p.py
```diff
@@ -0,0 +1,123 @@
...
+        p = self.current_params.copy()
+        p.update(self.static_params)
+
+        if self.time is None:
+            self.time = p['tc']
+
+        shloglr = hhloglr = 0
+        for ifo in self.sh:
+            fp, fc = self.det[ifo].antenna_pattern(p['ra'], p['dec'],
+                                                   p['polarization'],
```
ahnitz, line 94: It appears you are still getting polarization from the param dict. For marginalization, polarization really shouldn't be a parameter anymore, instead you should make the vector in the init of this function (with some option for how finely it is sampled). Just make a vector that spans  from 0 to 2pi with a configureable number of grid points.

ahnitz:
@shobhitrphy Also, it may be good to implement this as an option to the regular single template class, i.e. you set say marginalize_polarization=NUM_SAMPLES and then it does so as you've implemented, but otherwise it does what the code used to do and you can sample over polarization (or use a fixed value) as you like.

pycbc/inference/models/single_template_p.py
shobhitrphy: Alright. Will implement that.

ahnitz:
@shobhitrphy I think one of the Travis failures was just intermittent, if this passes the automated tests, I think this might be close to merging.

ahnitz (approved):

PR #3410, opened 2020-08-05
Added high-frequency cutoff to pycbc_single_template.

MarlinSchaefer:
Added a hard high-frequency cutoff to pycbc_single_tempalte.


bin/pycbc_single_template
```diff
@@ -315,6 +317,10 @@ with ctx:
                                     f_lower=template_flow, delta_f=delta_f,
                                     delta_t=gwstrain.delta_t)
             template = tc.multiply_and_add(tp, opt.u_val)
+    
+    kmax = int(opt.high_frequency_cutoff / template.delta_f)
```
ahnitz, line 321: @MarlinSchaefer You should put this in check that the option was given, since we don't want it to be required. 

i.e. 

```
if opt.high_frequency_foff:
      ....
```

MarlinSchaefer: Added the check

PR #3411, opened 2020-08-06
add option to throughput plots to weight results by duration of job

ahnitz:
If you allow a large variation in analysis times it would be interesting to compare the weighted distributions for these plots. This adds an option to do that.

### DIFF
--- a/bin/hdfcoinc/pycbc_plot_throughput
+++ b/bin/hdfcoinc/pycbc_plot_throughput
@@ -26,12 +26,14 @@ fig, (ax1, ax2, ax3) = pl.subplots(3,1,figsize=(10,10))
 for pa in args.input_file:
     f = h5py.File(pa, 'r')
     ifo = tuple(f.keys())[0]
+
     dur = f['%s/search/end_time' % ifo][:] - f['%s/search/start_time' % ifo][:]
     dur /= dur.mean()
     if args.duration_weighted:
         w = dur
     else:
-        w = 1
+        w = None
+
     if 'templates_per_core' in f['%s/search' % ifo].keys():
         tpc = f['%s/search/templates_per_core' % ifo][:]
     else:
@@ -44,6 +46,7 @@ for pa in args.input_file:
         stf = f['%s/search/setup_time_fraction' % ifo][:]
     else:
         stf = None
+
     if tpc is not None:
         label = str(ifo) + ': Harmonic mean  - ' + str(hmean(tpc))
         if args.duration_weighted:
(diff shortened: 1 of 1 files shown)


bin/hdfcoinc/pycbc_plot_throughput
```diff
@@ -37,28 +46,34 @@ for pa in args.input_file:
         stf = f['%s/search/setup_time_fraction' % ifo][:]
     else:
         stf = None
+
     if tpc is not None:
-        label = str(ifo) + ': Harmonic mean - ' + str(hmean(tpc))
-        ax1.hist(tpc, 100, color=ifo_color(ifo), alpha = 0.65, label = label)
+        label = str(ifo) + ': Harmonic mean  - ' + str(hmean(tpc))
+        if args.duration_weighted:
+            label = str(ifo) + ': Average  - ' + str((tpc*w).mean())
```
josh-willis, line 53: So, I'd like to request that our default still be to use the harmonic mean.  While you are adjusting the calculation for the detector time analyzed by each job, I think it is still also important to use the inverse weighting of job *runtime* that the harmonic mean implies.  Mostly, this is a reflection of what templates-per-core is: if you have that number, you should be able to multiply the total time analyzed by the size of the bank, and divide by the mean TPC to get the total CPU time required for the analysis.  So since we use TPC as a denominator, we should calculate the "average" TPC by averaging the reciprocal of each individual job's TPC.

In O1 and O2 analyses, where we did not sort the bank by mchirp, this gave computing usage estimates that were a few tens of percents closer to what was observed. Since O3, where we sort by mchirp *and* use the injection rejector, this has become much more important: many injection jobs will finish almost immediately, because the part of the bank they cover is just not sensitive to that injection set.  You get a very skewed view of what the rutime will be if you average in all of these very fast jobs with huge TPC, but if you do their harmonic mean, then it more correctly predicts the total analysis time, which is dominated by the jobs where that *doesn't* happen.

I believe that to do both things---harmonic mean of templates-per-core, but weighted by detector analysis time for each job---that what you would do is:
```!python
avg_tpc = 1.0/(w/tpc).mean()
```
except we'd have to worry about any vanishing TPC entries (I don't know that that happens in practice, but still). Probably a masked array could do this?

Thoughts? We could of course do all three: harmonic mean, weighted average, weighted harmonic mean. But that starts to be a lot.

ahnitz: @josh-willis I've updated with your suggestion. At the moment, I think we don't worry about zero valued entries. If that occurs the hmean from scipy would already have had an issue I think.

PR #3412, opened 2020-08-07
speed up trigger fitting 

ahnitz:
@GarethDaviesGW @tdent I've gotten around to pushing upstream some changes I made locally for the high mass ratio search I did. Due to the large number of templates the fit code was pretty slow in default configuration.  

This patch
* Speeds up fit_by_template generally by switching to faster versions of the same calls
* speeds up fit_over_param by many orders of magnitude in the specific case of a single dimension

### DIFF
--- a/bin/hdfcoinc/pycbc_fit_sngls_by_template
+++ b/bin/hdfcoinc/pycbc_fit_sngls_by_template
@@ -289,8 +289,15 @@ counts_above = []
 fits = []
 tpars = []
 
-for tnum in trange:
-    stat_in_template = stat[tid == tnum]
+tsort = tid.argsort()
+tid = tid[tsort]
+stat = stat[tsort]
+
+left = np.searchsorted(tid, trange, side='left')
+right = np.searchsorted(tid, trange, side='right')
+
+for j, tnum in enumerate(trange):
+    stat_in_template = stat[left[j]:right[j]]
     count_above = len(stat_in_template)
     count_total = count_in_template[tnum]
     if count_above == 0:
@@ -311,8 +318,12 @@ for tnum in trange:
                                                     (tnum - tmin, tmax - tmin))
 
 logging.info("Calculating median sigma for each template")
-median_sigma = [np.sqrt(np.median(trigf[args.ifo + '/sigmasq'][reg])) for reg in
-                    trigf[args.ifo + '/sigmasq_template'][:]]
+
+sigma_regions = trigf[args.ifo + '/sigmasq_template'][:]
+median_sigma = []
+for i, reg in enumerate(sigma_regions):
+    strigs = trigf[args.ifo + '/sigmasq'][reg]
+    median_sigma.append(np.median(strigs)**0.5)
 
 outfile = h5py.File(args.output, 'w')
 # store template-dependent fit output
--- a/bin/hdfcoinc/pycbc_fit_sngls_by_template
+++ b/bin/hdfcoinc/pycbc_fit_sngls_by_template
@@ -297,12 +297,13 @@ stat = stat[tsort]
 left = np.searchsorted(tid, trange, side='left')
 right = np.searchsorted(tid, trange, side='right')
 
+logging.info("Fitting ...")
 for j, tnum in enumerate(trange):
     stat_in_template = stat[left[j]:right[j]]
     count_above = len(stat_in_template)
     count_total = count_in_template[tnum]
     if count_above == 0:
-        # 'stupid' value to indicate no data, shouldn't hurt if 1/alpha is averaged
+        # default/sentinel value to indicate no data, shouldn't hurt if 1/alpha is averaged
         alpha = -100.
     else:
         alpha, sig_alpha = trstats.fit_above_thresh(
@@ -319,15 +320,13 @@ for j, tnum in enumerate(trange):
                                                     (tnum - tmin, tmax - tmin))
 
 logging.info("Calculating median sigma for each template")
-
 sigma_regions = trigf[args.ifo + '/sigmasq_template'][:]
 median_sigma = []
 for reg in sigma_regions:
     strigs = trigf[args.ifo + '/sigmasq'][reg]
-    median_sigma.append(np.median(strigs)**0.5)
+    median_sigma.append(np.median(strigs) ** 0.5)
 
 outfile = h5py.File(args.output, 'w')
-# store template-dependent fit output
 outfile.create_dataset("template_id", data=trange)
 outfile.create_dataset("count_above_thresh", data=counts_above)
 outfile.create_dataset("fit_coeff", data=fits)
@@ -346,4 +345,3 @@ outfile.attrs.create("analysis_time", data=total_time)
 
 outfile.close()
 logging.info('Done!')
-
--- a/bin/hdfcoinc/pycbc_fit_sngls_by_template
+++ b/bin/hdfcoinc/pycbc_fit_sngls_by_template
@@ -293,6 +293,7 @@ tsort = tid.argsort()
 tid = tid[tsort]
 stat = stat[tsort]
 
+# Get range of tid values which are the same
 left = np.searchsorted(tid, trange, side='left')
 right = np.searchsorted(tid, trange, side='right')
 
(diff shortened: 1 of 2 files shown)


bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -289,8 +289,15 @@ counts_above = []
...
-    stat_in_template = stat[tid == tnum]
+tsort = tid.argsort()
+tid = tid[tsort]
+stat = stat[tsort]
+
+left = np.searchsorted(tid, trange, side='left')
+right = np.searchsorted(tid, trange, side='right')
+
+for j, tnum in enumerate(trange):
+    stat_in_template = stat[left[j]:right[j]]
```
GarethDaviesGW, line 300: to ensure that I understand this:
- sort by template ID
- use searchsorted to find the extrema of indices of where template ID is that one
- use the extrema indices to select the stat values

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -311,8 +318,12 @@ for tnum in trange:
...
 
 logging.info("Calculating median sigma for each template")
-median_sigma = [np.sqrt(np.median(trigf[args.ifo + '/sigmasq'][reg])) for reg in
-                    trigf[args.ifo + '/sigmasq_template'][:]]
+
+sigma_regions = trigf[args.ifo + '/sigmasq_template'][:]
+median_sigma = []
+for i, reg in enumerate(sigma_regions):
+    strigs = trigf[args.ifo + '/sigmasq'][reg]
+    median_sigma.append(np.median(strigs)**0.5)
```
GarethDaviesGW, line 326: what is the `i` variable used for in this case? I'm not sure how this is faster than the list comprehension

bin/hdfcoinc/pycbc_fit_sngls_over_multiparam
```diff
@@ -117,13 +117,36 @@ nabove_smoothed = []
...
+
+    ntsum = ntotal.cumsum()
+    nasum = nabove.cumsum()
+    invsum = invalphan.cumsum()
+    num = right - left
+
+    if tcount: ntotal_smoothed = (ntsum[right] - ntsum[left]) / num
+    nabove_smoothed = (nasum[right] - nasum[left]) / num
+    invmean = (invsum[right] - invsum[left]) / num
+    alpha_smoothed = nabove_smoothed / invmean
```
GarethDaviesGW, line 139: looks good to me

bin/hdfcoinc/pycbc_fit_sngls_by_template
ahnitz: @GarethDaviesGW Yes, that's right.

bin/hdfcoinc/pycbc_fit_sngls_by_template
ahnitz: Ok, I think I might have used it for vebose timing at one part. Probably not needed anymore.

bin/hdfcoinc/pycbc_fit_sngls_over_multiparam
```diff
@@ -117,13 +117,36 @@ nabove_smoothed = []
...
+
+else:
+    for i in range(len(nabove)):
+        if len(parvals) > 1:
```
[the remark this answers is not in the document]
ahnitz: You are right, seems not to be needed.

bin/hdfcoinc/pycbc_fit_sngls_by_template
ahnitz: I've now removed this

bin/hdfcoinc/pycbc_fit_sngls_over_multiparam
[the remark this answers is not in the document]
ahnitz: Now gone

ahnitz:
@GarethDaviesGW Please re-check

tdent:
@ahnitz sorry, I haven't been able to look at this yet (weekend, travel, family visit)
can you give me a chance to see if it makes sense, ie is sufficiently explained?

ahnitz:
@tdent Yes, there is no rush. I'm just trying to clear out some of my personal backlog of hacks I've needed for searches but really should be pushed back to master.

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -289,8 +289,15 @@ counts_above = []
 fits = []
 tpars = []
 
-for tnum in trange:
-    stat_in_template = stat[tid == tnum]
+tsort = tid.argsort()
+tid = tid[tsort]
+stat = stat[tsort]
+
+left = np.searchsorted(tid, trange, side='left')
```
tdent, line 296: 1 line of comment here explaining that there will be multiple entries of the same tid in the sorted array and left/right catch the appropriate range? 
Also do we know if it works if/when some templates have zero triggers?

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -311,8 +318,12 @@ for tnum in trange:
                                                     (tnum - tmin, tmax - tmin))
 
 logging.info("Calculating median sigma for each template")
-median_sigma = [np.sqrt(np.median(trigf[args.ifo + '/sigmasq'][reg])) for reg in
-                    trigf[args.ifo + '/sigmasq_template'][:]]
+
+sigma_regions = trigf[args.ifo + '/sigmasq_template'][:]
+median_sigma = []
+for reg in sigma_regions:
```
tdent, line 324: is this expected to be faster than the list comprehension or just easier to absorb as more shorter lines of code?

bin/hdfcoinc/pycbc_fit_sngls_over_multiparam
```diff
@@ -117,13 +117,35 @@ nabove_smoothed = []
...
 rang = numpy.arange(0, len(nabove))
-for i in range(len(nabove)):
-    dsq = dist(i, rang)
-    l = (dsq < 1)
-    
-    if tcount: ntotal_smoothed.append(ntotal[l].mean())
-    nabove_smoothed.append(nabove[l].mean())
-    alpha_smoothed.append(nabove_smoothed[i] / invalphan[l].mean())
+
+if len(parvals) == 1:
```
tdent, line 121: add comment here explaining why we want to treat this case separately?

bin/hdfcoinc/pycbc_fit_sngls_over_multiparam
```diff
@@ -117,13 +117,35 @@ nabove_smoothed = []
...
+    sort = parvals[0].argsort()
+    parvals[0] = parvals[0][sort]
+    ntotal = ntotal[sort]
+    nabove = nabove[sort]
+    invalphan = invalphan[sort]
+
+    left = numpy.searchsorted(parvals[0], parvals[0] - args.smoothing_width[0])
```
tdent, line 128: add comment explaining what is being done (ie getting all templates within a sliding top hat parameter window around the template of interest)

bin/hdfcoinc/pycbc_fit_sngls_over_multiparam
```diff
@@ -117,13 +117,35 @@ nabove_smoothed = []
...
+    right = numpy.searchsorted(parvals[0], parvals[0] + args.smoothing_width[0]) - 1
+
+    ntsum = ntotal.cumsum()
```
tdent, line 131: again maybe 1 line of comment explaining what is being done (precomputing the sums that will be used for window averages)

bin/hdfcoinc/pycbc_fit_sngls_by_template
ahnitz: It works when there are zero triggers in my test. In that case, left and right end up the same value and if you follow the later logic you get empty arrays for the stat values.

ahnitz: This was not such an uncommon case in my use actually given the number of templates and how few triggers I kept, there were some templates without triggers. (thankfully the average fit was very smooth anyway over a large region).

bin/hdfcoinc/pycbc_fit_sngls_by_template
ahnitz: I think it's just easier to read in this case.

ahnitz:
@tdent Ok, I think this should have the requested comments now.

ahnitz:
Travis failure seems to be due to Tuesday maintenance at CIT taking down parts of GWOSC

PR #3413, opened 2020-08-11
Parallelize optimal_snr jobs

GarethDaviesGW:
Given certain approximants and amount of injections, optimal_snr jobs can take a long time causing jobs to be evicted/restarted in the workflow

This splits optimal_snr jobs in the same way as the findtrigs jobs, and recombines using a new executable, pycbc_optimal_snr_merge

This _will_ break current configs due to changing the way optimal_snr is used and optimal_snr_merge config requirements.

Newly required config tags:
- in `[executables]`, `optimal_snr_merge = ${which:pycbc_optimal_snr_merge}`
- in `[workflow-optimal-snr]`: `parallelization-factor = 5`
- add `[optimal_snr_merge]`

Tested in multiifo workflow, but as the 2-ifo workflow calls it in the same way, this should work there as well


pycbc/workflow/injection.py
```diff
@@ -50,19 +50,62 @@ def veto_injections(workflow, inj_file, veto_file, veto_name, out_dir, tags=None
...
+        opt_snr_exe = PyCBCOptimalSNRExecutable(workflow.cp, 'optimal_snr',
+                                                ifos=workflow.ifos,
+                                                out_dir=out_dir,
+                                                tags=tags + [str(i)])
+        node = opt_snr_exe.create_node(workflow, inj_file, precalc_psd_files,
+                                       group_str)
+        opt_snr_split_files += [node.output_files[0]]
+        workflow += node
+
+    opt_snr_merge_exe = PyCBCOptimalSNRMergeExecutable(workflow.cp,
```
ahnitz, line 101: Can you make this so that this job is bypassed if there is no split? Presumably just an if statment on whether factor is 1 or not.

pycbc/workflow/injection.py
```diff
@@ -50,19 +50,62 @@ def veto_injections(workflow, inj_file, veto_file, veto_name, out_dir, tags=None
...
         tags = []
 
-    node = Executable(workflow.cp, 'optimal_snr', ifos=workflow.ifos,
-                      out_dir=out_dir, tags=tags).create_node()
-    node.add_input_opt('--input-file', inj_file)
-    node.add_input_list_opt('--time-varying-psds', precalc_psd_files)
-    node.new_output_file_opt(workflow.analysis_time, '.xml', '--output-file')
-    workflow += node
-    return node.output_files[0]
+    factor = int(workflow.cp.get_opt_tags('workflow-optimal-snr',
```
ahnitz, line 85: I think this will fail if not provided. I don't think the option should be required, so can you check for the option and set factor to 1 if not provided?

pycbc/workflow/injection.py
GarethDaviesGW: Rather than setting factor to 1 and then checking it, I've wrapped it in a try/except (EAFP)

titodalcanton:
Instead of adding the new script `pycbc_optimal_snr_merge`, would it be possible to just use `ligolw_add`?

GarethDaviesGW:
> Instead of adding the new script `pycbc_optimal_snr_merge`, would it be possible to just use `ligolw_add`?

I'm not sure - I hadn't seen that functionality before
I'm not sure how it could work within a workflow, as surely it would still need an executable to wrap it in pegasus?

titodalcanton:
I am thinking about this:
```
[executables]
optimal_snr_merge = ligolw_add
```
I do not see an obvious reason why a specific script would be needed instead, as it is just merging tables.

ahnitz:
@GarethDaviesGW ligolw_add is an executable. I think @titodalcanton is saying why invent a new executable instead of using an existing one that likely already can do this. It takes options slightly differently, but otherwise, I think it should work.

GarethDaviesGW:
thanks - I hadnt seen the executable - I'll put that in rather than the optimal_snr_merge code

pycbc/workflow/injection.py
GarethDaviesGW: Done as suggested in the end

pycbc/workflow/injection.py
GarethDaviesGW: Done as suggested

GarethDaviesGW:
I've updated to use the ligolw_add executable as well and removed the optimal_snr_merge

GarethDaviesGW:
The last codeclimate and codebeat issues are ones which we have generally ignored before - the overwriting of create_node in particular looks to be something done a lot elsewhere

bin/pycbc_optimal_snr
```diff
@@ -94,6 +94,13 @@ class TimeVaryingPSD(object):
             self._curr_psd_index[curr_pid] = index
         return self._curr_psd[curr_pid]
 
+def parse_injection_range(num_inj, rangestr):
+    part = int(rangestr.split('/')[0])
+    pieces = int(rangestr.split('/')[1])
+    tmin =  int(num_inj / float(pieces) * part)
+    tmax =  int(num_inj / float(pieces) * (part+1))
+    return tmin, tmax
```
titodalcanton, line 102: Will this function guarantee that the final set of injections is identical to the starting set, i.e. is rounding/truncating going to be an issue in some edge cases? Would it be safer to specify instead directly the range of injection indices tmin/tmax?

GarethDaviesGW: tbh this is copy/pasted from how this is done in the parallelization of coinc jobs - I'm fairly sure there won't be an issue with this, as tmax from part n will be the same as tmin from part n+1, provided pieces is the same, so the indexing should be fine

GarethDaviesGW: specifying tmin and tmax directly wouldn't work in the case that you don't know a priori the number of injections

titodalcanton: Ah, I think I was unnecessarily worried by the floats. Looks like you could achieve the same just with integers though,
```python
    tmin =  num_inj * part // pieces
    tmax =  num_inj * (part+1) // pieces
```
which maybe is more clearly safe.

GarethDaviesGW: updated as suggested

pycbc/workflow/injection.py
```diff
@@ -50,19 +51,67 @@ def veto_injections(workflow, inj_file, veto_file, veto_name, out_dir, tags=None
...
+        opt_snr_split_files += [node.output_files[0]]
+        workflow += node
+
+    llwadd_exe = LigolwAddExecutable(workflow.cp, 'optimal_snr_merge',
+                                     ifos=workflow.ifos, out_dir=out_dir,
+                                     tags=tags)
+    merge_node = llwadd_exe.create_node(workflow.analysis_time,
+                                        opt_snr_split_files,
+                                        use_tmp_subdirs=False)
+    merge_node.add_opt('--ilwdchar-compat')
```
titodalcanton, line 111: Would it be better to keep this option in the config file? We may run the workflow under an environment with an older version of `ligolw_add` that does not yet have the option.

GarethDaviesGW: that sounds fair

titodalcanton: In fact it seems even fairly old CVMFS PyCBC envs do have the option, so this actually seems safer than I thought. Still, presumably at some point the option will be removed from `ligolw_add`...

titodalcanton:
This looks good to me, but I am a bit too rusty on the workflow code to feel comfortable approving that change. Maybe best if @spxiwh or @ahnitz give the final approval.

GarethDaviesGW:
tested with and without the parallelize option in config for workflow, and using the updated optimal_snr exe

PR #3414, opened 2020-08-11
Created option to set a high-frequency cutoff in pycbc_make_skymap

MarlinSchaefer:
Added an optional argument to pycbc_make_skymap.
This argument is passed to pycbc_single_template and SingleCoincForGraceDB.

For SingleCoincForGraceDB to accept a high-frequency cutoff changes to pycbc/io/live.py were required. I think the place I put these changes at don't interfere with anything else. I also think that implementing the high-frequency cutoff the way done here will pass it to Baystar.

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -271,6 +271,7 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, ifar, ifos, th
 
     kwargs = {'psds': {ifo: followup_data[ifo]['psd'] for ifo in ifos + follow_up},
               'low_frequency_cutoff': f_low,
+              'high_frequency_cutoff': f_upper,
               'followup_data': followup_data,
               'channel_names': channel_names}
 
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -61,7 +61,7 @@ def default_channel_name(time, ifo):
     raise ValueError('Interferometer {} not supported at time {}'.format(ifo, time))
     
 
-def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, ifar, ifos, thresh_SNR, ligolw_skymap_output='.', ligolw_psd_output=None, ligolw_event_output=None, window_bins=300, frame_types=None, channel_names=None, gracedb_server=None, test_event=True, custom_frame_files=None):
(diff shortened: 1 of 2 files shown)


pycbc/io/live.py
```diff
@@ -263,10 +265,16 @@ def __init__(self, ifos, coinc_results, **kwargs):
         for ifo in self.psds:
             psd = self.psds[ifo]
             kmin = int(kwargs['low_frequency_cutoff'] / psd.delta_f)
+            if kwargs['high_frequency_cutoff']:
```
ahnitz, line 268: @MarlinSchaefer @titodalcanton I'm not sure we need to modify the psd we give to bayestar do we? Assumin git covers at least the range intended?

bin/pycbc_make_skymap
```diff
@@ -143,6 +143,7 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, ifar, ifos, thresh_SNR,
         command.append("--spin1z"),command.append(str(spin1z))
         command.append("--spin2z"),command.append(str(spin2z))
         command.append("--low-frequency-cutoff"),command.append(str(f_low))
+        command.append("--high-frequency-cutoff"),command.append(str(f_upper))
```
ahnitz, line 146: Does this work if you don't provide the option? I.E. it will try to do str(None) and probably raise an error, won't it?

ahnitz: if this is the case, this should probably be in a if statement so that it is really optional.

titodalcanton: +1

bin/pycbc_make_skymap
```diff
@@ -341,6 +343,8 @@ if __name__ == '__main__':
     parser.add_argument('--spin2z', type=float, required=True)
     parser.add_argument('--f-low', type=float,
                         help="lower frequency cut-off (float)", default=20.0)
+    parser.add_argument('--f-upper', type=float,
+                        help="upper frequency cut-off (float)", default=None)
```
titodalcanton, line 347: `default=None` can be removed, since it is... The default :)

pycbc/io/live.py
titodalcanton: I do not think this is required either.

titodalcanton: @lpsinger does BAYESTAR care if the PSD extends past the `f_final` frequency?

ahnitz:
@MarlinSchaefer It looks like you've addressed the specific comments we had.

PR #3415, opened 2020-08-11
add options to pycbc_create_injections for use with offline search codes

ahnitz:
This allows one to set start / end times instead of a number of injections to draw. This is the most common scenario for search workflows and is sufficient to enable pycbc_create_injections to produce reasonable files for use with the offline analysis. There may be a couple distributions that should also be added for simplicity, but I think even there most things can be done with inline transforms, etc. 

Example config file and command line
```
[variable_params]
tc =

[static_params]
mass1 = 37
mass2 = 32
ra = 2.2
dec = -1.25
inclination = 2.5
coa_phase = 1.5
polarization = 1.75
distance = 100
f_ref = 20
f_lower = 18
approximant = IMRPhenomPv2
taper = start

[prior-tc]
name = uniform
min-tc = 180
max-tc = 200
```
```
pycbc_create_injections --verbose \
        --config-files test.ini \
        --gps-start-time 1000000000 \
        --gps-end-time   1000500000 \
        --seed 10 \
        --output-file injection.hdf \
        --variable-params-section variable_params \
        --static-params-section static_params \
        --dist-section prior \
        --force
```

### DIFF
--- a/bin/pycbc_create_injections
+++ b/bin/pycbc_create_injections
@@ -223,15 +223,21 @@ while True:
 
     # We are drawing until we hit a time so check if we've reached it
     if opts.gps_start_time:
+        if 'tc' in samples.fieldnames:
+            raise RuntimeError("A distribution was given for 'tc'. This is "
+                               "not compatible with providing start/end times.")
+
         if old_samples is not None:
             samples = numpy.concatenate([old_samples, samples])
+
         old_samples = samples
 
         # We have enough to cover out time span so we can fixup the times
         # and stop here
-        if samples['tc'].sum() > opts.gps_end_time - opts.gps_start_time:
-            samples = numpy.sort(samples, order='tc')
-            samples['tc'] = opts.gps_start_time + samples['tc'].cumsum()
+        if samples['time_delay'].sum() > opts.gps_end_time - opts.gps_start_time:
+            samples = numpy.sort(samples, order='time_delay')
+            tc = opts.gps_start_time + samples['time_delay'].cumsum()
+            samples = samples.add_fields([tc], ['tc'])
             samples = samples[samples['tc'] < opts.gps_end_time]
             logging.info('Total Injections: %s', len(samples))
             break
--- a/bin/pycbc_create_injections
+++ b/bin/pycbc_create_injections
@@ -122,6 +122,7 @@ from pycbc.io import record
 from pycbc.waveform import parameters
 from pycbc.workflow import configuration
 from pycbc.workflow import WorkflowConfigParser
+from numpy.random import uniform
 import h5py
 
 parser = argparse.ArgumentParser(description=__doc__,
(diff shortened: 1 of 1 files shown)


cdcapano (commented):
Thanks for finally adding this; it's been on the to-do list for way too long. What you did is clever. However, I'm not sure I like having tc being mangled like this. The meaning of tc in the config file is changing depending on what the command-line arguments are; that'll be a bit confusing for users I think. 

How about just making this all command line? So, add a time-step argument and a delta-time argument, and use those for generating the times? Also add a check for a tc prior in the config file, printing a warning that it's being ignored.

ahnitz:
@cdcapano I've just updated the code to raise an error if 'tc' is set in the config file. 

However, I do think we want the time delay distribution set this way, so as a compromise I've just changed the name to 'time_delay' in the config file to avoid confusion with tc.

ahnitz:
The config file would now look like

```
[variable_params]
time_delay =

[static_params]
mass1 = 37
mass2 = 32
ra = 2.2
dec = -1.25
inclination = 2.5
coa_phase = 1.5
polarization = 1.75
distance = 100
f_ref = 20
f_lower = 18
approximant = IMRPhenomPv2
taper = start

[prior-time_delay]
name = uniform
min-time_delay = 90
max-time_delay = 200
```

ahnitz:
@cdcapano I've uncommented the lop part, can this be merged now?

ahnitz:
@cdcapano Poke

bin/pycbc_create_injections
```diff
@@ -189,23 +200,56 @@ dists = distributions.read_distributions_from_config(cp, opts.dist_section)
...
+                               "not compatible with providing start/end times.")
+
+        if old_samples is not None:
+            samples = numpy.concatenate([old_samples, samples])
+
+        old_samples = samples
+
+        # We have enough to cover out time span so we can fixup the times
+        # and stop here
+        if samples['time_delay'].sum() > opts.gps_end_time - opts.gps_start_time:
```
cdcapano, line 237: This will fail with a non-obvious error if `time_delay` isn't included in the config file.

What's the reason for wanting this to be in the config file? Are you thinking of using a different distribution then uniform? I still find it a bit weird that you have to have half the functionality on the command line and half in the config file, but if you do think a non-uniform distribution will ever be useful, then fine. However, call it `delta_tc`; I think that's a bit more descriptive. And add a check that it's provided (just like you checked that `tc` is not provided).

cdcapano:
@ahnitz poke

ahnitz:
@cdcapano Please review again. I've moved everything to an option of the script directly as you've suggested to make it consistent. For now, I suspect this is good enough to cover most cases.

bin/pycbc_create_injections
```diff
@@ -162,6 +173,13 @@ if os.path.exists(opts.output_file) and not opts.force:
     raise OSError("output-file already exists; use --force if you wish to "
                   "overwrite it.")
 
+if opts.ninjections and (opts.gps_start_time is not None or
+                         opts.gps_end_time is not None or
+                         opts.time_step is not None or
+                         opts.time_window is not none):
```
cdcapano, line 179: Change `none` to `None`

cdcapano (approved):
Looks good. Only issues is you have `none` typo (see comment), which is what is causing one of the CI tests to fail. (I'm surprised code climate didn't catch it.) Anyway, once you fix that, feel free to merge.

PR #3418, opened 2020-08-12
Update documentation

titodalcanton:
* Remove statement that only Python 2 is supported.
* Document usage of XML segment files in the workflow segment module, as explained by @ahnitz on Slack.


ahnitz:
@titodalcanton Major kudos for improving the docs!!

PR #3419, opened 2020-08-12
improvements to pycbc_make_skymap

ahnitz:
1) Don't hard code template flow. It was forced to 27Hz always!!! Instead set to provided low frequency cutoff
2) Don't hard code segment length, instead base on mass parameters given and f_low
3) Don't hard code the approximant, allow to be selected

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -13,6 +13,8 @@ import logging
 from pycbc import filter
 from pycbc.io import live
 from pycbc.types import TimeSeries, FrequencySeries
+from pycbc.pnutils import nearest_larger_binary_number
+from pycbc.waveform.spa_tmplt import spa_length_in_time
 from pycbc import frame
 from glue.ligolw import utils as ligolw_utils
 from ligo.gracedb.rest import GraceDb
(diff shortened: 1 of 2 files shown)


bin/pycbc_make_skymap
```diff
@@ -91,41 +93,47 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, ifar, ifos, th
...
         "--autogating-taper","0.25",
         "--autogating-pad","16",
-        "--minimum-chisq-bins","3",
-        "--strain-high-pass","15.",
-        "--pad-data","8",
-        "--template-start-frequency","27",
-        "--chisq-bins",'0.72*get_freq("fSEOBNRv4Peak",params.mass1,params.mass2,params.spin1z,params.spin2z)**0.7']
+        "--strain-high-pass", str(highpass_frequency),
+        "--pad-data","0",
+        "--chisq-bins",'16']
```
titodalcanton, line 136: I think one of the original ideas behind this script was to generate a GraceDB upload as close as possible to how a search would upload it. So I would prefer to keep the expression for the chisq-bins as it was.

ahnitz: @titodalcanton Ok, though note that the other options weren't exactly the same as the search.

bin/pycbc_make_skymap
```diff
@@ -91,41 +93,47 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, ifar, ifos, th
...
         "--allow-zero-padding",
         "--autogating-threshold","100",
         "--autogating-cluster","0.5",
         "--autogating-width","0.25",
         "--autogating-taper","0.25",
         "--autogating-pad","16",
-        "--minimum-chisq-bins","3",
-        "--strain-high-pass","15.",
-        "--pad-data","8",
-        "--template-start-frequency","27",
```
titodalcanton, line 127: Woops. If the option is omitted, does the single template code just use `--low-frequency-cutoff`?

ahnitz: @titodalcanton Yes.

PR #3420, opened 2020-08-13
Support HDF injection files in pycbc_optimal_snr

titodalcanton:
This is an attempt to make `pycbc_optimal_snr` support HDF injection sets. My initial hope was to seamlessly support all possible input/output cases (XML to XML, HDF to HDF, XML to HDF, HDF to XML) but I think just same-format support is sufficient.

At the moment, this is not quite working when trying to **write** HDF injections, due to issues with the approximant string (see discussion). It seems to work when writing XML files. I am opening it to gather opinions on the former case.

The different name of the coalescence time field makes the code slightly more complicated than needed; see the `get_gc_end_time()` function. It would be great if HDF injections could somehow understand `geocent_end_time` to mean `tc`, although it would also be ugly to special-case that in the `FieldArray` implementation...

### DIFF
--- a/bin/pycbc_optimal_snr
+++ b/bin/pycbc_optimal_snr
@@ -164,7 +164,7 @@ if __name__ == '__main__':
     if not opts.time_varying_psds:
         pycbc.psd.verify_psd_options_multi_ifo(opts, parser, detectors)
 
-    log_level = logging.DEBUG if opts.verbose else logging.INFO
+    log_level = logging.INFO if opts.verbose else logging.WARN
     logging.basicConfig(level=log_level,
                         format='%(asctime)s %(message)s',
                         datefmt='%Y-%m-%d %H:%M:%S')
@@ -212,11 +212,14 @@ if __name__ == '__main__':
     def compute_optimal_snr(inj):
         for det, column in opts.snr_columns.items():
             # geocent time is robust to potentially incomplete sim tables
-            injection_time = inj.geocent_end_time
+            try:
+                injection_time = inj.geocent_end_time
+            except AttributeError:
+                injection_time = inj.tc
             psd = psds[det](injection_time)
             if psd is None:
                 continue
-            logging.info('Trying injection %s', inj.simulation_id)
+            logging.info('Trying injection %s at %s', inj.simulation_id, det)
             try:
                 wave = get_injection(injections, det, injection_time,
                                      simulation_id=inj.simulation_id)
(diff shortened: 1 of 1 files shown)


titodalcanton:
@ahnitz so this does not quite work as it fails to write the output HDF file:
```
  File "pycbc_optimal_snr", line 327, in <module>
    pycbc.inject.InjectionSet.write(opts.out_file, new_inj_table)
  File "/Users/tito/anaconda3/lib/python3.7/site-packages/pycbc/inject/inject.py", line 849, in write
    **metadata)
  File "/Users/tito/anaconda3/lib/python3.7/site-packages/pycbc/inject/inject.py", line 437, in write
    fp[field] = samples[field]
  File "/Users/tito/anaconda3/lib/python3.7/site-packages/h5py/_hl/group.py", line 387, in __setitem__
    ds = self.create_dataset(None, data=obj, dtype=base.guess_dtype(obj))
  File "/Users/tito/anaconda3/lib/python3.7/site-packages/h5py/_hl/group.py", line 136, in create_dataset
    dsid = dataset.make_new_dset(self, shape, dtype, data, **kwds)
  File "/Users/tito/anaconda3/lib/python3.7/site-packages/h5py/_hl/dataset.py", line 118, in make_new_dset
    tid = h5t.py_create(dtype, logical=1)
  File "h5py/h5t.pyx", line 1634, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1656, in h5py.h5t.py_create
  File "h5py/h5t.pyx", line 1717, in h5py.h5t.py_create
TypeError: No conversion path for dtype: dtype('<U10')
```
Looking at the dtype, the problem field is the approximant. Python 3 strings biting us again?

titodalcanton:
I think the issue is that `pycbc.inject.InjectionSet.write()` does not know that the approximant should be a static param, and tries to write it as a variable param. Is there a way to convince it to write it as a string attribute instead of a string dataset?

ahnitz:
@cdcapano @titodalcanton I think we need to figure out how to handle python3 strings in field arrays in general. I think this is the same issue that was hit earlier when trying to use the injection generation with different approximants in the same file (i.e. not as attributes). Certainly, storing as an attribute seems to work, but I don't think we want to pen ourselves to requiring that the approximant only be a static parameter.

cdcapano:
@titodalcanton I agree with @ahnitz, the write should be able to handle writing approximants in the dataset (i.e., not just as static params). I'm a bit unclear how you hit this though. Why was the approximant string unicode when you tried to write? I ask because generally this has only been an issue for me when trying to read/write in a python 3 environment using data from a file that was written in a python 2 environment. Could you post instructions on how to replicate the error?

In any event, I think to fix this, we'll just need to add a try/except to the write method to get it to convert unicode into strings before writing. That's a bit whack-a-mole, but I don't think there's any one universal fix we can do for things like this.

titodalcanton:
@cdcapano this is all within Python 3. See #3427 for a reduced example.

titodalcanton:
After fixing #3427 this now completes successfully. I want to do a few minimal checks before asking for approval.

Reading HDF and writing to XML works without errors, but the resulting XML files are not actually readable due to the `simulation_id` field not being written correctly. This may or may not fix itself when we move to `ligo.lw` with the new integer LIGOLW IDs, so I am sort of happy to ignore this use case for now.

tdent:
Where is the format specification for HDF injection files?

cdcapano (approved):
Looks good

PR #3421, opened 2020-08-14
Added ability to accept multiple frame files to pycbc_make_skymap.

MarlinSchaefer:
For certain purposes it might be useful to process multiple custom frames when using pycbc_make_skymap. I therefore added the ability for the code to handle a list of multiple frames per interferometer.

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -63,9 +63,9 @@ def default_channel_name(time, ifo):
     raise ValueError('Interferometer {} not supported at time {}'.format(ifo, time))
     
 
-def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, ifar, ifos, 
-         thresh_SNR, ligolw_skymap_output='.', ligolw_psd_output=None, 
-         ligolw_event_output=None, window_bins=300, 
+def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
+         ifar, ifos, thresh_SNR, ligolw_skymap_output='.',
+         ligolw_psd_output=None, ligolw_event_output=None, window_bins=300, 
          frame_types=None, channel_names=None, 
          gracedb_server=None, test_event=True, 
          custom_frame_files=None, approximant=None):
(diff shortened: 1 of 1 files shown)


ahnitz:
@MarlinSchaefer Were you able to test this by running by hand?

MarlinSchaefer:
@ahnitz I used it on a previous test case (just single file per detector) and it ran without issues. I haven't run a test with multiple files as I didn't have one handy. Will do so and add a comment once that is done.

MarlinSchaefer:
@ahnitz I now checked it for multiple files and there actually was an issue with checking the segment. I adjusted the issue (quite naively, buy just assuming the given files span a continuous stretch) and now the code fails at the stage where pycbc_single_template is called. It gives the error `OSError: /tmp/tmpgogm0q9d/PSD_271_H1.txt not found.`. The stderr-file gives:
```python
2020-08-14 14:03:53,456 Running with CPU support: 1 threads
2020-08-14 14:03:53,456 Reading Frames
XLAL Error - XLALFrStreamFileOpen (LALFrStream.c:128): No files in stream file cache
XLAL Error - XLALFrStreamFileOpen (LALFrStream.c:128): Invalid argument
XLAL Error - XLALFrStreamCacheOpen (LALFrStream.c:233): Internal function call failed: Invalid argument
  File "/home/marlin/environments/pycbc_issues/bin/pycbc_single_template", line 7, in <module>
    exec(compile(f.read(), __file__, 'exec'))
  File "/home/marlin/environments/pycbc_issues/src/pycbc/bin/pycbc_single_template", line 216, in <module>
    gwstrain = strain.from_cli(opt, pycbc.DYN_RANGE_FAC)
  File "/home/marlin/environments/pycbc_issues/src/pycbc/pycbc/strain/strain.py", line 231, in from_cli
    sieve=sieve)
  File "/home/marlin/environments/pycbc_issues/src/pycbc/pycbc/frame/frame.py", line 202, in read_frame
    stream = lalframe.FrStreamCacheOpen(cum_cache)
RuntimeError: Internal function call failed: Invalid argument
```
I'm not sure what I can get from that error.

MarlinSchaefer:
Tracked down the error to be caused by not properly passing the custom frame-files to pycbc_make_skymap. Afterwards discovered that the sample-rate for the data was hard coded to be 2048. Changed the sample-rate to be an option with default value 2048.

PR #3431, opened 2020-08-21
pycbc_make_skymap: fix a bug, some improvements, and code cleanup

titodalcanton:
Restore the original use case where the frame files are automatically guessed,
which was broken by the addition of the --custom-frame-files option.

Improve the logging in case of failure of pycbc_single_template.

Remove an unused option.

Massive code cleanup.

Given the large change I will test this thoroughly and probably add a few examples, so marking as WIP.

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -5,18 +5,15 @@ Runs a single-template matched filter on strain data from a number of detectors
 and calls BAYESTAR to produce a sky localization from the resulting set of SNR
 time series."""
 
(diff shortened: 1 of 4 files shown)


ahnitz:
@titodalcanton Is this still WIP?

titodalcanton:
It needs more testing. I did some basic ones but the differences I saw were a bit bigger than I expected, so I want to have a more careful check.

titodalcanton:
Ok, I think the main issue I wanted to fix is fixed and tested, and I understood why the GW170817 skymap was not what it should in my earlier tests. We will address that separately.

ahnitz (approved):

PR #3432, opened 2020-08-26
Add spin from pulsar frequency to conversions

cdcapano:
Adds a function to compute dimensionless spin given a pulsar's mass, radius, and rotation frequency. Just assumes the pulsar is a solid sphere when doing so. Example:
```
>>> from pycbc.conversions import spin_from_pulsar_freq
>>> spin_from_pulsar_freq(1.44, 11., 500.)
0.3975488979662235
```

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -658,10 +658,10 @@ def spin_from_pulsar_freq(mass, radius, freq):
     freq : float
         The spin frequency of the pulsar, in Hz.
     """
-    omega = 2 * numpy.pi * freq 
-    mt = mass * lal.MTSUN_SI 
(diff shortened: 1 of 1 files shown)


spxiwh (approved):
This looks good, other than things codeclimate has already mentioned: Don't use "I" as a variable name. mt also doesn't pass PEP-8 conventions for variable names (at least 3 characters long and isn't really self-descriptive ... But we can find plenty of such examples within PyCBC code ).

Just need to make sure that any users are aware that this will just be an *illustrative* measure of the moment of inertia (because of the assumptions clearly stated in the help text). I can't think of any use case where that is a problem though!

PR #3442, opened 2020-09-04
Add model with numerical marginalization of the polarization angle

ahnitz:
This adds a new model accessed by choosing 'marginalized_polarization'. It has one selectable option 'polarization_sampes' which sets how many integration points are used. It is hard coded to use a uniform in polarization prior from 0 to 2pi. 

In testing, the model seems to greatly improve convergence rate in at least some examples where the polarization angle and coa_phase had a complicated 2d histogram.


ahnitz:
@sum33it I think the remaining CC items we don't plan to fix here.

ahnitz:
@sum33it Ok, I think this is now good to go. I re-checked with a test that it continues to work.

ahnitz:
@sum33it Poke, can we merge this?

sum33it (approved):
It looks good to me.

PR #3444, opened 2020-09-09
make start file from dynesty output

ahnitz:


ahnitz:
This should be made to use the standard file handling so any file which can produce posterior samples can be used. Putting work in progress label in place until that is done.

bin/inference/pycbc_inference_start_from_dynesty
```diff
@@ -0,0 +1,46 @@
...
+use = choice(nsample, replace=False, size=ntemps * nwalkers)
+
+o = h5py.File(args.output_file, 'w')
+for k in f['samples']:
+    data = f['samples'][k][:][use]
+    o['samples/' + k] = data.reshape(ntemps, nwalkers, 1)
+
+o.attrs['filetype'] = 'emcee_pt_file'
+o.attrs['static_params'] = []
+o.attrs['variable_params'] = f['samples'].keys()
```
cdcapano, line 39: Set the output's `variable_params` to `f.variable_params`; otherwise you'll end up with things like the `loglikelihood` in there. That may or may not be an issue later on, but better to err on safety.

bin/inference/pycbc_inference_start_from_dynesty
```diff
@@ -0,0 +1,46 @@
...
+# each temperature and walker will get a random
+# point from the dynesty output
+
+ntemps = args.ntemps
+nwalkers = args.nwalkers
+
+import h5py
+f = h5py.File(args.input_file, 'r')
+
+nsample = len(f['samples'][f['samples'].keys()[0]])
```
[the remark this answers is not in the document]
ahnitz: Yeah, I agree. Hence my top-level comment above. I realized this would be a better way to go and would be needed anyway once the dyensty format changes.

ahnitz:
@cdcapano I believe all the comments have been addressed. Can you review again?

bin/inference/pycbc_inference_start_from_samples
```diff
@@ -0,0 +1,49 @@
...
+import numpy
+import h5py
+from numpy.random import choice
+from pycbc.inference.io import loadfile
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--input-file')
+parser.add_argument('--output-file')
+parser.add_argument('--sampler', default='emcee_pt_file',
+    help="The output sampler type, if none, we assume emcee_pt")
```
cdcapano, line 15: Make the default `emcee_pt`; i.e., drop the `_file` bit. You can get the file type name using the sampler class (see comment below).

bin/inference/pycbc_inference_start_from_samples
```diff
@@ -0,0 +1,49 @@
...
+
+f = loadfile(args.input_file)
+params = list(f.variable_params) + ['loglikelihood']
+samples = f.read_samples(params)
+nsample = len(samples)
+
+# These are the ids we'll use for the temps / walkers
+use = choice(nsample, replace=False, size=ntemps * nwalkers)
+
+o = loadfile(args.output_file, 'w', filetype=args.sampler)
```
cdcapano, line 36: Get the filetype using the sampler class:
```
filetype = samplers[args.sampler].io.name
```
(You'll need to add a `from pycbc.inference.sampler import samplers` for that to work)

bin/inference/pycbc_inference_start_from_samples
```diff
@@ -0,0 +1,49 @@
...
+for k in params:
+    data = samples[k][use]
+    o['samples/' + k] = data.reshape(ntemps, nwalkers, 1)
+
+o.attrs['filetype'] = 'emcee_pt_file'
```
cdcapano, line 41: Delete that line.

bin/inference/pycbc_inference_start_from_samples
```diff
@@ -0,0 +1,49 @@
...
+o.attrs['static_params'] = []
+o.attrs['variable_params'] = f.variable_params
+
+o['sampler_info/acceptance_fraction'] = numpy.zeros(nwalkers)
```
cdcapano, line 45: You shouldn't need to set the `acceptance_fraction`; it's not used for anything related to reading samples.

cdcapano (changes requested):
Thanks for generalizing. See comments.

cdcapano (approved):
Looks good, thanks

PR #3446, opened 2020-09-11
add multiband waveform to speed up slow / long waveform approximants

ahnitz:
This adds a 'multiband' waveform approximant which can use a base approximant which supports 'f_ref' and 'f_lower' / 'f_final' to speed up waveform generation. The waveform is generated at different delta_f's within frequency bands and then stitched back together. 

Example,
```
get_fd_waveform(approximant='multiband', base_approximant='IMRPhenomPv3', 
                            mass1=1.4, mass2=1.4,
                            f_lower=20, f_ref=20, f_final=1000,
                            delta_f=1.0/1024, spin1x=0.5, spin1z=0.5,
                            bands='50', lengths='50', overlap=1)
```

### DIFF
--- a/pycbc/waveform/multiband.py
+++ b/pycbc/waveform/multiband.py
@@ -1,0 +1,111 @@
+""" Tools and functions to calculate interpolate waveforms using multi-banding
+"""
+import numpy
+
+from pycbc.types import TimeSeries, zeros, FrequencySeries
+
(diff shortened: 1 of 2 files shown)


cdcapano (commented):
Looks fine, but some several of the code climate issues should be addressed.

ahnitz:
@cdcapano OK, I think the remaining CC issues are ignorable.

cdcapano (approved):

Issue #3054, opened 2019-12-10
pip install fails to build on macOS Catalina


geoffrey4444:
I'm using anaconda3 and macOS 10.15.1.

`pip install pycbc lalsuite` fails with this error:

~~~~~~~
Building wheel for pycbc (setup.py) ... error
  ERROR: Complete output from command /Users/geoffrey/anaconda3/bin/python -u -c 'import setuptools, tokenize;__file__='"'"'/private/var/folders/c_/lbf0gfvd56l_826nb5nf7zcm0000gn/T/pip-install-cqxa4mq3/pycbc/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' bdist_wheel -d /private/var/folders/c_/lbf0gfvd56l_826nb5nf7zcm0000gn/T/pip-wheel-gngg3k4c --python-tag cp37:
  ERROR: /Users/geoffrey/anaconda3/lib/python3.7/distutils/extension.py:131: UserWarning: Unknown Extension options: 'compiler_directives'
    warnings.warn(msg)
  running bdist_wheel
  running build
  running build_py
  creating build
  creating build/lib.macosx-10.7-x86_64-3.7
  creating build/lib.macosx-10.7-x86_64-3.7/pycbc
  copying pycbc/detector.py -> build/lib.macosx-10.7-x86_64-3.7/pycbc
  copying pycbc/transforms.py -> build/lib.macosx-10.7-x86_64-3.7/pycbc
  copying pycbc/mchirp_area.py -> build/lib.macosx-10.7-x86_64-3.7/pycbc
  copying pycbc/version.py -> build/lib.macosx-10.7-x86_64-3.7/pycbc
  copying pycbc/bin_utils.py -> build/lib.macosx-10.7-x86_64-3.7/pycbc
  copying pycbc/_version.py -> build/lib.macosx-10.7-x86_64-3.7/pycbc
  copying pycbc/boundaries.py -> build/lib.macosx-10.7-x86_64-3.7/pycbc
  copying pycbc/conversions.py -> build/lib.macosx-10.7-x86_64-3.7/pycbc
  copying pycbc/rate.py -> build/lib.macosx-10.7-x86_64-3.7/pycbc
[...]

PR #3447, opened 2020-09-15
patch to hack macs to work

ahnitz:
Attempt to add in fix recommend in #3054

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -212,15 +212,22 @@ cythonext = ['waveform.spa_tmplt',
 ext = []
 cython_compile_args = ['-O3', '-w', '-ffast-math',
                        '-ffinite-math-only']
+
 if platform.machine() == 'x86_64':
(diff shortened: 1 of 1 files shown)


titodalcanton:
I tested this change on an anaconda env, and at least it does not appear to break the install. I did not do extensive tests though.

ahnitz:
@titodalcanton @spxiwh Would either of you be happy to approve this? It seems @titodalcanton was able to verify this didn't break building on a mac and it would seem to resolve some mac circumstances. I can't really test further than that, unfortunately.

spxiwh (approved):
I'll approve this ... I think the problems reported were on up-to-date mac OS versions, which I don't have, but I don't see any harm in this.

titodalcanton:
My tests were on macOS 10.15.6 (up to date). No issues with merging this.

PR #3448, opened 2020-09-15
PyCBC Live: fix a bug arising from early-warning test

titodalcanton:
The time offset was not properly applied everywhere and was showing up in the trigger/SNR plot, so this fixes it along with some minor code cleanup.

It would be nice to have a clearer, earlier distinction between "end time" and "merger time" in the code, such that the GraceDB interface code does not need to know about it.

### DIFF
--- a/pycbc/io/live.py
+++ b/pycbc/io/live.py
@@ -347,7 +347,8 @@ class SingleCoincForGraceDB(object):
                 if ifo in self.ifos:
                     base = 'foreground/{}/'.format(ifo)
                     snr = self.coinc_results[base + 'snr']
-                    mt = self.coinc_results[base + 'end_time'] + self.time_offset
+                    mt = (self.coinc_results[base + 'end_time']
+                          + self.time_offset)
                     pl.plot([mt - ref_time], [snr], c=ifo_color(ifo),
                             marker='x')
 
--- a/pycbc/io/live.py
+++ b/pycbc/io/live.py
@@ -121,10 +121,10 @@ class SingleCoincForGraceDB(object):
                                       and coinc_results['HWINJ'])
 
         # Check if we need to apply a time offset (this may be permerger)
-        time_offset = 0
+        self.time_offset = 0
         rtoff = 'foreground/{}/time_offset'.format(ifos[0])
         if rtoff in coinc_results:
-            time_offset = coinc_results[rtoff]
+            self.time_offset = coinc_results[rtoff]
 
         if 'followup_data' in kwargs:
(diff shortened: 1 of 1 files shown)


pycbc/io/live.py
```diff
@@ -339,41 +337,42 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
                                                              '_psd.png')
-            pylab.figure()
+            pl.figure()
+            ref_time = int(self.merger_time)
             for ifo in sorted(self.snr_series):
                 curr_snrs = self.snr_series[ifo]
                 curr_snrs.save(snr_series_fname, group='%s/snr' % ifo)
-                pylab.plot(curr_snrs.sample_times, abs(curr_snrs),
-                           c=ifo_color(ifo), label=ifo)
+                pl.plot(curr_snrs.sample_times - ref_time, abs(curr_snrs),
```
ahnitz, line 345: @titodalcanton Just trying to understand what you've changed. Is this the only line that matter? There are other changes, but they all look like style choices.

titodalcanton: It is actually line 350. The change here just makes the plot's x-axis easier to read.

pycbc/io/live.py
```diff
@@ -339,41 +337,42 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
-            pylab.legend()
-            pylab.xlabel('GPS time (s)')
-            pylab.ylabel('SNR')
-            pylab.savefig(snr_series_plot_fname)
-            pylab.close()
-
-            pylab.figure()
+                    base = 'foreground/{}/'.format(ifo)
+                    snr = self.coinc_results[base + 'snr']
+                    mt = self.coinc_results[base + 'end_time'] + self.time_offset
```
ahnitz, line 350: I guess also here was changed.

ahnitz:
@titodalcanton The problem is that the anlaysis really doesn't care about the merger time i.e it is consistently relative to the data. It's only required anywhere since the gracedb postprocessing seems to require this. I  agree that this is messier than I'd like though. It's worth thinking about if it can be organized in a cleaner way.

ahnitz (approved):
@titodalcanton Well, I don't think I see any major problems, and since you've tested, I think this is fine to merge.

ahnitz:
@titodalcanton Travis will fail due to dcc being down, so you'll need to restart when that is back up. If this is more urgent, let me know.

ahnitz:
@titodalcanton Modulo that the one CC thing can be trivially fixed.

PR #3452, opened 2020-09-18
brute force marginalization

ahnitz:
Marginalize by simply calling the likelihood function multiple times. Sometimes useful for sampling.


cdcapano (approved):
Looks good.

PR #3453, opened 2020-09-18
Cache waveform generator

Richard-Sti:


pycbc/waveform/generator.py
```diff
@@ -135,8 +137,13 @@ def dostuff(self):
     def _generate_from_current(self):
         """Generates a waveform from the current parameters.
         """
+        values_hash = hash(frozenset(**self.current_params.values()))
```
cdcapano, line 140: Why the `frozenset`? I can you just do `hash(self.current_params.values()))`?

pycbc/waveform/generator.py
```diff
@@ -135,8 +137,13 @@ def dostuff(self):
...
+        if values_hash == self._cache['hash']:
+            return self._cache['waveform']
         try:
-            return self.generator(**self.current_params)
+            waveform = self.generator(**self.current_params)
+            self._cache.update({'hash': values_hash, 'waveform': waveform})
+            return waveform
         except RuntimeError as e:
```
cdcapano, line 147: This could be simplified...
```
try:
     return self._cache[values_hash]
except KeyError:
    try:
         waveform = self.generator(**self.current_params)
         self._cache.clear()
         self._cache[values_hash] = waveform
         return waveform
    except RunTimeError as e:
```

cdcapano (changes requested):
Thanks for doing this @Richard-Sti . I think it could be simplified a bit; see comments.

pycbc/waveform/generator.py
cdcapano: Actually, thinking about this some more, it would be better to only hash the generator's variable_args. Also, I realized that the hash of a generator doesn't give the correct thing. So:
```
values_hash = frozenset(self.current_params[p] for p in self.variable_args)
```
```

cdcapano (approved):
Looks good!

cdcapano:
@Richard-Sti You're going to need to fetch `gwastro` and rebase to `gwastro/master` to pick up Travis fix. Note that I had fixed a merge conflict from master. I'm not sure if that will show up in your branch or not. If not, make sure to fix the conflict before pushing. If you're not sure what the fix is, take a look at here `generator.py`. The conflict was Alex had added a block of code to record failed waveforms, which touches the same lines of code you did.

PR #3454, opened 2020-09-18
Update ringdown format_lmns to accept whitespace-separated string.

JulianWesterweck:
Making the list of desired ringdown modes a string prevents issues with conflicting dimensions in the FieldArray of parameters.

### DIFF
--- a/pycbc/waveform/ringdown.py
+++ b/pycbc/waveform/ringdown.py
@@ -68,29 +68,26 @@ def format_lmns(lmns):
     """ Checks if the format of the parameter lmns is correct, returning the
     appropriate format if not, and raise an error if nmodes=0.
     The required format for the ringdown approximants is a list of lmn modes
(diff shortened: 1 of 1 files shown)


micamu:
@JulianWesterweck Looks good to me! Have you tested it? I am not sure if we might need some `.decode()` for python3...

JulianWesterweck:
Yes, I tested it, both loading a model that uses the waveform, to make sure it is read correctly from a config file, and by just using the generator directly. So far that worked.

cdcapano (approved):
Looks good to me.

cdcapano:
@JulianWesterweck It's minor, but could you fix the whitespace issues that code climate picked up? Other than that, @micamu do you have any objections to this being merged?

JulianWesterweck:
@cdcapano @micamu Oh, yes, I missed that one. Done.

PR #3457, opened 2020-09-18
Add support for ptemcee

cdcapano:
Adds support for ptemcee.

Unfortunately, the released version of ptemcee (1.0.0) is incompatible with this, as I wrote it against ptemcee's current master. Given that there's hasn't been a release of ptemcee since 2015, and the last commit to it was in 2018, I don't think it'll be updated. So, to use this, you need to install from source.

Did a quick test on the normal distribution. Putting on hold until I have a chance to test on a GW.

### DIFF
--- a/companion.txt
+++ b/companion.txt
@@ -8,6 +8,7 @@ healpy
 ligo-gracedb
 
 # auxiliary samplers
+ptemcee
 cpnest
 pymultinest
 ultranest
--- a/companion.txt
+++ b/companion.txt
@@ -11,3 +11,4 @@ ligo-gracedb
 cpnest
 pymultinest
 ultranest
+https://github.com/willvousden/ptemcee/archive/master.tar.gz
--- a/companion.txt
+++ b/companion.txt
@@ -8,7 +8,6 @@ healpy
 ligo-gracedb
 
 # auxiliary samplers
-ptemcee
 cpnest
 pymultinest
 ultranest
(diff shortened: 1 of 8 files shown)


ahnitz:
@cdcapano If there is no new release, and you have to install from master, it might be good just to make a release ourselves of ptemcee (i.e. upload to pycbc under say pycbc-ptemcee) so that there is something people can install with pip. Not needed for this patch though, just would be nice if we start using this.

cdcapano:
@ahnitz Yeah, that's a good idea. How do you propose doing that exactly? Do you want to register a `pycbc-ptemcee` with PyPi, or just include a tar ball that can be downloaded from gwastro via pip, like we currently do with pegasus? I lean toward the latter. I don't think we want to be responsible for having to keep a registered version of ptemcee up to date.

ahnitz:
@cdcapano The latter is probably good enough.

ahnitz:
@cdcapano In which case you can just do it with the proper master repo of ptemcee when you add that to companion.txt.

cdcapano:
@ahnitz I didn't realize all repos have a master tar ball. Ok, I added it to companion.txt then.

ahnitz:
@cdcapano Some of the CC issues seem serious. This will also need to be rebased.

cdcapano:
@ahnitz I think the complaints about ptemcee not having particular methods are because code climate isn't picking up the master version of ptemcee. In any event, I realized I forgot to add functions to read the adaptation settings from the config file, so marking back as work in progress.

ahnitz (approved):

PR #3458, opened 2020-09-20
Fix issues with extracting samples from dynesty file

cdcapano:
Fixes a few bugs introduced by PR #3376. First, when reading raw samples from a dynesty file, the current code tries to extract a posterior for all samples rather than just the fields requested. 

Second, and more egregious, is that the random shuffle that's done to extract the posterior means that repeated calls to `read_raw_samples` are not deterministic. This is a particular problem for `plot_posterior`. If you are plotting scatter points colored by SNR, two calls are made to `read_raw_samples`: one to get the samples and one to get the SNR. The random shuffle means that the SNRs returned will not correspond to the samples that were obtained in the first call.

This fixes that by introducing a seed argument. The random generator's seed is set to the given value just before the shuffle is called, ensuring that repeated calls to `read_raw_samples` yield the same results. So as not to interfere with any other random number generation that might happen elsewhere in the code, a custom random generator is created solely for this purpose.

### DIFF
--- a/pycbc/inference/io/dynesty.py
+++ b/pycbc/inference/io/dynesty.py
@@ -72,10 +72,18 @@ class CommonNestedMetadataIO(object):
             act = parser.add_argument(
                 "--raw-samples", action='store_true', default=False,
                 help="Raw samples are the unweighted samples obtained from "
-                     "dynesty sampler."
-                     "Default value is False which means raw samples are"
-                     "weighted by log-weight array obtained from the sampler.")
+                     "the nested sampler. Default value is False, which means "
+                     "raw samples are weighted by the log-weight array "
+                     "obtained from the sampler, giving an estimate of the "
+                     "posterior.")
             actions.append(act)
+        if 'seed' not in skip_args:
+            act = parser.add_argument(
+                "--seed", type=int, default=0,
+                help="Set the random-number seed used for extracting the "
+                     "posterior samples. This is needed because the "
+                     "unweighted samples are randomly shuffled to produce "
+                     "a posterior. Default is 0.")
         return parser, actions
 
 class DynestyFile(CommonNestedMetadataIO, BaseNestedSamplerFile):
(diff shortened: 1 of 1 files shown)


ahnitz (approved):

ahnitz:
@cdcapano Looks like you used a really new numpy feature. The errors are real in travis
```
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/pycbc/inference/io/base_hdf.py", line 207, in read_samples
    samples = self.read_raw_samples(loadfields, **kwargs)
  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/pycbc/inference/io/dynesty.py", line 135, in read_raw_samples
    rng = numpy.random.default_rng(seed)
AttributeError: 'module' object has no attribute 'default_rng'
    FAILED!
```
Any reason you aren't just using the default random state infrastructure? I would think this is what it is intended for anyway.

cdcapano:
@ahnitz Since version 1.17 of numpy, using `default_rng` is the suggested way to create a random generator. The previous way, calling `RandomState`, is only left in as legacy support. The problem is that the new random module was one of the things numpy did not update when they dropped python 2.7 support. I've added a try/except `AttributeError` to fall back to calling `RandomState`, with a comment to delete once we drop 2.7. Hopefully that should fix it.

pycbc/inference/io/dynesty.py
```diff
@@ -132,7 +132,12 @@ def read_raw_samples(self, fields, raw_samples=False, seed=0):
                     i += 1
                 else:
                     j += 1
-            rng = numpy.random.default_rng(seed)
+            try:
+                rng = numpy.random.default_rng(seed)
+            except AttributeError:
```
ahnitz, line 137: This doesn't relate to python2.7 and it may lead us to trouble if you use a different random source with a try except. Please remove the try except and the comment.

ahnitz (changes requested):

pycbc/inference/io/dynesty.py
cdcapano: It relates to python 2.7 in the sense that you cannot use numpy version >= 1.17 if you are using python 2.7. When we drop 2.7 support, we can just bump the minimum numpy version to 1.17, and this issue goes away. I don't understand your second point about using a different random source. All this is being used for is to generate a one time random number generator for doing the shuffle, so as not to mess with the global generator that may be used elsewhere in whatever executable this is being used in. Are you worried that an AttributeError may be raised for another reason then just an old numpy version? If so, then I can add a check on the numpy version before trying `RandomState`; e.g.,
```
try:
    rng = numpy.random.default_rng(seed)
except AttributeError as e:
   if numpy.__version__ < 1.17:  # this is pseudo code
        rng = numpy.random.RandomState(seed)
   else:
        raise e
```
Would you prefer that?

Either way, I'd rather not just use `RandomState` since it states in the [numpy documentation](https://numpy.org/doc/stable/reference/random/legacy.html#numpy.random.RandomState) that it is only being kept for legacy purposes.

ahnitz: @cdcapano The main issue is that you have two different paths that give different results. If you can use the Generator interface *which matches the output of Randomstate* then I would not object, but I think your default setup that's not guaranteed. I don't want someone to get a different result depending on if they use python2 / 3 when that is not needed.

cdcapano: Ok, I understand your concern now. Yeah, with the current patch you'll end up with slightly different results if you use different versions of numpy. The only way to get exactly the same results is to use `RandomState` (you'll get different results from `RandomState` even if you use the MT19937 generator, even though it's nominally the same algorithm). But the current [numpy docs](https://numpy.org/doc/stable/reference/random/index.html) state that you should use `default_rng` for creating a generator. If they drop support for `RandomState` sometime in the future, then we'll need to change what's used... and at that point, you'll end up with different results if you use different versions of the code. Why not just do that now? At least the python 2.7/3.X thing is a clear deliminator. Besides, isn't expecting that you get exactly the same realization of random numbers regardless of what version of code you use a bit strict?

ahnitz (approved):

PR #3470, opened 2020-09-24
allow one to set the parameters for the first bounding region in dynesty

ahnitz:


ahnitz:
Why is this important one may ask? Dynesty can overconstrain the bounding region if it doesn't have a live point within all the modes of the posterior. This allows one to choose when it switches from sampling from the entire prior to a bounded region, which can help force it to sample from the entire prior for a longer period.

cdcapano (changes requested):
Looks fine. Could you add something about this to the `from_config` doc string. Actually, looking at that, I realize nothing is documented there. Could you add something about the other options that are parsed, a la [PTEmcee.from_config](http://pycbc.org/pycbc/latest/html/pycbc.inference.sampler.html#pycbc.inference.sampler.ptemcee.PTEmceeSampler.from_config). At the moment there's no way to know what can be set in the config file without looking at the code. Also, please address the PEP-8 issue.

ahnitz:
@cdcapano I've updated the documentation and included a small typo fix for ptemcee

cdcapano (approved):
Looks good. Thanks for fixing the issue in the ptemcee docs.

Issue #3427, opened 2020-08-20
`InjectionSet.write()` problem with approximant field on Python 3


titodalcanton:
Posting a reduced example to demonstrate the issue described in #3420 and possibly #3395.

I created an HDF injection set as follows:
```bash
pycbc_create_injections \
    --config-files injection_definition.ini \
    --variable-params-section variable_params \
    --static-params-section static_params \
    --dist-section prior \
    --ninjections 5 \
    --output-file injections.hdf \
    --force \
    --verbose
```
with `injection_definition.ini` being:
```
[variable_params]
tc =
mass1 =
mass2 =
ra =
dec =
inclination =
coa_phase =
polarization =
distance =

[static_params]
approximant = IMRPhenomD
f_lower = 12

[prior-tc]
name = uniform
min-tc = 1276524000
max-tc = 1276525000

[prior-mass1]
name = uniform
min-mass1 = 3
max-mass1 = 30

[prior-mass2]
name = uniform
min-mass2 = 3
max-mass2 = 30

[prior-ra+dec]
name = uniform_sky

[prior-inclination]
name = sin_angle

[prior-coa_phase]
name = uniform_angle

[prior-polarization]
name = uniform_angle

[prior-distance]
name = uniform_radius
min-distance = 20
max-distance = 500

[constraint-1]
name = custom
constraint_arg = q_from_mass1_mass2(mass1, mass2) <= 8
```
I then expected something like this to work:
```python
from pycbc.inject import InjectionSet

injs = InjectionSet('injections.hdf')
InjectionSet.write('test.hdf', injs.table)
```
However, I get this:
```
Traceback (most recent call last):
  File "test_injection_io.py", line 4, in <module>
    InjectionSet.write('test.hdf', injs.table)
  File "/Users/tito/anaconda3/lib/python3.7/site-packages/pycbc/inject/inject.py", line 849, in write
    **metadata)
  File "/Users/tito/anaconda3/lib/python3.7/site-packages/pycbc/inject/inject.py", line 437, in write
    fp[field] = samples[field]
  File "/Users/tito/anaconda3/lib/python3.7/site-packages/h5py/_hl/group.py", line 387, in __setitem__
    ds = self.create_dataset(None, data=obj, dtype=base.guess_dtype(obj))
  File "/Users/tito/anaconda3/lib/python3.7/site-packages/h5py/_hl/group.py", line 136, in create_dataset
[...]

PR #3471, opened 2020-09-24
Allow injection set to write string datatypes

cdcapano:
Fixes Issue #3427 

As described in there , `InjectionSet.write` fails if any of the injection parameters are strings. This can happen when writing an injection set that was loaded from another injection file. The reason is when the injections are loaded, any static arguments are expanded to fields. This is done using `numpy.repeat`. By default, numpy will store the python3 `str` as `U` type, but `h5py` doesn't know how to write that: if try to write the table back out, you'll get a `TypeError`. For example, say I have an injection file that has `approximant = IMRPhenomD` (as is the case in the example given in #3427). Since this is a static argument, it is stored in the injection file's `.attrs` as a `str`:
```
>>> injs = InjectionSet('injections.hdf')
>>> injs._injhandler.filehandler.attrs['approximant']
'IMRPhenomD'
>>> type(injs._injhandler.filehandler.attrs['approximant'])
str
```
The `InjectionSet`'s `.table` will therefore have an `approximant` field which is:
```
>>> injs.table['approximant']
array(['IMRPhenomD', 'IMRPhenomD', 'IMRPhenomD', 'IMRPhenomD',
       'IMRPhenomD'], dtype='<U10')
```
If you then try to do:
```
InjectionSet.write('test.hdf', injs.table)
```
You'll get a `TypeError`.

This fixes this by converting any arrays with `U` type to `S` types before writing. Since this means reading an injection file with strings will return byte arrays, I've also made it so that the read function converts all 'S' type arrays to 'U' type when loading. (If you don't do this, you'll get a byte type when you pull out the approximant, which will cause errors elsewhere.)

### DIFF
--- a/pycbc/inject/inject.py
+++ b/pycbc/inject/inject.py
@@ -434,7 +434,15 @@ class _HDFInjectionSet(object):
                     # try decoding it and writing
                     fp.attrs[arg] = str(val)
             for field in write_params:
-                fp[field] = samples[field]
+                try:
+                    fp[field] = samples[field]
+                except TypeError as e:
+                    # can get this in python 3 if the val was a numpy.str_ type
+                    # we'll try again as a string type
+                    if samples[field].dtype.char == 'U':
+                        fp[field] = samples[field].astype('S')
+                    else:
+                        raise e
 
 
 class CBCHDFInjectionSet(_HDFInjectionSet):
--- a/pycbc/inject/inject.py
+++ b/pycbc/inject/inject.py
@@ -361,6 +361,9 @@ class _HDFInjectionSet(object):
                 # otherwise, we can just repeat the value the needed number of
                 # times
                 arr = np.repeat(val, numinj)
+            # make sure any byte strings are stored as strings instead
+            if arr.dtype.char == 'S':
+                arr = arr.astype('U')
             injvals[param] = arr
         # make sure a coalescence time is specified for injections
         if 'tc' not in injvals:


ahnitz:
@cdcapano Thanks for fixing this! I was worried it might be more complicated.

titodalcanton:
Thanks @cdcapano. I quickly tried this on my test code described in #3427, and it seems to work as long as I generate the original HDF injection file *after* your patch. If, instead, I try to read and rewrite an old HDF file, I get this error:
```
  File "test_injection_io.py", line 3, in <module>
    injs = InjectionSet('injections.hdf')
  File "/home/tito/virtualenvs/pycbc_dev_py3/lib/python3.6/site-packages/pycbc/inject/inject.py", line 801, in __init__
    self._injhandler = get_hdf_injtype(sim_file)(sim_file, **kwds)
  File "/home/tito/virtualenvs/pycbc_dev_py3/lib/python3.6/site-packages/pycbc/inject/inject.py", line 748, in get_hdf_injtype
    return hdfinjtypes[ftype]
KeyError: b'cbc'
```
So I think this PR fixes the issue, but it will break reading old HDF files. If this is not a problem, I can approve this.

cdcapano:
@titodalcanton I think that's a bug separate from this. Was the file you tested on created in python 2? I can try to add a fix for that in this patch too.

titodalcanton (approved):
You are right, I only get that error if the initial file is generated on Python 2. So I am approving this, and feel free to keep that for a new PR.

PR #3472, opened 2020-09-24
Be able to read injection files written in python 2

cdcapano:
Fixes an issue where injection files written in python 2 could not be read in python 3. The problem was both the file type and static params names are saved as unicode in python 2.7. In python 3 this results in them being returned as bytes instead of strings, which breaks string comparisons. Fixes this by casting to string after loading from file.

@titodalcanton I think this should fix the problem you found while testing #3471 .

### DIFF
--- a/pycbc/inject/inject.py
+++ b/pycbc/inject/inject.py
@@ -343,7 +343,8 @@ class _HDFInjectionSet(object):
             numinj = tuple(injvals.values())[0].size
         # add any static args in the file
         try:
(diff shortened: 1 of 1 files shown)


titodalcanton (approved):
This fixes the problem in Python 3, does not break in Python 2, and the files written in Python 2 and Python 3 are almost identical (except for the different `CSET H5T_CSET_ASCII;` vs `CSET H5T_CSET_UTF8;`). So approving.

PR #3476, opened 2020-09-30
Removal of a manual list of trigger times

GarethDaviesGW:
If we have signals in a single detector, it causes significant little-dog background.

This change means we can remove these triggers manually (or a bad glitch, or whatever, just pointing to a time) from all backgrounds

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_coinc_statmap
+++ b/bin/hdfcoinc/pycbc_multiifo_coinc_statmap
@@ -68,6 +68,10 @@ parser.add_argument('--hierarchical-removal-against', type=str,
                          'triggers that are louder than either the "inclusive"'
                          ' (little-dogs-in) background, or the "exclusive" '
                          '(little-dogs-out) background. [default="none"]')
+parser.add_argument('--additional-event-times', type=float, nargs="+",
+                    help="Additional event times which will be removed from "
+                         "both inclusive and exclusive background "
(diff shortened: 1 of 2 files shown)


spxiwh (approved):
The code certainly looks like it should do what you say, but I think here the proof is really in your testing. Can you describe what has been tested here?

GarethDaviesGW:
> The code certainly looks like it should do what you say, but I think here the proof is really in your testing. Can you describe what has been tested here?

yes, sorry but testing this on proprietary data so can't post here. We have some large tails in the coincident background, which look like we would normally expect from the little-dogs of signals, which are coming from single-detector triggers. after this change, we remove the tails as the single-detector trigger is not forming little-dogs

bin/hdfcoinc/pycbc_multiifo_coinc_statmap
```diff
@@ -68,6 +68,10 @@ parser.add_argument('--hierarchical-removal-against', type=str,
                          'triggers that are louder than either the "inclusive"'
                          ' (little-dogs-in) background, or the "exclusive" '
                          '(little-dogs-out) background. [default="none"]')
+parser.add_argument('--additional-event-times', type=float, nargs="+",
+                    help="Additional event times which will be removed from "
```
tdent, line 72: if you want several times should they be space- or comma-separated? maybe indicate this in the help message?

PR #3478, opened 2020-10-01
Complete the range of non-bbh psd-threshold stats

tdent:


spxiwh (approved):

GarethDaviesGW (approved):
This looks like it will work (I got confused by the swapping of order of two of the statistics making it look like a lot more than it is)

tdent:
Yes, for some reason while editing the file I thought things would be better ordered that way .. codeclimate is the usual 'similar blocks of code' complaint

PR #3480, opened 2020-10-02
Fix typo in example workflow config and other cleanup

cdcapano:
Fixes a typo in the example workflow config files caught by Alex Brown. Source frame masses should be multiplied by the redshift to get detector frame mass, not divided. Also adds `--no-create-proxy` to the example `pycbc_submit_dax` command (that should of been there in the first place). Finally, this removes old workflow config files that are out of date and no longer used anywhere.


sum33it (approved):
Thanks for doing this @cdcapano! It looks good to me.

PR #3485, opened 2020-10-02
Core collapse supernovae 

chaitanyaafle:
This patch adds the core-collapse waveform model:

1. Added core bounce and post bounce waveform for core collapse supernovae. This uses a `.hdf` file containing principal component basis vectors to generate the waveforms. The principal components are generated using singular value decomposition of a catalog of numerical simulations. The waveform model is h = a_1 * x_1 + a_2 * x_2 + ... + a_n * x_n, where x_i are the principal component basis vectors and a_i are their coefficients. 

2. Added a constraint for the coefficients of the basis vectors that generate the core collapse waveform. The coefficients should be inside the two/three dimensional convex hull of the coefficients of the waveforms in a catalog.

3. For Bayesian inference, the coefficients are the variable parameters while the principal components  and the number of principal components  used are static parameters. Following is an example section of `.ini` file that would be used for inference:

```
[variable_params]
coeff_0 = 
coeff_1 = 
coeff_2 = 
coeff_3 = 

[static_params]
distance = 0.0081
no_of_pcs = 9
principal_components_file = /path/to/principal_components_file.hdf 

[constraint-hull]
name = supernovae_convex_hull
principal_components_file = ${static_params|principal_components_file}
hull_dimention = 3
constraint_arg =
```

### DIFF
--- a/pycbc/distributions/constraints.py
+++ b/pycbc/distributions/constraints.py
@@ -35,18 +35,17 @@ class Constraint(object):
         for kwarg in kwargs.keys():
             setattr(self, kwarg, kwargs[kwarg])
 
-
-        #For supernovae waveforms:
+        # For supernovae waveforms:
         if 'principal_components_file' in kwargs:
-            pc_file = kwargs['principal_components_file']
+            pc_filename = kwargs['principal_components_file']
             hull_dimention = numpy.array(kwargs['hull_dimention'])
             self.hull_dimention = int(hull_dimention)
-            f = h5py.File(pc_file, 'r')
-            pc_coefficients = numpy.array(f.get('coefficients'))
-            f.close()
+            pc_file = h5py.File(pc_filename, 'r')
+            pc_coefficients = numpy.array(pc_file.get('coefficients'))
+            pc_file.close()
             hull_points = []
             for dim in range(self.hull_dimention):
-                hull_points.append(pc_coefficients[:,dim])
+                hull_points.append(pc_coefficients[:, dim])
             hull_points = numpy.array(hull_points).T
             pc_coeffs_hull = Delaunay(hull_points)
             self._hull = pc_coeffs_hull
@@ -93,8 +92,8 @@ class ConvexHull(Constraint):
         points = numpy.array([params["coeff_0"],
                               params["coeff_1"],
                               params["coeff_2"]])
-        for ii in range(len(params["coeff_0"])):
-            point = points[:,ii][:self.hull_dimention]
+        for coeff_index in range(len(params["coeff_0"])):
+            point = points[:, coeff_index][:self.hull_dimention]
             output_array.append(self._hull.find_simplex(point)>=0)
         return numpy.array(output_array)
 
--- a/pycbc/distributions/constraints.py
+++ b/pycbc/distributions/constraints.py
@@ -18,6 +18,10 @@ This modules provides classes for evaluating multi-dimensional constraints.
 
 from pycbc import transforms
 from pycbc.io import record
+from scipy.spatial import Delaunay
+import numpy
+import h5py
+
 
 class Constraint(object):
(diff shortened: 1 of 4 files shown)


pycbc/distributions/constraints.py
```diff
@@ -31,9 +35,24 @@ def __init__(self, constraint_arg, transforms=None, **kwargs):
...
+            pc_file = h5py.File(pc_filename, 'r')
+            pc_coefficients = numpy.array(pc_file.get('coefficients'))
+            pc_file.close()
+            hull_points = []
+            for dim in range(self.hull_dimention):
+                hull_points.append(pc_coefficients[:, dim])
+            hull_points = numpy.array(hull_points).T
+            pc_coeffs_hull = Delaunay(hull_points)
+            self._hull = pc_coeffs_hull
+
```
cdcapano, line 52: I'd rather not change the base Constraint class. These things are `ConvexHall` specific. So, add an `__init__` function to `ConvexHull`, and put this block of code there. See comment below for more details.

pycbc/distributions/constraints.py
```diff
@@ -31,9 +35,24 @@ def __init__(self, constraint_arg, transforms=None, **kwargs):
...
     def __call__(self, params):
         """Evaluates constraint.
-        """
+        """''
```
cdcapano, line 55: You have an errant open/close quote here.

pycbc/distributions/constraints.py
```diff
@@ -60,7 +79,27 @@ def _constraint(self, params):
         return params[self.constraint_arg]
 
 
+class ConvexHull(Constraint):
+    """Pre defined constraint for core-collapse waveforms that checks
+    whether a given set of coefficients lie within the convex hull of
+    the coefficients of the principal component basis vectors.
+    """
+    name = "supernovae_convex_hull"
+    required_parameters = ["coeff_0", "coeff_1"]
```
cdcapano, line 88: Add an `__init__` function to execute the code you have above. You can use `super` to call the parent class `__init__`; e.g.:
```
def __init__(self, constraint_arg, transforms=None, **kwargs):
    super(SupernovaeConvexHull, self).__init__(constraint_arg, transforms=transforms, **kwargs) # <- this will execute Constraint.__init__, thereby setting all of the attributes that are set there
   [ your block of code from above ]
```

pycbc/distributions/constraints.py
```diff
@@ -60,7 +79,27 @@ def _constraint(self, params):
         return params[self.constraint_arg]
 
 
+class ConvexHull(Constraint):
```
cdcapano, line 82: Call this class `SupernoveConvexHull` to be more explicit. Just calling it `ConvexHull` is a little too general.

pycbc/waveform/supernovae.py
```diff
@@ -0,0 +1,51 @@
...
+    coefficients of the principal components and the distance. The number of
+    principal components used can also be varied.
+    """
+
+    try:
+        principal_components = _principal_components['principal_components']
+    except KeyError:
+        pc_file = h5py.File(kwargs['principal_components_file'], 'r')
+        principal_components = pc_file['principal_components']
+        _principal_components['principal_components'] = principal_components
```
cdcapano, line 23: Use a `with` here instead, as it will protect against dangling file handlers; i.e.:
```
with h5py.File(kwargs['principal_components_file'], 'r') as pc_file:
    principal_components = pc_file['principal_components']
    _principal_components['principal_components'] = principal_components
```

cdcapano (changes requested):
Thanks @chaitanyaafle . I have a few requests. Also, please address the PEP-8 issues picked up by code climate.

cdcapano (approved):
Looks good, thanks @chaitanyaafle

PR #3488, opened 2020-10-05
Additional gating veto

veronica-villa:
Implement an additional veto for the triggers near the gates to reduce the ammount of potential glitches in the area and improve the sensitivity of the search.

### DIFF
--- a/bin/hdfcoinc/pycbc_fit_sngls_by_template
+++ b/bin/hdfcoinc/pycbc_fit_sngls_by_template
@@ -77,8 +77,8 @@ parser.add_argument("--veto-segment-name", nargs='*', default=[], action='append
                     "given all segments for a given ifo will be used")
 parser.add_argument("--gating-veto-windows", nargs='+',
                     action=MultiDetOptionAction,
-                    help="Time in seconds to be vetoed before and after each "
-                         "gate. Given as detector-values pairs, e.g. "
+                    help="Seconds to be vetoed before and after the central time "
+                         "of each gate. Given as detector-values pairs, e.g. "
                          "H1:-1,2.5 L1:-1,2.5 V1:0,0")
 parser.add_argument("--output", required=True,
                     help="Location for output file containing fit coefficients"
@@ -205,21 +205,20 @@ if args.gating_veto_windows:
     if gveto_before > 0 or gveto_after < 0:
         raise ValueError("Gating veto window values must be negative before "
                          "gates and positive after gates.")
-    if gveto_before == 0 and gveto_after == 0:
-        pass
-    else:
+    if not (gveto_before == 0 and gveto_after == 0):
         gate_times = np.unique(trigf[args.ifo + '/gating/auto/time'][:])
         gveto_segs = events.veto.start_end_to_segments(gate_times + gveto_before,
                                                        gate_times + gveto_after).coalesce()
         all_segments -= gveto_segs
-        gveto_retain_idx = events.veto.indices_outside_times(time, gate_times + gveto_before,
+        gveto_retain_idx = events.veto.indices_outside_times(time,
+                                                             gate_times + gveto_before,
                                                              gate_times + gveto_after)
         stat = stat[gveto_retain_idx]
         tid = tid[gveto_retain_idx]
         time = time[gveto_retain_idx]
         if args.save_trig_param:
             tparam = tparam[gveto_retain_idx]
-        logging.info('%i trigs left after vetoing triggers near gates' %len(stat))
+        logging.info('%i trigs left after vetoing triggers near gates' % len(stat))
 
 total_time = abs(all_segments)
 
--- a/bin/hdfcoinc/pycbc_fit_sngls_by_template
+++ b/bin/hdfcoinc/pycbc_fit_sngls_by_template
@@ -22,6 +22,7 @@ import copy, numpy as np
 from pycbc import io, events
 from pycbc.events import triggers, trigger_fits as trstats
 from pycbc.events.stat import sngl_statistic_dict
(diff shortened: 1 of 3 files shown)


bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -191,6 +197,30 @@ for veto_file, veto_segment_name in zip(args.veto_file, args.veto_segment_name):
...
+if args.gating_veto_windows:
+    gating_veto = args.gating_veto_windows[args.ifo].split(',')
+    gveto_before = float(gating_veto[0])
+    gveto_after = float(gating_veto[1])
+    if gveto_before > 0 or gveto_after < 0:
+        raise ValueError("Gating veto window values must be negative before "
+                         "gates and positive after gates.")
+    if gveto_before == 0 and gveto_after == 0:
+        pass
+    else:
```
GarethDaviesGW, line 210: as you're passing in the first, case, this can be set as:
`if not (gveto_before == 0 and gveto_after == 0):`

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -191,6 +197,30 @@ for veto_file, veto_segment_name in zip(args.veto_file, args.veto_segment_name):
...
+                                                       gate_times + gveto_after).coalesce()
+        all_segments -= gveto_segs
+        gveto_retain_idx = events.veto.indices_outside_times(time, gate_times + gveto_before,
+                                                             gate_times + gveto_after)
+        stat = stat[gveto_retain_idx]
+        tid = tid[gveto_retain_idx]
+        time = time[gveto_retain_idx]
+        if args.save_trig_param:
+            tparam = tparam[gveto_retain_idx]
+        logging.info('%i trigs left after vetoing triggers near gates' %len(stat))
```
GarethDaviesGW, line 222: the way the logging module works, this can be changed to:
`...near gates', len(stat))`
This means that if logging isn't set it doesn't do the string operation

GarethDaviesGW: if not doing the above, then you should have whitespace both sides of the `%` operator

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -191,6 +197,30 @@ for veto_file, veto_segment_name in zip(args.veto_file, args.veto_segment_name):
...
+        raise ValueError("Gating veto window values must be negative before "
+                         "gates and positive after gates.")
+    if gveto_before == 0 and gveto_after == 0:
+        pass
+    else:
+        gate_times = np.unique(trigf[args.ifo + '/gating/auto/time'][:])
+        gveto_segs = events.veto.start_end_to_segments(gate_times + gveto_before,
+                                                       gate_times + gveto_after).coalesce()
+        all_segments -= gveto_segs
+        gveto_retain_idx = events.veto.indices_outside_times(time, gate_times + gveto_before,
```
GarethDaviesGW, line 215: could this be split into two lines on the comma between `time` and `gate_times`?

bin/hdfcoinc/pycbc_fit_sngls_by_template
```diff
@@ -74,6 +75,11 @@ parser.add_argument("--veto-file", nargs='*', default=[], action='append',
 parser.add_argument("--veto-segment-name", nargs='*', default=[], action='append',
                     help="Name(s) of veto segments to apply. Optional, if not "
                     "given all segments for a given ifo will be used")
+parser.add_argument("--gating-veto-windows", nargs='+',
+                    action=MultiDetOptionAction,
+                    help="Time in seconds to be vetoed before and after each "
+                         "gate. Given as detector-values pairs, e.g. "
+                         "H1:-1,2.5 L1:-1,2.5 V1:0,0")
```
GarethDaviesGW, line 82: This isn't quite clear - the time *isn't* before and after the gate, as the gate already has finite size from the way the autogate works, and you use the central time.

So the question is really, does 'before and after each gate' refer to the central time (which is what is used here) or the edges of the gate - which is how I would understand it

I would suggest 'before and after the central time of each gate', but that may be too verbose

bin/hdfcoinc/pycbc_multiifo_coinc_findtrigs
```diff
@@ -14,6 +15,11 @@ parser.add_argument("--veto-files", nargs='*', action='append', default=[],
                          "contained in the file are ignored")
 parser.add_argument("--segment-name", nargs='*', action='append', default=[],
                     help="Optional, name of veto segment in veto file")
+parser.add_argument("--gating-veto-windows", nargs='+',
+                    action=MultiDetOptionAction,
+                    help="Time in seconds to be vetoed before and after each "
+                         "gate. Given as detector-values pairs, e.g. "
+                         "H1:-1,2.5 L1:-1,2.5 V1:0,0")
```
GarethDaviesGW, line 22: help message question applies here, as above

GarethDaviesGW (commented):
This looks really good!
I have some (very pedantic) comments on minor improvements, but they can be ignored if wanted

bin/hdfcoinc/pycbc_fit_sngls_by_template
veronica-villa: I did it with the whitespace because it was done that way in the rest of the code

bin/hdfcoinc/pycbc_fit_sngls_by_template
veronica-villa: You are totally right, thanks for pointing at it!

GarethDaviesGW (approved):
looks good to me then!

PR #3490, opened 2020-10-06
Ringdown bugs

JulianWesterweck:
This corrects the order of arguments for the frequency domain ringdown waveform and ensures that the functions for the desired approximant are used.

### DIFF
--- a/pycbc/waveform/ringdown.py
+++ b/pycbc/waveform/ringdown.py
@@ -531,7 +531,7 @@ def multimode_base(input_params, domain, freq_tau_approximant=False):
             outcross.data += hcross.data
     else:
         raise ValueError('unrecognised domain argument {}; '
-                                   'must be either fd or td'.format(domain))
+                         'must be either fd or td'.format(domain))
 
     return norm * outplus, norm * outcross
 
--- a/pycbc/waveform/ringdown.py
+++ b/pycbc/waveform/ringdown.py
@@ -457,14 +457,32 @@ def fd_damped_sinusoid(f_0, tau, amp, phi, delta_f, f_lower, f_final, t_0=0.,
 #### Base multi-mode for all approximants
 ######################################################
 
-def multimode_base(input_params, fd_approximant=False, td_approximant=False,
-                   freq_tau_approximant=False):
+def multimode_base(input_params, domain, freq_tau_approximant=False):
     """Return a superposition of damped sinusoids in either time or frequency
     domains with parameters set by input_params.
+
+    Parameters
+    ----------
+    domain : string
+        Choose domain of the waveform, either 'td' for time domain
+        or 'fd' for frequency domain.
+    freq_tau_approximant : {False, bool}, optional
+        Choose choose the waveform approximant to use. Either based on
+        mass/spin (set to False, default), or on frequencies/damping times
+        of the modes (set to True).
+
+    Returns
+    -------
+    hplus : TimeSeries
+        The plus phase of a ringdown with the lm modes specified and
+        n overtones in the chosen domain (time or frequency).
+    hcross : TimeSeries
+        The cross phase of a ringdown with the lm modes specified and
+        n overtones in the chosen domain (time or frequency).
     """
     input_params['lmns'] = format_lmns(input_params['lmns'])
     amps, phis = lm_amps_phases(**input_params)
-    if freq_tau_approximant or 'final_mass' not in input_params.keys():
+    if freq_tau_approximant:
         freqs, taus = lm_freqs_taus(**input_params)
         norm = 1.
     else:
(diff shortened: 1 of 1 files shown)


pycbc/waveform/ringdown.py
```diff
@@ -457,23 +457,28 @@ def fd_damped_sinusoid(f_0, tau, amp, phi, delta_f, f_lower, f_final, t_0=0.,
...
-    else:
-        freqs, taus = lm_freqs_taus(**input_params)
-        norm = 1.
-
-    if 'delta_t' in input_params.keys():
+    if not (fd_approximant or td_approximant):
+        if 'delta_t' in input_params.keys():
+            td_approximant = True
+        elif 'delta_f' in input_params.keys():
+            fd_approximant = True
```
cdcapano, line 480: Get rid of this if block when you change to a single `domain` argument; then just evaluate the `domain` below, raising an error if it isn't either `fd` or `td`. That means that you won't be trying to guess what to use based on the input params, but I think that's for the best. Having it try to guess can lead to unexpected results.

pycbc/waveform/ringdown.py
```diff
@@ -457,23 +457,28 @@ def fd_damped_sinusoid(f_0, tau, amp, phi, delta_f, f_lower, f_final, t_0=0.,
...
-def multimode_base(input_params):
+def multimode_base(input_params, fd_approximant=False, td_approximant=False,
+                   freq_tau_approximant=False):
     """Return a superposition of damped sinusoids in either time or frequency
     domains with parameters set by input_params.
     """
     input_params['lmns'] = format_lmns(input_params['lmns'])
     amps, phis = lm_amps_phases(**input_params)
-    if 'final_mass' in input_params.keys():
+    if freq_tau_approximant or 'final_mass' not in input_params.keys():
```
cdcapano, line 467: Just base it on the `freq_tau_approximant`.

pycbc/waveform/ringdown.py
```diff
@@ -457,23 +457,28 @@ def fd_damped_sinusoid(f_0, tau, amp, phi, delta_f, f_lower, f_final, t_0=0.,
 #### Base multi-mode for all approximants
 ######################################################
 
-def multimode_base(input_params):
+def multimode_base(input_params, fd_approximant=False, td_approximant=False,
+                   freq_tau_approximant=False):
     """Return a superposition of damped sinusoids in either time or frequency
     domains with parameters set by input_params.
```
cdcapano, line 463: Add a `Parameters` section to the doc string to explain what each keyword argument does. Also, you could combine this into a single, required argument `domain` that must be set to either `fd` or `td`. That'll simplify the logic below, and make it more straightforward to explain how to use the function.

pycbc/waveform/ringdown.py
```diff
@@ -457,23 +457,28 @@ def fd_damped_sinusoid(f_0, tau, amp, phi, delta_f, f_lower, f_final, t_0=0.,
...
-        freqs, taus = lm_freqs_taus(**input_params)
-        norm = 1.
-
-    if 'delta_t' in input_params.keys():
+    if not (fd_approximant or td_approximant):
+        if 'delta_t' in input_params.keys():
+            td_approximant = True
+        elif 'delta_f' in input_params.keys():
+            fd_approximant = True
+    if td_approximant:
```
cdcapano, line 481: Change to `if domain == 'td':`

pycbc/waveform/ringdown.py
```diff
@@ -499,15 +504,16 @@ def multimode_base(input_params):
             else:
                 outplus.data += hplus
                 outcross.data += hcross
-    elif 'delta_f' in input_params.keys():
+    elif fd_approximant:
```
cdcapano, line 507: `elif domain == 'fd':`

pycbc/waveform/ringdown.py
```diff
@@ -499,15 +504,16 @@ def multimode_base(input_params):
...
         outplus, outcross = fd_output_vector(freqs, taus,
                             input_params['delta_f'], input_params['f_final'])
```
cdcapano, line 509: Add here an
```
else:
    raise ValueError('unrecognized domain argument {}; '
                               'must be either fd or td'.format(domain))
```

pycbc/waveform/ringdown.py
```diff
@@ -577,7 +583,7 @@ def get_td_from_final_mass_spin(template=None, **kwargs):
 
     input_params = props(template, mass_spin_required_args, td_args, **kwargs)
 
-    return multimode_base(input_params)
+    return multimode_base(input_params, td_approximant=True)
```
cdcapano, line 586: With the above changes, here replace `td_approximant=True` with `'td'`.

pycbc/waveform/ringdown.py
```diff
@@ -637,7 +643,7 @@ def get_fd_from_final_mass_spin(template=None, **kwargs):
 
     input_params = props(template, mass_spin_required_args, fd_args, **kwargs)
 
-    return multimode_base(input_params)
+    return multimode_base(input_params, fd_approximant=True)
```
cdcapano, line 646: And here `fd_approximant=True` -> `'fd'`. Ditto for the other places below.

cdcapano (changes requested):
Thanks for fixing this Julian. Just a few requests related to how the `fd_approximant` and `td_approximant` are specified; see below.

micamu (commented):
I am not quite sure if these comments I had written yesterday got to you, so clicking an extra button now in case they didn't. They seem to be in agreement with some of Collin's comments

micamu:
And I just saw the recent changes, I approve now :)

micamu (approved):

JulianWesterweck:
Indeed, I did not see your comments until now - thanks!
I will look at fixing the codeclimate complaint.

cdcapano (approved):

cdcapano:
Looks good, thank @JulianWesterweck

PR #3491, opened 2020-10-07
coinc_lim_for_thresh for BBH statistics

GarethDaviesGW:
the mchirp weighting got missed off in the coinc_lim_for_thresh stuff, which means that some background coincs were dismissed when they shouldn't have been

### DIFF
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -1621,6 +1621,14 @@ class ExpFitSGPSDFgBgNormBBHStatistic(ExpFitSGFgBgNormNewStatistic):
         logr_s += numpy.log((self.curr_mchirp / 20.0) ** (11./3.0))
         return logr_s
 
+    def coinc_multiifo_lim_for_thresh(self, s, thresh, limifo,
+                                      **kwargs): # pylint:disable=unused-argument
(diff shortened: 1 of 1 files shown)


tdent (approved):

connor-mcisaac:
I guess this assumes that you have run the "single" method before this to get the right chirp mass, which is definitely the case for the findtrigs code, not sure where else this code is used.

GarethDaviesGW:
> I guess this assumes that you have run the "single" method before this to get the right chirp mass, which is definitely the case for the findtrigs code, not sure where else this code is used.

I had considered trying to make this so that it uses the minimum chirp mass from the bank, using the "assume best case scenario" principle, but think that would have been overcomplicating things

connor-mcisaac:
> > I guess this assumes that you have run the "single" method before this to get the right chirp mass, which is definitely the case for the findtrigs code, not sure where else this code is used.
> 
> I had considered trying to make this so that it uses the minimum chirp mass from the bank, using the "assume best case scenario" principle, but think that would have been overcomplicating things

Since the findtrig code only does 1 template at a time it's fine to use the actually mchirp, and as you say it makes thing a bit simpler.

connor-mcisaac (approved):

GarethDaviesGW:
This has been tested by comparing coinc_findtrigs output with and without this change to coinc_findtrigs output with the coinc_lim_for_thresh disabled. This fixes the problem

PR #3492, opened 2020-10-07
Add lmn_from_other_lmn conversions to list; fix bug

JulianWesterweck:
The lmn_from_other_lmn conversions were missing from the list. 

These conversions claim to accept floats, but only work with numpy arrays. Now both should work with giving the same output type as the input. 
However, if instead of a float, integers are given as input, the output will instead be changed to an array. 
Should that be dealt with also?

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -1119,11 +1119,16 @@ def freqlmn_from_other_lmn(f0, tau, current_l, current_m, new_l, new_m):
         correspond to an unphysical Kerr black hole mass and/or spin,
         ``numpy.nan`` will be returned.
     """
-    mass = final_mass_from_f0_tau(f0, tau, l=current_l, m=current_m)
-    spin = final_spin_from_f0_tau(f0, tau, l=current_l, m=current_m)
+    mass = numpy.array(final_mass_from_f0_tau(f0, tau, l=current_l,
+                                              m=current_m))
+    spin = numpy.array(final_spin_from_f0_tau(f0, tau, l=current_l,
+                                              m=current_m))
 
     mass[mass < 0] = numpy.nan
     spin[numpy.abs(spin) > 0.9996] = numpy.nan
+
+    if type(f0) is float and type(tau) is float:
+        mass, spin = float(mass), float(spin)
 
     return freq_from_final_mass_spin(mass, spin,
                                      l=new_l, m=new_m, nmodes=1)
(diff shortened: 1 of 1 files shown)


pycbc/conversions.py
```diff
@@ -1119,12 +1119,17 @@ def freqlmn_from_other_lmn(f0, tau, current_l, current_m, new_l, new_m):
         correspond to an unphysical Kerr black hole mass and/or spin,
         ``numpy.nan`` will be returned.
     """
-    mass = final_mass_from_f0_tau(f0, tau, l=current_l, m=current_m)
-    spin = final_spin_from_f0_tau(f0, tau, l=current_l, m=current_m)
+    mass = numpy.array(final_mass_from_f0_tau(f0, tau, l=current_l,
```
micamu, line 1122: Maybe you could use the function `ensurearray`, like it is done here https://github.com/gwastro/pycbc/blob/master/pycbc/conversions.py#L1045. So leaving these two lines as they were and then adding
`mass, spin, _ = ensurearray(mass, spin)`

pycbc/conversions.py
```diff
@@ -1119,12 +1119,17 @@ def freqlmn_from_other_lmn(f0, tau, current_l, current_m, new_l, new_m):
...
+                                              m=current_m))
+    spin = numpy.array(final_spin_from_f0_tau(f0, tau, l=current_l,
+                                              m=current_m))
 
     mass[mass < 0] = numpy.nan
     spin[numpy.abs(spin) > 0.9996] = numpy.nan
 
+    if type(f0) is float and type(tau) is float:
```
micamu, line 1130: What you could do here instead is `if len(mass) == 1`, that would solve the issue with the ints. And we should not add in this loop `and len(spin) == 1`, we should  just raise an error above if mass and spin don't have the same lengt. 
But I am not quite convinced about this, is it that important to not return an array if the input is not an array?

micamu (changes requested):

pycbc/conversions.py
JulianWesterweck: Maybe it is not an issue - but after the problems we had with the dimensions of arrays in the past, I was worried this might be an issue, too. I expect that depends on the functions using the output of the conversion.

micamu: I just realized, one would hardly get mass and spin with different lengths, what the error should be raised on is if f0 and tau do not have the same length

pycbc/conversions.py
JulianWesterweck: Yes, we could just do it in the same way as was used in that function - then the behaviour would be consistent, and it also already takes care of the output type.
I will add that.

JulianWesterweck:
@micamu Now it is using `ensurearray` to cast the input to arrays. The output of the functions now is consistent with what `final_mass_from_f0_tau` does. The conversions for the QNMs are not all consistent regarding the output being turned into an array - but that probably does not matter for now.

micamu:
@JulianWesterweck For consistency with other functions in the module, it might make sense to use formatreturn with the input_is_array returned from ensurearray. 
https://github.com/gwastro/pycbc/blob/master/pycbc/conversions.py#L75
It might not be strictly necessary, but cleaner

micamu (approved):

JulianWesterweck:
@cdcapano Do you have a preference here?
I think using `format_return` for these conversions is not necessary (as it is called already in on of the functions used by the conversion).
But as Miriam said, it might be cleaner in the code to do it anyway?

cdcapano:
@JulianWesterweck Just realized I was responsible for this. Yes, use `format_return` as @micamu suggested. That'll make sure that if a float/int (array) was input, a float/int (array) is returned.

cdcapano (approved):

cdcapano:
@JulianWesterweck Rebase to master to get the builds to start.

PR #3493, opened 2020-10-09
prepare for release v1.16.11

GarethDaviesGW:


spxiwh (approved):

tdent:
Note that the 'gate veto' option hasn't yet been incorporated into the split hist trigger plots, so these will (currently) not be useful as diagnosis of the pipeline coinc outputs. I am checking with @veronica-villa whether this can be done quickly.

GarethDaviesGW:
> Note that the 'gate veto' option hasn't yet been incorporated into the split hist trigger plots, so these will (currently) not be useful as diagnosis of the pipeline coinc outputs. I am checking with @veronica-villa whether this can be done quickly.

Okay - I'll hold off for a little while

PR #3498, opened 2020-10-09
add script to monitor an inference job and generate updated plots and results page when it updates

ahnitz:
This script monitors an inference output file. When that file updates, it generates a series of plots and makes a combined results page. See example here 

https://www.atlas.aei.uni-hannover.de/work/ahnitz/projects/tpage/results/html/checkpoints/1/

The plots generated are controlled by a config file. Multiple executables can be used as long as they use the the '--input-file' and '--output-file' options and multiple plots can be made for each by using subsections of their executable keyname. I.E. a second plot here could be made with a 'posterior-2' section. 

```
[workflow]
[results_page]
template-file = templates/orange.html

[executables]
results_page = ${which:pycbc_make_html_page}
posterior = ${which:pycbc_inference_plot_posterior}

[posterior]
[posterior-1]
z-arg = snr
```

The monitor is then started with
```
pycbc_inference_monitor \
--verbose \
--inference-file single.hdf \
--config-files plot.ini  \
--output-dir results
```

### DIFF
--- a/bin/inference/pycbc_inference_monitor
+++ b/bin/inference/pycbc_inference_monitor
@@ -6,10 +6,12 @@ import os, sys, time, glob, shutil, logging, argparse, pycbc
 import pycbc.workflow as wf
 
 parser = argparse.ArgumentParser()
-parser.add_argument('--inference-file')
-parser.add_argument('--check-interval', default=10, type=int)
-parser.add_argument('--output-dir')
-parser.add_argument('--allow-failure', action='store_true')
+parser.add_argument('--inference-file', help="The name of the inference file")
+parser.add_argument('--check-interval', default=10, type=int,
+                    help="Polling interval to check for file changes (s)")
+parser.add_argument('--output-dir', help="Output plots / results directory")
+parser.add_argument('--allow-failure', action='store_true',
+                    help="Allow for a failure in plot generation")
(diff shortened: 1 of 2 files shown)


bin/inference/pycbc_inference_monitor
```diff
@@ -0,0 +1,81 @@
+#!/bin/env python
```
cdcapano, line 1: Should you add a copywrite?

bin/inference/pycbc_inference_monitor
```diff
@@ -0,0 +1,81 @@
...
+
+import os, sys, time, glob, shutil, logging, argparse, pycbc
+import pycbc.workflow as wf
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--inference-file')
+parser.add_argument('--check-interval', default=10, type=int)
+parser.add_argument('--output-dir')
+parser.add_argument('--allow-failure', action='store_true')
+parser.add_argument('--verbose', action='store_true')
```
cdcapano, line 13: Add some help messages.

cdcapano (approved):
Looks fine. Could you just add some help messages to the parser before merging?

cdcapano:
The Travis failure looks unrelated.

Issue #3395, opened 2020-07-24
Need to decode approximant strings in pycbc_condition_strain 


maxtrevor:
Currently pycbc_condition_strain does not recognize approximants after reading them from the variable parameters table of an injection hdf. For example, when approximant 'SEOBNRv4' is stored as a variable parameter in an injection hdf, reading the hdf gives back a string b'SEOBNRv4' which is not recognized as being the same thing as 'SEOBNRv4'

Running the pycbc live test contained in my pull request  https://github.com/gwastro/pycbc/pull/3322 causes an error when pycbc_condition_strain gets to Applying Injections

2020-07-24 12:14:04,373 Applying injections
Traceback (most recent call last):
  File "/home/max.trevor/dev_env/bin/pycbc_condition_strain", line 4, in <module>
    __import__('pkg_resources').run_script('PyCBC===e9f3da', 'pycbc_condition_strain')
  File "/home/max.trevor/dev_env/lib/python3.6/site-packages/pkg_resources/__init__.py", line 667, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/max.trevor/dev_env/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1464, in run_script
    exec(code, namespace, namespace)
  File "/home/max.trevor/dev_env/lib/python3.6/site-packages/PyCBC-e9f3da-py3.6-linux-x86_64.egg/EGG-INFO/scripts/pycbc_condition_strain", line 87, in <module>
    precision=args.output_precision)
  File "/home/max.trevor/dev_env/lib/python3.6/site-packages/PyCBC-e9f3da-py3.6-linux-x86_64.egg/pycbc/strain/strain.py", line 392, in from_cli
    inj_filter_rejector=inj_filter_rejector)
  File "/home/max.trevor/dev_env/lib/python3.6/site-packages/PyCBC-e9f3da-py3.6-linux-x86_64.egg/pycbc/inject/inject.py", line 505, in apply
    detector_name, f_lower=f_l, distance_scale=distance_scale)
  File "/home/max.trevor/dev_env/lib/python3.6/site-packages/PyCBC-e9f3da-py3.6-linux-x86_64.egg/pycbc/inject/inject.py", line 557, in make_strain_from_inj_object
    **self.extra_args)
[...]

PR #3502, opened 2020-10-16
Fix error with approximant strings in HDF injections

titodalcanton:
This is a proposal for fixing #3395. Apparently the HDF injection code has trouble when the approximant field (or any other string field) is not static. The fix is based on the approach already used for handling static arguments.

I am not sure if this will break existing use cases, and I have not tested it with Python 2, but opening for consideration.


cdcapano (approved):
Look fine. Storing string values for injection datasets hasn't really been used much previously, so I think this should be ok.

PR #3504, opened 2020-10-16
Fix case where one detector combo has no foreground triggers

connor-mcisaac:
The add statmap code is breaking when finding the most significant trigger for hierarchical removal if there are no triggers in a given detector combination.

In the dictionary comprehension it does argmax for each detector combo, if one is empty this produces an error.
This can be fixed by including an if statement in the comprehension to test that the array is non-zero in size.

### DIFF
--- a/bin/hdfcoinc/pycbc_multiifo_add_statmap
+++ b/bin/hdfcoinc/pycbc_multiifo_add_statmap
@@ -453,9 +453,11 @@ while True:
                     combined_fg_data.data[key][:]
 
     max_each_combo = {combo: sep_fg_data[combo].data['ifar'][:].argmax()
-                      for combo in all_ifo_combos}
+                      for combo in all_ifo_combos
+                      if len(sep_fg_data[combo].data['ifar'][:]) > 0}
     max_ifars = {combo: sep_fg_data[combo].data['ifar'][:][max_each_combo[combo]]
-                    for combo in all_ifo_combos}
+                 for combo in all_ifo_combos
+                 if len(sep_fg_data[combo].data['ifar'][:]) > 0}
 
     logging.info('Maximum IFAR values per combination:')
     for k in max_ifars:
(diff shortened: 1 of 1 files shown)


GarethDaviesGW (commented):
This looks good and fixes the issue - as its doing the same loop twice it _might_ be preferable to have a normal loop which adds to a dictionary, but dont think it really matters

Have you tested that nothing later in the code relies on the keys of either dictionary for the full list of combinations?

bin/hdfcoinc/pycbc_multiifo_add_statmap
```diff
@@ -453,9 +453,11 @@ while True:
                     combined_fg_data.data[key][:]
 
     max_each_combo = {combo: sep_fg_data[combo].data['ifar'][:].argmax()
-                      for combo in all_ifo_combos}
+                      for combo in all_ifo_combos
+                      if len(sep_fg_data[combo].data['ifar'][:]) > 0}
     max_ifars = {combo: sep_fg_data[combo].data['ifar'][:][max_each_combo[combo]]
-                    for combo in all_ifo_combos}
+                 for combo in all_ifo_combos
```
GarethDaviesGW, line 459: If this is updated to `for combo in max_each_combo.keys()` that would make it a bit more readable

connor-mcisaac:
> This looks good and fixes the issue - as its doing the same loop twice it _might_ be preferable to have a normal loop which adds to a dictionary, but dont think it really matters
> 
> Have you tested that nothing later in the code relies on the keys of either dictionary for the full list of combinations?

After this point these 2 dicts are used to get the loudest trigger and then are not used after that.

bin/hdfcoinc/pycbc_multiifo_add_statmap
connor-mcisaac: I've updated to use the keys and value of the previous dict to make it simpler

GarethDaviesGW (approved):
Looks good, thanks!

PR #3505, opened 2020-10-16
Write config file to inference file

cdcapano:
Stores the config file used for pycbc inference to the output hdf file.

The config file is stored as a pickled String buffer to `config_file/{index}` in the hdf file. The `{index}` is an integer that keeps track of the write. If a job is restarted from a checkpoint, the config file will be stored as the last index + 1. This also allows you to track any changes that were made to the config file between restarts. (I suppose if I wanted to be super efficient I'd only store a diff on each, but since each config file is only a few hundred bytes, it's not worth it.)

A convenience method is added to the BaseInferenceFile to read the config file. By default it will load the last one saved, and will return it as a WorkflowConfigParser object. You can have it just return a string buffer though using the `return_cp` kwarg. Example reading:
```
>>> fp = io.loadfile('inference.hdf', 'r')
>>> cp = fp.read_config_file()
>>> cp.get('model', 'name')
'gaussian_noise'
```

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -33,6 +33,7 @@ from pycbc.waveform import generator
 from pycbc import __version__
 from pycbc import inference
 from pycbc.inference import (models, burn_in, option_utils)
+from pycbc.inference.io import loadfile
 from pycbc.workflow import configuration
 
 # command line usage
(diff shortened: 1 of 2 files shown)


ahnitz (approved):
Looks good, once travis passes, please merge

cdcapano:
Looks like the issue with python 2.7 is legit:
```

  File "/home/travis/virtualenv/python2.7.15/bin/pycbc_inference", line 144, in <module>

    fp.write_config_file(cp)

  File "/home/travis/virtualenv/python2.7.15/lib/python2.7/site-packages/pycbc/inference/io/base_hdf.py", line 869, in write_config_file

    cp.write(out)

  File "/opt/python/2.7.15/lib/python2.7/ConfigParser.py", line 407, in write

    fp.write("[%s]\n" % section)

TypeError: unicode argument expected, got 'str'
```
Damn unicode... I'll try and fix in the morning.

cdcapano:
Using `StringIO.StringIO` for python 2.7 seems to have worked, so merging.

PR #3506, opened 2020-10-16
Script to calculate stats, write to file

cdcapano:
Adds a script that will load samples from a file, recalculate the likelihood and prior for each point, then write it back out to a new file, along with any additional auxiliary statistics (e.g., SNR in each detector, max likelihood phase) that the model saves. Mostly useful for the nested samplers and emcee_pt/ptemcee, which do not store blob data, but can also be used to recalculate likelihood for a different model.

Depends on #3505.


cdcapano:
Rebased now that #3505 is merged. This is ready for review now.

ahnitz (approved):

PR #3509, opened 2020-10-21
 Correcting git clone 

AlexBrown6291:
git clone command in the documentation has a typo.  This fixes the typo.


titodalcanton (approved):
Good catch and thanks for fixing it!

PR #3510, opened 2020-10-22
Fix empty plots in minifollowups

MPillas:
We noticed that, in rare occasions, the single-template plots in the injection minifollowups are empty. This appears to be due to the fact that pycbc_injection_minifollowup compares the injection merger time to the science segments (instead of the analyzed segments) when setting up single-template plots. This pull request fixes the issue, and it also uses the injection merger time at the detector instead of the geocenter.


spxiwh (approved):
Thanks Marion! This is good to merge. There seems to be some weirdness that Travis hasn't started here. I think this needs to be rebased and pushed again to force Travis to try again. (Check with Tito if that didn't make sense!)

PR #3518, opened 2020-11-02
coinc_mergetrigs: deduplicate livetime segments

titodalcanton:
@MPillas and I noticed that `HDF_TRIGGER_MERGE` files store duplicated live time segments (datasets `search/start_time` and `search/end_time`) when the inspiral jobs use bank splitting. This appears to be correctly taken into account by some of the codes reading `HDF_TRIGGER_MERGE` files (for example `pycbc_fit_sngls_by_template`), but not all (for example `pycbc_plot_throughput`). This patch deduplicates the segments right away when creating the merged triggers, so as to avoid confusion and errors downstream.

### DIFF
--- a/bin/hdfcoinc/pycbc_coinc_mergetrigs
+++ b/bin/hdfcoinc/pycbc_coinc_mergetrigs
@@ -85,8 +85,8 @@ for filename in args.trigger_files:
         logging.error('Cannot open %s', filename)
         raise e
     ifo_data = data[ifo]
-    segments.add((ifo_data['search/start_time'][:],
-                  ifo_data['search/end_time'][:]))
+    segments.add((float(ifo_data['search/start_time'][:]),
+                  float(ifo_data['search/end_time'][:])))
 
     if 'templates_per_core' in ifo_data['search'].keys():
         tpc = numpy.append(tpc, ifo_data['search/templates_per_core'][:])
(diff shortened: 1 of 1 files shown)


bin/hdfcoinc/pycbc_coinc_mergetrigs
```diff
@@ -86,8 +85,8 @@ for filename in args.trigger_files:
         logging.error('Cannot open %s', filename)
         raise e
     ifo_data = data[ifo]
-    s, e = ifo_data['search/start_time'][:], ifo_data['search/end_time'][:]
-    start, end = numpy.append(start, s), numpy.append(end, e)
+    segments.add((float(ifo_data['search/start_time'][:]),
```
tdent, line 88: looks OK. I would just put a comment eg 'Don't add segments that duplicate existing ones'

bin/hdfcoinc/pycbc_coinc_mergetrigs
```diff
@@ -107,8 +106,11 @@ for filename in args.trigger_files:
                 if not gk in gating:
                     gating[gk] = numpy.array([], dtype=numpy.float64)
                 gating[gk] = numpy.append(gating[gk], gk_data[:])
-    data.close()    
-f['%s/search/start_time' % ifo], f['%s/search/end_time' % ifo] = start, end
+    data.close()
+
+segments = numpy.array(sorted(segments), dtype=numpy.float64)
```
tdent, line 111: is it clear what sorting will do to 'segments' (a set of tuples) ? is it desirable?

titodalcanton: It will convert the set to a list [sorted by the first item of each tuple](https://docs.python.org/3/reference/expressions.html#value-comparisons), i.e. the start time of the segment, which seemed desirable to me.

tdent (approved):

bin/hdfcoinc/pycbc_coinc_mergetrigs
titodalcanton: Added a comment with the intended behavior for good measure…

PR #3522, opened 2020-11-06
update retention levels for opt_snr jobs in parallelised cases

GarethDaviesGW:
the default for LigolwAddExecutable is to _not_ store the output file, but for optimal_snr jobs the default _is_ to store these - this is the opposite of what is useful in most cases when the optimal_snr job is parallelised

The optimal_snr jobs in the parallelised case are not used individually anywhere, so intermediate_product is appropriate


titodalcanton (approved):
Looks good to me, I will just ask to fix the line-too-long issue.

PR #3524, opened 2020-11-11
Unpin numpy, but avoid 1.19.0

duncanmmacleod:
The issue reported in https://github.com/numpy/numpy/issues/16660 was fixed in https://github.com/numpy/numpy/pull/16666 and released as part of numpy 1.19.1, so pycbc should now be able to use those newer versions, and just avoid 1.19.0.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,4 +2,4 @@
 requires = ["setuptools",
             "wheel",
             "cython>=0.29",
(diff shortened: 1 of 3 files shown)


ahnitz:
@duncanmmacleod Please rebase from master and I believe this should now pass

ahnitz (approved):

PR #3525, opened 2020-11-11
Store max likelihood loglr in polarization and brute force model

cdcapano:
Changes the polarization marginalization model to store the maxL log likelihood ratio. Also has the brute force marginalization model use this instead of the marginalized log likelihood ratio, as it is currently.


ahnitz (approved):
Looks OK I think.

ahnitz:
@cdcapano Rebase from master so travis will pass

ahnitz:
@cdcapano Please rebase, so this can be merged.

PR #3528, opened 2020-11-19
fix bug for mean mchirp m_tot in coinc inspiral table

GarethDaviesGW:
Mchirp and mtot are averaged from the singles to get values in the coinc table - I assume that this is in case we don't require exact match in the future.

This mean used the wrong number of ifos as the denominator; the number of operating rather than involved IFOs

This fixes the bug

### DIFF
--- a/pycbc/io/hdf.py
+++ b/pycbc/io/hdf.py
@@ -906,17 +906,16 @@ class ForegroundTriggers(object):
             coinc_id = lsctables.CoincID(idx)
 
             # Set up sngls
-            # FIXME: As two-ifo is hardcoded loop over all ifos
             sngl_mchirps = []
             sngl_mtots = []
             net_snrsq = 0
-            involved_ifos = []
+            triggered_ifos = []
             for ifo in ifos:
                 # If this ifo is not participating in this coincidence then
                 # ignore it and move on.
                 if not sngl_col_vals['snr'][ifo][1][idx]:
                     continue
-                involved_ifos += [ifo]
+                triggered_ifos += [ifo]
                 event_id = lsctables.SnglInspiralID(sngl_event_count)
                 sngl_event_count += 1
                 sngl = return_empty_sngl()
@@ -937,8 +936,8 @@ class ForegroundTriggers(object):
                 sngl.mchirp, _ = pnutils.mass1_mass2_to_mchirp_eta(
                         sngl.mass1, sngl.mass2)
                 sngl.eff_distance = (sngl.sigmasq)**0.5 / sngl.snr
-                # If we ever decide to not do exact match, then mchirp
-                # needs combined from the different singles
+                # If exact match is not used, then take mean
+                # masses from the single triggers
                 sngl_mchirps += [sngl.mchirp]
                 sngl_mtots += [sngl.mtotal]
 
@@ -951,16 +950,16 @@ class ForegroundTriggers(object):
                 coinc_map_row.event_id = event_id
                 coinc_event_map_table.append(coinc_map_row)
 
-            sngl_combined_mchirp = np.array(sngl_mchirps).mean()
-            sngl_combined_mtot = np.array(sngl_mtots).mean()
+            sngl_combined_mchirp = np.mean(sngl_mchirps)
+            sngl_combined_mtot = np.mean(sngl_mtots)
 
             # Set up coinc inspiral and coinc event tables
             coinc_event_row = lsctables.Coinc()
             coinc_inspiral_row = lsctables.CoincInspiral()
             coinc_event_row.coinc_def_id = coinc_def_id
-            coinc_event_row.nevents = len(involved_ifos)
-            coinc_event_row.instruments = ','.join(involved_ifos)
-            coinc_inspiral_row.set_ifos(involved_ifos)
+            coinc_event_row.nevents = len(triggered_ifos)
+            coinc_event_row.instruments = ','.join(triggered_ifos)
+            coinc_inspiral_row.set_ifos(triggered_ifos)
             coinc_event_row.time_slide_id = time_slide_id
             coinc_event_row.process_id = proc_id
             coinc_event_row.coinc_event_id = coinc_id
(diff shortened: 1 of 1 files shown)


titodalcanton:
Thanks for fixing this @GarethDaviesGW. Some comments:
* I see you changed "combined" to "mean" in the var names, which is kind of what I was complaining about yesterday. However, Tom's comment about "combined" in the var names is actually important here. Before the final divisions, the variables do not contain the means, so "combined" is actually more appropriate.
* However, I think the code is even clearer if you accumulate the single chirp and total masses into two lists (`single_mchirps` and `single_mtotals`?), and take the means from the lists at the end, instead of calculating the sum as you go.
* Is `n_involved_ifos` always going to be equal to `sngl_event_count`? If so, you should use a single variable.
* Would it be easy to also fix the incorrectly reported list of interferometers?
* Before merging, could you do a quick test of this on gracedb-playground?

GarethDaviesGW:
> Thanks for fixing this @GarethDaviesGW. Some comments:
> 
> * I see you changed "combined" to "mean" in the var names, which is kind of what I was complaining about yesterday. However, Tom's comment about "combined" in the var names is actually important here. Before the final divisions, the variables do not contain the means, so "combined" is actually more appropriate.
> * However, I think the code is even clearer if you accumulate the single chirp and total masses into two lists (`single_mchirps` and `single_mtotals`?), and take the means from the lists at the end, instead of calculating the sum as you go.

Thanks - I've changed back to combined _and_ done the list thing

> * Is `n_involved_ifos` always going to be equal to `sngl_event_count`? If so, you should use a single variable.

I'm not sure on that, but have used numpy's mean method instead of an explicit number in order to avoid issues like this happening again

> * Would it be easy to also fix the incorrectly reported list of interferometers?

I have done, it's updated on coinc table on the upload. I think the variables around it are also based on this, so have done those as well

> * Before merging, could you do a quick test of this on gracedb-playground?

See https://gracedb-playground.ligo.org/events/T362836/view/

GarethDaviesGW:
(just saw i forgot to add tom to the reviewer list)

lpsinger:
> Mchirp and mtot are averaged from the singles to get values in the coinc table - I assume that this is in case we don't require exact match in the future.

What about asserting that they are all the same?

Localization would not work correctly if the templates are different in different detectors.

pycbc/io/hdf.py
```diff
@@ -907,14 +907,16 @@ def to_coinc_xml_object(self, file_name):
...
+            sngl_mchirps = []
+            sngl_mtots = []
             net_snrsq = 0
+            involved_ifos = []
             for ifo in ifos:
                 # If this ifo is not participating in this coincidence then
                 # ignore it and move on.
                 if not sngl_col_vals['snr'][ifo][1][idx]:
                     continue
+                involved_ifos += [ifo]
```
titodalcanton, line 919: I would use `involved_ifos.append(ifo)` but this is fine too.

pycbc/io/hdf.py
```diff
@@ -947,16 +951,16 @@ def to_coinc_xml_object(self, file_name):
                 coinc_map_row.event_id = event_id
                 coinc_event_map_table.append(coinc_map_row)
 
-            sngl_combined_mchirp = sngl_combined_mchirp / len(ifos)
-            sngl_combined_mtot = sngl_combined_mtot / len(ifos)
+            sngl_combined_mchirp = np.array(sngl_mchirps).mean()
+            sngl_combined_mtot = np.array(sngl_mtots).mean()
```
titodalcanton, line 955: Note that the shorter/simpler `np.mean(sngl_mchirps)` works too.

titodalcanton:
> > Mchirp and mtot are averaged from the singles to get values in the coinc table - I assume that this is in case we don't require exact match in the future.
> 
> What about asserting that they are all the same?
> 
> Localization would not work correctly if the templates are different in different detectors.

@lpsinger I think the existing code was intentionally designed to generically support the possibility of a search using different templates in different detectors (maybe @spxiwh can chime in here). Also note that this code is used to upload offline events, not low-latency ones, so BAYESTAR will typically not be used here.

tdent:
It's not clear to me if the list of ifos *observing* should be passed to gracedb in the coinc 'instruments' field - this would be the relevant answer for localization as all observing ifos are used.  

If this code is exclusively for uploading offline events then I guess the usage proposed here, setting it to the *triggered* ifos, is OK.

pycbc/io/hdf.py
```diff
@@ -907,14 +907,16 @@ def to_coinc_xml_object(self, file_name):
 
             # Set up sngls
             # FIXME: As two-ifo is hardcoded loop over all ifos
```
tdent, line 909: remove the 'FIXME' !

pycbc/io/hdf.py
```diff
@@ -907,14 +907,16 @@ def to_coinc_xml_object(self, file_name):
...
-            sngl_combined_mchirp = 0
-            sngl_combined_mtot = 0
+            sngl_mchirps = []
+            sngl_mtots = []
             net_snrsq = 0
+            involved_ifos = []
```
tdent, line 913: to be clear, I would rather call this 'triggered_ifos'

pycbc/io/hdf.py
```diff
@@ -935,8 +937,10 @@ def to_coinc_xml_object(self, file_name):
                 sngl.mchirp, _ = pnutils.mass1_mass2_to_mchirp_eta(
                         sngl.mass1, sngl.mass2)
                 sngl.eff_distance = (sngl.sigmasq)**0.5 / sngl.snr
-                sngl_combined_mchirp += sngl.mchirp
-                sngl_combined_mtot += sngl.mtotal
+                # If we ever decide to not do exact match, then mchirp
+                # needs combined from the different singles
```
tdent, line 941: rewrite comment : 'If exact match is not used, take mean masses over different triggers'

pycbc/io/hdf.py
GarethDaviesGW: that makes more sense

tdent (approved):

PR #3529, opened 2020-11-20
Improve pycbc_plot_vt_ratio

titodalcanton:
* Expand `--help` description to make it clear where the necessary files come from.
* Open the input files read-only.
* Replace assertion with a more appropriate argparse error.
* Add metadata to the output file as done in other plotting scripts.
* Add support for writing metadata to PDF images.
* Code cleanup.

### DIFF
--- a/bin/plotting/pycbc_plot_vt_ratio
+++ b/bin/plotting/pycbc_plot_vt_ratio
@@ -1,44 +1,57 @@
 #!/usr/bin/python
 
+"""
+This script compares the sensitivities (VTs) of two searches having a common
+set of injections. It reads two HDF files produced by pycbc_page_sensitivity's
+--hdf-out option, and plots the ratios of their VTs at various IFARs.
+"""
+
 import h5py
 import matplotlib
 matplotlib.use('Agg')
 import matplotlib.pyplot as plt
 import numpy as np
 import argparse
+from pycbc.results import save_fig_with_metadata
 
-parser = argparse.ArgumentParser()
-parser = argparse.ArgumentParser(usage="",
-    description="Plot ratios of VTs calculated at various IFARs using pycbc_page_sensitivity for two analyses")
+
+parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument('--vt-file-one', required=True,
-                    help='HDF file containing VT curves, first set of data for comparison')
+                    help='HDF file containing VT curves, '
+                         'first set of data for comparison')
 parser.add_argument('--vt-file-two', required=True,
-                    help='HDF file containing VT curves, second set of data for comparison')
+                    help='HDF file containing VT curves, '
+                         'second set of data for comparison')
 parser.add_argument('--desc-one',  required=True,
-                    help='descriptor tag for first set of data (short, for use in subscript)')
+                    help='Descriptor tag for first set of data '
+                         '(short, for use in subscript)')
 parser.add_argument('--desc-two', type=str, required=True,
-                    help='descriptor tag for second set of data (short, for use in subscript)')
+                    help='Descriptor tag for second set of data '
+                         '(short, for use in subscript)')
 parser.add_argument('--outfile', type=str, required=True,
                     help='Output file to save to')
 parser.add_argument('--ifars', type=float, required=True, nargs='+',
-                    help='IFAR values to plot VT ratio for. Note that the plotted values '
-                         'will be the closest values available from the VT files')
-parser.add_argument('--log-x', action='store_true', default=False,
+                    help='IFAR values to plot VT ratio for. Note that the '
+                         'plotted values will be the closest values available '
+                         'from the VT files')
+parser.add_argument('--log-x', action='store_true',
                     help='Use logarithmic x-axis')
-parser.add_argument('--log-y', action='store_true', default=False,
+parser.add_argument('--log-y', action='store_true',
                     help='Use logarithmic y-axis')
 args = parser.parse_args()
 
 # Load in the two datasets
-ffirst_set = h5py.File(args.vt_file_one)
-fsecond_set = h5py.File(args.vt_file_two)
+ffirst_set = h5py.File(args.vt_file_one, 'r')
+fsecond_set = h5py.File(args.vt_file_two, 'r')
 
 # Find the index closest to the given IFAR value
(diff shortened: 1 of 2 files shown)


bin/plotting/pycbc_plot_vt_ratio
```diff
@@ -1,44 +1,58 @@
 #!/usr/bin/python
 
+"""
+This script compares the sensitivities (VTs) of two searches having a common
+set of injections. It reads two HDF files produced by pycbc_page_sensitivity's
```
GarethDaviesGW, line 5: Common set of injections isn't required, and it's actually assumed that the injections are independent (This makes errors in comparisons look worse)

bin/plotting/pycbc_plot_vt_ratio
```diff
@@ -127,10 +137,11 @@ ax_mi.set_ylabel(r'$\frac{VT(\mathrm{' + args.desc_one +'})}{VT(\mathrm{' +
                  args.desc_two +'})}$')
 plt.tight_layout()
 
-# write out to file, save as pdf if requested
-if args.outfile.split('.')[-1] == 'pdf':
-    fig_mi.savefig(args.outfile, format='PDF')
-else:
-    fig_mi.savefig(args.outfile)
+caption = 'Sensitivity comparison between {} and {}.'.format(
```
GarethDaviesGW, line 140: Sensitivity --> VT sensitivity

pycbc/results/metadata.py
```diff
@@ -80,18 +80,31 @@ def save_png_with_metadata(fig, filename, fig_kwds, kwds):
...
+    """
+    # https://stackoverflow.com/a/17462125
+    from matplotlib.backends.backend_pdf import PdfPages
+
+    with PdfPages(filename) as pdfp:
+        fig.savefig(pdfp, format='pdf', **fig_kwds)
+        metadata = pdfp.infodict()
+        for key in kwds:
+            metadata[str(key)] = str(kwds[key])
+
```
GarethDaviesGW, line 94: This is a bit more than improving the vt_ratio script, and I've not looked at this stuff before, but from the documentation this looks like it will do the right thing

Do you have an example output?

GarethDaviesGW (commented):
A few comments - I guess this is a work in progress as the help messages haven't changed?

bin/plotting/pycbc_plot_vt_ratio
titodalcanton: Would "consistent sets" of injections be correct? You still need them to be drawn from the same population, right?

pycbc/results/metadata.py
titodalcanton: PDF before/after:
https://ldas-jobs.ligo.caltech.edu/~tito.canton/pycbc/test_pr3529/vt_ratio_igwn.pdf
https://ldas-jobs.ligo.caltech.edu/~tito.canton/pycbc/test_pr3529/vt_ratio.pdf

PNG before/after:
https://ldas-jobs.ligo.caltech.edu/~tito.canton/pycbc/test_pr3529/vt_ratio_igwn.png
https://ldas-jobs.ligo.caltech.edu/~tito.canton/pycbc/test_pr3529/vt_ratio.png

I had no issues with these PDFs. The only caveat is that the metadata keywords I am inserting are not the "official" ones in the PDF standard, so I get an annoying warning like
```
lib/python2.7/site-packages/matplotlib/backends/backend_pdf.py:1579: UserWarning: Unknown infodict keyword: version
  warnings.warn('Unknown infodict keyword: %s' % k)
```
There seems to be a more modern way to add metadata to PDFs which matplotlib does not seem to support yet.

titodalcanton:
> A few comments - I guess this is a work in progress as the help messages haven't changed?

I actually changed the description to avoid repeating the same thing twice.

bin/plotting/pycbc_plot_vt_ratio
GarethDaviesGW: that makes sense

GarethDaviesGW: that makes sense

pycbc/results/metadata.py
GarethDaviesGW: Okay, thanks. I _may_ have had to rely on the pdf metadata being stored in plaintext in the raw bytes (apparently it is!), to check that it worked. These look good

titodalcanton: I added a couple lines that convert `title` to `Title`, the official PDF keyword for title. One warning less, and the title actually shows up in Preview and Firefox!

titodalcanton:
Woops, I just noticed that we have
```
bin/plotting/pycbc_plot_vt_ratio
bin/workflow_comparisons/offline_search/pycbc_plot_vt_ratio_vs_ifar
```
should I have modified the latter instead?

josh-willis:
> Woops, I just noticed that we have
> 
> ```
> bin/plotting/pycbc_plot_vt_ratio
> bin/workflow_comparisons/offline_search/pycbc_plot_vt_ratio_vs_ifar
> ```
> 
> should I have modified the latter instead?

The latter is one that I wrote, and which was used in the review of the GPU implementation.  It plots the VT  ratio as a function of IFAR rather than at certain specified values; this was specifically requested by the reviewers.

I only wrote it to be compatible with the two-IFO workflow; because the file formats are different for multi-IFO, it would have to be modified to work with it (though I am happy for someone to do that).

GarethDaviesGW (approved):

titodalcanton:
> > Woops, I just noticed that we have
> > ```
> > bin/plotting/pycbc_plot_vt_ratio
> > bin/workflow_comparisons/offline_search/pycbc_plot_vt_ratio_vs_ifar
> > ```
> > 
> > 
> > should I have modified the latter instead?
> 
> The latter is one that I wrote, and which was used in the review of the GPU implementation. It plots the VT ratio as a function of IFAR rather than at certain specified values; this was specifically requested by the reviewers.
> 
> I only wrote it to be compatible with the two-IFO workflow; because the file formats are different for multi-IFO, it would have to be modified to work with it (though I am happy for someone to do that).

Thanks. I see the two scripts share some code. I will merge this PR now, but it would be good to merge the two scripts at some point.

PR #3531, opened 2020-11-23
Allow for single float input in format_lmns.

JulianWesterweck:
If only a single lmns identifier is given (e.g. lmns = 220), `pycbc_create_injections` casts this to a float. This was not recognised by format_lmns.
Now a float is cast to an integer, then to a string to correct this.


micamu (approved):

cdcapano (approved):

cdcapano:
@JulianWesterweck I'm not sure why the builds are stuck. Try rebasing to master.

PR #3548, opened 2020-12-05
add extra options to dynesty walk method

ahnitz:


cdcapano (approved):
Thanks... that was easy. Did you intend to change the example config file for the single template from `emcee_pt` to dynesty? If so, go ahead and merge.

PR #3549, opened 2020-12-05
add method option to project wave to use alternate approximations

ahnitz:
This adds two additional methods to the 'project_wave' method of the Detector class. These allow for more approximant ways to project the polarizations into the detector frame. The purpose is for testing different approximations in likelihoods and like-for-like cross comparisons. The two methods are (1) a basic apply constant factor for the polarizations (2) allow fp/fc to vary over time. Neither includes doppler corrections like the default 'lal' method does.

 For reference (1) seems to be the dominant correction for signals of length several hours. I am seeing the mismatch between the result from the 'lal' method and the 'constant' decreased by an order of magnitude when applying only the time varying polarization corrections. 

I am not doing so now, but I may consider adding in another version with some doppler corrections for the earth rotation, if it turns out interesting to include in likelihoods.

### DIFF
--- a/pycbc/detector.py
+++ b/pycbc/detector.py
@@ -290,7 +290,7 @@ class Detector(object):
             Declination of source location
         polarization: float
             Polarization angle of the source
-        method: str, Optional
+        method: {'lal', 'constant', 'vary_polarization'}
             The method to use for projecting the polarizations into the
             detector frame. Default is 'lal'.
         reference_time: float, Optional
@@ -326,9 +326,10 @@ class Detector(object):
             if method == 'constant':
                 time = rtime
             elif method == 'vary_polarization':
-                if not isinstance(hp, TimeSeries):
-                    raise TypeError('Polarizations must be given as time'
-                                    ' series for this method')
+                if (not isinstance(hp, TimeSeries) or
+                    not isinstance(hc, TimeSeries)):
+                    raise TypeError('Waveform polarizations must be given'
+                                    ' as time series for this method')
 
                 # this is more granular than needed, may be optimized later
                 # assume earth rotation in ~30 ms needed for earth ceneter
--- a/pycbc/detector.py
+++ b/pycbc/detector.py
@@ -181,7 +181,7 @@ class Detector(object):
                 fcross = (x * dy + y * dx).sum(axis=0)
             else:
                 fplus = (x * dx - y * dy).sum()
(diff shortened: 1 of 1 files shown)


pycbc/detector.py
```diff
@@ -270,18 +270,81 @@ def time_delay_from_detector(self, other_detector, right_ascension,
...
+            Cross polarization of the GW
+        ra: float
+            Right ascension of source location
+        dec: float
+            Declination of source location
+        polarization: float
+            Polarization angle of the source
+        method: str, Optional
+            The method to use for projecting the polarizations into the
+            detector frame. Default is 'lal'.
```
cdcapano, line 295: Could you list the options here? I think the numpy syntax for parameters that take a set of options is to do:
```
method : {'lal', 'constant', 'vary_polarization'}
```

pycbc/detector.py
```diff
@@ -270,18 +270,81 @@ def time_delay_from_detector(self, other_detector, right_ascension,
...
+                # the given time series. If not provided, we'll choose
+                # the midpoint time.
+                rtime = float(hp.end_time + hp.start_time) / 2.0
+
+            if method == 'constant':
+                time = rtime
+            elif method == 'vary_polarization':
+                if not isinstance(hp, TimeSeries):
+                    raise TypeError('Polarizations must be given as time'
+                                    ' series for this method')
```
cdcapano, line 331: Change this to "Waveforms must be given..." Saying polarizations made me think that you had to provide the polarization angle as a time series. Also, shouldn't you check both hp and hc here?

cdcapano (commented):
Thanks for doing this. Just a couple minor requests related to doc string and error checking.

It would be good if this were also made to work with frequency series. For the lal or vary polarization that would mean transforming to time domain then back to frequency domain. For time shifting, that would mean using the fd time shift. Then the FDomainDetFrame generator could be pointed to this, which would allow us to also use the time-changing response function for the templates. It'd be slow, but at least it would work as a test. Since that'll require some more work though, I'm ok with doing that in a separate PR.

ahnitz:
@cdcapano Please check again

cdcapano (approved):
Looks good to me

cdcapano:
@ahnitz Ah, good point. I forgot about the wraparound issue. Ok, well, something to mark for later.

PR #3550, opened 2020-12-05
add option to set detector projection method in the injection file

ahnitz:
This enables a 'detector_projection_method' option to set the option used in project wave. It can be done on an injection by injection basis as any other configuration coming from an hdf config file.


ahnitz:
Depends on #3549

ahnitz:
@cdcapano Please review again, now that this has been rebased.

cdcapano (approved):
Looks good

PR #3551, opened 2020-12-05
allow relative model to be able to handle time changing antenna patterns

ahnitz:
This updates the relative model to allow it to handle changes in the antenna pattern as a function of time which can greatly affect the PE of very long duration sources, such as those observed by CE and ET. This includes only the polarization factor corrections and not the doppler shift. It is implemented only as a constant correction on the same scale as the frequency binning. Since it is applied in the frequency domain this should not be used with models with higher order modes or wherever there is signfiicant confusion about what time a particular frequency maps to.

### DIFF
--- a/pycbc/inference/models/gaussian_noise.py
+++ b/pycbc/inference/models/gaussian_noise.py
@@ -182,16 +182,16 @@ class BaseGaussianNoise(BaseDataModel):
         # store the psds and whiten the data
         self.psds = psds
 
-    @property 
+    @property
     def high_frequency_cutoff(self):
         """The high frequency cutoff of the inner product."""
         return self._f_upper
-        
-    @property 
+
+    @property
     def low_frequency_cutoff(self):
         """The low frequency cutoff of the inner product."""
         return self._f_lower
-        
+
     @property
     def kmin(self):
         """Dictionary of starting indices for the inner product.
--- a/pycbc/inference/models/gaussian_noise.py
+++ b/pycbc/inference/models/gaussian_noise.py
@@ -123,7 +123,7 @@ class BaseGaussianNoise(BaseDataModel):
         self.ignore_failed_waveforms = ignore_failed_waveforms
         # check if low frequency cutoff has been provided for every IFO with
         # data
-        for ifo in self.data.keys():
+        for ifo in self.data:
             if low_frequency_cutoff[ifo] is None:
                 raise ValueError(
                     "A low-frequency-cutoff must be provided for every "
@@ -151,6 +151,8 @@ class BaseGaussianNoise(BaseDataModel):
         self.low_frequency_cutoff = self._f_lower = low_frequency_cutoff
 
         # set upper frequency cutoff
+        if high_frequency_cutoff is None:
+            high_frequency_cutoff = {ifo: None for ifo in self.data}
         self.high_frequency_cutoff = self._f_upper = high_frequency_cutoff
 
         # Set the cutoff indices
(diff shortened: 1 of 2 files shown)


ahnitz:
depends on #3550 and #3549

ahnitz:
This adds an option to the 'relative' model. It is disabled by default (so current behavior). To enable
```
[model]
vary_polarization = 
```

ahnitz:
@dfinstad This also includes a minor fix. You had inadvertently hard coded the waveform to never be used below 10 Hz.

pycbc/inference/models/gaussian_noise.py
```diff
@@ -160,53 +165,25 @@ def __init__(self, variable_params, data, low_frequency_cutoff, psds=None,
...
-        if high_frequency_cutoff is not None and bool(high_frequency_cutoff):
-            for det in self._data:
-                if det in high_frequency_cutoff:
-                    self._f_upper[det] = high_frequency_cutoff[det]
-                else:
-                    self._f_upper[det] = None
-        else:
-            for det in self._data.keys():
-                self._f_upper[det] = None
-
```
cdcapano, line 209: Thanks for getting rid of the setter, it was needlessly complicated and didn't actually do anything useful. However, please keep `high_frequency_cutoff` as a property that returns `_f_upper` rather than an attribute. I don't want the user thinking they can change the high frequency cutoff by setting that. The reason this isn't settable after the init is the weights would need to be recalculated. By making it an attribute, the user might think they can change it by settting `high_frequency_cutoff`, when it actually doesn't do anything.

I just realized you had also changed `low_frequency_cutoff` from a property to an attribute in an earlier PR. Please switch that back too. So, change `self._f_lower = self.low_frequency_cutoff = low_frequency_cutoff` in the init to `self._f_lower = low_frequency_cutoff`. Then add:
```
@property 
def low_frequency_cutoff(self):
    """The low frequency cutoff of the inner product."""
    return self._f_lower
```
Ditto for `high_frequency_cutoff`.

cdcapano (changes requested):
The projection bit looks fine to me. I have a request for the changes you made to the `BaseGaussianNoise` class though; see below.

pycbc/inference/models/relbin.py
```diff
@@ -225,6 +236,18 @@ def __init__(self, variable_params, data, low_frequency_cutoff,
                      self.df)
         self.sdat = self.summary_data()
 
+        # Calculate the times to evaluate fp/fc
+        if vary_polarization is not False:
```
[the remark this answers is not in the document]
ahnitz: For technical reasons that doesn't catch everything. Parameters from the configparse in the PE code don't always get transformed to the obvious thing for true/ false statements. i.e. something =, doesn't return true, it gets made a string which won't evaluate quite the way you are thinking.

pycbc/inference/models/gaussian_noise.py
ahnitz: @cdcapano Ok, done.

ahnitz:
@cdcapano Please review again

pycbc/inference/models/gaussian_noise.py
cdcapano: Thanks!

pycbc/inference/models/relbin.py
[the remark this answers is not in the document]
cdcapano: Ok

cdcapano (approved):
👍

cdcapano:
@ahnitz Code climate is complaining about some whitespace.

ahnitz:
@dfinstad I'll wait on your sign-off as this changes some of your code.

dfinstad (approved):
The changes look good to me. Thanks for doing this!

PR #3559, opened 2020-12-14
Cleanup PyCBC Live example

titodalcanton:
This is a cleanup PR paving the way for less trivial upcoming changes to the PyCBC Live example/test. Most of the PR is just white space and formatting stuff which I did not want to mix with less trivial changes. However, I also noticed that each injection was being done twice (effectively halving its luminosity distance). I fixed that by actually doing a single injection at the right distance.


spxiwh (approved):
Thanks

PR #3560, opened 2020-12-15
pycbc_process_sngls: subtract vetoed time from live time

titodalcanton:
As discussed previously, this corrects the live time saved by `pycbc_process_sngls` in case a veto file is used. It also fixes a recently introduced variable name error.

Still needs test.

### DIFF
--- a/bin/pycbc_process_sngls
+++ b/bin/pycbc_process_sngls
@@ -2,9 +2,12 @@
 
 """Reads in and vetoes single ifo triggers, cuts, reranks and clusters them"""
 
-import h5py, numpy, argparse, logging
-from pycbc.io import hdf
-from pycbc.events import stat, coinc, ranking
+import argparse
+import logging
+import numpy
+import h5py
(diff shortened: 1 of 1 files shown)


tdent:
Looks OK.  Is there anywhere in the output that records the total livetime?  If not would it make sense to add one e.g. as an attribute?

titodalcanton:
Yes, I will add it as an attribute of the `search` group.

tdent (approved):
Should be OK, I would though check if any other search output file has a field called 'livetime' or similar so that we are not stuck with differently spelt versions ('livetime' vs 'live_time' etc)

titodalcanton:
`pycbc_live_single_trigger_fits` also creates an attribute called "live_time", while the offline coinc code uses "foreground_time" attributes. I have a (very mild) preference for "live_time" here, but I am happy to follow stronger opinions.

titodalcanton:
I tested this by checking the reported live time against the summary pages of a few boxes, with and without vetoes.

PR #3562, opened 2020-12-15
add method to add custom detector

ahnitz:
This adds a method to add a custom detector on the earth defined by its lat/lon and the orientation defined as the azimuthal angle of the y-arm (so 0 is y-arm pointing due north). The patch also makes lalsimulation optional for this module.

### DIFF
--- a/pycbc/detector.py
+++ b/pycbc/detector.py
@@ -107,6 +107,7 @@ def add_detector_on_earth(name, longitude, latitude, orientation=0, height=0):
                                       'response': response,
                                       }
 
+
 class Detector(object):
     """A gravitational wave detector
     """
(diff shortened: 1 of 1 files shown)


ahnitz:
This PR does not address how to access this functionality from existing codes / scripts. I'll think about that and come back with another PR. Suggestions welcome.

spxiwh (approved):
I think the interface looks good to me. I think the real question would be how to access this in an executable, as discussed. I think having the information in a config file would be best, but then how to send it on?

I guess either as a global command-line group sent to almost every executable, or perhaps as an environment variable, that then every code (and things like jupyter notebooks) could easily access.

I think this does need to be hooked up with frDetector, although happy for that to be a separate patch.

ahnitz:
rebasing so the tests will pass

PR #3563, opened 2020-12-15
correction of minor bugs in make_skymap and detector.py

PAduverne:
Corrections of two minor bugs.
Add a line to remove the temporary directories created by pycbc_make_skymap.
Changed a line in detector.py in the function project_wave() to avoid a Domain error when calculating the default reference_time.


titodalcanton:
Just to clarify, the `detector.py` change is necessary to fix an error caused by summing two `LIGOTimeGPS` objects with times in the present.

ahnitz (approved):

PR #3565, opened 2020-12-16
apt update

ahnitz:


ahnitz:
@spxiwh This fixes some issues with installing the software when the repos are modified.

spxiwh (approved):

PR #3569, opened 2020-12-17
Add more info to offline GraceDB uploads

titodalcanton:
This tags each offline GraceDB upload with a PyCBC version string and install path, and the absolute path to the input file which provided the candidate. It also adds a `--log-message` option which can be used to add an extra message to the log (e.g. `--log-message 'Candidate from hyperbank search on HLV C01 data'`).

Example: https://gracedb-playground.ligo.org/events/T368754/view/

### DIFF
--- a/bin/pycbc_upload_xml_to_gracedb
+++ b/bin/pycbc_upload_xml_to_gracedb
@@ -49,6 +49,8 @@ parser.add_argument("--psd-files", nargs='+', required=True,
                     help='HDF file(s) containing the PSDs to upload')
 parser.add_argument('--input-file', required=True, type=str,
                     help='Input LIGOLW XML file of coincidences.')
+parser.add_argument('--log-message', type=str, metavar='MESSAGE',
+                    help='Add a log entry to each upload with the given message')
 parser.add_argument('--testing', action="store_true", default=False,
                     help="Upload event to the TEST group of gracedb.")
 parser.add_argument('--min-ifar', type=float, metavar='YEARS',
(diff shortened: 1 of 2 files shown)


bin/pycbc_upload_xml_to_gracedb
```diff
@@ -160,10 +163,26 @@ for event in coinc_table:
...
+    # add info for tracking code version
+    gracedb_tag_with_version(gracedb, r['graceid'])
+
+    # document the absolute path to the input file
+    input_file_str = 'Candidate uploaded from ' \
+        + os.path.abspath(args.input_file)
+    gracedb.writeLog(r['graceid'], input_file_str)
+
+    # add the custom log message, if provided
+    if args.log_message is not None:
```
GarethDaviesGW, line 182: is the 'is not None' needed? i.e. is default None?

GarethDaviesGW (approved):
Approved: Everything looks good to me, one small question on default values but its more info than requested changes

bin/pycbc_upload_xml_to_gracedb
titodalcanton: In this case `if args.log_message is not None:` is in practice equivalent to `if args.log_message:`. The default is None if the option is not given. I have a slight preference for the former version, because it asks more explicitly if the option is given or not, rather than asking if `bool(args.log_message)` is True, which depends not only on the option being given or not, but also on what string is being given. For the present case the only difference is noticeable if you pass the silly argument `--log-message ''`: it will create an empty comment with the first version, and not create anything with the second.

GarethDaviesGW: okay that sounds fair

PR #3576, opened 2021-01-07
add effective distance as method of detector class

ahnitz:


ahnitz:
@spxiwh I actually need this merged before the hdf stuff.

spxiwh (approved):
Sorry, I missed this one, but was happy with this change in the child PR.

PR #3577, opened 2021-01-07
fix inclination conventions

ahnitz:
As someone noted, I mixed up some of the inclination conventions. This was a mistake in single_template and the pycbc_dtphase code. In both, it doesn't really affect anything (other than causing the polarization angle to now be inconsistently define). I've tested both codes give the same results before / after. However, we should bring this in line to avoid confusion and further mistakes.

To be clear this wasn't an issue because the polarization angle in both cases was treated as a nuisance angle so you can transform from 'broken' to the 'fixed' convention by a rotation. If someone wanted to look at measuring the polarization angle, it would be inconsistent with the literature though.


spxiwh (approved):
Thanks

PR #3581, opened 2021-01-11
always run with hdf injection format internally

ahnitz:
This converts xml to hdf and enables inputting hdf format in the workflow.


ahnitz:
will depend on #3578

PR #3582, opened 2021-01-12
Included an option to choose the scipy optimizer

ArthurTolley:
Also includes the relevant arguments that are needed to customise the optimizers. Changed the default differential_evolution parameters from (100,200) to (50,100).

Within the compute_network_snr function, the addition was needed for the shgo optimizer.
The default optimizer is still differential_evolution, used the try & except but unsure if this is necessary.

### DIFF
--- a/bin/pycbc_optimize_snr
+++ b/bin/pycbc_optimize_snr
@@ -142,10 +142,10 @@ parser.add_argument('--cores', type=int,
 parser.add_argument('--optimizer', type=str, default='differential_evolution',
                     choices=['differential_evolution', 'shgo'],
                     help='The optimizer to use, differential_evolution or shgo')
-parser.add_argument('--DI-maxiter', type=int, default=50,
+parser.add_argument('--di-maxiter', type=int, default=50,
                     help='Only relevant for --optimizer differential_evolution: '
                     'The maximum number of generations over which the entire population is evolved.')
-parser.add_argument('--DI-popsize', type=int, default=100,
+parser.add_argument('--di-popsize', type=int, default=100,
                     help='Only relevant for --optimizer differential_evolution: '
                          'A multiplier for setting the total population size.')
 parser.add_argument('--shgo-samples', type=int, default=76,
@@ -206,8 +206,8 @@ logging.info('Starting optimization')
 
 if args.optimizer == 'differential_evolution':
     results = differential_evolution(compute_network_snr, bounds,
-                                 maxiter=args.DI_maxiter, workers=(args.cores or -1),
-                                 popsize=args.DI_popsize, mutation=(0.5,1),
+                                 maxiter=args.di_maxiter, workers=(args.cores or -1),
+                                 popsize=args.di_popsize, mutation=(0.5,1),
                                  recombination=0.7, callback=callback_func,
                                  args=extra_args)
 
--- a/bin/pycbc_optimize_snr
+++ b/bin/pycbc_optimize_snr
@@ -140,19 +140,20 @@ parser.add_argument('--output-path', required=True,
 parser.add_argument('--cores', type=int,
                     help='Restrict calculation to given number of CPU cores')
 parser.add_argument('--optimizer', type=str, default='differential_evolution',
-                    help='The optimizer to use, shgo or differential_evolution')
-parser.add_argument('--maxiter', type=int, default='50',
-                    help='Differential Evolution: The maximum number of '
-                         'generations over which the entire population is evolved.')
-parser.add_argument('--popsize', type=int, default='100',
-                    help='Differential Evolution: A multiplier for setting the '
-                         'total population size.')
-parser.add_argument('--iters', type=int, default='3',
-                    help='shgo: Number of sampling points used in the construction '
-                         'of the simplicial complex. ')
-parser.add_argument('--n', type=int, default='76',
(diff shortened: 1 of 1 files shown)


bin/pycbc_optimize_snr
```diff
@@ -186,12 +203,22 @@ bounds = [(minchirp, maxchirp), (0.01, 0.2499), (-0.9, 0.9), (-0.9, 0.9)]
...
                                  recombination=0.7, callback=callback_func,
                                  args=extra_args)
 
+    elif args.optimizer == 'shgo':
+        results = shgo(compute_network_snr, bounds=bounds, args=extra_args, 
+                iters=args.iters, n=args.n, sampling_method="sobol")
+               
+except RuntimeError: 
+    print("You haven't specified an optimizer, either shgo or differential_evolution")
+
```
titodalcanton, line 221: I am not sure a try/except statement is the way to go here, or that it will even do what you want. I recommend using the `choices` argument of `parser.add_argument()` instead, which will ensure that `args.optimizer` will have a valid value.

bin/pycbc_optimize_snr
```diff
@@ -137,6 +139,21 @@ parser.add_argument('--output-path', required=True,
...
+parser.add_argument('--maxiter', type=int, default='50',
+                    help='Differential Evolution: The maximum number of '
+                         'generations over which the entire population is evolved.')
+parser.add_argument('--popsize', type=int, default='100',
+                    help='Differential Evolution: A multiplier for setting the '
+                         'total population size.')
+parser.add_argument('--iters', type=int, default='3',
+                    help='shgo: Number of sampling points used in the construction '
+                         'of the simplicial complex. ')
+parser.add_argument('--n', type=int, default='76',
```
titodalcanton, line 153: I suggest looking for more descriptive names for these options, for example:
* Prefix the DI-relevant options with "di", and the SHGO-relevant options with "shgo". 
* Is there a better (but still short) name than "n"? Does it make sense to recycle the --maxiter option for this parameter?
* Please clarify in the help strings that some options are only relevant to DI and SHGO (e.g. "Only relevant with --optimizer differential_evolution").

titodalcanton: * When the argument is an int, you do not need quotes for the `default` kwarg, since it is an int.

titodalcanton (changes requested):
Thanks @ArthurTolley! I left a couple specific requests for changes above. In addition, note that the new options will not be usable unless you also modify `pycbc_live` to actually pass those options to `pycbc_optimize_snr`. In other words, `pycbc_live` will probably also need extra options. You can make that change in this PR or in a future one, however, I have no preference.

bin/pycbc_optimize_snr
ArthurTolley: With regards to the second point, I had mixed up the two parameters 🤦‍♂️. I can't think of a shorter term for n, with it's definition being 'Number of sampling points used in the construction of the simplicial complex.'. I think maybe I will go with 'shgo_samples'.

titodalcanton: Then `--shgo-num-points` seems a reasonable choice to me, as does `--shgo-samples` (but please stick to dashes).

It seems I was the one mixing up stuff! I meant to propose reusing `--maxiter` in place of `--iters`, not `--n`.

ArthurTolley:
@titodalcanton I've included the changes you recommended, with regards to the `pycbc_live` file I will wait to make those changes in a different PR but I will definitely do them.

titodalcanton:
@ArthurTolley two more comments:
* I do not see an import for the `shgo()` function. Have you had a chance to actually test this PR, both with DI and SHGO?
* I just noticed that the `shgo()` documentation also points out a `maxiter` argument. Does it make sense to pass `args.maxiter` to it as well (and thus remove the "di" prefix from the option)?

ArthurTolley:
Hi @titodalcanton, your comments are much appreciated.

To be fully honest, I might have had the wrong assumption that the checks in the PR will do this. I will test it now and add the changes. I was slightly afraid of accidentally uploading a trigger (which I don't think I would ever have the permission to do) but I will endeavor to make sure I have the write arguments to prevent this.

The `maxiter` argument is optional for `shgo()` which I haven't played around with so i'm unsure what effect it might have to pass this argument. Again, this is something I can test in the future.

titodalcanton:
Hah, the tests *should* ideally take care of it, but they are not magic: humans still have to create them! For the specific case of `pycbc_optimize_snr`, the only test in place is actually that it does not fail when given the `--help` argument. Which is close to not having a test at all.

If we did have a less trivial test, *you* (the PR author) would be the one teaching it how to call `pycbc_optimize_snr` with the new options you added. It would be part of this PR.

As there is no such test, unfortunately you have to actually run it and see for the time being.

ArthurTolley:
Okay, the run makes it through the optimization but there are errors related to `KeyError: "Unable to open object (object 'mc_area_args' doesn't exist)"`.

titodalcanton:
That error seems due to running recent code on an old event, predating the introduction of the source classification. As long as the same error happens with DI and SHGO, and after the optimization, it should be harmless.

ArthurTolley:
Yes, that is the case. It happens twice after both of the optimizations complete.

titodalcanton (approved):

PR #3583, opened 2021-01-14
PyCBC Live: use Astropy to get a more precise current GPS time

titodalcanton:
`lal.GPSTimeNow()` does not return the fractional part, which is a bit inconvenient when monitoring the O(1 s) lag of the early-warning configuration of PyCBC Live. So here I am switching to Astropy to get the current GPS. This is probably not the most precise method (there are ongoing [issues](https://github.com/astropy/astropy/issues/6190) with GPS times in Astropy) and @duncanmmacleod suggested using GWPy as a more precise alternative. Opinions are welcome, given that GWPy would be an extra dependency and I do not think the Astropy precision issue is critical here.

This PR also simplifies the logging a bit by pre-inserting the MPI rank in the format string, so we do not have to carry it along every time.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -11,7 +11,6 @@ See https://arxiv.org/abs/1805.11174 for an overview."""
 
 import sys
 import argparse, numpy, pycbc, logging, cProfile, h5py, lal, json
-from time import time
 import os.path
 import itertools
 import platform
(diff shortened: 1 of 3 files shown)


ahnitz (approved):
Using astropy is the correct approach. I don't think that depending on gwpy makes sense, given that much of its functionality already exists in PyCBC. It would be better to integrate more closely with lower level libraries such as astropy.

duncanmmacleod:
AFAIR, the issue with precision in astropy is related to combining multiple `Time` objects, so if you're just using astropy to get the current GPS time and then storing it internally as a LIGOTimeGPS or a float, you shouldn't see any issues.

PR #3589, opened 2021-01-22
Move interpolation into new parent class of WorkflowConfigParser

ahnitz:
This moves the interpolating features added to ConfigParser to a separate class which can be called outside of a workflow context (and without partially pulling in a Pegasus dependency).  This also drops the glue.pipeline dependence within the core pycbc library (some old exes still remain).


spxiwh (approved):
I'm happy with moving this as described .... Obviously the tests must pass :-).

PR #3590, opened 2021-01-22
MCMC history in workflow

cdcapano:
Adds MCMC history plots as an optional diagnostic in the inference workflow. Adds caption to the plots for displaying on results page. Also adds markers to the plots to make them more legible when they are only 1 or fewer points. Example can be found in section 6, [here](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/projects/pycbc_inference/gw150914_example/gw150914_test-emcee/).


ahnitz (approved):
Looks trivial enough, if tests pass and no cc issues, go ahead and merge

PR #3592, opened 2021-01-23
Add SEOBNRv4HM_ROM to dictionary of filter time lengths

sudughonge:
Closes https://github.com/gwastro/pycbc/issues/3591

### DIFF
--- a/pycbc/waveform/waveform.py
+++ b/pycbc/waveform/waveform.py
@@ -846,7 +846,7 @@ def imrphenomhm_length_in_time(**kwargs):
 def seobnrv4hm_length_in_time(**kwargs):
     """ Estimates the duration of SEOBNRv4HM waveforms that include higher modes.
     """
-    # Default maximum node number for SEOBNRv4HM is 5 
+    # Default maximum node number for SEOBNRv4HM is 5
     # The relevant lower order approximant here is SEOBNRv4
     return get_hm_length_in_time('SEOBNRv4', 5, **kwargs)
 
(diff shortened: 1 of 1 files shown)


ahnitz:
@sudughonge Thank you for the PR. Is this really the right length estimator though? The purpose here is that it should not be too short so that the waveform is not truncated or wrapped when you translate it from fd to td domain. The length estimate you've used should be OK for the dominant 2-2 mode, but what about for the higher modes of this waveform? If they start from the same frequency, they will not have the same duration as the 2-2 mode. Have you considered this?

sudughonge:
Hi @ahnitz that's a fair point and I have not considered this, and this isn't the correct way to do it. Do you think a strategy similar to the one used to compute the filter length for imrphenomhm waveforms [here](https://github.com/sudughonge/pycbc/blob/3591/pycbc/waveform/waveform.py#L839) would be a good start?

I see why this is sort of complicated for SEOBNRv4HM_ROM: the SEOBNRv4HM waveform is in the time domain so doesn't need a filter length computation and SEOBNRv4_ROM is a frequency domain waveform that has the same approx length as SEOBNRv4.

sudughonge:
@ahnitz just added a function to calculate the SEOBNRv4HM length. Since the max mode number is 5 (https://arxiv.org/abs/1512.02248), I set `maxm = 5` here.

pycbc/waveform/waveform.py
```diff
@@ -852,6 +852,22 @@ def imrphenomhm_length_in_time(**kwargs):
     kwargs['f_lower'] = flow * 2./maxm
     return get_imr_length("IMRPhenomD", **kwargs)
 
+def seobnrv4hm_length_in_time(**kwargs):
```
ahnitz, line 855: Can this be combined with the imr phenom one to avoid code duplication?

sudughonge: Sure, let me see what I can do.

sudughonge: @ahnitz just added a couple of changes. I've defined a function `get_hm_length_in_time` that takes input the (2,2) approximant name for the relevant HM approximant (`lor_approximant`) and its default `max ` and the makes the call to `lalsimulation.get_imr_length`.

ahnitz:
@sudughonge This looks good to me now. There are some whitespace issues you can see in the code climeate report. If you can address those and assuming none of the unit tests fail, then this can be merged.

ahnitz (approved):

sudughonge:
@ahnitz just fixed the linting errors.

PR #3596, opened 2021-01-26
Reimplement PhaseTDStatistic

### LINKED ISSUE #3580: PyCBC Live test: generate PTA files and use a PTA-based ranking
The current example/test of PyCBC Live uses a simplistic `--ranking-statistic quadsum` which is not what was used in production during O3. Modify the example to create minimal PTA histogram files and use a ranking that includes the PTA term, such as `phasetd`.

Requires fixing #3584.

### LINKED ISSUE #3584: PyCBC Live: re-enable the O3 coinc ranking

titodalcanton:
Reimplement `PhaseTDStatistic` based on the code that was in place before #3535, and modify the PyCBC Live test to use it. Some extra changes are needed to fix some errors related to a `from_cli()` method modifying some of the arguments and breaking further calls to the same function.

Closes #3580. Closes #3584.

I have not tested that the resulting values are actually correct or equivalent to the old code, though, so marking as WIP.

### DIFF
--- a/pycbc/events/coinc.py
+++ b/pycbc/events/coinc.py
@@ -850,21 +850,20 @@ class LiveCoincTimeslideBackgroundEstimator(object):
     @classmethod
     def from_cli(cls, args, num_templates, analysis_chunk, ifos):
         from . import stat
+
         # Allow None inputs
-        if args.statistic_files is None:
-            args.statistic_files = []
-        if args.statistic_keywords is None:
-            args.statistic_keywords = []
+        stat_files = args.statistic_files or []
+        stat_keywords = args.statistic_keywords or []
 
         # flatten the list of lists of filenames to a single list (may be empty)
-        args.statistic_files = sum(args.statistic_files, [])
+        stat_files = sum(stat_files, [])
 
-        kwargs = stat.parse_statistic_keywords_opt(args.statistic_keywords)
+        kwargs = stat.parse_statistic_keywords_opt(stat_keywords)
 
         return cls(num_templates, analysis_chunk,
                    args.ranking_statistic,
                    args.sngl_ranking,
-                   args.statistic_files,
+                   stat_files,
                    return_background=args.store_background,
                    ifar_limit=args.background_ifar_limit,
                    timeslide_interval=args.timeslide_interval,
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -354,12 +354,11 @@ class PhaseTDStatistic(QuadratureSumStatistic):
                 if num != len(ifos):
                     continue
 
-                match = [ifo in name for ifo in ifos]
+                match = [ifo in ifokey for ifo in ifos]
                 if False in match:
                     continue
-                else:
-                    selected = name
-                    break
+                selected = name
+                break
 
         if selected is None:
             raise RuntimeError("Couldn't figure out which stat file to use")
@@ -618,27 +617,25 @@ class PhaseTDStatistic(QuadratureSumStatistic):
         return self.single(single_info[1])
 
     def rank_stat_coinc(self, sngls_list, slide, step, to_shift,
-                        **kwargs): # pylint:disable=unused-argument
+                        **kwargs):  # pylint:disable=unused-argument
         """
         Calculate the coincident detection statistic.
         """
         rstat = sum(s[1]['snglstat'] ** 2 for s in sngls_list)
-        cstat = rstat + 2. * self.logsignalrate(
-            {d: s for d, s in sngls_list},
-            slide * step,
-            to_shift
-        )
+        cstat = rstat + 2. * self.logsignalrate(dict(sngls_list),
+                                                slide * step,
+                                                to_shift)
         cstat[cstat < 0] = 0
         return cstat ** 0.5
 
     def coinc_lim_for_thresh(self, sngls_list, thresh, limifo,
-                             **kwargs): # pylint:disable=unused-argument
+                             **kwargs):  # pylint:disable=unused-argument
         """
         Optimization function to identify coincs too quiet to be of interest
         Calculate the required single detector statistic to exceed
         the threshold for each of the input triggers.
         """
-        if self.hist is None:
+        if not self.has_hist:
             self.get_hist()
         lim_stat = [b['snglstat'] for a, b in sngls_list if a == limifo][0]
         s1 = thresh ** 2. - lim_stat ** 2.
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -619,7 +619,8 @@ class PhaseTDStatistic(QuadratureSumStatistic):
     def rank_stat_coinc(self, sngls_list, slide, step, to_shift,
                         **kwargs):  # pylint:disable=unused-argument
         """
-        Calculate the coincident detection statistic.
+        Calculate the coincident detection statistic, defined in Eq 2 of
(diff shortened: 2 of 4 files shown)


spxiwh (approved):
This looks good to me. As long as you/others are happy that the new code in stat.py reproduces what was done with the old 2-det coinc code then please merge.

pycbc/events/coinc.py
```diff
@@ -850,21 +850,20 @@ def pick_best_coinc(cls, coinc_results):
     @classmethod
     def from_cli(cls, args, num_templates, analysis_chunk, ifos):
         from . import stat
+
         # Allow None inputs
-        if args.statistic_files is None:
-            args.statistic_files = []
-        if args.statistic_keywords is None:
-            args.statistic_keywords = []
+        stat_files = args.statistic_files or []
```
tdent, line 855: I haven't quite seen what is the point of the change here. If no input is given `stat_files` will be None, is this OK?

pycbc/events/coinc.py
```diff
@@ -850,21 +850,20 @@ def pick_best_coinc(cls, coinc_results):
...
+        stat_keywords = args.statistic_keywords or []
```
tdent, line 856: it seems the parse_statistic_keywords_opt function requires a list, so does it crash if given None?

pycbc/events/stat.py
```diff
@@ -616,26 +614,35 @@ def rank_stat_single(self, single_info):
...
-    def rank_stat_coinc(self, s, slide, step, to_shift,
-                        **kwargs): # pylint:disable=unused-argument
+    def rank_stat_coinc(self, sngls_list, slide, step, to_shift,
+                        **kwargs):  # pylint:disable=unused-argument
         """
         Calculate the coincident detection statistic.
         """
-        err_msg = "Sorry! No-one has implemented this method yet! "
-        raise NotImplementedError(err_msg)
+        rstat = sum(s[1]['snglstat'] ** 2 for s in sngls_list)
```
tdent, line 624: I think a bit more documentation here would be good, the reference will be this 
https://iopscience.iop.org/article/10.3847/1538-4357/aa8f50#apjaa8f50eqn2 
ie eq.2 of the O2-statistic paper.

pycbc/events/coinc.py
titodalcanton: Two changes happened here:
1. I stored the result of the operation into a local variable (`stat_files`) instead of altering `args`. This was necessary because altering `args` was messing with downstream code which also read command-line arguments.
2. I switched to a more compact implementation, using `or` instead of an `if` statement. This is not completely equivalent in general, but it is for the practical purpose of this piece of code (i.e. making sure that `stat_files` is an empty list instead of `None`).

To answer your question, if no input is given then `stat_files` will be an empty list.

pycbc/events/coinc.py
titodalcanton: That is exactly the point of this code, making sure that `parse_statistic_keyword_opt()` receives an empty list instead of `None`.

pycbc/events/stat.py
titodalcanton: Good point, will address.

pycbc/events/coinc.py
tdent: See previous comments. There doesn't seem to be a case where args.statistic_keywords could evaluate to False although a valid option had been given, so OK.  (Default values of these two options seems to be [] anyway.)

titodalcanton:
Hmm, interesting failure in PyCBC Live
```
>> [Thu Jan 28 10:00:54 UTC 2021] Checking results
2021-01-28 10:00:55,585 No triggers found in V1
2021-01-28 10:00:55,585 Single Trigger Test Failed
2021-01-28 10:00:55,585 2 coincident trigger(s) detected
2021-01-28 10:00:55,586 Checking trigger output/coinc-1272790100.1669922-zimpiz.xml.gz
2021-01-28 10:00:55,668 IFO SNRs: [9.41339   7.7795067]
2021-01-28 10:00:55,669 Network SNR: 12.211988
2021-01-28 10:00:55,669 IFO End Time: 1272790100.000000
2021-01-28 10:00:55,669 Mass 1: 290.929321
2021-01-28 10:00:55,669 Mass 2: 3.675545
2021-01-28 10:00:55,669 Spin1z: 0.993485
2021-01-28 10:00:55,669 Spin2z: 0.927135
2021-01-28 10:00:55,669 Checking trigger output/coinc-1272790260.0966797-nrxeea.xml.gz
2021-01-28 10:00:55,749 IFO SNRs: [8.146549 6.880121]
2021-01-28 10:00:55,750 Network SNR: 10.663130
2021-01-28 10:00:55,750 IFO End Time: 1272790260.000000
2021-01-28 10:00:55,750 Mass 1: 1.133169
2021-01-28 10:00:55,750 Mass 2: 1.010624
2021-01-28 10:00:55,750 Spin1z: 0.029544
2021-01-28 10:00:55,750 Spin2z: 0.020994
2021-01-28 10:00:55,750 Test Failed
    FAILED!
```
???

titodalcanton:
For reference, the failing test was due to #3603.

titodalcanton:
Here is a very simple test using the current example. I compared the rankings of the recovered injections between this code version and 503873c95654b67171377a3737222aaa1a46be54 (just before the stat refactor).

```
2020_05_06/H1L1V1-Live-1272790096.069336-8.hdf
  Difference in coa_phase_H1: -1.1920928955078125e-07
  Difference in coa_phase_V1: -2.384185791015625e-07
  Difference in coinc_rank: 0.40607452392578125
  Difference in coinc_rank_squared: 10.332295827902271
  Difference in end_time_H1: 0.0
  Difference in end_time_V1: 0.0
  Difference in ifos: set()
  Difference in snr_H1: 4.76837158203125e-06
  Difference in snr_V1: -7.152557373046875e-06
2020_05_06/H1L1V1-Live-1272790256.069336-8.hdf
  Difference in coa_phase_H1: -1.1920928955078125e-07
  Difference in coa_phase_V1: -3.5762786865234375e-07
  Difference in coinc_rank: 0.42845726013183594
  Difference in coinc_rank_squared: 10.54476247247294
  Difference in end_time_H1: 0.0
  Difference in end_time_V1: 0.0
  Difference in ifos: set()
  Difference in snr_H1: -3.814697265625e-06
  Difference in snr_V1: 9.5367431640625e-07
```

The difference in coinc ranks seems a bit high (both injections have coinc rank ~12). Part of this difference may be due to the necessary change in the format of the phase-time-amplitude files, however.

titodalcanton:
I have now poked at the old and new phase-time-amplitude files for some time (see plots [here](https://ldas-jobs.ligo.caltech.edu/~tito.canton/pycbc/live/test_pr3596/pta_files/)) and I convinced myself that the ~100x difference in expected signal rate comes from the phase-time-amplitude files themselves and not from this code change. I might look into this further, but I would like to merge this so we can have a working O3-like configuration of PyCBC Live.

tdent:
@titodalcanton have you had any feedback from @ahnitz on the change in PTA files and resulting change in statistic?

titodalcanton:
@ahnitz and I discussed this and agreed that the rate is expected to have differences due to the PTA file format change.

PR #3597, opened 2021-01-26
add detectors with config file

ahnitz:
This adds the capability to expand the pycbc detector set with a config file. The config file can either be read using the 'load_detector_config' function manually (i.e. in a custom script), or it can be autoloaded using the environment variable "PYCBC_DETECTOR_CONFIG". 

I've written it in such a way that we can add other methods in the future for describing detectors. The thinking is that we may want different formats for things like space-based detectors. 

Example config file
```
[detector-r1]
method = earth_normal
longitude = -2.0840567691659397
latitude = 0.810795263791696
yangle = 4.084080696105957
```

### DIFF
--- a/pycbc/detector.py
+++ b/pycbc/detector.py
@@ -28,9 +28,11 @@
 """This module provides utilities for calculating detector responses and timing
 between observatories.
 """
+import os
 import numpy as np
 import lal
 from pycbc.types import TimeSeries
+from pycbc.types.config import InterpolatingConfigParser
 from astropy.time import Time
 from astropy import constants, coordinates, units
 from astropy.coordinates.matrix_utilities import rotation_matrix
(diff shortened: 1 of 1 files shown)


titodalcanton:
Looks mostly alright to me, just two comments:
* "Detector on Earth" could in principle mean different kinds of detectors, which would be defined by different parameters (interferometers, resonant bars, resonant spheres). But I think the name you chose to go in the config file, "Earth normal", effectively means "interferometer lying on the ground", and I think 99% of the use cases will refer to interferometers anyway, so this is certainly not an issue for now.
* Can the angle between the arms be specified as well? It was not immediately clear from the function.

ahnitz:
@titodalcanton Yes, this is only for detectors lying on the ground (i.e. tangent to it). They must be two arm interferometers, however, the arm angles can be anything. The 'xangle' and 'yangle' sets the angle (from clockwise north) of the arms.

ahnitz:
@titodalcanton So you can simulate the 3 independent ET combinations for example by making arms with 60 degree opening angle and rotating them with respect to each other.

titodalcanton (approved):

PR #3603, opened 2021-01-28
fix plive test

ahnitz:
* set the amount of data to create to a multiple of 32 so that every file is the same size
* fix a rounding issue in time slice
    * found edge case where the index was within floating point precision of an exact index value. This occured at exactly an integer start time which causes condition strain (and write_frame) to fail due to needing to output second boundary frame files. This case is also not what was intended by the rounding mode option to time slice, whereby you might ask for a time *in between* sample rates and that's not clear we want to round to the nearest (hence the option). However, if we give the exact right time for a given index value, we do want it to use that index. 
* fix issue in rounding of factor between internal raw data sample rate and the strain data 
    * found edge case where this would round to an incorrect factor.

### DIFF
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -184,6 +184,12 @@ class TimeSeries(Array):
         start_idx = float(start - self.start_time) * self.sample_rate
         end_idx = float(end - self.start_time) * self.sample_rate
 
+        if _numpy.isclose(start_idx, round(start_idx)):
+            start_idx = round(start_idx)
+
+        if _numpy.isclose(end_idx, round(end_idx)):
+            end_idx = round(end_idx)
+
         if mode == 'floor':
             start_idx = int(start_idx)
             end_idx = int(end_idx)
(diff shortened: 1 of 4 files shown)


titodalcanton (changes requested):
Please also update the GPS times in `check_results.py`. Yup, that duplication is not good!

pycbc/types/timeseries.py
```diff
@@ -184,6 +184,12 @@ def time_slice(self, start, end, mode='floor'):
         start_idx = float(start - self.start_time) * self.sample_rate
         end_idx = float(end - self.start_time) * self.sample_rate
 
+        if _numpy.isclose(start_idx, round(start_idx)):
+            start_idx = round(start_idx)
+
+        if _numpy.isclose(end_idx, round(end_idx)):
+            end_idx = round(end_idx)
+
```
titodalcanton, line 192: Could you update the PR description with a sentence about this change? Does it fix something triggered by your tests?

ahnitz: @titodalcanton See updated description.

titodalcanton (approved):

PR #3604, opened 2021-01-28
Adding the functionality of passing the approximants into BAYESTAR

ArthurTolley:
Passing the approximant argument into `bayestar-localize-coincs`.

BAYESTAR uses the `TaylorF2` waveform instead of `SPAtmplt` so I had to add an if statement there.

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -332,7 +332,6 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
     approximant = waveform.bank.parse_approximant_arg(opt.approximant, row)[0]
 
     # BAYESTAR uses TaylorF2 instead of SPAtmplt
-
     if approximant == 'SPAtmplt':
         approximant = 'TaylorF2'
 
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -9,11 +9,11 @@ import h5py, pycbc, subprocess, argparse, tempfile, shutil
 import numpy as np
 import logging
 from pycbc.filter import sigmasq
-from pycbc.io import live
+from pycbc.io import live, WaveformArray
 from pycbc.types import TimeSeries, FrequencySeries, MultiDetOptionAppendAction
 from pycbc.pnutils import nearest_larger_binary_number
 from pycbc.waveform.spa_tmplt import spa_length_in_time
-from pycbc import frame
+from pycbc import frame, waveform
 from pycbc.psd import interpolate
 from ligo.gracedb.rest import GraceDb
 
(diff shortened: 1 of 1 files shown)


ArthurTolley:
I think I messed up assigning by assigning Assignees instead of Reviewers haha

titodalcanton:
@lpsinger could you remind us what BAYESTAR's `--waveform` option does when using the SNR time series provided by the search? How important is it that the waveform matches what the search used?

ArthurTolley:
If I can add my reasoning behind it, if you don't have the `SEOBNRv4_ROM` approximant (a 400MB install) you can use `IMRPhenomD` instead if you add this `--waveform` argument to `bayestar-localize-coincs`.

bin/pycbc_make_skymap
```diff
@@ -321,11 +321,27 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
...
+	    spin1z=opt.spin1z,
+	    spin2z=opt.spin2z)
+
+    approximant = waveform.bank.parse_approximant_arg(opt.approximant, row)[0]
+
+    # BAYESTAR uses TaylorF2 instead of SPAtmplt
+
+    if approximant == 'SPAtmplt':
+	approximant = 'TaylorF2'
+
```
titodalcanton, line 338: Just a minor request for indentation: line 337 should be indented with 4 spaces relative to the previous one.

bin/pycbc_make_skymap
```diff
@@ -321,11 +321,27 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
...
     # run BAYESTAR to generate the skymap
 
     cmd = ['bayestar-localize-coincs',
            ligolw_file_path,
            ligolw_file_path,
+	   '--waveform', str(approximant),
```
titodalcanton, line 344: Indentation: line 344 should line up with the rest of the block.

bin/pycbc_make_skymap
titodalcanton: Personal taste, I would also remove the empty line 335, as the comment refers specifically to a very short block of code.

ArthurTolley:
I think I've resolved those issues, I'm not sure if it was me or my text editor showing me slightly different things to what was pushed 🤷‍♂️

titodalcanton (approved):
This removes the need for `SEOBNRv4_ROM`, as pointed out, and is also more correct in general, so I am ok with it.

lpsinger:
> @lpsinger could you remind us what BAYESTAR's `--waveform` option does when using the SNR time series provided by the search? How important is it that the waveform matches what the search used?

It used to be pretty important, but now that we are using SNR time series, the only thing that it impacts is the calculated horizon distance, which impacts the distance estimation only.

I have been repeatedly begging search pipelines to populate the relevant fields in the sngl_inspiral table so that BAYESTAR doesn't need to know the waveform or even the PSD.

PR #3606, opened 2021-02-01
pycbc_dtphase: fix inclination convention, UI improvements and cleanup

titodalcanton:
Applies #3577 to `pycbc_dtphase`, makes `pycbc_dtphase` a bit more self-explanatory, and removes a couple unused imports.

### DIFF
--- a/bin/all_sky_search/pycbc_dtphase
+++ b/bin/all_sky_search/pycbc_dtphase
@@ -166,8 +166,8 @@ for ifo0 in args.ifos:
         dec = np.arccos(uniform(-1., 1., size=size)) - np.pi/2
         inc = np.arccos(uniform(-1., 1., size=size))
         pol = uniform(0, 2 * np.pi, size=size)
-        ip = np.cos(inc)
-        ic = 0.5 * (1.0 + ip * ip)
+        ic = np.cos(inc)
+        ip = 0.5 * (1.0 + ic * ic)
 
         # calculate the toa, poa, and amplitude of each sample
         data = {}
--- a/bin/all_sky_search/pycbc_dtphase
+++ b/bin/all_sky_search/pycbc_dtphase
@@ -1,28 +1,31 @@
 #!/bin/env python
-""" Create a file containing the time, phase and amplitude
-correlations between two or more detectors for signals by
-doing a simple monte-carlo
+
+"""
+Create a file containing the time, phase and amplitude correlations between two
+or more detectors for signals by doing a simple monte-carlo.
 
 Output is the relative amplitude, time, and phase as compared to a reference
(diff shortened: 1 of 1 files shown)


bin/all_sky_search/pycbc_dtphase
```diff
@@ -60,7 +63,9 @@ parser.add_argument('--param-bin-dtype', default='int32',
                     choices=['int8', 'int16', 'int32'])
 args = parser.parse_args()
 
-assert len(args.relative_sensitivities) == len(args.ifos)
+if len(args.relative_sensitivities) != len(args.ifos):
+    parser.error('--relative-sensitivities requires one numerical argument '
+                 'for each detector')
```
GarethDaviesGW, line 68: More a question than a comment - would making --relative-sensitivities a MultiDetOption be useful here?

bin/all_sky_search/pycbc_dtphase
```diff
@@ -1,36 +1,39 @@
...
-from numpy.random import normal, uniform, power
+
+import argparse, h5py, numpy as np, pycbc.detector, logging
+from numpy.random import uniform
 from scipy.stats import norm
 from copy import deepcopy
 
-parser = argparse.ArgumentParser()
+
+parser = argparse.ArgumentParser(description=__doc__)
```
GarethDaviesGW, line 22: Well that's a useful thing I never knew about!

GarethDaviesGW:
I'll approve once the fp/fc changes from #3577 are made here

bin/all_sky_search/pycbc_dtphase
titodalcanton: I do love that!

bin/all_sky_search/pycbc_dtphase
titodalcanton: Is the benefit being able to do `--relative-sensitivities 1` to assume equally sensitive detectors?

GarethDaviesGW: It would still have to be `--relative-sensitivities 1 1 1` (in a 3 ifo network), but I think that the advantage of being clearer in the input arguments, and not relying on the orders being the same would be an advantage

titodalcanton: I see `args.relative_sensitivities` is stored as an attribute in the output HDF file. Would changing that have any consequence downstream? Or I suppose I could keep the output format identical to what is now.

GarethDaviesGW: good point - I think keeping it as it is should make sense then

GarethDaviesGW (approved):

PR #3613, opened 2021-02-04
re-normalize the cumulative weights to 1 to avoid numerical issues when resampling weighted samples

ahnitz:
@sum33it @cdcapano I encountered a case where the weight sum is nearly, but not quite 1, and where there are sufficient samples so that a point may want to access higher than this value in the distribution tail. This occurs because of numerical error in the sampler weights. The immediate problem is that this prevents resampling to an equal-weight distribution.


cdcapano (approved):
Looks fine

sum33it (approved):

PR #3616, opened 2021-02-05
add test model to build posterior prob from samples

ahnitz:


pycbc/inference/models/analytic.py
```diff
@@ -212,3 +214,48 @@ def _loglikelihood(self):
...
+        idx = numpy.random.choice(idx, size=int(nsamples), replace=False)
+        samples = samples[:, idx]
+
+        logging.info('making kde with %s samples', samples.shape[-1])
+        self.kde = stats.gaussian_kde(samples)
+        logging.info('done initializing test posterior model')
+
+    def _loglikelihood(self):
+        """Returns the log pdf of the Rosenbrock function.
+        """
```
cdcapano, line 258: Fix the doc string here.

cdcapano (approved):
Looks good; just fix the doc string.

PR #3617, opened 2021-02-06
Various fixes to PyCBC Live's test

titodalcanton:
* Fix the injection recovery check in PyCBC Live's test, which was effectively disabled.
* Inject the two signals at different sky locations, to make BAYESTAR's results more interesting.
* The injected right ascension and declination were given in degrees, while they take radians (not a problem per se, but it got me confused for a moment when I plotted BAYESTAR's results).
* Remove some duplication in the simulation parameters.

### DIFF
--- a/examples/live/check_results.py
+++ b/examples/live/check_results.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python
 
 import sys
+import argparse
 import glob
 import logging as log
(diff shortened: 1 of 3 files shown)


spxiwh (approved):
Changes are all confined to `examples` and I think everything looks good. Thanks!

PR #3618, opened 2021-02-08
gracedb upload script - add options to make output without upload

GarethDaviesGW:
Add options in pycbc_upload_to_gracedb:
- to check if the event already exists in gracedb before uploading it (this becomes default behaviour)
- to not upload anything
- to choose to use a naming convention for the XML file https://dcc.ligo.org/LIGO-T050017/public

### DIFF
--- a/bin/pycbc_upload_xml_to_gracedb
+++ b/bin/pycbc_upload_xml_to_gracedb
@@ -37,6 +37,26 @@ from glue.ligolw import utils as ligolw_utils
 from ligo.segments import segment, segmentlist
 from pycbc.io.live import gracedb_tag_with_version
 
+def check_gracedb_for_event(gdb_handle, query, far):
+    """
+    Check if there is an event in gracedb with the queried string
+    which matches the FAR given
+    """
+    gdb_events_match_query = list(gdb_handle.events(query=query))
+    ifar = 1. / lal.YRJUL_SI / far
+    for gdb_event in gdb_events_match_query:
+            # Test each gracedb event to see if the FAR matches this event
+            if gdb_event['far'] == far:
+                # If an event has been found which matches the FAR
+                logging.info('Event already exists in GraceDb server with '
+                             'time %.3f and IFAR %.3e: %s',
+                             gdb_event['gpstime'], ifar,
+                             gdb_event['graceid'])
+                return True
+    # If no event has been found, log this, and return False
+    logging.info('No event found in GraceDb with IFAR %.3e when using '
+                 'query %s', ifar, query)
+    return False
 
 class LIGOLWContentHandler(ligolw.LIGOLWContentHandler):
     pass
@@ -58,6 +78,23 @@ parser.add_argument('--min-ifar', type=float, metavar='YEARS',
 parser.add_argument('--production-server', action="store_true", default=False,
                     help="Upload event to production graceDB. If not given "
                          "events will be uploaded to playground server.")
+parser.add_argument('--force-overwrite', action='store_true', default=False,
+                    help="GraceDb instance will be checked for if an event "
+                         "with the same event time and FAR already exist. "
+                         "If so, event will not be uploaded")
+parser.add_argument('--query-string', default='pycbc',
+                    help="If not using --force-overwrite, add a string to "
+                         "gracedb query to further filter events. "
+                         "Default='pycbc'")
+parser.add_argument('--filename-T050017',
+                    help="If given use T050017-v1 naming convention for "
+                         "output XML filename. Argument is string for "
+                         "search identifier in filename, e.g. 'AllSky' "
+                         "would give H1L1V1-PYCBC_AllSky-1234567890-1.xml.")
+parser.add_argument('--no-upload', action='store_true',
+                    help="Flag used to indicate that we are not uploading to "
+                         "GraceDb.")
+
 args = parser.parse_args()
 
 if args.production_server:
@@ -113,6 +150,13 @@ for event in coinc_table:
             coinc_inspiral_table_curr[0].combined_far > 1./args.min_ifar/lal.YRJUL_SI:
         continue
 
+    if not args.force_overwrite:
+        time = coinc_inspiral_table_curr[0].end_time
+        far = coinc_inspiral_table_curr[0].combined_far
+        query = args.query_string + ' %.3f .. %.3f' % (time - 1, time + 1)
+        if check_gracedb_for_event(gracedb, query, far):
+            continue
+
     sngl_ids = []
     for coinc_map in coinc_event_map_table:
         if coinc_map.coinc_event_id == event.coinc_event_id:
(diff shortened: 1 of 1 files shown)


bin/pycbc_upload_xml_to_gracedb
```diff
@@ -153,22 +198,44 @@ for event in coinc_table:
...
-    ligolw_utils.write_filename(xmldoc, "tmp_coinc_xml_file.xml")
+
+    if args.filename_T050017:
+        ifos = sorted([sngl.ifo for sngl in sngl_inspiral_table_curr])
+        ifos_str = ''.join(ifos)
+        id_str = args.filename_T050017
+        filename_xml = "{}-PyCBC_{}-{:d}-1.xml".format(ifos_str, id_str, time)
+        filename_psd = "{}-PyCBC_{}-{:d}-1_psd.xml.gz".format(ifos_str, id_str, time)
+    else:
+        print("Using default filenames")
```
titodalcanton, line 209: I am not sure this logging is necessary given that you do not have a corresponding output in the other branch of the `if`, but if you decide to keep it, replace `print()` with `logging.info()`.

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -153,22 +198,44 @@ for event in coinc_table:
     xmldoc.childNodes[-1].appendChild(coinc_inspiral_table_curr)
     xmldoc.childNodes[-1].appendChild(coinc_event_map_table_curr)
     xmldoc.childNodes[-1].appendChild(sngl_inspiral_table_curr)
-    ligolw_utils.write_filename(xmldoc, "tmp_coinc_xml_file.xml")
+
+    if args.filename_T050017:
+        ifos = sorted([sngl.ifo for sngl in sngl_inspiral_table_curr])
+        ifos_str = ''.join(ifos)
```
titodalcanton, line 204: Is `sum(ifos)` easier to read?

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -153,22 +198,44 @@ for event in coinc_table:
...
+        id_str = args.filename_T050017
+        filename_xml = "{}-PyCBC_{}-{:d}-1.xml".format(ifos_str, id_str, time)
+        filename_psd = "{}-PyCBC_{}-{:d}-1_psd.xml.gz".format(ifos_str, id_str, time)
```
titodalcanton, line 207: If you want to stick to the naming convention, I guess `filename_psd` should be something like `H1L1-PYCBC_PSD_FOOBAR-1234-1.xml.gz` instead? And maybe to make things more clear, `filename_xml` could then be `H1L1-PYCBC_TRIGGER_FOOBAR-1234-1.xml.gz` instead.

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -153,22 +198,44 @@ for event in coinc_table:
...
+        xmldoc.childNodes[-1].removeChild(sngl_inspiral_table_curr)
+        continue
+
     if args.testing:
-        r = gracedb.createEvent("Test", "pycbc", "tmp_coinc_xml_file.xml",
+        r = gracedb.createEvent("Test", "pycbc", filename_xml,
                                 search="AllSky", offline=True).json()
     else:
-        r = gracedb.createEvent("CBC", "pycbc", "tmp_coinc_xml_file.xml",
+        r = gracedb.createEvent("CBC", "pycbc", filename_xml,
```
titodalcanton, line 230: While you are at it, may I ask to reduce these several lines to something like this:
```python
group_tag = 'Test' if args.testing else 'CBC'
r = gracedb.createEvent(group_tag, 'pycbc', filename_xml…
```

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -153,22 +198,44 @@ for event in coinc_table:
...
                                 search="AllSky", offline=True).json()
     logging.info("Uploaded event %s.", r["graceid"])
 
     # upload PSD
     gracedb.writeLog(
         r["graceid"], "PyCBC PSD estimate from the time of event",
-        "psd.xml.gz", open("tmp_psd.xml.gz", "rb").read(), "psd").json()
-    logging.info("Uploaded file psd.xml.gz to event %s.", r["graceid"])
+        "psd.xml.gz", open(filename_psd, "rb").read(), "psd").json()
```
titodalcanton, line 237: Should you also replace `"psd.xml.gz"` with `filename_psd`? Genuine question actually, I forget if downstream software expects it to be literally named `psd.xml.gz`!

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -37,6 +37,26 @@ from glue.ligolw import utils as ligolw_utils
...
+def check_gracedb_for_event(gdb_handle, query, far):
+    """
+    Check if there is an event in gracedb with the queried string
+    which matches the FAR given
+    """
+    gdb_events_match_query = list(gdb_handle.events(query=query))
+    ifar = 1. / lal.YRJUL_SI / far
+    for gdb_event in gdb_events_match_query:
+            # Test each gracedb event to see if the FAR matches this event
+            if gdb_event['far'] == far:
```
titodalcanton, line 49: Are FARs not rounded or truncated by GraceDB? I think some tolerance is needed here.

bin/pycbc_upload_xml_to_gracedb
GarethDaviesGW: Ah, I hadn't thought of the downstream software relying on things.

I could make it so that the naming convention can only be used with --no-upload?

bin/pycbc_upload_xml_to_gracedb
GarethDaviesGW: good point - my use case was only to avoid events being uploaded more than once. I'll put in a small difference

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -37,6 +37,26 @@ from glue.ligolw import utils as ligolw_utils
...
+                # If an event has been found which matches the FAR
+                logging.info('Event already exists in GraceDb server with '
+                             'time %.3f and IFAR %.3e: %s',
+                             gdb_event['gpstime'], ifar,
+                             gdb_event['graceid'])
+                return True
+    # If no event has been found, log this, and return False
+    logging.info('No event found in GraceDb with IFAR %.3e when using '
+                 'query %s', ifar, query)
```
titodalcanton, line 58: Suggest quoting the query string in the logging call.

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -37,6 +37,26 @@ from glue.ligolw import utils as ligolw_utils
...
 
+def check_gracedb_for_event(gdb_handle, query, far):
+    """
+    Check if there is an event in gracedb with the queried string
+    which matches the FAR given
+    """
+    gdb_events_match_query = list(gdb_handle.events(query=query))
+    ifar = 1. / lal.YRJUL_SI / far
+    for gdb_event in gdb_events_match_query:
+            # Test each gracedb event to see if the FAR matches this event
```
titodalcanton, line 48: Please stick to 4-space indentation.

bin/pycbc_upload_xml_to_gracedb
GarethDaviesGW: I think the "trigger" bit might not be allowed according to how people downstream are interpreting the naming convention at the moment

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -58,6 +78,24 @@ parser.add_argument('--min-ifar', type=float, metavar='YEARS',
...
+parser.add_argument('--query-string', default='pycbc',
+                    help="If not using --force-overwrite, add a string to "
+                         "gracedb query to further filter events. "
+                         "Default='pycbc'")
+parser.add_argument('--filename-T050017',
+                    help="If given use T050017-v1 naming convention for "
+                         "output XML filename. Argument is string for "
+                         "search identifier in filename, e.g. 'AllSky' "
+                         "would give H1L1V1-PYCBC_AllSky-1234567890-1.xml.")
+# https://dcc.ligo.org/LIGO-T050017/public
```
titodalcanton, line 94: This comment probably goes before the previous `add_argument` call, or maybe even in the help string for the option?

titodalcanton (changes requested):
Thanks Gareth, see comments inline, and a more general one: can we just use the T050017 convention all the time and not have the option?

bin/pycbc_upload_xml_to_gracedb
titodalcanton: This sounds like something you want to test quite a bit on gracedb-playground.

GarethDaviesGW:
> Can we just use the T050017 convention all the time and not have the option?

That would be nice - but i think your point on downstream naming requirements means I should have a look at how it is used/does it break things

bin/pycbc_upload_xml_to_gracedb
titodalcanton: I guess I was trying to say that, with the change you are currently proposing, the file will still be named `psd.xml.gz` on GraceDB, even when using `--format-T050017`. But it sounds like you are thinking of using `--format-T050017` in combination with `--no-upload`, so maybe you are ok with that.

GarethDaviesGW: looking at the writelog documentation [here](https://ligo-gracedb.readthedocs.io/en/latest/api.html?highlight=writelog#ligo.gracedb.rest.GraceDb.writeLog)

> filecontents (file, optional) – handler pointing to a file to be read and uploaded. If this argument is specified, the contents will be saved as filename on the server.

The bytes object which we get from the read method is the fie contents, and the psd.xml.gz is the filename it is saved as - so it's accidentally (i.e. I forgot to change that bit) the right answer!

bin/pycbc_upload_xml_to_gracedb
GarethDaviesGW: Apparently it doesn't work :shrug:

bin/pycbc_upload_xml_to_gracedb
GarethDaviesGW: Looking at the gstlal uploads filenames (i.e. if the requirements are the same, the filenames must be the same), the psds do require the same filename; `psd.xml.gz`, but the coinc xmls do not

bin/pycbc_upload_xml_to_gracedb
GarethDaviesGW: I did a fair bit of testing but removing the actual upload and replacing with a print statement instead of actual upload and this did work in the == case.

I've added a tolerance of 1e-16 (approx 1e-7 in IFAR) to be safe and will retest.

bin/pycbc_upload_xml_to_gracedb
GarethDaviesGW: I've removed the option to have default filenames now, so this has been removed

GarethDaviesGW:
I've updated to use the T050017 naming convention by default, and allow a different output directory for results files.

titodalcanton (approved):
Thanks, looks good.

Issue #3619, opened 2021-02-08
workflow ConfigParser import error


a-r-williamson:
It appears that the changes in #3589 have broken a few imports, specifically two places where we attempt `from pycbc.workflow import ConfigParser`

- https://github.com/gwastro/pycbc/blob/f0fd8babb749f86783c6cfdeb7dbfe2b611da2b6/pycbc/distributions/__init__.py#L20
- https://github.com/gwastro/pycbc/blob/f0fd8babb749f86783c6cfdeb7dbfe2b611da2b6/pycbc/inference/sampler/base_mcmc.py#L37

PR #3620, opened 2021-02-08
Changing two broken imports for ConfigParser

a-r-williamson:
These chages are intended to address the issue raised in #3619, where these imports no longer worked after #3589


ahnitz (changes requested):
@a-r-williamson I think the import was for the ConfigParser class which is contained in the configparser (lower case) module, so this doesn't seem to be quite the right imports.

ahnitz (approved):

PR #3622, opened 2021-02-10
try to make relative model independent of detector

ahnitz:
I've encountered dynamic range issues when trying to push ET + CE analyses. One reason for this is due to extending the CE analysis artifically down to 2 Hz where the PSD is very large. The current code does not allow setting different options per detector for the relative model. This patch attempts to do that.

### DIFF
--- a/examples/inference/relative/relative.ini
+++ b/examples/inference/relative/relative.ini
@@ -1,7 +1,7 @@
 [model]
 name = relative
-low-frequency-cutoff = H1:30.0 L1:25.0 V1:40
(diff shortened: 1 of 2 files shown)


cdcapano (approved):
Looks ok to me, but @dfinstad should really check first.

dfinstad (approved):
Thanks very much for doing this. The changes look reasonable, and certainly the added flexibility for the model is fantastic. My only slight reservation is that this must introduce a little bit of extra overhead for the case where all detectors will use the same settings, but I assume that's small so I'll go ahead and approve this.

PR #3625, opened 2021-02-12
try building wheels

ahnitz:
This PR is an attempt to start building PyCBC binary wheel files. It would be nice for many users not to have to build pycbc directly but to be able to install a pre-built binary. This is an attempt to create those binaries using the 'manylinux' container which is the standard for redistributable linux wheels.

### DIFF
--- a/.github/workflows/basic-tests.yml
+++ b/.github/workflows/basic-tests.yml
@@ -71,6 +71,7 @@ jobs:
         SINGLE_COMMIT: true
   build_wheel:
     name: Build manylinux wheels for later possible upload
+    runs-on: ubuntu-20.04
     steps: 
       - uses: RalfG/python-wheels-manylinux-build@v0.3.3
         with:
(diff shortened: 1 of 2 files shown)


ahnitz:
Todo:
*  upload wheels to pypi on release
* build wheels for MacOSX

ahnitz:
@spxiwh I'll leave figuring out macos to the future, otherwise, I think this is ready to be reviews. Note that I've moved pypi to its own workflow and out of the basic-tests ones. This means I can't require those pass before this can run, however, as we only merge requests to master and make releases on commits where those tests have already passed, I think this is still safe enough for our purposes.

ahnitz:
Ok, this now *does* build macos wheels as I've moved it to the community standard cibuildwheel tool. 

https://cibuildwheel.readthedocs.io/en/stable/

spxiwh (approved):
I still don't fully understand all of github actions, but this all looks sane. I guess if problems arise if we try a release it can be resolved afterwards. Thanks.

PR #3631, opened 2021-02-15
enable basic mac tests

ahnitz:


ahnitz:
Depends on #3632 and #3629

ahnitz:
Putting this on hold until a bunch of prereqs in other PRs are merged.

ahnitz:
@spxiwh I found one really weird thing here. "TaylorT2" is return NaNs but only in python3.8 in the mac environment. This was causing the autochisq test to fail. Switching to SpinTaylorT4 and adjusting the test allows it to pass. However, I'm somewhat uncomfortable with this. 

@spxiwh The github env is supposed to be mac os 10.15. Do you know if TaylorT2 works for you in a 3.8 env and that OS?

spxiwh (approved):
Thanks! This looks good. At the moment I don't have this OS, or this version of python3 installed on my mac, so won't be able to test this immediately. So approving now anyway.

SpinTaylorT5 is more used anyway ... But it doesn't really make sense why p37 -> p38 would cause differences in *one* lalsim approximant on a mac. (I might suspect something beyond just python changes)

ahnitz:
@spxiwh Yes, I agree. I suspect that lalsuite is either picking up (or has bundled with it) a different system library there which is failing.

PR #3632, opened 2021-02-15
unpin h5py and fix issue in descriptor usage of compund types

ahnitz:
Newer version of h5py return a dtype descriptor with metadata, this caused our functions for handling record arrays to fail when constructing compound types as the return product had an unexpected format. This patch attempts to coerce it to the required format for constructing a compound type with numpy. 

This should hopefully allows us to move forward with h5py and when we can drop python2.7 support enable us to use h5py > 3 as a base which has a number of useful improvements.

### DIFF
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -177,5 +177,5 @@ if 'analysis_time' in fits.attrs:
     outfile.attrs['analysis_time'] = fits.attrs['analysis_time']
 
 # add a magic file attribute so that coinc_findtrigs can parse it
-outfile.attrs['stat'] = ifo + b'-fit_coeffs'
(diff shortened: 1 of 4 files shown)


ahnitz:
I've tested this locally addresses the first identified problem, however, I have not yet run the full test suite. :crossed_fingers:

ahnitz:
Hopefully should address #2937

ahnitz:
Including @spxiwh to ensure LSC members are aware this would affect H5py versions.

ahnitz:
Ok, I got the test workflow to run through with h5py 3.1.0 now. @spxiwh I couple places I needed to revert some decodes. I don't think we can maintain compatibility for things generated in a different python version (at least not between the 2/3 divide). This should be self-consistent though on either side and still work.

spxiwh (approved):
Thanks. The python2/3 cross incompatibility in h5py is annoying, but if we now start thinking that we are dropping python2 support in the near future then I think removing the decodes is fine.

PR #3637, opened 2021-02-16
Add script to estimate source probabilities from trigger files

veronica-villa:


bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+GPS_START_TIME = args.trigger_file.split('-')[2]
+GPS_START_TIME_NS = args.trigger_file.split('-')[3].split('.')[0]
+
+dir_path = 'source_probability_results/CHUNK_%s-%s/' % (GPS_START_TIME,
+                                                        GPS_START_TIME_NS)
+if not os.path.exists(dir_path):
+    try:
+        os.makedirs(dir_path)
+        logging.info('Created directory %s' % dir_path)
+    except OSError:
```
tdent, line 53: I'm not sure why you need to have a specially coded try-except switch here - what is the problem if you just write os.makedirs and if there is some sort of error the program should stop and report the error anyway

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+              (args.bank_file).split('/')[-1],
+              (',').join([name.split('/')[-1] for name
+                          in args.single_detector_triggers])))
+
+GPS_START_TIME = args.trigger_file.split('-')[2]
+GPS_START_TIME_NS = args.trigger_file.split('-')[3].split('.')[0]
+
+dir_path = 'source_probability_results/CHUNK_%s-%s/' % (GPS_START_TIME,
+                                                        GPS_START_TIME_NS)
+if not os.path.exists(dir_path):
```
tdent, line 49: If I look for the documentation on os.makedirs it seems to have an option `exist_ok=False` which you can change to True to allow for the case that the dir already exists. So I think if you use this it's not necessary to check if it already exists.

Maybe you might want to warn the user that they could be overwriting existing results or have a special option to control that, but that is a separate functionality which at the moment I don't think we need

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+    try:
+        os.makedirs(dir_path)
+        logging.info('Created directory %s' % dir_path)
+    except OSError:
+        print('ERROR creating directory %s' % dir_path)
```
tdent, line 54: The problem here is that the program prints that there is an error but doesn't halt, it still goes on and does the calculations and tries to save the output plots ..

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+
+mass1 = fortrigs.get_bankfile_array('mass1')[idx]
+mass2 = fortrigs.get_bankfile_array('mass2')[idx]
+mchirp,_ = mass1_mass2_to_mchirp_eta(mass1, mass2)
+end_time = fortrigs.get_end_time()[idx]
+sngl_snr = fortrigs.get_snglfile_array_dict('snr')
+sngl_sigmasq = fortrigs.get_snglfile_array_dict('sigmasq')
+
+for event in tqdm.trange(len(ifar)):
+    ifos_event = [ifo for ifo in sngl_snr.keys() if sngl_snr[ifo][1][event]]
```
tdent, line 78: maybe we need a comment here to explain what is sngl_snr[ifo] so we can see why you are indexing on the 1 or 0 element

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+parser.add_argument('--ifar-threshold', type=float, default=None,
+                    help='Select only candidate events with IFAR '
+                         'above threshold.')
+parser.add_argument('--verbose', action='count')
+parser.add_argument("--version", action="version",
+                    version=pycbc.version.git_verbose_msg)
+mchirp_area.insert_args(parser)
+args = parser.parse_args()
+
+mc_area_args = mchirp_area.from_cli(args)
```
tdent, line 32: maybe you want to check here (ie `assert`) that the NS/mass gap boundaries are the same number, otherwise if they are not then p(MG) could be nonzero and then we would get incorrect results by removing p(MG)

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+import h5py, logging
+import numpy as np
+import json
+import tqdm
+from pycbc.io import hdf
+from pycbc.pnutils import mass1_mass2_to_mchirp_eta
+from pycbc import mchirp_area
+import pycbc.version
+
+parser = argparse.ArgumentParser()
```
titodalcanton, line 17: You could pass the `description=__doc__` argument to the constructor, so that the docstring shows up in the help message.

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+parser.add_argument("--version", action="version",
+                    version=pycbc.version.git_verbose_msg)
+mchirp_area.insert_args(parser)
+args = parser.parse_args()
+
+mc_area_args = mchirp_area.from_cli(args)
+
+if args.verbose:
+    log_level = logging.INFO
+    logging.basicConfig(format='%(asctime)s : %(message)s', level=log_level)
```
titodalcanton, line 36: I suggesting using the function `pycbc.init_logging()` instead of these 3 lines.

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+
+logging.info('Using files: %s, %s, %s' %
```
titodalcanton, line 38: Do not use the `%` formatter in `logging`. Pass the arguments as simple arguments to `logging.info()` instead.

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+import json
+import tqdm
+from pycbc.io import hdf
+from pycbc.pnutils import mass1_mass2_to_mchirp_eta
+from pycbc import mchirp_area
+import pycbc.version
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--trigger-file')
+parser.add_argument('--bank-file')
```
titodalcanton, line 19: Please mark every required option with the `required=True` arg.

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+    ifos_event = [ifo for ifo in sngl_snr.keys() if sngl_snr[ifo][1][event]]
+    snrs_event = [sngl_snr[ifo][0][event] for ifo in ifos_event]
+    coinc_snr = sum([sngl_snr**2 for sngl_snr in snrs_event]) ** 0.5
+    min_eff_dist = min([(sngl_sigmasq[ifo][0][event])**0.5
+                       / sngl_snr[ifo][0][event] for ifo in ifos_event])
+    probs = mchirp_area.calc_probabilities(mchirp[event], coinc_snr,
+                                           min_eff_dist, mc_area_args)
+    probs.pop("Mass Gap")
+    ifo_names = ''.join(sorted(ifos_event))
+    end_time_s = str(end_time[event]).split('.')[0]
```
titodalcanton, line 87: Please use a more readable way of rounding the number,
```
end_time_s = '{:.0f}'.format(end_time[event])
```

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,92 @@
...
+
+    out_name = dir_path + '%s-%s-%s-1.json' % (ifo_names, args.search_tag,
+                                               end_time_s)
```
titodalcanton, line 90: Related to the above comment: in fact you could just do the rounding here and remove line 87.

bin/pycbc_source_probability_offline
veronica-villa: Thanks for bringing this up. I thought also on the possibility of having an extra argument like '--include-mass-gap' in case anyone wants the full set (?)

tdent: ok - that would be more complicated than we want to do now in this script. maybe to allow for this in future you could define a variable `include_mass_gap` which is set to False and then have this check performed only if the variable is False and also remove the MG probability only in this case. 
.. i.e. this would allow for the option to be added later.

bin/pycbc_source_probability_offline
veronica-villa: It seems the `exist_ok` option is only available for python3... so I think maybe we can stay with the `if not os.path.exists(dir_path)` option for the moment

titodalcanton: You could also use the custom `makedir()` function we have in `pycbc/__init__.py`, which I think does what you want.

bin/pycbc_source_probability_offline
veronica-villa: I did not want to round the number but truncate it so I just used the int(number) function

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,96 @@
+#!/usr/bin/env python
+"""
+Compute source probabilities using mchirp estimation method for all events
+in a chunk with an IFAR above certain threshold.
+"""
+import os
```
tdent, line 6: do you use os ?

bin/pycbc_source_probability_offline
```diff
@@ -0,0 +1,96 @@
...
+import pycbc
+from pycbc.io import hdf
+from pycbc.pnutils import mass1_mass2_to_mchirp_eta
+from pycbc import mchirp_area
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument('--trigger-file', required=True)
+parser.add_argument('--bank-file', required=True)
+parser.add_argument('--single-detector-triggers', nargs='+', required=True)
+parser.add_argument('--search-tag', required=True,
```
tdent, line 22: maybe explain the help a bit more, eg 'String to add to the output file names, eg (...)'

tdent (changes requested):
I think the import os can be removed, otherwise approved!

tdent (approved):

PR #3638, opened 2021-02-16
fetch full git history so version is set correctly when building wheel on github

ahnitz:


ahnitz:
@spxiwh Looking at the output of https://github.com/gwastro/pycbc/pull/3638/checks, I believe the wheels are now sanely named again after this change. Once the build has finished, it should also upload the wheels so I can double check them all.

ahnitz:
@spxiwh Ok, looking at the artifact uploaded at the link above, I believe the files are now properly named once the git history was not removed.

spxiwh (approved):
Thanks!

PR #3640, opened 2021-02-19
pycbc_make_skymap: add support for generating skymaps from simulated data & injections

ArthurTolley:
These changes let pycbc_make_skymap to take an injection file (in the form of an xml file) and produce a skymap using pycbc_single_template functionality for generating simulated data for an injection.

Things that need changing/improving:
- Currently only takes H1, L1, V1 as the detectors. There is a temporary print statement which tells you when you cant use an ifo you've inputted but ideally this would somehow disregard the entire pycbc_single_template subprocess that has been partially built for that ifo so far.
- I've tested this for simulated skymaps and for a single example I have locally, I can't find any examples which I can test this on to see if the functionality remains the same (the single example I have is a GWOSC set of frame files).
- Some descriptions could be updated to fit better

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -27,6 +27,8 @@ from ligo.gracedb.rest import GraceDb
 def default_frame_type(time, ifo):
     """Sensible defaults for frame types based on interferometer and time.
     """
+    if opt.injection_file is not None:
+        return ifo + '_FAKE_DATA'
     if time < 1137254517:
         # O1
         if ifo in ['H1', 'L1']:
@@ -48,6 +50,8 @@ def default_frame_type(time, ifo):
 def default_channel_name(time, ifo):
     """Sensible defaults for channel name based on interferometer and time.
     """
+    if opt.injection_file is not None:
+        return ifo + ':FAKE_DATA'
     if time < 1137254517:
         # O1
         if ifo in ['H1', 'L1']:
@@ -72,7 +76,7 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
          frame_types=None, channel_names=None,
          gracedb_server=None, test_event=True,
          custom_frame_files=None, approximant=None, detector_state=None,
-         veto_definer=None):
+         veto_definer=None, injection_file=None):
 
     if not test_event and not gracedb_server:
         raise RuntimeError('a GraceDB URL must be specified if not a test event.')
@@ -155,7 +159,8 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
         "--autogating-pad","16",
         "--strain-high-pass", str(highpass_frequency),
         "--pad-data", "8",
-        "--chisq-bins",'0.72*get_freq("fSEOBNRv4Peak",params.mass1,params.mass2,params.spin1z,params.spin2z)**0.7']
+        "--chisq-bins",'0.72*get_freq("fSEOBNRv4Peak",params.mass1,params.mass2,params.spin1z,params.spin2z)**0.7',
+        "--injection-file", str(opt.injection_file)]
 
         coinc_results['foreground/'+ifo+'/end_time'] = float(trig_time)
         coinc_results['foreground/'+ifo+'/mass1'] = mass1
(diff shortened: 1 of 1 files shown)


ArthurTolley:
I apologise that I didn't do this on a branch within my fork, I must've completely forgotten to.

bin/pycbc_make_skymap
```diff
@@ -448,6 +461,8 @@ if __name__ == '__main__':
                         action=MultiDetOptionAppendAction,
                         help='Lists of local frame files, e.g., '
                              'H1:/path/to/frame/file L1:/path/to/frame/file')
+    parser.add_argument('--injection-file', type=str,
+                        help='The injection file as an .xml.')
```
titodalcanton, line 465: Does anything actually prevent the use of HDF injection files? If not, please update the help text to something less specific, like "Optional path to an injection file".

bin/pycbc_make_skymap
```diff
@@ -155,7 +159,8 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
         "--autogating-pad","16",
         "--strain-high-pass", str(highpass_frequency),
         "--pad-data", "8",
-        "--chisq-bins",'0.72*get_freq("fSEOBNRv4Peak",params.mass1,params.mass2,params.spin1z,params.spin2z)**0.7']
+        "--chisq-bins",'0.72*get_freq("fSEOBNRv4Peak",params.mass1,params.mass2,params.spin1z,params.spin2z)**0.7',
+        "--injection-file", str(opt.injection_file)]
```
titodalcanton, line 163: What happens here if I do *not* want to use an injection file?

ArthurTolley: Wow, okay, I didn't update my code locally correctly. I will do some more testing and get back to you.

titodalcanton:
> I apologise that I didn't do this on a branch within my fork, I must've completely forgotten to.

This should not affect this PR, but it now prevents you from doing `git pull` and updating your working copy to the upstream master. So you will have to do a hard reset of your master branch to bring it back in sync with the upstream master after this PR is merged.

bin/pycbc_make_skymap
ArthurTolley: I'm using `pycbc_create_injections` which, looking back to the example, does create a `injections.hdf` file, not an xml which I think I'm forcing it to do. Might explain some of the strange things it was doing when I was using it

bin/pycbc_make_skymap
ArthurTolley: I've used `if opt.injection_file is not None`, there must be a more elegant way of writing this?

titodalcanton:
It looks like you want to couple the ability to provide an injection file with the ability to use simulated data. This also includes hardcoding the model PSDs. It is true that `pycbc_make_skymap` already hardcodes a lot of stuff, but I am a bit hesitant to put even more hardcoding into it.

Have you considered instead using `pycbc_condition_strain` to generate the fake data with injections, and then running `pycbc_make_skymap` on the generated data? This is what people have used for previous similar studies. It does mean running two commands instead of one, and having to go through a few temporary files. However, this is not really a problem if you write the temporary files to `/dev/shm`, and you get complete flexibility in your options without having to edit any code. Does this approach not work for some reason?

ArthurTolley:
Hi @titodalcanton, I hope you had a relaxing weekend!

I agree that the hardcoding is something that isn't preferable, I have tried to find alternative methods of doing this but I have ran out of ideas.

I haven't tried using `pycbc_conditioned_strain` (I didn't know this program existed and so have not tried this approach), this is something I will consider however it would be nice to have the simulated data all wrapped up into a single program because `pycbc_single_template` does currently perform the functionality that  `pycbc_conditioned_strain` would.

I'm all ears when it comes to suggestions on how to improve the code that I've written.

titodalcanton:
@ArthurTolley fair enough, here are some requests for changes:
* Do not tie together the injection file and fake data, i.e. let the user add an injection to real *or* simulated data.
* Instead of hardcoding the PSD models, create a new multidetector command-line option that works like this: `--fake-strain H1:aLIGOZeroDetSuperPowers L1:aLIGOZeroDetSuperPowers V1:AdvVirgoAmazingSensitivity K1:KAGRAUnconceivablyGreatPSD` (have a look at similar options for examples, such as `--detector-status`). The value for each detector will have to be picked out and sent to the corresponding `--fake-strain` option passed to `pycbc_single_template`.
* You will also want the ability to specify the random seeds for the simulated strain, so create an additional multidetector option like this: `--fake-strain-seed H1:1234 L1:2345 V1:3456`. Again, each value will have to be sent to the `pycbc_single_template` call corresponding to the right detector.
* Other `--frame-strain-*` options can probably be omitted, and if `pycbc_single_template` wants them explicitly, it is probably fine to hardcode them to something sensible (but *only* if `--fake-strain` has been given).

ArthurTolley:
These are all great suggestions, I will work on them slowly, most likely updating this PR as I go (if that's okay) to keep a diary of sorts for the progress.

PR #3646, opened 2021-02-23
unpin numpy version

ahnitz:


ahnitz:
@spxiwh I think we can now unpin numpy again with the latest lalsuite release. It should avoid the segfault issues that were observed before.

ahnitz:
@spxiwh To help with this I've always switched out pyproject.toml to use scipy's metapackage for setting which version of numpy to build against. 

see
https://pypi.org/project/oldest-supported-numpy/

ahnitz:
Ok, maybe using the 'oldest-supported-numpy' package won't work. Darn.

ahnitz:
@spxiwh I also took the opportunity to fix an issue with the pyproject.toml. If someone is installing from there, we should use the oldest version of numpy possible which is compatible / available in wheel form. This is to ensure the best compatibility if someone uses an older numpy that it happens to build against. I tried using scipy's metapackage for this, but we require newer versions in some places, so I've put the information in by hand with a note to future self on what to remember if we need to modify this file again.

spxiwh (approved):

PR #3647, opened 2021-02-23
add option to set final taper size for pre-merger waveform

ahnitz:


ahnitz:
@titodalcanton This allows one to enable and set the size of a FD taper window for the preTaylorF2 approximant. The new option is 'final_taper' which tables a number in Hz. It applies a taper end at the 'f_final' option you give.

titodalcanton (approved):
Looks good.

PR #3649, opened 2021-02-24
updates to relative model

ahnitz:


ahnitz:
@dfinstad @cdcapano This makes a few improvements to the relative model to improve performance. Though I'll note it still can be much faster. 

1) only generate waveform for ifos that have different frequency samples. Originally, the loop was written so the waveform was calculated for every ifo, which is an uneaded duplication.
2) various math rearrangement to reduce inner loop count. For example, one division that was always done was moved to the precalculated summary data. 
3) Move the inner math of the likelihood calculation to a c-compiled function in 'relbin_cpu.pyx'. This includes two functions, one where fp/fc/dtc are frequency dependent and one where they are constants. 

There is also one scientific change. 
1) Supply the time as a function of frequency vector *also* to the time shift function. This effectively adds in the doppler shift due to the earth rotation (or another way to think about it is really the time delay change from earth center since we consider the approximation where the earth's center is non-accelerating anyway). 


Notes:
I said there is still major speedups beyond this to be had. In particular in order roughly of importance
1) the next obvious improvement would be to speed up the detector time delay and antenna patter methods. 
2) find way to speed up the waveform transformations being applied
3) speed up prior calculation.

ahnitz:
@dfinstad @cdcapano So you know where things stand *after* this patch. The antenna pattern stuff is the ~ 18% listed as internal to the loglr function but not specifically named. You can see that the transform are then quite expense after that. 
test

ahnitz:
also requires #3655

cdcapano (approved):
Looks fine to me. What kind of speed up did you get with this? Also, I noticed you added unit tests for some of the models; thanks for doing that.

cdcapano:
Looks fine to me, @dfinstad?

ahnitz:
@cdcapano Depends on the configuration. For mine ~ 2x. There are lots of places to optimize still though.

ahnitz:
@cdcapano Yes,  put in a test for the phase-marginalized models. It would be good to have a set in the future for the other models (i.e. the pol marg ones and the regular one)

PR #3650, opened 2021-02-25
Fix bug in final_spin_from_initial

micamu:
In the conversions module, the lalsim function called in final_spin_from_initial is currently giving an error
"TypeError: in method 'SimIMREOBFinalMassSpin', argument 3 of type 'REAL8'"

This PR fixes the bug by ensuring the masses are floats.


cdcapano (commented):
Thanks for fixing... since you've done this, should ensure the spins are floats also? Easiest way to do it would be:
```
spin1 = list(map(float, [spin1x[ii], spin1y[ii], spin1z[ii]]))
```
Ditto for spin2.

micamu:
@cdcapano can this be merged now?

cdcapano (approved):
Yes, sorry forgot about this.

PR #3651, opened 2021-02-28
add new unittest for inference models

ahnitz:
This adds a unittest which checks a few common inference models. More testing can be done and would be a welcome addition. To get things started this checks that the 'marginized phase', 'single template', and 'relative' model give similar likelihood values GW170817 at a chosen sample point.


ahnitz:
Depends on #3649

ahnitz:
I've decided to just put this directly into #3649

PR #3655, opened 2021-03-01
speed up field array access when using eval

ahnitz:
As I mentioned in #3649, the record array access can be slow when using an eval. This helps to speed that up in the case where you do the same kind of access repeatedly. 

It is ~ 7x faster in the simple case of doing something like arr["x + 1"] where x is a named field. 

In my benchmark  case it reduced the useage of this method from 25 % of the runtime to 13%, so still not great, but certainly better than before ~2x for the case I was doing which includes 3 waveform transforms to handle source frame analysis with comoving volume prior. I think to do much better would require some reworking of getattr possibly and my guess is could only afford another factor of 2 in the best case without dropping the use of eval entirely. I think the latter is not worth it at the moment, but should be kept in mind that this is an area which could be improved.


cdcapano (approved):
Thanks for doing this, looks good.

cdcapano:
@ahnitz I forgot to mention, please address the two code climate issues before merging.

I don't know why the build is failing; the run workflow output isn't very descriptive. It seems like it should be unrelated to this patch.

PR #3661, opened 2021-03-08
choose wheel preferentially

ahnitz:


ahnitz:
@spxiwh Recent PRs such as #3655 are failing due to lacking hdf header files. This occurred because the last release did not include wheels for older platforms (python < 3.7). This changes our install so that wheels are preferred and an older version will be installed in favor of trying to build from source.

ahnitz:
Ack, this option doesn't exist on python2.7!

PR #3665, opened 2021-03-09
update h5py requirements

ahnitz:


ahnitz:
Revert this and unpin for python 3.9 in favor of #3661 when we drop python 2.7 support. This should do for now.

spxiwh (approved):
No objections from me, but hopefully we can remove python2 in the coming months, so we might want to keep an eye on things to be cut out after that happens!

ahnitz:
@spxiwh The test failures here are due to upstream packaging issues (i.e. packages dropping python2.7 support but not yet fully declaring as such so they get picked up in the wrong environment).

ahnitz:
at some point we may want to follow https://numpy.org/neps/nep-0029-deprecation_policy.html

as much of our ecosystem is likely to follow this schedule roughly speaking. We should consider however, if this will cause us issues on our supported production ecosystems. In which case, it may not be possible to do so.

lpsinger:
I've urged the LSC software change control board (SCCB) to follow NEP 29.

ahnitz:
@lpsinger Thanks that's good to know. The outcome will be important here. The main concern we'd have is to make sure we don't break compatibility with a platform that is required for production analyses. Typically that has been limited by the oldest LSC compute cluster / reference OS.

PR #3666, opened 2021-03-09
add intended python requirement

ahnitz:
We should be in the habit of proactively declaring the python version supported so that incompatible version are not installed through pypi. Currently, this is 2.7, and >=3.6.


ahnitz:
By doing this now, we save some trouble later when we ourselves drop python2.7 support in the near future.

spxiwh (approved):

PR #3667, opened 2021-03-09
Support tagging for additional minifollowup jobs

dethodav:
This PR fixes an inconsistency with how tags were propagated to mini followup jobs as part of the coinc workflow. Currently, tags are only propagated for foreground follow up jobs: https://github.com/gwastro/pycbc/blob/f8f57150a96439c12720961796e47e53dce2b075/pycbc/workflow/minifollowups.py#L102-L103

 This PR incorporates the same tagging code for the injection and single detector follow ups.


spxiwh (approved):
Thanks!

PR #3669, opened 2021-03-11
add option to make saving data optional in inference files

ahnitz:
When doing lots of inference jobs on fake data and with possibly many detectors, the output file set can be huge mostly due to the size of the data set stored internally. This allows one to skip that by adding the following option to the model

```
[model]
no-save-data = 
...
...
```


cdcapano (approved):
Looks good

PR #3670, opened 2021-03-11
fix loop on failed attribute

ahnitz:
@cdcapano I found one more case that was a problem with field array. It could now (after the past patch) get into a loop due to how geattribute and getitem both refer to each other!. This resolves that so there can't be a loop when an attribute fails to be read, instead it will up-report the problem.


ahnitz:
@cdcapano BTW, the fundamental problem here is that broken virtualfields are allowed to exist. This would never have been an issue if virtualfields could be relied upon to list properties that actually worked. (like maybe when the are added it checks if it can actually be called?).

cdcapano (approved):

PR #3673, opened 2021-03-12
Changed the gating to use the default method of the time series.

MarlinSchaefer:
Added a new option --autogating-method to choose between different
available methods to apply the gates.


ahnitz:
@MarlinSchaefer I think our code is failing at your name having an umlaut. We should probably fix that. See the error here 

```
  File "setup.py", line 206, in <module>
    VERSION = get_version_info()
  File "setup.py", line 146, in get_version_info
    f.write('git_author = \'%s\'\n' % vinfo.author)
UnicodeEncodeError: 'ascii' codec can't encode character u'\xe4' in position 24: ordinal not in range(128)
```

This is occuring in python2.7. Can you make a 27 env so you can reproduce? Presumably we can fix with an appropriate use of encode on the string.

ahnitz:
@MarlinSchaefer Can you also verify that the code produces the same gates when you set the option corresponding to the older method? Perhaps use pycbc_condition_strain on a known time that should be gated and compare the time series produced?

titodalcanton:
Would this option affect just the auto-gates, or all gates? In the latter case, it should probably be called `--gating-method`. Also, you may want to use the `choices` kwarg.

ahnitz:
@MarlinSchaefer @titodalcanton This probably should be for all gates. @MarlinSchaefer That means this spot here also needs to use the method. https://github.com/gwastro/pycbc/blob/c153ecdb6655dfc26cf9e3cf45056b943190a745/pycbc/strain/strain.py#L326

titodalcanton:
Probably easier to modify the function `gate_data()` to make it take the method argument, in that case?

MarlinSchaefer:
@ahnitz I get the same error (UnicodeEncodeError) when calling `get_version_info` in a local Python 2.7 environment. Slapping an `.encode('utf-8')` behind many of the `vinfo.something` removes the error. However, I have no idea about encodings, so I'm not really sure what I'm doing there.

ahnitz:
@MarlinSchaefer Let's see what you've had to change. Can you post the changes?

MarlinSchaefer:
Changes are at `git_author`, `git_commiter`, `git_builder` and `git_verbose_msg`.
```python
def get_version_info():
    """Get VCS info and write version info to version.py
    """
    from pycbc import _version_helper

    class vdummy(object):
        def __getattr__(self, attr):
            return ''

    # If this is a pycbc git repo always populate version information using GIT
    try:
        vinfo = _version_helper.generate_git_version_info()
    except:
        vinfo = vdummy()
        vinfo.version = '1.18.dev1'
        vinfo.release = 'False'

    with open('pycbc/version.py', 'w') as f:
        f.write("# coding: utf-8\n")
        f.write("# Generated by setup.py for PyCBC on %s.\n\n"
                % vinfo.build_date)

        # print general info
        f.write('version = \'%s\'\n' % vinfo.version)
        f.write('date = \'%s\'\n' % vinfo.date)
        f.write('release = %s\n' % vinfo.release)
        f.write('last_release = \'%s\'\n' % vinfo.last_release)

        # print git info
        f.write('\ngit_hash = \'%s\'\n' % vinfo.hash)
        f.write('git_branch = \'%s\'\n' % vinfo.branch)
        f.write('git_tag = \'%s\'\n' % vinfo.tag)
        f.write('git_author = \'%s\'\n' % vinfo.author.encode('utf-8'))
        f.write('git_committer = \'%s\'\n' % vinfo.committer.encode('utf-8'))
        f.write('git_status = \'%s\'\n' % vinfo.status)
        f.write('git_builder = \'%s\'\n' % vinfo.builder.encode('utf-8'))
        f.write('git_build_date = \'%s\'\n' % vinfo.build_date)
        f.write('git_verbose_msg = """Version: %s\n'
                'Branch: %s\n'
                'Tag: %s\n'
                'Id: %s\n'
                'Builder: %s\n'
                'Build date: %s\n'
                'Repository status is %s"""\n' %(
                                               vinfo.version.encode('utf-8'),
                                               vinfo.branch.encode('utf-8'),
                                               vinfo.tag,
                                               vinfo.hash.encode('utf-8'),
                                               vinfo.builder.encode('utf-8'),
                                               vinfo.build_date.encode('utf-8'),
                                               vinfo.status.encode('utf-8')))
        f.write('from pycbc._version import *\n')
        version = vinfo.version

    from pycbc import version
    version = version.version
    return version
```

ahnitz:
@MarlinSchaefer Can you commit this and see if it fixes the unittests? Also I suspect you don't need the encoding everywhere, just  where a name would show up.

MarlinSchaefer:
@ahnitz Sure, I'll commit it in a second. The encodings are required though everywhere, where I put them. I tried to remove any of them and the code crashed. The reason there is no encoding on `vinfo.tag` is because the tag is `None` and has no encoding method.

MarlinSchaefer:
Okay, it's now those encodings that get me.

ahnitz:
@spxiwh You are bit more experienced with this encoding stuff. What's the python2/3 compatible solution here?

spxiwh:
Yeah, don't encode this. This is *not* a binary file, so you can't write `bytes` into it. python2 might let you get away with that, but it's not a good idea.

What is the original error. If I try to reproduce this with a minimal example on python2 I see:
```In [1]: fp = open ('test.DELETE','w')

In [2]: fp.write('üasd')

```
and it's fine.

spxiwh:
Screen Shot 2021-03-12 at 17 31 08
I did manage to reproduce this (though I'm still not sure why my example above actually works). Here's what you need

titodalcanton:
If you look at https://github.com/gwastro/pycbc/blob/master/pycbc/_version_helper.py#L67, there the output of the `git log` command (which is bytes, I think) is assumed to be UTF-8 and is decoded into a regular string. So I think what you want to do here is *reencode* it back to bytes (which you did) and then write the `version.py` file in `'wb'` mode, rather than `'w'`.

MarlinSchaefer:
@ahnitz  I've now checked the conditioned strain on some time interval that contains a gated section. Comparing the two methods visually produces no difference, the condition `old == new`, however, evaluates to `False`. The difference `abs(new - old)` on the other hand has a maximum of `1e-32`. The data is of the scale `1e-20`. So there is a difference of 12 orders of magnitude and most of that is in the parts of the data that are not altered by the gating method. Is that difference small enough or should I go looking for reasons?

ahnitz:
@MarlinSchaefer Thanks, that's within the expected floating point uncertainty we should expect.

ahnitz (approved):

PR #3675, opened 2021-03-15
add option to not save state with dynesty

ahnitz:


sum33it:
@ahnitz it looks fine to me. Do you think that we should also have a warning message about using this option? Like when one use this option then one can not make debugging plots for dynesty. As it will make workflows to be failed when both options are selected. @cdcapano any thoughts on this?

ahnitz:
@sum33it I'm not sure we can guess every possible combination of incompatible options. It's not the default, so  user has to copy this option to their config file.

sum33it (approved):

PR #3676, opened 2021-03-16
Fix definition of inclination and add thetajn to parameter names

cdcapano:


ahnitz (approved):

ahnitz:
@cdcapano Job failured looked like an issue on the github side. It might just be transient (or temporary at least)

PR #3680, opened 2021-03-17
Fix bug requiring inclination always

JulianWesterweck:
Inclination is always required to be set.
This prevents _not_ using spherical harmonics.
Now, spherical harmonics will be ignored if inclination is not given for the ...fromFreqTau approximants, as intended.


micamu:
Thanks for fixing this. Just being extra picky here, could we call this "inc" parameter "inclination" or "iota" to make it clear what it is? Thanks!

micamu:
@JulianWesterweck is this PR now obsolete, given the changes in PR #3713?

JulianWesterweck:
@ahnitz Yes, this is not needed anymore, thanks for the reminder. Closed.

PR #3685, opened 2021-04-01
pycbc_create_injections: Updating description from 'distribution' to 'prior'

ArthurTolley:
The current description of pycbc_create_injections references using a 'distribution' section for each variable parameter. 
This should be a 'prior' section for each parameters, this PR reflects this change.

The description still refers to a 'distributions' module which I believe is what it is still called so no changes there.


cdcapano (approved):
This looks good, thank you for fixing this @ArthurTolley

PR #3686, opened 2021-04-02
test pol marg models in unittest

ahnitz:
Add test to compare some of the polarization marginalized models. This looks at the single template and the brute force + marginalized_pol models (in combination, not individually).


cdcapano (approved):

ahnitz:
@cdcapano There is a bug here, but it is only on macosx and python 3.8. Can you reproduce on a mac system?

ahnitz:
Depends on #3692

Issue #3687, opened 2021-04-03
PyCBC currently broken with decorator >= 5.0


ahnitz:
With the last release of decorator, the behavior was changed so that default keywords are no longer being passed in the same way. 

See discussion here https://github.com/micheles/decorator/issues/104

This affects a lot of methods in our type classes which have default arguments.

ahnitz:
This looks like a mess to fix, so my suggestion atm is to hard require decorator < 5 until either we can sort out all the places that might need to be addressed (and how to do so) or there is a fix upstream.

ahnitz:
Example error on our side. 
```.tox/py-unittest/lib/python3.8/site-packages/pycbc/waveform/waveform.py:527: in get_td_waveform
    return wav_gen(**input_params)
.tox/py-unittest/lib/python3.8/site-packages/pycbc/waveform/waveform.py:703: in get_td_waveform_from_fd
    hp = wfutils.fd_to_td(hp, left_window=(nparams['f_lower'],
.tox/py-unittest/lib/python3.8/site-packages/pycbc/waveform/utils.py:531: in fd_to_td
    return htilde.to_timeseries(delta_t=delta_t)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

args = (<pycbc.types.frequencyseries.FrequencySeries object at 0x7f2ef38835e0>,)
kw = {'delta_t': None}

    def fun(*args, **kw):
>       return caller(func, *(extras + args), **kw)
E       TypeError: _noreal() got an unexpected keyword argument 'delta_t'

.tox/py-unittest/lib/python3.8/site-packages/decorator.py:214: TypeError
```

PR #3688, opened 2021-04-03
pin upper bound on decorator

ahnitz:
See #3687 which is caused by the most recent decorator release. My hope is that the behavior of decorator will revert back to what it did before so that we can unpin this requirement at some point. See discussion in https://github.com/micheles/decorator/issues/104.


ahnitz:
@spxiwh Depending on the resolution, we may need to add unpinning this to the list of things to do only after we fully drop 2.7 support. Otherwise, there may not be a dual python way (that is not hacky) to address this.

ahnitz:
I think this has been addressed upstream. Closing

PR #3689, opened 2021-04-06
avoid astropy issues in last point release

ahnitz:


requirements.txt
```diff
@@ -1,6 +1,6 @@
 # requirements for most basic library use
 astropy>=2.0.3,<3.0.0; python_version <= '2.7'
-astropy>=2.0.3; python_version >= '3.0'
+astropy>=2.0.3!=4.2.1,!=4.0.5; python_version >= '3.0'
```
titodalcanton, line 3: Is a comma needed between `2.0.3` and `!=`?

ahnitz: Yes, see updated commit.

titodalcanton (approved):
Just for reference, this is a workaround for [this error](https://git.ligo.org/lscsoft/ligo.skymap/-/issues/26).

ahnitz:
Here is the corresponding astropy report as well https://github.com/astropy/astropy/issues/11486

titodalcanton:
@ahnitz the latest release of ligo.skymap includes a workaround for astropy 4.2.1. How do you feel about reverting this pin?

ahnitz:
@titodalcanton It's not pinned, it just excludes the buggy astropy version. Newer version of astropy should be fine as they've fixed it on their end. Is there a reason we need to run 4.2.1?

lpsinger:
What are the issues with astropy 4.2.1?

ahnitz:
@lpsinger See the connected astropy issue. It has already been fixed upstream in astropy.

lpsinger:
Well, as @titodalcanton said, there is also a workaround in ligo.skymap. So if you update to the latest version of ligo.skymap, then you can remove that astropy pin.

ahnitz:
@lpsinger Is ligo.skymap pinned to 4.2.1? If so, then yep, go ahead and remove the exclusion to allow the latest ligo.skymap to be run for testing. I'm not sure when the next astropy release (with their fix) would be.

titodalcanton:
I was asking because the latest version of ligo.skymap requires astropy 4.2.1 (in order to fix other bugs) and there has not been an astropy 4.2.2 in a while.

ahnitz:
@titodalcanton Ok, I hadn't realized ligo.skymap was so restrictive. In which case, yeah, go ahead ,we weren't excluding 4.2.1 for any other reason I believe.

lpsinger:
> @lpsinger Is ligo.skymap pinned to 4.2.1? If so, then yep, go ahead and remove the exclusion to allow the latest ligo.skymap to be run for testing. I'm not sure when the next astropy release (with their fix) would be.

https://git.ligo.org/lscsoft/ligo.skymap/-/blob/v0.5.3/setup.cfg#L42

PR #3690, opened 2021-04-09
Adding helper message for 'psd-abort-distance' used by PyCBC Live.

rmacas:
Improving PyCBC Live arg `max/min-psd-abort-distance` helper message as per @titodalcanton request.

Added explanation for `max/min-psd-abort-distance` as follows:

> Safety BNS horizon distance (in Mpc) above/below which a detector's data is discarded.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -576,7 +576,7 @@ parser.add_argument('--max-batch-size', type=int, default=2**27)
 parser.add_argument('--store-loudest-index', type=int, default=0)
 parser.add_argument('--max-psd-abort-distance', type=float, default=numpy.inf,
                     help="Safety BNS horizon distance (in Mpc) above which a "
-                    "detector's data is discarded.".)
+                    "detector's data is discarded.")
 parser.add_argument('--min-psd-abort-distance', type=float, default=-numpy.inf,
                     help="Safety BNS horizon distance (in Mpc) below which a "
                     "detector's data is discarded.")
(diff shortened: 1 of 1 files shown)


bin/pycbc_live
```diff
@@ -574,8 +574,12 @@ parser.add_argument('--output-background-n-loudest', type=int, default=10000,
 parser.add_argument('--newsnr-threshold', type=float, default=0)
 parser.add_argument('--max-batch-size', type=int, default=2**27)
 parser.add_argument('--store-loudest-index', type=int, default=0)
-parser.add_argument('--max-psd-abort-distance', type=float, default=numpy.inf)
-parser.add_argument('--min-psd-abort-distance', type=float, default=-numpy.inf)
+parser.add_argument('--max-psd-abort-distance', type=float, default=numpy.inf,
+                    help="Safety BNS horizon distance (in Mpc) above which a "
+                    "detector's data is discarded.".)
```
titodalcanton, line 579: Thanks for the contribution @rmacas! There is a typo here (extra period after the end of the string) which is making the tests fail.

titodalcanton (changes requested):

bin/pycbc_live
rmacas: Thanks @titodalcanton, well spotted. Fixed the typo, now the PR passes the checks.

titodalcanton (approved):
Here you go: :cake: ☕

PR #3691, opened 2021-04-12
Improve of area calculation on mchirp_area.py

veronica-villa:
[`calc_areas`](https://github.com/gwastro/pycbc/blob/master/pycbc/mchirp_area.py#L90) function of `mchirp_area.py` presented some repetitive behaviour for each of the CBC categories. Since this repetition makes it quite a long function, we defined a new function to compute the area in the chirp mass contour inside a m1m2 region, common to all CBC cases (`get_area`).

A sanity check was performed applying both codes to the same set of events and configuration: [run_file](https://github.com/veronica-villa/source_probabilities_estimation_pycbclive/blob/master/sanity_check.py), [sanity_check_results](https://github.com/veronica-villa/source_probabilities_estimation_pycbclive/blob/master/mchirp_area_improved_sanity_check.txt).

### DIFF
--- a/pycbc/mchirp_area_improved.py
+++ b/pycbc/mchirp_area_improved.py
@@ -1,0 +1,204 @@
+# Module with utilities for estimating candidate events source probabilities
+# Initial code by A. Curiel Barroso, August 2019
+# Modified by V. Villa-Ortega, January 2020, March 2021
+
+"""Functions to compute the area corresponding to different CBC on the m1 & m2
+plane when given a central mchirp value and uncertainty.
+It also includes a function that calculates the source frame when given the
+detector frame mass and redshift.
+"""
+
+import math
+import numpy as np
+from pycbc.conversions import mass2_from_mchirp_mass1 as m2mcm1
+from scipy.integrate import quad
+from pycbc.cosmology import _redshift
+
+
+def insert_args(parser):
+    mchirp_group = parser.add_argument_group("Arguments for estimating the "
+                                             "source probabilities of a "
+                                             "candidate event using the snr, "
+                                             "mchirp, and effective distance.")
+    mchirp_group.add_argument('--src-class-mass-range', type=float, nargs=2,
+                              metavar=('MIN_M2', 'MAX_M1'),
+                              default=[1.0, 45.0],
+                              help="Minimum and maximum values for the mass "
+                                   "of the binary components, used as limits "
+                                   "of the mass plane when computing the area "
+                                   "corresponding to different CBC sources.")
+    mchirp_group.add_argument('--src-class-mass-gap', type=float, nargs=2,
+                              metavar=('MAX_NS', 'MIN_BH'), default=[3.0, 5.0],
+                              help="Limits of the mass gap, that correspond "
+                                   "to the maximum mass of a neutron star "
+                                   "and the minimum mass of a black hole. "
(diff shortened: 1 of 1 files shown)


pycbc/mchirp_area_improved.py
```diff
@@ -0,0 +1,204 @@
...
+    uncertainty and computes the source mass and its uncertainty.
+    """
+    msrc = mdet / (1. + z)
+    del_msrc = msrc * ((del_mdet / mdet) ** 2.
+                       + (del_z / (1. + z)) ** 2.) ** 0.5
+    return (msrc, del_msrc)
+
+
+def intmc(mc, x_min, x_max):
+    """Returns the integral of a component mass as a function of the mass of
```
tdent, line 84: maybe the help message could be improved - eg 'returns the integral of m2 over m1 between x_min and x_max, assuming that mchirp is fixed

pycbc/mchirp_area_improved.py
```diff
@@ -0,0 +1,204 @@
...
+       the other component, taking mchirp as an argument.
+    """
+    integral = quad(lambda x, mc: m2mcm1(mc, x), x_min, x_max, args=mc)
+    return integral[0]
+
+
+def get_area(trig_mc, lim_h1, lim_h2, lim_v1, lim_v2):
+    """Returns the area of the chirp mass contour in each region of the m1m2
```
tdent, line 92: the help string here is also a bit vague - do you know what it means by 'the area of the chirp mass contour' ? 
maybe the area *under* the contour?

tdent: also, the help message should say that trig_mc has to be a tuple or list of two values, the first value represents the central estimate of mchirp while the second is the uncertainty, otherwise it isn't clear what is happening

pycbc/mchirp_area_improved.py
```diff
@@ -0,0 +1,204 @@
...
+       plane taking horizontal and vertical limits of the region as arguments.
+    """
+    mcb = trig_mc[0] + trig_mc[1]
```
tdent, line 95: do you have an idea why these variables are called 'mcb' and 'mcs' ? maybe one should rename them to be more obvious?

veronica-villa: I'm not sure, because they were named like that at Andres' code, but I guess it could be something as mchirp big and mchirp small (?) I also agree in the rename.. I'm not sure if something like mc_top, mc_bottom or something like that?

titodalcanton:
I noticed that you are creating a brand new file `pycbc/mchirp_area_improved.py` and leaving the old `pycbc/mchirp_area.py` as is. Is the idea to create a separate new module with similar functionality, or should you apply the improvement to `mchirp_area.py`?

pycbc/mchirp_area_improved.py
tdent: mc_min / mc_max seems obvious .. ?

veronica-villa:
Hi @titodalcanton, we are going through a review of the source probability estimations for O3a-final and O3b-catalog papers, where `mchirp_area.py` was used. We are waiting for `mchirp_area_improved.py` to be reviewed, so we can rename it as `mchirp_area` and remove the actual `mchirp_area` module or rename it as `mchirp_area_old.py`

tdent (approved):

PR #3694, opened 2021-04-15
numpy style docstring for get_area

tdent:


tdent:
This is 'optional' but indicates a general upgrade of docstrings we might want to make depending on how review progresses.

veronica-villa (approved):

PR #3700, opened 2021-04-21
Add option for segment server when making sky maps

bpiotrzkowski:
This came as a request from admins at the Livingston cluster and @titodalcanton to reduce the queries on the standard segments server (https://segments.ligo.org), and instead to use the backup server (https://segments-backup.ligo.org) which has a higher bandwith. This change allows us to choose this server natively.


spxiwh (approved):
This looks good. Thanks!

PR #3701, opened 2021-04-21
Replace all references of numpy.int with int

duncanmmacleod:
This PR replaces all instances of `numpy.int` with the builtin `int`.

This resolves a deprecation warning from numpy 1.20, see
<https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated>.

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_hdfinjfind
+++ b/bin/all_sky_search/pycbc_coinc_hdfinjfind
@@ -202,7 +202,7 @@ for trigger_file, injection_file in zip(args.trigger_files,
     found_after_vetoes = numpy.array([i for i in found_within_time
                                       if i not in vetoed_all])
     missed_after_vetoes = numpy.array([i for i in missed_within_time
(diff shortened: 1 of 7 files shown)


lpsinger:
Some of these should probably be `np.intp`, not `int` or `np.int`.

duncanmmacleod:
> Some of these should probably be `np.intp`, not `int` or `np.int`.

This PR is designed to have no impact other than fixing the `DeprecationWarning`s, functional changes like your suggestion should be probably posted in an issue to not be forgotten, and then addressed in a future PR.

spxiwh (approved):
Leo is probably right that many of these should be 64-bit np ints. However, I think in all cases if integers got large enough that that mattered, the code would break anyway.

Approving and merging for now as fixing the Deprecation warning without changing functionality is good. Thanks Duncan!

lpsinger:
Well, the issue is that you should use `np.intp` for anything that you might use to index other Numpy arrays.

PR #3704, opened 2021-04-27
handle 170608 in a nicer way

ahnitz:
This pulls in a hack I used for O2, but mainlines it in a nicer way. The idea is to allow for the analysis of the 170608 time in a more uniform way as the other runs. The key factor is that the Hanford data was released as a separate run name by GWOSC so we need to be able to retrieve this name.


spxiwh (approved):
Looks good to me.

Issue #3698, opened 2021-04-15
Interactive plot features from mpld3 not working


dethodav:
The interactive features of plots in pycbc result pages haven’t been working. Impacted plotting scrips include: 

*  `pycbc_page_segplot`
* `pycbc_plot_gating`
 
The issue looks to be that the relevant pycbc code is no longer compatible with the most recent version of mpld3. 

The specific function that appears to be the issue is below:

https://github.com/gwastro/pycbc/blob/ba74cf52d14aa13acc33226f458d533c0fcbbacd/pycbc/results/mpld3_utils.py#L37-L110

This issue could be addressed in the short term by encoring `mpld3==0.3` in the requirements or using a different, similar plugin that is compatible with recent versions of `mpld3`. For example, the following line could be removed from `pycbc_page_segplot` and replaced with `mpld3.plugins.connect(fig, mpld3.pluginsZoom())`:

https://github.com/gwastro/pycbc/blob/ba74cf52d14aa13acc33226f458d533c0fcbbacd/bin/plotting/pycbc_page_segplot#L200

To restore full functionality with recent version sof `mpld3`, this custom script will likely need to be rewritten.

GarethCabournDavies:
Solved by #3705, just not closed the issue

PR #3705, opened 2021-05-03
Restore interactive plotting

dethodav:
This PR resolves the issue noted in #3698 by removing use of a custom `mpld3` plugin that is no longer functional with the most recent version of `mpld3`. An example output plot is [uploaded here](https://github.com/gwastro/pycbc/files/6417353/H1L1-PAGE_SEGPLOT_SUMMARY-EXAMPLE.zip)

Note that not all of the original functionality is restored with this PR. This plot original supported an interactive feature where the user could hover over a specific segment to display information (such as the start and end times) for that segment and used a custom plugin to only zoom and move on the x-axis.


ahnitz (approved):

ahnitz:
Thanks @dethodav !

PR #3710, opened 2021-05-19
More informative error messages in `array.py`

titodalcanton:
Makes debugging a bit easier and removes some code duplication.


ahnitz (approved):
This looks like a nice removal of duplicate code. Thanks!

PR #3711, opened 2021-05-19
Updates to plot posterior

cdcapano:
Some updates for plot posterior that I've been meaning to commit to master for awhile. Does the following:
 * Adds the ability to plot the maximum likelihood point with argument `--plot-maxl`. If turned on, an `x` is added to the 2D marginals showing where the maxL point is.
 * If `--use-kombine-kde` is turned on, will now call kombine's `optimized_kde` method. That uses Bayesian inference criteria to determine if there are multiple modes. It's slow, but can be useful for distributions that have more than one mode.
 * Adds a `--kde-args` to the command line. These are passed through to the KDE. Needed for to limit the number of points given to kombine's KDE (more than a few thousand and it will crawl to a halt), but also can be used to modify what scipy's KDE uses for bandwidths.
 * Adds the ability to turn off contour labels (`--no-contour-labels`), change the contour linestyles (`--contour-linestyles`) and turn off percentile lines in the marginal percentile plots (`--no-marginal-lines`) via the command line.
 * If 1 or 2 parameters are being plotted from multiple input files, the legend will be put in the 1D/2D axis, instead of the upper left of the figure.

### DIFF
--- a/bin/inference/pycbc_inference_plot_posterior
+++ b/bin/inference/pycbc_inference_plot_posterior
@@ -293,6 +293,7 @@ for (i, s) in enumerate(samples):
                     plot_marginal_lines=not opts.no_marginal_lines,
                     plot_maxl=opts.plot_maxl,
                     marginal_percentiles=opts.marginal_percentiles,
+                    marginal_title=not opts.no_marginal_titles,
                     plot_scatter=opts.plot_scatter,
                     zvals=zvals[i] if zvals is not None else None,
                     show_colorbar=opts.z_arg is not None,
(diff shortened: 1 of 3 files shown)


micamu (commented):
That's great! I assume you have tested all these new options and they work fine?
Since you are fiddling with the legend: I worry that placing it in the 1D/2D axis might be suboptimal when plotting 2D densities or if the posterior is railing against the edge where the legend is being placed. It might be worth adding an optional argument "legend-location" or similar, that takes in the matplotlib accepted location parameters (either coordinates or strings, or maybe just coordinates). That way one has the freedom to move the legend if the code's default placement does not work well.

cdcapano:
@micamu poke

cdcapano:
@micamu poke

ahnitz (approved):

PR #3714, opened 2021-05-21
python3 fixes + bugfix in pycbc_add_statmap

ahnitz:
@GarethDaviesGW This fixes a few python3 issues that come up, and also an indentation bug.


GarethDaviesGW (approved):
Looks simple enough, so long as it's tested

PR #3716, opened 2021-05-21
allow more general single stat in coinc statistic

ahnitz:
The expfitfgbg statistic was somewhat hardcoded in the way it chose a single detector statistic, now any can be used in its construction.

### DIFF
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -1159,7 +1159,7 @@ class PhaseTDExpFitStatistic(PhaseTDStatistic, ExpFitCombinedSNR):
         return s1
 
 
-class ExpFitSGBgRateStatistic(ExpFitStatistic):
+class ExpFitBgRateStatistic(ExpFitStatistic):
     """
     Detection statistic using an exponential falloff noise model.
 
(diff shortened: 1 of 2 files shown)


pycbc/events/stat.py
```diff
@@ -1546,7 +1543,8 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
         """
 
         # Safety against subclassing and not rethinking this
-        allowed_names = ['ExpFitSGFgBgNormStatistic']
+        allowed_names = ['ExpFitSGFgBgNormStatistic',
+                         'ExpFitSGPSDFgBgNormBBHStatistic']
```
GarethDaviesGW, line 1547: I'll have to take a bit more time to look through this, but this looks like PSD and BBH are mutually required, is that correct/desired?

ahnitz: It is not actually correct anymore, the names are somewhat misleading atm. Would you suggest renaming? I could do that.

ahnitz: i.e. you can use any single stat now there in either of the multi-det coinc classes

GarethDaviesGW: In that case removing both "SG" and "PSD" should make sense

ahnitz: OK, I can do that. I'll need to update the test config files as well. And you all should note that for your internal configs as well if  using these classes.

ahnitz: @GarethDaviesGW Ok, I've renamed everything by dropping the sg / psd parts everywhere.

GarethDaviesGW (approved):

PR #3718, opened 2021-05-21
Add LAL cosmology option to mchirp_area_improved.py

veronica-villa:
Since the cosmology model which is currently being use for LVC production results differs from the default cosmology of the `pycbc.cosmology.redshift` function, we need to introduce the option to use it for source probabilities estimation. 
```
pycbc_default_cosmology = astropy.cosmology.FlatLambdaCDM(H0=67.74, Om0=0.3075)

lal_default_cosmology = astropy.cosmology.FlatLambdaCDM(H0=67.90, Om0=0.3065)
```

### DIFF
--- a/pycbc/mchirp_area_improved.py
+++ b/pycbc/mchirp_area_improved.py
@@ -13,7 +13,7 @@ import numpy as np
 from pycbc.conversions import mass2_from_mchirp_mass1 as m2mcm1
 from scipy.integrate import quad
 from pycbc.cosmology import _redshift
-
+from astropy.cosmology import FlatLambdaCDM
 
 def insert_args(parser):
     mchirp_group = parser.add_argument_group("Arguments for estimating the "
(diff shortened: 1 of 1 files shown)


pycbc/mchirp_area_improved.py
```diff
@@ -67,8 +72,27 @@ def from_cli(args):
...
+            'mass_gap': args.src_class_mass_gap_separate,
+            'lal_cosmology': args.src_class_lal_cosmology}
+
+def redshift_estimation(distance, distance_std, lal_cosmology):
+    """Takes values of distance and its uncertainty and returns a
+       dictionary with estimates of the redshift and its uncertainty.
+       If the argument 'lal_cosmology' is True, it uses Planck15 cosmology
+       model as defined in lalsuite instead of the astropy default.
+    """
+    if lal_cosmology:
```
tdent, line 84: can you add a comment pointing to the line of code defining this in pesummary? 
I think it is this currently 
https://git.ligo.org/lscsoft/pesummary/-/blob/master/pesummary/gw/cosmology.py#L43
& you should mention the function name Planck15_lal_cosmology()

tdent (approved):

PR #3720, opened 2021-05-25
Revert "avoid astropy issues in last point release (#3689)"

titodalcanton:
This reverts commit 4022136a9c515d97d23a970bbcb349230e6d20c6.

As discussed at the bottom of PR #3689, ligo.skymap now includes
a workaround for the Astropy bug which originally prompted that PR,
so we can remove this restriction.


ahnitz:
@titodalcanton Have we already put in a minimum ligo.skymap version then?

titodalcanton:
No, and apparently an older version of ligo.skymap is still being picked for some reason.

titodalcanton:
I think we are starting to hit problems with Python 3.6. ligo.skymap [dropped it](https://git.ligo.org/lscsoft/ligo.skymap/-/blob/master/CHANGES.rst) at 0.5.1, and Astropy [dropped it](https://github.com/astropy/astropy/blob/main/CHANGES.rst#other-changes-and-additions-1) at 4.2. So the 3.6 test is falling back to the buggy Astropy 4.0.5, somehow decides to pick ligo.skymap 0.5.1, but not a newer version (maybe due to matplotlib… 🤯). Then the PyCBC Live test fails again due to the known bug. I wonder if the targets with higher Python versions would pick the right versions and succeed.

So… I am tempted to leave things as they are for now and put this on hold until we drop the 3.6 test target. What do you think @ahnitz?

ahnitz:
@titodalcanton Leave as is. It's not as if the version is pinned anyway, it just excludes the known broken ones.

PR #3744, opened 2021-07-01
simplify use of rvs

ahnitz:
If we have a cdfinv, then we should just use it for the rvs function rather than duplicating code. 

@cdcapano One thing I didn't touch was the qnm.py distribution however I don't understand some of the things it is doing. Why is it dealing with constraints and renormalization? The join distirbution should do both those things? What is going on there?

### DIFF
--- a/pycbc/distributions/angular.py
+++ b/pycbc/distributions/angular.py
@@ -254,6 +254,7 @@ class SinAngle(UniformAngle):
             numpy.log(self._dfunc(
                 numpy.array([kwargs[p] for p in self._params]))).sum()
 
+
 class CosAngle(SinAngle):
     r"""A cosine distribution. This is the same thing as a sine distribution,
     but with the domain shifted to `[-pi/2, pi/2]`. See SinAngle for more
--- a/pycbc/distributions/angular.py
+++ b/pycbc/distributions/angular.py
@@ -254,39 +254,6 @@ class SinAngle(UniformAngle):
             numpy.log(self._dfunc(
                 numpy.array([kwargs[p] for p in self._params]))).sum()
 
-
-    def rvs(self, size=1, param=None):
-        """Gives a set of random values drawn from this distribution.
-
-        Parameters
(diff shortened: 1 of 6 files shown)


cdcapano:
The UniformF0Tau distribution was meant for RD frequency and damping times in which you can exclude certain masses and spins (assuming Kerr), with the default excluding negative masses and naked singularities. The only way to do that is with constraints, since there's no simple analytic formula for the pdf. You could do it with the JointDistribution (and so write out the constraints in the config file), but if I remember right, this could lead to large constraint sections in the config file that were prone to typos. Hence the distribution. It's also faster this way since the constraint is hard coded rather than having to evaluate a string on each likelihood call.

ahnitz:
@cdcapano Poke

cdcapano (commented):
This looks ok, but could you double check that the CosAngle and SinAngle rvs functions still provide the right distribution of random draws when bounds are included? I think they should, but not certain looking at the cdfinv function (if not, this would be a bug in the cdfinv).

ahnitz:
@cdcapano Looks OK to me

```
from pycbc.distributions import SinAngle, CosAngle
import numpy, pylab

x = SinAngle(x=[0, numpy.pi])
vals = x.rvs(size=100000)['x']
_ = pylab.hist(vals, bins=100)

x = CosAngle(x=[-numpy.pi/2.0, numpy.pi/2.0])
vals = x.rvs(size=100000)['x']
_ = pylab.hist(vals, bins=100)

x = SinAngle(x=[0, numpy.pi/2.0])
vals = x.rvs(size=100000)['x']
_ = pylab.hist(vals, bins=100)

x = CosAngle(x=[-numpy.pi/2.0, 0])
vals = x.rvs(size=100000)['x']
_ = pylab.hist(vals, bins=100)
```

download (87)

cdcapano (approved):
Ok, thanks for making those plots. Looks good.

PR #3754, opened 2021-07-20
Added arguments to new multi detector gating option forgotten in #3752

jakeb245:
This adds some important arguments to make the gating method option work properly with multi-detector searches that I missed in #3752.


spxiwh (approved):
Great. Thanks!

PR #3758, opened 2021-07-28
Fix bug that can break dynesty parallelization

cdcapano:
Currently, `choose_pool` will incorrectly set `pool.size = 1` if you use python multiprocessing instead of MPI and you have `mpi4py` installed in your environment. This is a serious problem if you are using dynesty: `pool.size` is used by dynesty to determine how many parallel proposal points to generate (it sets the sampler's `queue_size`; see [this function](https://github.com/joshspeagle/dynesty/blob/master/py/dynesty/sampler.py#L340)). If pool.size is 1, it will only ever generate one proposal point at a time, rather than a batch. The result is that the pool is setup, but then only one process in that pool is only used, making it no faster than just running on a single core. This fixes this by appropriately setting `pool.size` if `use_mpi` is False.

This also ensures that `queue_size` and other pool-related sampler attributes are not included in dynesty's checkpoint. Without this, dynesty will continue to use whatever was used in the checkpoint, regardless of how many cores you are currently using. This can be a problem if, e.g., you start a run using a smaller set of cores, then increase it at a later point.

### DIFF
--- a/pycbc/inference/sampler/dynesty.py
+++ b/pycbc/inference/sampler/dynesty.py
@@ -162,7 +162,9 @@ class DynestySampler(BaseSampler):
                           'propose_point',
                           'update_proposal',
                           '_UPDATE', '_PROPOSE',
-                          'evolve_point']
+                          'evolve_point', 'use_pool', 'queue_size',
+                          'use_pool_ptform', 'use_pool_logl',
+                          'use_pool_evolve']
 
     def run(self):
         diff_niter = 1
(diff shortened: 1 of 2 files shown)


cdcapano:
Assigning @spxiwh since @ahnitz on vacation.

sum33it:
Looks good to me. I have also done a test with `nprocesses=10` and it gives improvement of at least factor of 2 compared to the installation without this fix..

sum33it (approved):

PR #3760, opened 2021-08-05
Avoid div by 0 error if magnitude is exactly 0

a-r-williamson:
Passing a spin vector that is exactly zero in cartesian components to `cartesian_to_spherical_polar` would cause a divide by zero error. This should probably just return zero components instead.

### DIFF
--- a/pycbc/coordinates.py
+++ b/pycbc/coordinates.py
@@ -83,7 +83,7 @@ def cartesian_to_spherical_polar(x, y, z):
         return numpy.arccos(z / rho) if rho else 0.0
     else:
         return numpy.arccos(numpy.divide(z, rho, out=numpy.ones_like(z),
-                                         where=rho!=0))
+                                         where=rho != 0))
 
 
 def cartesian_to_spherical(x, y, z):
--- a/pycbc/coordinates.py
+++ b/pycbc/coordinates.py
@@ -79,7 +79,8 @@ def cartesian_to_spherical_polar(x, y, z):
         The polar angle.
     """
     rho = cartesian_to_spherical_rho(x, y, z)
-    return numpy.arccos(z / rho) if rho else 0.0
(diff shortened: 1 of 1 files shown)


pycbc/coordinates.py
```diff
@@ -78,7 +78,8 @@ def cartesian_to_spherical_polar(x, y, z):
     theta : {numpy.array, float}
         The polar angle.
     """
-    return numpy.arccos(z / cartesian_to_spherical_rho(x, y, z))
+    rho = cartesian_to_spherical_rho(x, y, z)
+    return numpy.arccos(z / rho) if rho else 0.0
```
ahnitz, line 82: @a-r-williamson This doesn't look like it would work with vector import, right? 

@cdcapano Can verify, but I think where possibly we want to avoid assuming scalar or vector input on these kind of generic functions.

a-r-williamson: Yes I agree, I'll rethink this. Perhaps something like:
```
return numpy.arccos(np.divide(z, rho, out=np.zeros_like(a), where=rho!=0.0))
```

a-r-williamson:
The latest commit gives the following output:
```
>>> x = y = z = np.arange(2) * 0.2
>>> coords.cartesian_to_spherical(x, y, z)
(array([0.        , 0.34641016]), array([0.        , 0.78539816]), array([0.        , 0.95531662]))
>>> x = y = z = 0
>>> coords.cartesian_to_spherical(x, y, z)
(0.0, 0.0, 0.0)
>>> x = y = z = 0.2
>>> coords.cartesian_to_spherical(x, y, z)
(0.3464101615137755, 0.7853981633974483, 0.9553166181245093)
```

cdcapano (approved):
Looks good, thanks for making it vector friendly.

PR #3763, opened 2021-08-10
Fix analysis time

yi-fan-wang:
Try to relax the constraint that data from all detectors should have the same length. Work in progress

2021.11.08 Update: This module adds support to `gaussian_noise`, `marginalized_phase` and `marginalized_polarization` for analyzing different data segment lengths and/or sampling rates for different IFOs.

Future optimization: add support for more models, handle the case when some of the IFOs use the same settings (in that case I don't need to generate a waveform generator for each one of IFO

### DIFF
--- a/pycbc/inference/models/marginalized_gaussian_noise.py
+++ b/pycbc/inference/models/marginalized_gaussian_noise.py
@@ -213,13 +213,25 @@ class MarginalizedPolarization(BaseGaussianNoise):
             variable_params, data, low_frequency_cutoff, psds=psds,
             high_frequency_cutoff=high_frequency_cutoff, normalize=normalize,
             static_params=static_params, **kwargs)
-        # create the waveform generator
-        self.waveform_generator = create_waveform_generator(
-            self.variable_params, self.data,
-            waveform_transforms=self.waveform_transforms,
-            recalibration=self.recalibration,
-            generator_class=generator.FDomainDetFrameTwoPolGenerator,
-            gates=self.gates, **self.static_params)
+        # Determine if all data have the same sampling rate and segment length
+        if self.all_ifodata_same_rate_length == True:
+            # create a waveform generator for all ifos
+            self.waveform_generator = create_waveform_generator(
+                self.variable_params, self.data,
+                waveform_transforms=self.waveform_transforms,
+                recalibration=self.recalibration,
+                generator_class=generator.FDomainDetFrameTwoPolGenerator,
+                gates=self.gates, **self.static_params)
+        else:
+            # create a waveform generator for each ifo respestively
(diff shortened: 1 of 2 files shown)


ahnitz (changes requested):
@yi-fan-wang On first glance this seems largely OK. 

One thing to fix is that this doesn't affect the marginalized_phase model at the moment. I think the easiest way to do that would actually be to have the marginalized_phase class inherit from the gaussianNoise (rather than basegaussiannoise) and have it *not* redo the __init__. Does that make sense to you and if you look at what both __init__'s are doing, do you agree that would work as intended? The marginalized polarization needs a different init because it has to use a different waveform generator due to needing both polarizations returned, however. 

For future reference, this could be sped up slightly in the case where some of the detectors use the same setting, but we can leave that for a future PR.

pycbc/inference/models/marginalized_gaussian_noise.py
```diff
@@ -262,7 +274,12 @@ def _loglr(self):
         """
         params = self.current_params
         try:
-            wfs = self.waveform_generator.generate(**params)
+            if self.all_ifodata_same_rate_length:
```
ahnitz, line 277: Marginalized phase might also need this bit.

ahnitz:
@yi-fan-wang Besides for my comment, do your tests show that this is working as expected? If so, we should try to address the final issues and get this merged.

@sum33it Any other comments?

ahnitz (approved):
@yi-fan-wang Please merge if you are satisfied this is correct (i.e. works in your tests) and the unittests pass

yi-fan-wang:
@ahnitz I'll do another PE run to make sure the `marginalized_phase` is working as well and then I'll merge it.

pycbc/inference/models/marginalized_gaussian_noise.py
```diff
@@ -28,7 +28,7 @@
 from .gaussian_noise import (BaseGaussianNoise, create_waveform_generator)
```
ahnitz, line 28: @yi-fan-wang You forgot to import GaussianNoise when you made this change

ahnitz:
@yi-fan-wang Please check this over and fix the issues so this can be merged.

pycbc/inference/models/marginalized_gaussian_noise.py
yi-fan-wang: Added for marginalized phase

pycbc/inference/models/marginalized_gaussian_noise.py
yi-fan-wang: Fixed

Issue #3762, opened 2021-08-09
Infinite loop in ordering of waveform transforms


a-r-williamson:
When applying the rigid rotational transform being worked on in #3756 I hit an infinite loop in the `order_transforms` function. This happened in a case when I defined priors on the cartesian spin components, which are both input to and output by the transform, and this was the only transform being applied. As a result [this line](https://github.com/gwastro/pycbc/blob/59e2d627330d63bfc29f4aea70e0874f15b43059/pycbc/transforms.py#L2275) resolves as False and we get stuck.

I would be interested in hearing suggestions about how to overcome this. Is this a bug that could recur in other cases that we should fix? (There look to be other transforms that have parameters common between their inputs and outputs, [e.g.](https://github.com/gwastro/pycbc/blob/59e2d627330d63bfc29f4aea70e0874f15b43059/pycbc/transforms.py#L789-L791)). Or would it be preferable to change [our new transform](https://github.com/gwastro/pycbc/blob/a7ede7aaa0fe30e9d1e3e4887053725b873fa508/pycbc/transforms.py#L621) to avoid this?

cdcapano:
Interesting... I forgot we had transforms that can have shared input and output parameters (we really haven't used the other spin transforms; they were written as a part of a master's student's project, then never really put to use). The ordering function should be able to handle that. I think the easiest solution is to modify `order_transforms` to pull out new outputs rather than all outputs. In other words, change [line 2268](https://github.com/gwastro/pycbc/blob/59e2d627330d63bfc29f4aea70e0874f15b43059/pycbc/transforms.py#L2268) to:
```
outputs = set().union(*[set(t.outputs)-set(t.inputs) for t in transforms])
```
I think that should do it.

PR #3764, opened 2021-08-10
Some transforms have same params in inputs/outputs

a-r-williamson:
This change is to fix the issue in #3762.


cdcapano (approved):
Looks good

PR #3766, opened 2021-08-17
add functions to get rate from redshift  or distance from rate

ahnitz:
As opposed to setting a cosmological volume, it would be better to set a population model explicitly and be able to transform from the resulting rate to distance horizon, etc. In this way we could use explicitly merger rate models as priors for example. 

This adds a few functions in this area along with an example function which can be used to set a rate distribution (following the stellar-formation rate). 

The key function is 'rate_from_redshift' which takes in a redshift and a 'rate_desnity' function. The rate density should take in a redshift and return the rate of occurances of something (might be merger rate, need not be though) per unit volume. The function then can integrate this over the comoving volume out to this redshift. The cosmological time dilation is also taken into account in this integral, so the assumption is that the rate_density function should be in terms of local quantities

This function is then used to provide the additional helper functions such as converting to luminosity distance (important for PE / injections / waveform generation).

### DIFF
--- a/pycbc/cosmology.py
+++ b/pycbc/cosmology.py
@@ -31,7 +31,7 @@ Any other distance measure is explicitly named; e.g., ``comoving_distance``.
 
 import logging
 import numpy
-from scipy import interpolate
+from scipy import interpolate, integrate
 import astropy.cosmology
 from astropy import units
 from astropy.cosmology.core import CosmologyError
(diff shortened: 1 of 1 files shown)


cdcapano (approved):
Looks good. Thanks for adding this. Can you just address the code climate issues before merging (just need to add some blank lines)?

PR #3768, opened 2021-08-24
prep for release

GarethDaviesGW:
Following instructions of https://pycbc.org/pycbc/latest/html/release.html

Do we still need the pycbc-config release?


ahnitz (approved):

GarethDaviesGW:
Any ideas on the failed test? just try again?

ahnitz:
@GarethDaviesGW Vaguely looks like some connection / http issues, so yes, I'd just restart. If that still fails, we can look into more detail to see if something is up.

PR #3771, opened 2021-08-25
add test of pycbc tutorials to CI

ahnitz:
This does the same test that the pycbc tutorials run themselves to ensure that any functionality used in the tutorials doesn't cause breakage there.


cdcapano (approved):
This is great!

PR #3713, opened 2021-05-19
ringdown update

JulianWesterweck:
Adds ability to use arbitrary polarisations for the ringdown modes.
Adds option to use spheroidal harmonics instead of spherical harmonics.
Includes -m modes.

PR #3712, opened 2021-05-19
Add gated Gaussian noise model, using inpainting for the gated times.

JulianWesterweck:
Co-authored-by: Collin Capano <collin.capano@aei.mpg.de>
Co-authored-by: Shilpa Kastha <shilpa.kastha@aei.mpg.de>

PR #3777, opened 2021-08-27
Add gated gaussian model with marginalized polarization

cdcapano:
Adds a version of the `GatedGaussian` that numerically marginalizes over polarization. Requires #3712 and #3713.

### DIFF
--- a/pycbc/inference/models/gated_gaussian_noise.py
+++ b/pycbc/inference/models/gated_gaussian_noise.py
@@ -20,60 +20,45 @@ method to fill the removed part such that it does not enter the likelihood.
 
 import logging
 import numpy
+from abc import abstractmethod
+from scipy import special
 
 from pycbc.waveform import (NoWaveformError, FailedWaveformError)
 from pycbc.types import FrequencySeries
 from pycbc.detector import Detector
 from pycbc.pnutils import hybrid_meco_frequency
 from pycbc.waveform.utils import time_from_frequencyseries
+from pycbc.waveform import generator
 from .gaussian_noise import (BaseGaussianNoise, create_waveform_generator)
 from .base_data import BaseDataModel
 from .data_utils import fd_data_from_strain_dict
 
 
-class GatedGaussianNoise(BaseGaussianNoise):
-    r"""Model that applies a time domain gate, assuming stationary Gaussian
-    noise.
+class BaseGatedGaussian(BaseGaussianNoise):
+    r"""Base model for gated gaussian.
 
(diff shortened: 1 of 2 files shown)


pycbc/inference/models/gated_gaussian_noise.py
```diff
@@ -207,24 +316,25 @@ def get_gate_times_hmeco(self):
...
         except NoWaveformError:
             return self._nowaveform_logl()
         except FailedWaveformError as e:
             if self.ignore_failed_waveforms:
                 return self._nowaveform_logl()
             raise e
-
+        # gate input for ringdown analysis which consideres a start time
+        # and an end time
+        dgate = params['gate_window']
```
ahnitz, line 330: This is definitely a bug, no? Youa re reading from params before it is generated on the next line. I suspect this hasn't been run in exactly this state. Perhaps some patches missing? @cdcapano

pycbc/inference/models/gated_gaussian_noise.py
```diff
@@ -154,6 +141,128 @@ def _loglr(self):
...
+
+        Returns
+        -------
+        dict :
+            Dictionary of detector names -> FrequencySeries.
+        """
+        wfs = self.get_waveforms()
+        out = {}
+        for det, h in wfs.items():
+            invpsd = self._invpsds[det]
```
ahnitz, line 217: CC is reporting this as unused, and it certainly looks like it is to me.

ahnitz (changes requested):
See the two comments (noted from the CC report as well as potential problems)

ahnitz:
@micamu Since @cdcapano is on leave, and after you have a chance to look at it, I would be OK to approve if someone can double check they can run this model and we manually apply the fixes to the two obvious minor errors.

pycbc/inference/models/gated_gaussian_noise.py
cdcapano: Yeah, that's a bug. I switched the order of the lines to fix it. We haven't actually been using this function, which is why I didn't catch it.

pycbc/inference/models/gated_gaussian_noise.py
cdcapano: Fixed.

cdcapano:
@ahnitz I addressed the two issues you caught, plus a couple other that code climate was complaining about.

ahnitz (approved):

ahnitz:
@cdcapano I've appoved this. Note the github action failures are due to a problem with dqsegdb, not with this PR.

PR #3787, opened 2021-09-04
fix to ids for HDF inj in injfilterrejector, allow multiple use

ahnitz:
This patch addresses a couple issues, one is a bug. 

1) I don't think the ids were used correclty (probably my fault originall) for injfliterrejector when comparing to the precalculated injection short version of the waveform. This makes it more explicit. 

2) This enables one to use the injfilterrejector multiple times (i.e. with multiple calls to the injection code) although still only if the same injection file is used.

### DIFF
--- a/bin/pycbc_inspiral
+++ b/bin/pycbc_inspiral
@@ -244,7 +244,7 @@ def template_triggers(t_num):
         # determine whether
         # to filter this template/segment if injections are present.
         if not inj_filter_rejector.template_segment_checker(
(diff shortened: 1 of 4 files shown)


ahnitz:
injfilter rejector is one of the ares that could be simplified a lot if we could drop xml support.

spxiwh (approved):
Looks good, thanks. I don't see a reason to continue XML support here, so feel free to strip it out as needed (in a separate PR) .... Perhaps can be added to tasks for hackathon?

PR #3788, opened 2021-09-06
disable dqsegdb to fix test suite

ahnitz:
The tests are broken due to dqsegdb being unable to install. See https://github.com/ligovirgo/dqsegdb/issues/79. 

We can re-enable whenever the next release is that is able to fix this issue.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ dynesty
 epsie>=0.6
 
 # For LDG service access
-dqsegdb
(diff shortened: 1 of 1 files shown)


ahnitz:
@spxiwh Ok, I changed this to use a fixed fork of dqsegdb. This should be reverted on the next release of dqsegdb if something like https://github.com/ligovirgo/dqsegdb/pull/80

spxiwh (approved):

PR #3795, opened 2021-09-12
Highpass waveforms in gated gaussian if the data were highpassed

cdcapano:
If the data were highpassed for the gated gaussian model, then you can get spurious large likelihoods for time domain waveforms if they are not also highpassed. This is due to excess large low frequency power in the waveform that's not downweighted by the PSD. This fixes that by highpass waveforms if the `strain-high-pass` is set the data section of the config file. For example, 220+221 ringdown waveform [without highpass filtering](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/projects/gated_gaussian/gw190521/test_model_updates/fdomain-220_221-nohighpass.png); same [with highpass filtering](https://www.atlas.aei.uni-hannover.de/~work-cdcapano/projects/gated_gaussian/gw190521/test_model_updates/fdomain-220_221-withhighpass.png). The former gives a log likelihood ratio of 327 (much too large for the parameters and the injection considered) while the latter gives -62.

@ahnitz Right now I'm using the same IIR highpass filter that is used by the strain module. I'm not sure if it'd better to use the FIR filter instead, or just do something directly in the frequency domain. Timing this, it's 25% slower with the highpass for ringdown waveforms with a 2s gate (~300ms without highpassing, ~400ms with). Let me know if you think that'd be better.

### DIFF
--- a/pycbc/inference/models/gated_gaussian_noise.py
+++ b/pycbc/inference/models/gated_gaussian_noise.py
@@ -29,6 +29,7 @@ from pycbc.detector import Detector
 from pycbc.pnutils import hybrid_meco_frequency
 from pycbc.waveform.utils import time_from_frequencyseries
 from pycbc.waveform import generator
+from pycbc.filter import highpass
 from .gaussian_noise import (BaseGaussianNoise, create_waveform_generator)
 from .base_data import BaseDataModel
 from .data_utils import fd_data_from_strain_dict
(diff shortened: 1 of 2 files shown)


ahnitz:
@cdcapano No objections here, but one questions. Does the likelihood not have a low frequency cutoff as well? Or is that taken to be a very low value (i.e. fixed to the same as used to generate the gate -> 0? ).

ahnitz (approved):

PR #3796, opened 2021-09-14
Solve mchirp_area.py module duplication

veronica-villa:
Due to code repetition in the existing `mchirp_area.py` module, we introduced some functions to make the code more readable in a second module called `mchirp_area_improved.py`. The changes were made in the PR #3691, #3718 and #3721. After checking that both modules produce the same results (a review of the method and code can be seen [here](https://wiki.ligo.org/CBC/Searches/PyCBC/PyCBCLiveSourceProbabilities)), the old one can be deleted since is no longer needed.

Also, the default estimation coefficients that were hard-coded in the old version of the module were outdated. The current estimation coefficients were also reviewed, and we decided to remove the defaults to ensure the latest values are passed as arguments.


tdent (approved):

tdent:
@veronica-villa since we got no further comments or requests, I think you can 'squash and merge'

PR #3799, opened 2021-09-15
added offline FGMC modules

tdent:


ahnitz:
@tdent A few quick notes. 

1) I think this looks like the FGMC code that had floated around a bit. Can you make sure the copyright is correct is that the proper authors get credit (even if it is just you in case you had reimplemented it). 

2) Can you put in the population folder? I'd like to reorganize a couple things into there which might make sense and keep things a bit more structured. Separately I was thinking maybe to move the rates (essentially utilties) module and maybe also sensitivity.py. 

3) The code could do with some minor cleanup. Maybe pass through 'black' as a first step to handle the easy style stuff. Then I'd ask if all the plotting stuff makes sense in the core module or if that could be excluded?

tdent:
ah, I had meant to put the files into pycbc/population but somehow that didn't get done before

tdent:
The plotting is designed (if we can call it 'designed' at all) to be run as a stand-alone command line script, so at present not part of a pipeline. Indeed a fair amount of work would be needed to convert it to standard pycbc 'pipeline format'.

tdent:
Concerning the copyright request, the current location of the source of some of the 'old FGMC code' is noted (though it may not be non-LVC-visible) - it didn't have an original copyright or authorship statement, though I could ask the 'original author' what they would prefer. 
Is there a specific recommendation for copyright statements?

tdent:
Ping on this .. I think the remaining improvements that can be done without too much effort are splitting off the plotting functions from fgmc_laguerre and adding 'more formal' copyright if wanted

pycbc/population/__init__.py
```diff
@@ -1,2 +1,4 @@
 from pycbc.population.rates_functions import *
 from pycbc.population.scale_injections import *
+from pycbc.population.fgmc_functions import *
```
ahnitz, line 3: It would be good if you do this to explicitly have an __all__ in the modules you import from to control what actually gets up-imported (instead of *everything*)

ahnitz:
@tdent Yes, it would be good to add the copyright. Also, yes, I would recommend splitting off the plotting and summary printing parts of the module. If they are used in a specific executable, perhaps that's where they'd make sense in say a follow-up PR?

tdent:
Anything else on this?  Next to be added will be an executable and then I think it will make sense to switch to logging for print/debug statements etc

ahnitz:
@tdent I think this is mostly OK, but can you take a pass through the CC and fix any issue where possible?

pycbc/population/__init__.py
```diff
@@ -1,2 +1,5 @@
...
+from pycbc.population.fgmc_laguerre import *
+from pycbc.population.fgmc_plots import *
```
ahnitz, line 5: I might skip this import here, as importing matplotlib things at the top level can sometimes cause unforeseen problems (i.e. conflicts with other matplotlib settings that depend on who calls first)

ahnitz (approved):
Try to reduce the CC issues and see the one import comment. After that please merge at your convenience.

pycbc/population/__init__.py
tdent: Sounds reasonable.

tdent:
I have already spent quite a lot of time (>hour) on CC complaints. 
Many of the remaining ones are lines slightly longer than 80. How important are these to 'fix'? 
For the remaining ones, are any particularly important?

PR #3803, opened 2021-09-21
fix mpi for ultranest  

ahnitz:
and maybe pymultinest?


spxiwh (approved):

ahnitz:
@spxiwh While I could test that ultranest worked again with this patch, can you verify if this also worked for multinest?

PR #3804, opened 2021-09-22
Addition of new data-quality ranking statistic

dethodav:
This pull request add in support for a new ranking statistic that incorporates arbitrary data quality information into the PyCBC noise model. This is done by adding a time-dependent term into the calculated log-noise rate. This allows previous data quality products (e.g. data quality flags) to be used as part of the ranking statistic in addition to supporting new types of data quality products. 

This PR only includes commits that add support for the statistic itself. Additional code to support this ranking statistic in the coinc search workflow will be added in a separate PR. 

A summary of the main changes and the impacted files are as follows: 

- Edits to existing files to support new ranking statistic
  - `pycbc/events/stat.py`
    - The vast majority of edits are to include the new ranking statistic class. One minor edit is made to a base ranking statistic class function to allow the new ranking statistic to use
  - `pycbc/events/coinc.py`
    - The bank binning code has been edited to work in python 3 and to support additional binning methods
- New executables
  - `bin/all_sky_search/pycbc_calculate_dq`
    - Executable that ingests and processes data quality information for use by PyCBC. DQ data is, in general, downsampled and then processed with a maximization or blrms function over each time step. This executable is for DQ products that are time series. 
  - `bin/all_sky_search/pycbc_calculate_dqflag`
    - This executable has the same purpose as `pycbc_calculate_dq`, but ingests segment data such as DQ flags
  - `bin/all_sky_search/pycbc_bin_trigger_rates_dq`
    - Executable that calculates the relative trigger rate during each DQ bin. 
  - `bin/all_sky_search/pycbc_rerank_dq`
    - This executable uses the information from `pycbc_bin_trigger_rates_dq` to create a time series containing the reranking information that is included as part of the ranking statistic
- Addition of lowpass filter to strain cli
  - `pycbc/filter/resample.py`
    - Add a lowpass filter function based on `lal` lowpass method (same as the highpass method)
  - `pycbc/strain/strain.py`
    - Add support for lowpass method in `strain.from_cli`

This PR does not introduce any known issues with back-compatibility. 

A description of this new ranking statistic can be found here: https://dcc.ligo.org/LIGO-T2100367 (currently requires LVK credentials)


dethodav:
#3805 is a related PR that adds support for this ranking statistic in the coinc search workflow that should be merged after this PR.

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -0,0 +1,164 @@
...
+import logging
+import argparse
+import pycbc
+import pycbc.events
+from pycbc.events import stat as pystat
+import numpy as np
+import h5py as h5
+import matplotlib
+matplotlib.use('Agg')
+import matplotlib.pyplot as plt
```
titodalcanton, line 13: Watch out for unused imports (applies to the entire PR).

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -0,0 +1,164 @@
...
+                                                        data)
+    del data
+    locs_names = [b.split(':')[0] for b in args.background_bins]
+else:
+    locs_dict = {'all_bin': np.arange(0, len(bank['mass1'][:]), 1)}
+    locs_names = ['all_bin']
+
+logging.info('Reading trigger file...')
+ifo = args.ifo
+trig_file = h5.File(args.trig_file,'r')
```
titodalcanton, line 60: Try to use a `with h5.File(…) as trig_file:` block here, and remove the explicit `close()` call below. Also applies to similar cases further down.

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -0,0 +1,164 @@
...
+        del stat
+trig_file.close()
+
+trig_times_int = trig_times.astype('int')
+
+dq_times = np.array([])
+dq_logl = np.array([])
+
+for filename in args.dq_file:
+    logging.info('Reading DQ file %s...'%filename)
```
titodalcanton, line 84: This should be
```python
logging.info('Reading DQ file %s...', filename)
```
Also applies to similar cases further down.

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -0,0 +1,164 @@
...
+    if mean_rate > 0.:
+        rates = rates / mean_rate
+    
+    logging.info('Writing rates to output file %s...'%args.output_file)
+    grp = f.create_group(bin_name)
+    grp['rates']=rates
+    grp['locs']=locs_dict[bin_name]
+ 
+f.attrs['names'] = locs_names
+f.close()
```
titodalcanton, line 164: Finish by logging "Done" here, as you did for other scripts. When looking at logs to debug problems, I find it very helpful when a process reports explicitly that it has completed successfully.

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env python
+""" Ingest and process DQ data
+"""
+import logging, argparse, numpy, h5py, multiprocessing, time, copy
+from six.moves import (range, zip_longest)
```
titodalcanton, line 5: Do we still need `six` now that we moved to Python 3?

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,100 @@
...
+import pycbc, pycbc.strain
+from pycbc.version import git_verbose_msg as version
+from pycbc.fft.fftw import set_measure_level
+from pycbc.workflow import resolve_td_option
+from ligo.segments import segmentlist, segment
+set_measure_level(0)
```
titodalcanton, line 11: (Why) Is this needed?

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,100 @@
...
+    argstmp.gps_start_time = int(seg[0])
+    argstmp.gps_end_time = int(seg[1])
+    tmp_segment = segment([argstmp.gps_start_time, argstmp.gps_end_time])
+    argstmp.channel_name = resolve_td_option(args.channel_name, tmp_segment)
+
+    try:
+        gwstrain = pycbc.strain.from_cli(argstmp)
+    except RuntimeError:
+        time.sleep(10)
+        return get_dq(seg)
```
titodalcanton, line 54: I think this `try` block warrants a comment to explain why it is needed.

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,100 @@
...
+                                                    dq_step_size)])
+    elif dq_method=="blrms":
+        # Calculate the blrms over given interval
+        val_dq = numpy.array([numpy.mean(gwstrain.numpy()
+                                          [int(n*dq_step_size):
+                                           int((n+1)*dq_step_size)]**2)**0.5
+                               for n in numpy.arange(0, len(gwstrain.numpy()) /
+                                                     dq_step_size)])
+    else:
+        logging.info("%s is not an available dq method", dq_method)
```
titodalcanton, line 79: What happens then, is it ok to just fall through? Or should an exception be raised here?

bin/all_sky_search/pycbc_rerank_dq
```diff
@@ -0,0 +1,124 @@
...
+            dq_percentiles = numpy.percentile(log_like_vals,percentiles)[1:]
+
+            rates_bin = numpy.array([n_bins - \
+                              len(dq_percentiles[dq_percentiles >= dq_ll]) \
+                              for dq_ll in log_like_vals])
+            log_like_dict[bin_name] = log_rates[rates_bin.astype('int')]
+
+        f.close()
+
+    else: logging.error('Could not find binned trigger rates file')
```
titodalcanton, line 91: Should this be an exception instead? Is it expected that the user (or workflow) might provide an invalid file path here?

pycbc/filter/resample.py
```diff
@@ -292,6 +295,50 @@ def highpass(timeseries, frequency, filter_order=8, attenuation=0.1):
     return TimeSeries(lal_data.data.data, delta_t = lal_data.deltaT,
                       dtype=timeseries.dtype, epoch=timeseries._epoch)
 
+def lowpass(timeseries, frequency, filter_order=8, attenuation=0.1):
```
titodalcanton, line 298: Maybe this change should be split to a different PR? It seems fairly independent from the rest of the PR.

pycbc/strain/strain.py
```diff
@@ -368,6 +368,10 @@ def from_cli(opt, dyn_range_fac=1, precision='single',
         logging.info("Highpass Filtering")
         strain = highpass(strain, frequency=opt.strain_high_pass)
 
+    if opt.strain_low_pass:
+        logging.info("Lowpass Filtering")
+        strain = lowpass(strain, frequency=opt.strain_low_pass)
```
titodalcanton, line 373: Is this also done when simulating fake noise? Should it be?

titodalcanton (changes requested):
I need more time to fully understand this, but here are a few simple and minor things to consider in the mean time.

pycbc/strain/strain.py
dethodav: I believe this does apply when generating fake noise - I'll check that explicitly

bin/all_sky_search/pycbc_bin_trigger_rates_dq
dethodav: Done

bin/all_sky_search/pycbc_calculate_dq
dethodav: This has been removed

bin/all_sky_search/pycbc_bin_trigger_rates_dq
dethodav: Done

bin/all_sky_search/pycbc_bin_trigger_rates_dq
dethodav: Done

pycbc/filter/resample.py
dethodav: I think that makes sense - I'll make a separate PR

bin/all_sky_search/pycbc_rerank_dq
dethodav: This has been made an exception

bin/all_sky_search/pycbc_calculate_dq
dethodav: This has been made an exception

bin/all_sky_search/pycbc_calculate_dq
dethodav: This has been removed

bin/all_sky_search/pycbc_calculate_dq
dethodav: This is copied over from `pycbc_calcuclate_psd`:
https://github.com/gwastro/pycbc/blob/90503e2bf6ec60715b3b04139d7c77f4d6e05046/bin/all_sky_search/pycbc_calculate_psd#L53-L59
I'm happy to also note the reason here. But more generally, do you think this try statement is needed?

dethodav:
@titodalcanton Thanks for taking a look so quickly! I've gone through and addressed most of your minor comments above

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,98 @@
...
+    times_dq = numpy.arange(gwstrain.start_time,gwstrain.end_time,
+                            1./dq_sample_rate)
+
+    return val_dq, times_dq
+
+ifo = args.channel_name[0:2]
+
+dq = get_dq((args.gps_start_time,args.gps_end_time))
+
+f = h5py.File(args.output_file, 'w')
```
titodalcanton, line 89: Use a `with` block here too.

bin/all_sky_search/pycbc_calculate_dqflag
```diff
@@ -0,0 +1,74 @@
...
+    flag = flag_list[1]
+
+# Get science segments from file
+# to do: make this a configurable option
+segs = SegFile.from_segment_xml(args.science_segments
+                                 ).segment_dict["%s:SCIENCE_AVAILABLE" % ifo]
+# Create data quality time series
+dq = make_dq_ts(segs, args.dq_segments, ifo, flag)
+
+f = h5py.File(args.output_file, 'w')
```
titodalcanton, line 66: Use a `with` block here too.

pycbc/filter/resample.py
```diff
@@ -177,6 +177,9 @@ def resample_to_delta_t(timeseries, delta_t, method='butterworth'):
 
 _highpass_func = {numpy.dtype('float32'): lal.HighPassREAL4TimeSeries,
                  numpy.dtype('float64'): lal.HighPassREAL8TimeSeries}
+_lowpass_func = {numpy.dtype('float32'): lal.LowPassREAL4TimeSeries,
+                 numpy.dtype('float64'): lal.LowPassREAL8TimeSeries}
+
```
titodalcanton, line 182: I suggest you remove this change now that it has been split to its own PR, to avoid conflicts when rebasing.

titodalcanton (changes requested):

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,98 @@
...
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument('--version', action='version', version=version)
+parser.add_argument('--verbose', action="store_true")
+parser.add_argument("--dq-sample-rate", default=1, type=int,
+                    help="Sample rate to calculate dq timeseries, "
+                         "in Hz [default=1Hz]")
+parser.add_argument("--dq-sample-method", required=True,
+                    help="Method used to calculate dq timeseries. "
+                         "Options are 'max' or 'blrms'")
```
titodalcanton, line 18: Use the `choices` kwarg to properly limit the possible values.

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env python
+""" Ingest and process DQ data
```
titodalcanton, line 2: Can this docstring be made a bit more descriptive? What kind of DQ data does this script take as input, and what does it do when it "processes" it?

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,98 @@
...
+    logging.info('Getting dq values for %.1f-%.1f (%.1f s)', seg[0],
+                 seg[1], abs(seg[1]-seg[0]))
+    argstmp = copy.deepcopy(args)
+    argstmp.gps_start_time = int(seg[0])
+    argstmp.gps_end_time = int(seg[1])
+    tmp_segment = segment([argstmp.gps_start_time, argstmp.gps_end_time])
+    argstmp.channel_name = resolve_td_option(args.channel_name, tmp_segment)
+
+    try:
+        gwstrain = pycbc.strain.from_cli(argstmp)
```
titodalcanton, line 48: If I understand correctly the purpose of this script, the idea is to use the `pycbc.strain.from_cli()` infrastructure to actually read a DQ time series (as opposed to h(t) strain data). In that case, `gwstrain` is probably not the best variable name here. Would `dq_data` be more appropriate?

bin/all_sky_search/pycbc_calculate_dq
titodalcanton: Hmm, if we think a retry mechanism is beneficial, why not put it in `pycbc.strain.from_cli()` itself? Otherwise we need to duplicate this `try`-`except` code every time we want to read some data.

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,98 @@
...
+    elif dq_method=="blrms":
+        # Calculate the blrms over given interval
+        val_dq = numpy.array([numpy.mean(gwstrain.numpy()
+                                          [int(n*dq_step_size):
+                                           int((n+1)*dq_step_size)]**2)**0.5
+                               for n in numpy.arange(0, len(gwstrain.numpy()) /
+                                                     dq_step_size)])
+    else:
+        raise TypeError("%s is not an available dq method. " 
+                        "Options are 'max' or 'blrms'", dq_method)
```
titodalcanton, line 77: If you use the `choices` kwarg when setting up the `--dq-sample-method` option, you should never end up in this situation, so this `else` block can be removed.

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,98 @@
...
+
+    # figure out list of times
+    times_dq = numpy.arange(gwstrain.start_time,gwstrain.end_time,
+                            1./dq_sample_rate)
+
+    return val_dq, times_dq
```
titodalcanton, line 83: Why not returning a `TimeSeries` object here, as opposed to two separate arrays for value and time?

titodalcanton (changes requested):

bin/all_sky_search/pycbc_calculate_dq
```diff
@@ -0,0 +1,98 @@
...
+
+dq = get_dq((args.gps_start_time,args.gps_end_time))
+
+f = h5py.File(args.output_file, 'w')
+start, end = [args.gps_start_time], [args.gps_end_time]
+
+f[ifo + '/start_time'] = numpy.array(start, dtype=numpy.uint32)
+f[ifo + '/end_time'] = numpy.array(end, dtype=numpy.uint32)
+f[ifo + '/dq_val'] = numpy.array(dq[0], dtype=numpy.float32)
+f[ifo + '/times'] = numpy.array(dq[1], dtype=numpy.uint32)
```
titodalcanton, line 95: Looks like you basically just want to save a time series to an HDF file here, right? In this case it would be sufficient to make `get_dq()` return a `TimeSeries` object as I wrote above (making sure to give it the right metadata, in particular the start time) and then just do
```python
dq = get_dq((args.gps_start_time,args.gps_end_time))
dq.save(args.output_file)
```
(this would also allow the user to save the data in ASCII and Numpy formats).

pycbc/filter/resample.py
dethodav: These edits have now been removed

bin/all_sky_search/pycbc_rerank_dq
```diff
@@ -0,0 +1,120 @@
...
+len_log_like = len(log_like)
+
+times = times[log_like_ranking]
+log_like_vals = numpy.array(log_like[log_like_ranking])
+
+log_like_dict = {}
+percent_dict = {}
+
+if not args.rate_file or args.bank_file:
+    raise TypeError('Must supply either a bank or dq rates file')
```
titodalcanton, line 59: This check looks a bit strange. Is the idea that you need one of the two options, but not both? If so, try using a [mutually exclusive option group](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) in your argument parser. Then this explicit check should not be needed.

bin/all_sky_search/pycbc_calculate_dq
dethodav: I've now switched to using `TimeSeries` objects throughout (when possible)

bin/all_sky_search/pycbc_rerank_dq
dethodav: Yes, basically if a DQ rate file is provided, then information from that file is used. Otherwise the bank file is used to apply an analytic model. I've implemented your suggestion

bin/all_sky_search/pycbc_calculate_dq
dethodav: @ahnitz originally wrote this part of `pycbc_calculate_psd` - perhaps this try statement is no longer necessary in general? I'm open to moving/deleting this part of the code.

titodalcanton: I suggest removing the try-except block from here. If we run into I/O issues that warrant putting it back, then we look for a solution at the level of the `pycbc.strain` module.

bin/all_sky_search/pycbc_rerank_dq
```diff
@@ -0,0 +1,124 @@
...
+else:
+    # Use analytic model from Godwin et al. arXiv:2010.15282 
+    bin_names = ['all_bin']
+    with h5py.File(args.bank_file, 'r') as bank:
+        locs_dict = {'all_bin': numpy.arange(0, len(bank['mass1'][:]), 1)}
+
+    log_like = numpy.arange(len(log_like))*100./len(log_like)
+
+    p_min = int(len_log_like/2.)
+    p_max = int(len_log_like/(1.+numpy.e**-args.max_stat))
```
titodalcanton, line 101: Unless there is a numerical reason to prefer the explicit power, I suggest using `numpy.exp()` here.

bin/all_sky_search/pycbc_calculate_dqflag
```diff
@@ -0,0 +1,70 @@
...
+pycbc.init_logging(args.verbose)
+
+
+def make_dq_ts(dq_segs, start, end, ifo, flag):
+    """ Create a data quality timeseries
+    """
+    logging.info('Creating data quality timeseries for flag %s', 
+                 flag)
+    # Keep just times which belong to science_segments
+    dq_times = numpy.arange(float(start), float(end)).astype(int)
```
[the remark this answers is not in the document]
dethodav: I've switched these to just ints

bin/all_sky_search/pycbc_calculate_dq
dethodav: I've now removed this try block. I agree that makes sense in the short term.

dethodav:
@titodalcanton Are there an additional checks/changes needed before this can be merged? There seems to be a couple of suggestions from code climate, but I'm not sure if these need to be addressed.

titodalcanton:
@dethodav I think I am close to approving this, modulo two things:
* Please rebase on today's master, it should hopefully clear most remanining codeclimate issues. I will not hold this up for the issues related to code complexity.
* I have not had a chance to carefully go through your code and see if I have bigger-picture concerns yet. Can I ask for a couple more days to clear higher priority stuff?

dethodav:
@titodalcanton No worries - I just wanted to make sure you weren't waiting on us for anything. Let me know if you have any more suggestions/questions!

dethodav:
@titodalcanton I think I've now officially implemented all of your suggestions. Thanks again for all of these helpful comments!

PR #3807, opened 2021-09-23
Remove more pre-Python-3.7 stuff

titodalcanton:
Remove more pre-Python-3.7 stuff, in particular `from six.moves import range`. Also removes a few unused imports.

Edit: it also re-enables the tests for PyCBC Live and `pycbc_make_skymap`, which were accidentally disabled in #3719.

### DIFF
--- a/bin/all_sky_search/pycbc_calculate_psd
+++ b/bin/all_sky_search/pycbc_calculate_psd
@@ -2,7 +2,7 @@
 """ Calculate psd estimates for analysis segments
 """
 import logging, argparse, numpy, h5py, multiprocessing, time, copy
-from six.moves import (range, zip_longest)
+from six.moves import zip_longest
 import pycbc, pycbc.psd, pycbc.strain, pycbc.events
 from pycbc.version import git_verbose_msg as version
 from pycbc.fft.fftw import set_measure_level
--- a/bin/bank/pycbc_aligned_bank_cat
+++ b/bin/bank/pycbc_aligned_bank_cat
@@ -28,7 +28,6 @@ import argparse
 import numpy
 import pycbc.version
 import h5py
-from six.moves import range
 from glue.ligolw import ligolw
 from glue.ligolw import lsctables
 from glue.ligolw import utils
--- a/bin/bank/pycbc_bank_verification
+++ b/bin/bank/pycbc_bank_verification
@@ -37,7 +37,6 @@ import copy
 import logging
 import numpy
 import h5py
-from six.moves import range
 from glue.ligolw import ligolw, table, lsctables, utils as ligolw_utils
 from pycbc import tmpltbank, psd, strain, pnutils
 import matplotlib
--- a/bin/bank/pycbc_geom_aligned_2dstack
+++ b/bin/bank/pycbc_geom_aligned_2dstack
@@ -32,7 +32,6 @@ import copy
 import numpy
 import logging
 import h5py
-from six.moves import range
 import pycbc.tmpltbank
 import pycbc.version
 from pycbc import pnutils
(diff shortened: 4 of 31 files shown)


titodalcanton:
Apparently the PyCBC Live test is failing, but it does not seem related to what I changed here. It seems to arise from a missing command line option related to source classification.

```
  File "/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/runner/work/pycbc/pycbc/.tox/py-search/lib/python3.9/site-packages/mpi4py/__main__.py", line 7, in <module>
    main()
  File "/home/runner/work/pycbc/pycbc/.tox/py-search/lib/python3.9/site-packages/mpi4py/run.py", line 198, in main
    run_command_line(args)
  File "/home/runner/work/pycbc/pycbc/.tox/py-search/lib/python3.9/site-packages/mpi4py/run.py", line 47, in run_command_line
    run_path(sys.argv[0], run_name='__main__')
  File "/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/runpy.py", line 268, in run_path
    return _run_module_code(code, init_globals, run_name,
  File "/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/runpy.py", line 97, in _run_module_code
    _run_code(code, mod_globals, init_globals,
  File "/opt/hostedtoolcache/Python/3.9.7/x64/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/runner/work/pycbc/pycbc/.tox/py-search/bin/pycbc_live", line 676, in <module>
    mc_area_args=mchirp_area.from_cli(args))
  File "/home/runner/work/pycbc/pycbc/.tox/py-search/lib/python3.9/site-packages/pycbc/mchirp_area.py", line 72, in from_cli
    'b0': args.src_class_lum_distance_to_delta[0],
TypeError: 'NoneType' object is not subscriptable
```

@veronica-villa any idea?

veronica-villa:
Hi @titodalcanton, I think this is because we removed the default value for the `--src-class-lum-distance-to-delta` argument that used to be `default=[-0.449, -0.342]` to be sure that the most recent values were passed as arguments. Should we change this to `default=[]` to ensure that is not a 'NoneType' object?

titodalcanton:
Here is what I think happened:
* A few months ago, #3719 effectively disabled the PyCBC Live test due to a syntax error in `pycbc_test_suite.sh`. This was my fault, sorry!
* Unfortunately, the syntax error did not cause the test to fail (Bash happily returns 0 after a syntax error!) so nobody noticed.
* In #3796, we effectively made `--src-class-lum-distance-to-delta` a required option for `pycbc_live`. However, the example script was not updated to use the option. This broke the test, but we did not notice.
* This PR fixed the syntax error in `pycbc_test_suite.sh`, so we only see the failure now.

@veronica-villa I suggest you open a PR to make `--src-class-lum-distance-to-delta` a required option, and add it to `/examples/live/run.sh` (the same applies to any other option which is effectively required). Adding back the default value would also work, but I seem to understand that you prefer not to do that, which is fine.

titodalcanton:
> Should we change this to `default=[]` to ensure that is not a 'NoneType' object?

Just to answer this question @veronica-villa: `default=[]` will not be enough, because your code expects `args.src_class_lum_distance_to_delta` to contain two values. So either you provide a default iterable containing two values, or the user has to use the correct option (and the option must be marked as required, so we get a clear error message when it is missing).

ahnitz (approved):

PR #3808, opened 2021-09-23
Add required arguments to source classification

veronica-villa:
As noted in #3807,  some required arguments for running `mchirp_area.py` were not added to the pycbc-live test script (`/examples/live/run.sh`). I added the required flag for the arguments to the module, and their latest values to the script.


titodalcanton (approved):
I confirm this fixed the example script. Thanks @veronica-villa.

PR #3809, opened 2021-09-23
Do not install OSG 3.6

josh-willis:
See if this change fixes the Docker build problems...


ahnitz:
To see where OSG 3.5 is already installed in the base el8 igwn image here https://hub.docker.com/layers/igwn/base/el8/images/sha256-6649ac3643aebe230349a05eb58b0136228066fb8fde241f8f5dd2f0d3f16bd5?context=explore

josh-willis:
So this seems to have been an upstream packaging issue.  If the other tests succeed, we can go ahead and merge this since right now everything other PR is blocked.  But I'll try to keep an eye on things and see when the testing RPM gets moved into production, so that we can revert that change and go back to pulling everything from production.

However, for now I think we want to stick with OSG-3.5; OSG-3.6 drops support for gridftp proxies and I just don't think we have reason to believe our workflows would actually run yet in that scenario.  But we will need to get there soonish (early next year?  Possibly later?).  So the first commit in this PR I don't think will be reverted in the near future.

ahnitz (approved):

PR #3818, opened 2021-10-08
Added option to return phase offset for match

alanknee:
This adds a simple option to `pycbc.filter.match` to output the phase shift that gives the best match between two waveforms. So one can get the phase-aligned waveforms through a rotation e^(i*phi) in the complex plane.

### DIFF
--- a/pycbc/filter/matchedfilter.py
+++ b/pycbc/filter/matchedfilter.py
@@ -1386,7 +1386,7 @@ def match(vec1, vec2, psd=None, low_frequency_cutoff=None,
         id_shift, maxsnr = quadratic_interpolate_peak(left, middle, right)
         max_id = max_id + id_shift
 
-    if max_phase:
+    if return_phase:
         rounded_max_id = int(round(max_id))
         phi = numpy.angle(snr[rounded_max_id])
         return maxsnr * snr_norm / sqrt(v2_norm), max_id, phi


ahnitz:
@alanknee I think this is ready to merge after passing the automated checks (:crossed_fingers:). I made a couple small edits. The argument I changed to 'return_phase', this is somewhat more in line with conventions in numpy/scipy for example for optional return arguments. 

The other is that I verified and in fact the phase did have the wrong sign originally. It should not have had the minus sign. You can see what I would expect to be returned by way of the following examples (and now it does)

```
import numpy
from pycbc.waveform import get_fd_waveform
from pycbc.filter import match

hp, hc = get_fd_waveform(approximant="TaylorF2", mass1=5, mass2=5,
                         f_lower=30, delta_f=1.0/64,
                        )

hp2, hc2 = get_fd_waveform(approximant="TaylorF2", mass1=5, mass2=5,
                         f_lower=30, delta_f=1.0/64, coa_phase=1.5
                        )

m, i, p = match(hp, hp2, return_phase=True)
print("signal phase {}, orbital phase {}".format(p, p / 2.0))

m, i, p = match(hp * numpy.exp(1.0j * p), hp2, return_phase=True)
print("signal phase {}, orbital phase {}".format(p, p / 2.0))
```
```
signal phase 3.0, orbital phase 1.5
signal phase -2.533730551580047e-17, orbital phase -1.2668652757900235e-17
```

PR #3821, opened 2021-10-12
Use MarkupPy.markup instead of glue.markup

duncanmmacleod:
This PR modifies the `legacy_grb.py` module to use the 'official' release of the `markup.py` module, instead of the one that @duncan-brown copied over to `glue`.

This adds a new runtime requirement on [`MarkupPy`](https://pypi.org/project/MarkupPy/) but I'm not sure where to declare that since it's only used in this one 'legacy' module.

### DIFF
--- a/pycbc/results/legacy_grb.py
+++ b/pycbc/results/legacy_grb.py
@@ -26,7 +26,7 @@ import re
 import os
 from argparse import ArgumentParser
 from ligo import segments
-from glue import markup
+from MarkupPy import markup
 
 def initialize_page(title, style, script, header=None):
     """


pannarale:
> @pannarale @a-r-williamson can comment, but my understanding is that this code is being replaced with the same html generation code as elsewhere in PyCBC. I've seen prototypes of these pages, so assume this is close. I'm happy to _not_ have this declared as an explicit dependency, as long as things actually work (e.g. try/except around import so if it's not there you only get failures if trying to use this code, but can import fine).

Yes, this is about to become obsolete.  I agree with the `try...except` idea.  We have done this elsewhere to provisionally provide support (that we know we will drop) for xml files.

pannarale:
This PR is obsolete by now and can be closed.  `pycbc/results/legacy_grb.py` no longer needs `glue` or `markup`.

PR #3823, opened 2021-10-12
mkl test should deactivate if it finds no MKL installed

ahnitz:


ahnitz:
@duncanmmacleod This should the mkl issue on arm64, but as I don't have such a machine, any chance you can double check? 

@spxiwh I've updated this to only turn off if arm64 is the architecture. If MKL every appears on arm64 we can of course revert this. I changed it back also to hard fail if you are on any other architecture and mkl isn't installed.

PR #3831, opened 2021-10-16
set up phasetd histograms at instantiation but allow to delay

ahnitz:
I'm not sure I remember why I did it this way, but at the moment we delay setting up a histogram for the phasetd class until it is immediately needed for use. In findtrigs with multiple cores this has some serious downsides that effectively cause large memory use.

In case the old behavior is needed by someone, I've added an option to allow it.


ahnitz:
@tdent I've updated the PR to use 'pregenerate_hist' instead.

PR #3852, opened 2021-10-29
only analyze templates which have triggers

ahnitz:
This is a speed-up to pycbc_coinc_findtrigs in the case where not all templates could even possibly produce a trigger. In other words, a template can be quickly skipped if there is zero triggers from a template in *any* of the required detectors. There is a path that could partly help with this before, but this patch pre-culls the set of templates which is much more efficient. 

 This occurs for injection sets with heavy use of injection filtering, for example. In that case, the performance improvement can be quite large (orders of magnitude if the bank is large enough and each injection set is narrow).

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_findtrigs
+++ b/bin/all_sky_search/pycbc_coinc_findtrigs
@@ -110,7 +110,7 @@ def exit_cleaning():
 atexit.register(exit_cleaning)
 
 # If some templates have no triggers one or more ifos, there can be no
-# coincs so we track which templates contains triggers in all ifos
+# coincs: so, track which templates contain triggers in all ifos
 tids_with_trigs = None
 
 for i in range(len(args.trigger_files)):
--- a/bin/all_sky_search/pycbc_coinc_findtrigs
+++ b/bin/all_sky_search/pycbc_coinc_findtrigs
@@ -110,7 +110,7 @@ def exit_cleaning():
 atexit.register(exit_cleaning)
 
 # Which templates might have triggers, initially we'll assume they all might
-tids_with_trigs = numpy.arange(0, num_templates)
+tids_with_trigs = None
 
 for i in range(len(args.trigger_files)):
     if args.stage_input:
@@ -174,7 +174,8 @@ else:
 
 # Only analyze templates which might have triggers based on
 # our 'tids_with_trigs'
-template_ids = numpy.intersect1d(tids_with_trigs, template_ids)
+if tids_with_trigs is not None:
+    template_ids = numpy.intersect1d(tids_with_trigs, template_ids)
 
 # 'data' will store output of coinc finding
 # in addition to these lists of coinc info, will also store trigger times and
--- a/bin/all_sky_search/pycbc_coinc_findtrigs
+++ b/bin/all_sky_search/pycbc_coinc_findtrigs
@@ -109,7 +109,8 @@ def exit_cleaning():
             pass
 atexit.register(exit_cleaning)
 
-# Which templates might have triggers, initially we'll assume they all might
+# If some templates have no triggers one or more ifos, there can be no
+# coincs so we track which templates contains triggers in all ifos
 tids_with_trigs = None
 
 for i in range(len(args.trigger_files)):
@@ -174,8 +175,7 @@ if args.randomize_template_order:
 else:
     template_ids = range(tmin, tmax)
 
-# Only analyze templates which might have triggers based on
-# our 'tids_with_trigs'
+# Only analyze templates which might have coincs
 if tids_with_trigs is not None:
     template_ids = numpy.intersect1d(tids_with_trigs, template_ids)
 
--- a/bin/all_sky_search/pycbc_coinc_findtrigs
+++ b/bin/all_sky_search/pycbc_coinc_findtrigs
@@ -109,7 +109,9 @@ def exit_cleaning():
             pass
 atexit.register(exit_cleaning)
 
-unique_tids = numpy.arange(0, num_templates)
+# Which templates might have triggers, initially we'll assume they all might
+tids_with_trigs = numpy.arange(0, num_templates)
+
 for i in range(len(args.trigger_files)):
     if args.stage_input:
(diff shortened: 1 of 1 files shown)


bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -109,6 +109,7 @@ def exit_cleaning():
             pass
 atexit.register(exit_cleaning)
 
+unique_tids = numpy.arange(0, num_templates)
```
tdent, line 112: comment what this is going to be for?

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -127,6 +128,13 @@ for i in range(len(args.trigger_files)):
                             args.gating_veto_windows)
     ifo = reader.ifo
     trigs.ifos.append(ifo)
+    
+    # We don't have that many triggers compared to num templates so
+    # likely many zeros which can be pre-culled
+    if len(reader.file[ifo]['template_id']) < num_templates * 10:
```
tdent, line 134: the *10 factor feels hacky .. it doesn't really tell us if there are lots of unoccupied templates. I think numpy.unique should be fairly fast, so why not always do that and decide based on the actual number of different templates?  Or if unique/intersect1s aren't always fast enough, decide based on the length of the template_id dataset (ie whatever determines if it will be slow)?

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -163,6 +171,8 @@ if args.randomize_template_order:
 else:
     template_ids = range(tmin, tmax)
 
+template_ids = numpy.intersect1d(unique_tids, template_ids)
```
tdent, line 174: add comment to explain what is being done

ahnitz: Ok, done

bin/all_sky_search/pycbc_coinc_findtrigs
ahnitz: Ok, done

bin/all_sky_search/pycbc_coinc_findtrigs
ahnitz: Well, the implicit assumption I guess is that triggers aren't say all clustered in a single template. I guess that could be wrong. Ok, I've switched to  a fixed threshold

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -163,6 +172,10 @@ if args.randomize_template_order:
 else:
     template_ids = range(tmin, tmax)
 
+# Only analyze templates which might have triggers based on
+# our 'tids_with_trigs'
+template_ids = numpy.intersect1d(tids_with_trigs, template_ids)
```
GarethCabournDavies, line 177: Does this bit add significant effort in the case where we _don't_ want to only analyse certain templates, i.e. if `len(reader.file[ifo]['template_id']) > 2**27`

Can we add a flag in that case and therefore skip this intersect operation?

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -127,6 +130,12 @@ for i in range(len(args.trigger_files)):
                             args.gating_veto_windows)
     ifo = reader.ifo
     trigs.ifos.append(ifo)
+    
+    # We don't have that many triggers, see if we can skip some templates
+    if len(reader.file[ifo]['template_id']) < 2**27:
+        uniq = numpy.unique(reader.file[ifo]['template_id'][:])
```
GarethCabournDavies, line 136: How are the efficiency tests on this? - numpy.unique on (2 ** 27 - 1) templates feels like it will be slow

Apparently pandas.unique is a lot faster (O(N) rather than O(NlogN), but I've not checked that (https://stackoverflow.com/a/15637512)

ahnitz: Seems to take just a few seconds when I run by hand, I think that's fast enough that it shouldn't be any significant impact to overall runtime.

ahnitz: i.e. the worst case is that everything is unique. Seems good enough that performance is not a convern.

```
In [11]: %timeit numpy.unique(numpy.arange(0, 2**27, 1))
1.94 s ± 37.7 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

bin/all_sky_search/pycbc_coinc_findtrigs
ahnitz: Ok, done.

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -109,6 +109,9 @@ def exit_cleaning():
             pass
 atexit.register(exit_cleaning)
 
+# Which templates might have triggers, initially we'll assume they all might
```
tdent, line 112: The assignment of 'None' here is a bit confusing in the light of 'initially we'll assume they all might' .. suggest eg 'If some templates have no triggers, tids_with_trigs will contain only the ones that do'

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -163,6 +174,11 @@ if args.randomize_template_order:
 else:
     template_ids = range(tmin, tmax)
 
+# Only analyze templates which might have triggers based on
+# our 'tids_with_trigs'
```
tdent, line 178: is it 'might have triggers' or 'do have triggers' ?  I guess what you want to say is 'might have coincs'

Also cut 'based on our tids_with_trigs' since it's obvious from code

ahnitz: Done

bin/all_sky_search/pycbc_coinc_findtrigs
ahnitz: Ok, done

tdent (approved):

PR #3856, opened 2021-11-02
make grid tools / proxy checks optional in start script

ahnitz:
This is one of the PRs needed for non-LDG users to use the standard shell script tools made when using the '--submit-now' options. 

The change is to only do the proxy checks if you request so in the profile information i.e. so the current behavior would be replicated by adding 

```
[pegasus_profile
pycbc|check_grid = 
```
to one's config files. Otherwise, no check is performed by default.


spxiwh (approved):
Makes sense .... Now need to remember to add this to LVC configs going forward!

PR #3860, opened 2021-11-08
fix typo in pycbc_inspiral

ahnitz:
This would cause an error I think if all no templates were generated. That might happen in some injection cases with lots of filtering.


ahnitz:
Also, sorry, since I introduced this typo to begin with

spxiwh (approved):

PR #3862, opened 2021-11-11
remove normalisation by mean count above threshold in fit_over codes

GarethCabournDavies:


GarethCabournDavies:
### Relevant slack conversation:
Gareth Cabourn Davies: @ahnitz  @spxiwh I'm looking at the template fitting codes and I'm confused by this line: https://github.com/gwastro/pycbc/blob/master/bin/all_sky_search/pycbc_fit_sngls_over_multiparam#L104 - dividing by the mean has no effect on the relative weights in the smoothing, but the count above thresh is divided by constant factor which I don't think is being dealt with later?

Alexander Harvey Nitz  7:25 PM
@GarethCabournDavies @tdent  might remember faster, note that this is the same convention as in https://github.com/gwastro/pycbc/blob/master/bin/all_sky_search/pycbc_fit_sngls_over_param

Alexander Harvey Nitz  7:59 PM
@GarethCabournDavies Note, I'm not saying it's correct, it might be a mistake.

Thomas Dent  12:03 AM
@GarethCabournDavies That may be correct, it’s stuff we originally did I think .. we do need the absolute rate of triggers for the calculation at some point

Gareth Cabourn Davies  9:13 AM
I think that it makes no difference relative to the other events in the same coinc type, but does make a difference when comparing them to one another

So I think it was correct, but now it isn't (using @ahnitz 's wording of correct vs incorrect)

tdent:
In the 2-ifo case, the normalization was done 'for convenience' in order to make the template-dependent factors going into the expfit/pta stat of order 1 (ie log of them is ~0 on average).  There is already a non-normalized trigger count per template output (count_in_template) which was put in in the early stages of developing the multi-ifo stat, with the thought that this would be used instead of count_above_thresh, but in the end that change was not made, but the necessary un-normalization of count_above_thresh was also not done.

ahnitz (commented):
I don't have any special objection to this. I think assuming @tdent is happy, I won't object.

tdent:
@GarethCabournDavies As suggested in the slack channel it would be good to see this tested, eg with an O3a chunk, in case it causes unexpected shifts in the detection statistic absolute values.

GarethCabournDavies:
Comparison of the count of louder events vs background event statistic in each coincidence type:

normalised_count_comparison

The one with the divide-by-mean is the one with higher statistic value, and the lower values are with this line removed
The 3-ifo line is affected a bit more, as expected, but not as significantly as I'd expected

tdent:
> 
> The one with the divide-by-mean is the one with higher statistic value, and the lower values are with this line removed The 3-ifo line is affected a bit moree, as expected, but not as significantly as I'd expected

The 2-ifo lines shift by about -4 in stat value, so I think we expect the 3-ifo line to shift by about -6, which it more or less does. 

We do lose precision in the 3-ifo background, I can't remember if we ever looked into making different decimation levels for 3-ifo vs 2-ifo coinc - I suppose it's possible though maybe clunky in the ini file.

GarethCabournDavies:
We never did look at the decimation, and I don't think that anything downstream relies on the decimation settings being the same (I think there's one plot which can have the decimation factors separated included in the ini file)

I think this can now be approved?

tdent (approved):
Can you open an issue or MR for analysis ini files to make the change in decimation settings? Otherwise this should be OK

tdent:
I guess we should also check that the sensitivity plots are at least unchanged, if not improved.

GarethCabournDavies:
I've added the new decimation into the example ini files, I had an issue with git, so somehow lost the changes but have re-added them

tdent:
The ini changes look OK, but I don't know why the 'run small search' test is failing (could be entirely unrelated). 
Could you check the sensitivity plots from your chunk run?

GarethCabournDavies:
> The ini changes look OK, but I don't know why the 'run small search' test is failing (could be entirely unrelated). Could you check the sensitivity plots from your chunk run?

They look almost exactly the same. The foundmissed plot looks the same, but the sensitivities do have a _slight_ jiggle

tdent:
We do expect some jiggle if the background decimation has changed a bit (ie random sampling differences in FAR).

tdent:
I think the 'small search' failure is known, so OK to merge

PR #3864, opened 2021-11-11
add distance marginalization 

ahnitz:
This add distance marginalization as an option in pycbc inference models. It can either directly calculate the integration numerically on a fixed grid, or can try to precalculate this (valid over a range of SNRs if grid is precalculation interpolation is dense enough). The former is more accurate and shouldn't slow down likelihoods which are 'slow' if no other brute-force marginalizations are enabled, however, the latter should be fast enough to use in conjunction with other marginalizations (like polarization).

### DIFF
--- a/examples/inference/single/run.sh
+++ b/examples/inference/single/run.sh
@@ -1,6 +1,6 @@
-`which python` -m cProfile -o log /home/ahnitz/projects/pycbc/bin/inference/pycbc_inference \
+pycbc_inference \
 --config-file `dirname "$0"`/single.ini \
---nprocesses=1 \
+--nprocesses=4 \
 --output-file single.hdf \
 --seed 0 \
 --force \
--- a/examples/inference/single/run.sh
+++ b/examples/inference/single/run.sh
@@ -1,4 +1,4 @@
-pycbc_inference \
+`which python` -m cProfile -o log /home/ahnitz/projects/pycbc/bin/inference/pycbc_inference \
 --config-file `dirname "$0"`/single.ini \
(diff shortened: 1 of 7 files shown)


ahnitz:
@yi-fan-wang @cdcapano I think this is ready for review, can either of you take a look?

ahnitz:
To enable for models which support you add the following options (some optional) to the model section. This is an example where the prior is defined in terms of comoving volume. If it is in distance, the option can be omitted. The final three options have default values (same as shown here). 

```
marginalize_distance = True
marginalize_distance_param = comoving_volume
marginalize_distance_interpolator = True
marginalize_distance_snr_range = 1, 50
marginalize_distance_density = 1000, 1000
marginalize_distance_samples = 1000
```

cdcapano (commented):
Thanks for doing this. It looks good, but please verfiy that you haven't broken any of the current marginalizations. Also, there's several code climate issues that should be addressed: some unused imports, and formatting issues. Code climate is also complaining about the instance variables `DistMarg` sets not being in the `__init__`. I realize that is just a helper class, but you can silence that be defining the variables as class variables, all set to `None`.

ahnitz:
@cdcapano There are unittests for the existing likelihoods (that check the values even) so I am confident those are unchanged in output. I'll look what can be done with reducing some of the CC issues (some might be silenceable).

cdcapano (approved):
Looks good. I don't know why the small search test run is failing. Seems unrelated (maybe GWOSC servicing?), so approving.

PR #3866, opened 2021-11-15
Move custom LIGOLW content handler classes to `pycbc.io.ligolw`

titodalcanton:
In many places throughout the code base, we have code like this:
```python
class ContentHandler(ligolw.LIGOLWContentHandler):
    pass
lsctables.use_in(ContentHandler)
```
This PR moves the definition of this class to a single place (`pycbc.io.ligolw`), so that it can be just imported where needed. This should make it easier to maintain the code that uses the `ligo.lw` module.

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_hdfinjfind
+++ b/bin/all_sky_search/pycbc_coinc_hdfinjfind
@@ -4,19 +4,15 @@ files.
 """
 
 import argparse, h5py, logging, types, numpy, os.path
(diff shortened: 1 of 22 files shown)


spxiwh (approved):
Looks good, thanks Tito. Check the remaining CodeClimate things, which might be simple fixes, but otherwise merge when ready!

titodalcanton:
The "small search" failure is due to a recent matplotlib/mpld3 issue, unrelated to this PR.

PR #3874, opened 2021-11-24
Add multi fgmc executable and one missing pop module import

tdent:


tdent:
@ahnitz I expect it will be preferable to use logging rather than `if args.verbose: print()`, can you point to a 'good' example of this?  Eg is pycbc inspiral a good example?

ahnitz:
@tdent Yes, it should be a reasonable example.

tdent:
What's the intention behind the custom signal handler in pycbc `init_logging` btw?

ahnitz:
@tdent It allows you to enable logging for a process that is currently running. That's pretty useful for long-running jobs when you want to get debug info.

titodalcanton:
@tdent you activate verbose logging for a process that is already running in non-verbose mode with something like this: `pkill -USR1 pycbc_inspiral`.

tdent:
Oh I see, it's actually during process execution.  (I thought command line options are also effectively 'at runtime' ..)

bin/population/pycbc_multiifo_pastro
```diff
@@ -0,0 +1,396 @@
...
+                    help='Make diagnostic background PDF plots')
+parser.add_argument('--plot-inj-hist', action='store_true',
+                    help='Make diagnostic injection PDF plots')
+parser.add_argument('--plot-dir', required=True, help='Destination for plots, '
+                    'must include final "/"')
+parser.add_argument('--plot-tag', help='String for plot filenames')
+
+args = parser.parse_args()
+
+assert args.plot_dir.endswith('/')
```
GarethCabournDavies, line 110: it would be better to either dynamically add this (i.e. if not endswith('/'), += '/'), or use os.path.join

GarethCabournDavies: also an `os.path.exists(arg.plot_dir)` check would be useful to fail here straight away if the output directory doesn't exists

bin/population/pycbc_multiifo_pastro
```diff
@@ -0,0 +1,396 @@
...
+    if args.vt_err is None:
+        logging.warn('NB : no VT error provided, will use 0!')
+        args.vt_err = 0.
+    logging.info('Rate in units of ' + args.rate_units + ' ...')
+    print('Rate without VT uncertainty :')
+    print('%.1f_{-%.1f}^{+%.1f}' % (
+        cmed / args.vt_mean, abs(cminus) / args.vt_mean, cplus / args.vt_mean))
+    import rate_functions
+    fracerr = args.vt_err / args.vt_mean
+    # attr that can arbitrarily rescale counts in rate calculation : don't do this
```
GarethCabournDavies, line 296: "don't do this"?

bin/population/pycbc_multiifo_pastro
```diff
@@ -0,0 +1,396 @@
...
+    raise RuntimeError("Can't supply VT error without a VT mean!")
+
+
+# if injections are to be evaluated 'like zerolag', read them in
+if args.rank_injections:
+    assert args.p_astro_inj is not None
+    inj = utils.ForegroundEvents(args, args.coinc_times,
+                                 bin_lo=args.min_mchirp, bin_hi=args.max_mchirp)
+
+    # assume that we use the same bank all through
```
GarethCabournDavies, line 317: Is this a safe assumption? If this is a safe assumption, should this be implemented for zerolag?

GarethCabournDavies (commented):
I think this looks the same/similar to what I've reviewed already. I have a couple of comments, but these are more code implementation than validity of the analysis

bin/population/pycbc_multiifo_pastro
tdent: Don't rescale by having a 'scale' different from 1 ...

bin/population/pycbc_multiifo_pastro
tdent: maybe ...

GarethCabournDavies (approved):

PR #3877, opened 2021-11-27
edit example seach config to enanble decimation again

ahnitz:


titodalcanton:
So this is the fix for the thing that was causing the small search to fail, right? I think it is best if @GarethCabournDavies can double check the updated config here.

ahnitz:
@titodalcanton Yes, it failed because the settings were incorrectly labelled so it turned off any decimation. That caused the coinc job memory useage to balloon far beyond the memory limit of the machine, hence all jobs slowed to a crawl as they probably were stuck swapping to disk.

GarethCabournDavies:
This seems good for the small search config change, but we will still need to fix the issue as a whole

GarethCabournDavies (approved):

ahnitz:
@GarethCabournDavies Yeah, agree. It's a problem that we can't use negative values.

PR #3878, opened 2021-11-29
Allow decimation thresholds in any order and with negative values

### LINKED ISSUE #3876: Can't use negative values for --loudest-keep-values in pycbc_coinc_findtrigs
One can't do the following 

```
pycbc_coinc_findtrigs --loudest-keep-values 10:10 0:10 -5:10
```
The following also does not work
```
pycbc_coinc_findtrigs --loudest-keep-values "10:10 0:10 -5:10"
pycbc_coinc_findtrigs --loudest-keep-values "10:10" "0:10" "-5:10"
```
In the former case it (naturally) gets treated as a single argument. In the latter, you get an error that argparse isn't expecting an option -5:10. The quotes don't appear to do anything. 

See related problem https://bugs.python.org/issue9334

Thoughts are to allow a token stand in place of '-' or to revierse the order of the threshold vs decimation factor (which does work). The issue seems to be anytime there is a leading hyphon irrespective of any earlier quotes.

GarethCabournDavies:
Helps to fix #3876 , using @tdent 's suggested syntax

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_findtrigs
+++ b/bin/all_sky_search/pycbc_coinc_findtrigs
@@ -45,7 +45,8 @@ parser.add_argument("--loudest-keep-values",
                          " decimation to coincs with stat value below the"
                          " given thresholds. Supply as a comma-separated list"
                          " of threshold:decimation value pairs surrounded by"
-                         " square brackets."
+                         " square brackets. Decimation values must be"
+                         " positive integers."
                          " Ex. [9:10,8.5:30,8:30,7.5:30]."
                          " Default: no decimation")
 parser.add_argument("--template-fraction-range", default="0/1",
@@ -197,6 +198,10 @@ loudest_keep_vals = args.loudest_keep_values.strip('[]').split(',')
 
 for decstr in loudest_keep_vals:
     thresh, factor = decstr.split(':')
+    if float(factor) % 1:
+        raise RuntimeError("Non-integer decimation is not supported") 
+    if int(factor) < 1:
+        raise RuntimeError("Negative or zero decimation does not make sense")
     if int(factor) == 1:
         continue
     threshes.append(float(thresh))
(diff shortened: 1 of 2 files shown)


examples/search/analysis.ini
```diff
@@ -179,10 +179,10 @@ statistic-files = ${resolve:./statHL.hdf} ${resolve:./statLV.hdf} ${resolve:./st
 timeslide-interval = 0.1
 
 [coinc-full_data-2det]
-loudest-keep-values = 0:25
+loudest-keep-values = [0:25]
```
ahnitz, line 182: Maybe try with a negative value also now to demonstrate this works? (and have it be tested now)

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -187,16 +191,25 @@ for ifo in trigs.ifos:
...
 
-total_factors = [1]
+factors = [1]
 threshes = [numpy.inf]
-for decstr in args.loudest_keep_values:
+loudest_keep_vals = args.loudest_keep_values.strip('[]').split(',')
+
+for decstr in loudest_keep_vals:
     thresh, factor = decstr.split(':')
     if int(factor) == 1:
```
tdent, line 200: it seems we want some check for non-integer factors .. eg if someone used '1.5' as a decimation factor 
I guess at first order we don't want to allow non-integer, so I would say eg `if int(factor) != float(factor): raise RuntimeError("Sorry, non-integer decimation not supported")

tdent: Could be that some part of the code does already fail for factor strings like '1.5' - not to mention '-3' ... - but if so it would be nice to see that commented explicitly. Ideally we'd want to explicitly limit the factors to be positive integers

tdent (commented):
See comments - I think it's worth putting in an input check against 'silly' factors to defend against possible future c&p errors/typos

examples/search/analysis.ini
GarethCabournDavies: I've put it in with negative values to check it

GarethCabournDavies: That check has passed now

tdent (approved):

PR #3883, opened 2021-11-29
sanitize newlines in configparser input to spaces

ahnitz:
This should re-enable use of multiline input in config parsers within inference workflows. The change to pegasus5 was accompanied by a change to using Pegasus style string formatting rather than condor. The pegasus style is supperior as it no longer requires us to worry about quoting to get things to be passed as we'd have wanted. The downside is that newlines get passed *exactly* as we've specified which breaks inference jobs that used that format. In general we can just auto-convert these to spaces and they should run again.


ahnitz:
@yi-fan-wang Did this work? Can it be merged?

yi-fan-wang (approved):
looks good

PR #3890, opened 2021-11-30
fix memory leak in field array access using expressions

ahnitz:


ahnitz:
Test failures due to LIGO servers not being up for some files.

PR #3895, opened 2021-12-06
optimize SNR : few cleanups, allow for no GDB id

tdent:
As far as I can see the SNR threshold option is not implemented, so I have turned this into a warning.

### DIFF
--- a/bin/pycbc_optimize_snr
+++ b/bin/pycbc_optimize_snr
@@ -24,14 +24,15 @@ from pycbc.io.hdf import load_hdf5_to_dict
 from pycbc.detector import Detector
 from pycbc.psd import interpolate
 
-try: 
+try:
     import pyswarms as ps
 except:
     ps = None
 
-
+    
 Nfeval = 0
 start_time = time.time()
+
 
 def callback_func(Xi, convergence=0):
     global Nfeval
@@ -40,7 +41,8 @@ def callback_func(Xi, convergence=0):
         return True
     Nfeval += 1
 
-# Define the function for computing network snr
+
+# Function for computing network snr
 def compute_network_snr_core(v, *argv):
     data = argv[0]
     coinc_times = argv[1]
@@ -98,15 +100,18 @@ def compute_network_snr_core(v, *argv):
         snr_series = snr[onsource_slice] * norm
         snr_series_dict[ifo] = snr * norm
         snr_series_dict['sigmasq_' + ifo] = sigmasq
-        network_snrsq += max(abs(snr_series._data))**2
-    return (network_snrsq**0.5), snr_series_dict
+        network_snrsq += max(abs(snr_series._data)) ** 2.
+
+    return network_snrsq ** 0.5, snr_series_dict
+
 
 def compute_network_snr(v, *argv):
     if len(argv) == 1:
         argv = argv[0]
     nsnr, _ = compute_network_snr_core(v, *argv)
     return -nsnr
- 
+
+
 def compute_network_snr_pso(v, *argv, **kwargs):
     argv = kwargs['args']
     nsnr_array = numpy.array([-compute_network_snr_core(v_i, *argv)[0] for v_i in v])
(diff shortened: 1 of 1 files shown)


bin/pycbc_optimize_snr
```diff
@@ -24,23 +24,25 @@ from pycbc.io.hdf import load_hdf5_to_dict
...
 from pycbc.psd import interpolate
 
-try: 
+try:
     import pyswarms as ps
 except:
     ps = None
 
-
+
```
titodalcanton, line 32: Looks like a few extra whitespaces are being added here.

bin/pycbc_optimize_snr
```diff
@@ -120,21 +125,21 @@ parser.add_argument('--verbose', action='store_true')
...
                          'Live')
-parser.add_argument("--data-files", type=str, nargs='+',
+parser.add_argument('--data-files', type=str, nargs='+',
                     action=MultiDetOptionAction,
                     metavar='IFO:DATA_FILE',
-                    help="Locations of the overwhitened data files produced "
-                         "by PyCBC Live.")
-parser.add_argument("--psd-files", type=str, nargs='+',
+                    help='Locations of the overwhitened data files produced '
+                         'by PyCBC Live.")
```
titodalcanton, line 132: Hmm I guess the tests are failing because of the quotes here.

bin/pycbc_optimize_snr
```diff
@@ -120,21 +125,21 @@ parser.add_argument('--verbose', action='store_true')
...
                     metavar='IFO:PSD_FILE',
-                    help="Locations of the PSD files produced "
-                         "by PyCBC Live.")
-parser.add_argument("--approximant", required=True,
-                    help="Waveform approximant string.")
-parser.add_argument("--snr-threshold", default=4.0,
-                    help="If the SNR in ifo X is below this threshold do not "
-                         "consider it part of the coincidence.")
+                    help='Locations of the PSD files produced '
+                         'by PyCBC Live.")
```
titodalcanton, line 137: …and here.

bin/pycbc_optimize_snr
```diff
@@ -120,21 +125,21 @@ parser.add_argument('--verbose', action='store_true')
...
+parser.add_argument('--approximant', required=True,
+                    help='Waveform approximant string.')
+parser.add_argument('--snr-threshold', default=4.0,
+                    help='If the SNR in ifo X is below this threshold do not '
+                         'consider it part of the coincidence. Not implemented')
```
titodalcanton, line 142: I forget what is going on here, but I vote for just dropping the option if it is not used. @spxiwh?

bin/pycbc_optimize_snr
```diff
@@ -179,14 +185,18 @@ parser.add_argument('--pso-c2', type=float, default=2.0,
 parser.add_argument('--pso-w', type=float, default=0.01,
                     help='Only relevant for --optimizer pso: '
                     'The hyperparameter w: the intertia parameter.')
-                    
 
+# Input checking
 args = parser.parse_args()
+if args.snr_threshold != 4:
+    raise NotImplementedError("Sorry, the SNR threshold option doesn't '
+                              "work yet")
```
titodalcanton, line 193: If you end up keeping the unused option, please use
```python
parser.error('Sorry, the SNR threshold option does not work yet')
```
instead of an exception.

bin/pycbc_optimize_snr
```diff
@@ -196,18 +206,25 @@ for ifo in args.data_files.keys():
 
 fp = h5py.File(args.params_file, 'r')
 ifos = list(data.keys())
-original_gid = fp['gid'][()]
+try:
```
titodalcanton, line 209: Since the lack of `gid` seems an "expected" occurrence (I guess it happens if the search was told not to upload to GraceDB?) I would use an explicit check here (`if 'gid' in fp`) instead of a try-except.

bin/pycbc_optimize_snr
```diff
@@ -196,18 +206,25 @@ for ifo in args.data_files.keys():
...
+    original_gid = fp['gid'][()]
+except KeyError as gdberr:
+    if args.enable_gracedb_upload:
+        print('Need original gracedb ID to upload followup!', gdberr)
```
titodalcanton, line 213: Use `logging.error()` here, not `print()`.

bin/pycbc_optimize_snr
```diff
@@ -249,16 +266,16 @@ if args.optimizer == 'differential_evolution':
                                  args=extra_args)
 
 elif args.optimizer == 'shgo':
-    results = shgo(compute_network_snr, bounds=bounds, args=extra_args, 
+    results = shgo(compute_network_snr, bounds=bounds, args=extra_args,
                 iters=args.shgo_iters, n=args.shgo_samples, sampling_method="sobol")
-                
+
```
titodalcanton, line 271: I would even drop the empty line here.

bin/pycbc_optimize_snr
titodalcanton: If you do drop it, make sure `pycbc_live` does not use it!

bin/pycbc_optimize_snr
tdent: is that this? https://docs.python.org/3/library/argparse.html?highlight=parser%20error#argparse.ArgumentParser.error

tdent:
btw, do we know why the compute optimal snr function returns `(network_snrsq**0.5)` as a single-element tuple rather than a scalar?

bin/pycbc_optimize_snr
tdent: Looks like it doesn't.

titodalcanton:
@tdent are you referring to the return value of `compute_network_snr_core()`? `(network_snrsq**0.5)` is actually not a tuple (you would need an extra comma to make it one).

tdent:
> @tdent are you referring to the return value of `compute_network_snr_core()`? `(network_snrsq**0.5)` is actually not a tuple (you would need an extra comma to make it one).

I should have said network not optimal .. so yeah, I assume that this parenthesis around a single value has no effect.

titodalcanton (approved):
Looks good.

PR #3896, opened 2021-12-07
Error message from missing buffer size

tdent:


ahnitz (approved):
@tdent Yeah, if we don't have a length predictor, it is indeed an error.

PR #3897, opened 2021-12-08
Improve error when strain data is not available

titodalcanton:
When using `Merger(name).strain(detector)` and the strain data does not actually exist, the current code raises a rather vague error like this:
```
UnboundLocalError: local variable 'url' referenced before assignment
```
This PR makes the error clearer and cleans up the code slightly.


ahnitz (approved):
Seems like a nice QOL improvement. Thanks!

PR #3900, opened 2021-12-11
Don't hardcode event ID format

tdent:
Tested, seems to work


titodalcanton (approved):
The `event_id` is set by `SingleCoincForGraceDB`'s constructor, so this is fine.

PR #3902, opened 2021-12-21
made pycbc_make_inference_plots_workflow compatible with Pegasus 5 an…

ShilpaKastha:
made pycbc_make_inference_plots_workflow compatible with Pegasus 5 made the plot_prior optional in inference_followups.py


ahnitz:
@ShilpaKastha Can you fix the Codeclimeate errors. the mixed tab/whitespace one is the most problematic, but I think all of them can be addressed.

ahnitz:
@ShilpaKastha Otherwise, I think this is probably fine.

ahnitz (approved):

PR #3905, opened 2021-12-22
don't set profile within jobsetup

ahnitz:
@spxiwh This removes the couple instance where we explicitly set profile information in jobs setup. 

After the P5 changes, we lost the ability to override these with the config options. This occured because the code was removed which checked if a profile profile was set and only set a new one if a 'force' option was given. This allowed the config file to take precedence over things in the code. 

I'm not sure if we want that functionality back, but we absolutely do need the ability to set this information in the config file. By removing these lines, the config file is no longer ignored.

### DIFF
--- a/pycbc/workflow/jobsetup.py
+++ b/pycbc/workflow/jobsetup.py
@@ -507,7 +507,6 @@ class PyCBCInspiralExecutable(Executable):
             set_submit_subdir=False
         )
         self.cp = cp
-        self.set_memory(2000)
         self.injection_file = injection_file
         self.ext = '.hdf'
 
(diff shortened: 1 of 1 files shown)


spxiwh (approved):
This does all seem fair. I think that it is strange that things set here could be overridden by the config file. Maybe we could do something like "if memory is not already set, set it to 2000" in jobsetup.py, but I think if we run a `set_memory` command then memory should be set.

I don't see a use-case where these things don't belong in the config files though!

ahnitz:
@spxiwh Yeah, I agree the config file seems the most sensible, hence why I just opted to make sure these actually can be set by config file rather than being silently ignored in preference of the hardcoded options (which are much too small for some injection jobs for example).

PR #3915, opened 2022-01-21
Allow .h5 and .hdf5 extensions for TemplateBank output files

duncanmmacleod:
This PR modifies `TemplateBank.write_hdf` to allow using `.h5` and `.hdf5` file extensions for output files, since `.hdf` is ambiguous with HDF4 files.


spxiwh (approved):
Thanks, this looks straightforward and unobjectionable.

PR #3916, opened 2022-01-21
Modernize `pycbc_losc_segment_query`

titodalcanton:
`pycbc_losc_segment_query` has a number of issues and I suspect it is not really used nowadays. However, it is still part of the `--help` sanity tests, and it fails on Python 3.10 due to its use of the abandoned `glue.segmentdb` module. This change removes the LIGOLW writing code from the script in favor of using the SegFile class. The script functionality is more or less preserved, apart from losing the comments in the LIGOLW tables. I also do some cleanup of the script, for example turning all [LOSC](https://www.losc.fr/) references to GWOSC. I think some additional cleanup/fixing is due but I did not spend further time on it, given that I am not sure how much this is really used nowadays.

As part of this change, I created another wrapper function in `pycbc.io.ligolw` to handle the LIGOLW process table, so we can remove a few more imports and quirks from the main code. Not all calls to `register_to_xmldoc()` are removed though.

Finally, I think there was a bug in `SegFile.from_segment_list()`, so I also fix that.

### DIFF
--- a/bin/pycbc_losc_segment_query
+++ b/bin/pycbc_losc_segment_query
@@ -1,50 +1,43 @@
 #!/usr/bin/env python
 
 import os
-import pwd
 import logging
 import json
 import argparse
 import shutil
 from urllib.request import urlopen
-
 import ligo.segments
-from glue import git_version
-from ligo.lw import ligolw
-from ligo.lw import utils
-from ligo.lw.utils import process
-from glue.segmentdb import segmentdb_utils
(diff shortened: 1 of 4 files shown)


ahnitz:
Is there any reason not to merger this? Btw, does anyone use the segment query script anymore? The main workflows I think won't do so, but if this script is in use, then I guess updating is better than removing.

spxiwh (approved):
Thanks!

PR #3919, opened 2022-01-25
Use shutil.which instead of distutils.spawn.find_executable

duncanmmacleod:
This PR replaces all instances of `distutils.spawn.find_executable` with [`shutil.which`](https://docs.python.org/3/library/shutil.html#shutil.which), distutils is [deprecated](https://www.python.org/dev/peps/pep-0632/).


titodalcanton (approved):
Seems a fairly trivial change. Thanks @duncanmmacleod.

PR #3923, opened 2022-01-28
Fix in from_ligolw_table

a-r-williamson:
As discussed in [#3921](https://github.com/gwastro/pycbc/issues/3921#issuecomment-1023359795), changes in ligo.lw mean tables have columns like process:process_id but corresponding row attributes are only the post-colon part, e.g. process_id. This caused an error when reading in from xml via `from_ligolw_table`.

I haven't been able to find any other places where this same issue will arise, hopefully it's only here.

### DIFF
--- a/pycbc/io/record.py
+++ b/pycbc/io/record.py
@@ -1181,7 +1181,7 @@ class FieldArray(numpy.recarray):
         # get the values
         if _default_types_status['ilwd_as_int']:
             input_array = \
-                [tuple(getattr(row, col) if dt != 'ilwd:char'
+                [tuple(getattr(row, col.split(':')[-1]) if dt != 'ilwd:char'
                        else int(getattr(row, col))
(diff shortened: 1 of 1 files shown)


titodalcanton:
Could you add a comment to explain the need for the split:
```python
# columns like `process:process_id` must be accessed in Python as just `process_id`
```
And I think a similar change may be needed a few lines below? Not sure.

a-r-williamson:
I don't think the change is needed below as well. If I understood what's going on, `_default_types_status['ilwd_as_int']` will be True when handling tables in the new format, since it's a check for whether ilwd:chars are strings or ints. @spxiwh or @ahnitz is that right?

titodalcanton (approved):
I was hesitating on further comments as I am not super familiar with the io.record module, but as far as I can tell this looks ok and simple enough.

PR #3925, opened 2022-01-31
Workaround for pip 22.0.x issue

titodalcanton:
Trying to work around this problem: https://github.com/pypa/pip/issues/10851.


ahnitz (approved):
If this fixes the issue, then go ahead and merge. It looks like this is the solution others seems to be going with as well for now.

PR #3926, opened 2022-01-31
update page_foreground hdf output to match 4-ogc summary file

GarethCabournDavies:
The pycbc_page_foreground output has an option to output in hdf format. I thought it best to match the 4-OGC output [here](https://github.com/gwastro/4-ogc/blob/master/search/4-OGC_small.hdf), as it isn't too different to what we already do

I think that we should try to align formats of released event information, and this could be the basis of future data releases

### DIFF
--- a/bin/all_sky_search/pycbc_add_statmap
+++ b/bin/all_sky_search/pycbc_add_statmap
@@ -451,7 +451,6 @@ while True:
         logging.info("Reached hierarchical removal limit of %d" %
                      args.max_hierarchical_removal)
         break
-    logging.info("Hierarchical iteration %d" % h_iterations)
 
     max_each_combo = {combo: sep_fg_data[combo].data['ifar'][:].argmax()
                       for combo in all_ifo_combos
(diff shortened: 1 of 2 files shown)


bin/all_sky_search/pycbc_add_statmap
```diff
@@ -451,7 +451,6 @@ while True:
         logging.info("Reached hierarchical removal limit of %d" %
                      args.max_hierarchical_removal)
         break
-    logging.info("Hierarchical iteration %d" % h_iterations)
```
GarethCabournDavies, line 454: This is a leftover change from #3914

ahnitz:
@GarethCabournDavies I guess I forgot this code had an hdf output option. THis makes sense then.

ahnitz (approved):

GarethCabournDavies:
> @GarethCabournDavies I guess I forgot this code had an hdf output option. THis makes sense then.

Only six months old so easily forgettable - not used in the workflow either.

PR #3929, opened 2022-02-03
Moving multi_inspiral utils to coherent module

a-r-williamson:
As development on `pycbc_multi_inspiral` has progressed, the executable has gathered a lot of utility functions to do with trigger identification and coherent SNR calculation that would be better placed elsewhere. I suggest moving them to a coherent module.

### DIFF
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -96,6 +96,8 @@ if __name__ == '__main__':
     parser.add_argument("--coinc-threshold", type=float, default=0.0,
                         help="Triggers with coincident/coherent snr below "
                              "this value will be discarded.")
+    parser.add_argument("--do-null-cut", action='store_true',
+                        help="Apply a cut based on null SNR.")
     parser.add_argument("--null-min", type=float, default=5.25,
                         help="Triggers with null_snr above this value will be"
                              " discarded.")
@@ -455,7 +457,7 @@ if __name__ == '__main__':
                         null, rho_coh, rho_coinc, coinc_idx, coinc_triggers =\
                             coh.null_snr(
                                 rho_coh, rho_coinc, snrv=coinc_triggers,
-                                index=coinc_idx)
+                                index=coinc_idx, apply_cut=args.do_null_cut)
                         if len(coinc_idx) != 0:
                             logging.info("Max null SNR = %.2f", max(null))
                         logging.info(
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -381,7 +381,7 @@ if __name__ == '__main__':
                 for ifo in args.instruments:
                     # Move on if this segment has no data
                     if len(snr_dict[ifo])==0:
-                        raise ValueError(
+                        raise RuntimeError(
                             'The SNR triggers dictionary is empty. This '
                             'should not be possible.')
                     # Time delay is applied to indices
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -23,6 +23,7 @@ import argparse
 from pycbc import vetoes, psd, waveform, events, strain, scheme, fft,\
     DYN_RANGE_FAC
 from pycbc.vetoes import sgchisq
+from pycbc.events import coherent as coh, EventManagerCoherent
 from pycbc.filter import MatchedFilterControl
 from pycbc.types import TimeSeries, zeros, float32, complex64
 from pycbc.types import MultiDetOptionAction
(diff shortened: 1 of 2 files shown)


titodalcanton:
@a-r-williamson since you are introducing a new file, this is a good chance of applying the [Black autoformatter](https://github.com/gwastro/pycbc/wiki/How-to-satisfy-CodeClimate) without polluting the history of that file with code reformatting. Could you have a look?

a-r-williamson:
> @a-r-williamson since you are introducing a new file, this is a good chance of applying the [Black autoformatter](https://github.com/gwastro/pycbc/wiki/How-to-satisfy-CodeClimate) without polluting the history of that file with code reformatting. Could you have a look?

Having just spent some time going through trying to clean up a lot of the formatting by hand... I'll check that out, thanks!

bin/pycbc_multi_inspiral
```diff
@@ -16,56 +16,56 @@
...
-import time
+from pycbc.vetoes import sgchisq
 
 time_init = time.time()
 
-parser = argparse.ArgumentParser(usage='',
-    description="Find multiple detector gravitational-wave triggers.")
+parser = argparse.ArgumentParser(
+    usage="", description="Find multiple detector gravitational-wave triggers."
+)
```
titodalcanton, line 38: I suggest putting the description in the docstring at the top of the file, and then doing
```python
```
See for example `pycbc_optimal_snr` or other scripts.

bin/pycbc_multi_inspiral
```diff
@@ -93,275 +93,64 @@ parser.add_argument("--user-tag", type=str, metavar="TAG", help="""
...
+opt.insert_optimization_option_group(parser)
+inject.insert_injfilterrejector_option_group_multi_ifo(parser)
 sgchisq.SingleDetSGChisq.insert_option_group(parser)
-opt = parser.parse_args()
+args = parser.parse_args()
 
-# Use the time in the middle of the segment to calculate the antenna patterns.
-t_gps = opt.trigger_time
+log_level = logging.DEBUG if args.verbose else logging.INFO
+logging.basicConfig(format='%(asctime)s : %(message)s', level=log_level)
```
titodalcanton, line 127: Use `pycbc.init_logging()` here.

bin/pycbc_multi_inspiral
```diff
@@ -567,167 +360,190 @@ with ctx:
...
+            nifo = len(args.instruments)
 
-            for ifo in opt.instruments:
+            for ifo in args.instruments:
                 # Move on if this segment has no data
                 if len(snr[ifo])==0:
-                    raise ValueError('The SNR triggers dictionary is empty.\
-                                     This should not be possible.')
+                    raise ValueError('The SNR triggers dictionary is empty. '
+                                     'This should not be possible.')
```
titodalcanton, line 402: A `RuntimeError` may be more appropriate than a `ValueError` here.

bin/pycbc_multi_inspiral
```diff
@@ -736,25 +552,27 @@ with ctx:
...
-            cluster_window
-            )
+            "time_index", "coherent_snr", cluster_window)
         event_mgr.finalize_template_events()
-event_mgr.write_events(opt.output)
+event_mgr.write_events(args.output)
 
 logging.info("Finished")
-logging.info("Time to complete analysis: %s" % str(time.time() - time_init))
+logging.info("Time to complete analysis: %d", int(time.time() - time_init))
```
titodalcanton, line 578: This is a bit shorter and properly rounded:
```python
logging.info("Time to complete analysis: %.0f s", time.time() - time_init)
```

pycbc/events/coherent.py
```diff
@@ -0,0 +1,371 @@
...
+        # later.
+        if len(idx_dict[ifo]) != 0:
+            coinc_list = np.hstack(
+                [coinc_list, idx_dict[ifo] - time_delay_idx[ifo]]
+            )
+    # Search through coinc_idx for repeated indexes. These must have been loud
+    # in at least 2 detectors.
+    coinc_idx = np.unique(coinc_list, return_counts=True)[0][
+        np.unique(coinc_list, return_counts=True)[1] > 1
+    ]
```
titodalcanton, line 62: This may be more readable like this:
```python
something = np.unique(coinc_list, return_counts=True)
coinc_idx = something[0][something[1] > 1]
```

pycbc/events/coherent.py
```diff
@@ -0,0 +1,371 @@
...
+        for ifo in ifos
+    }
+    coinc_snr2 = sum(snr2.values())
+    snr2_ratio = {ifo: snr2[ifo] / coinc_snr2 for ifo in ifos}
+    net_chisq = sum([chisq_per_dof[ifo] * snr2_ratio[ifo] for ifo in ifos])
+    return net_chisq
+
+
+def reweighted_snr(netwk_snr, netwk_chisq, idx1=3, idx2=1.0 / 6.0):
+    """Return the chi-squeared re-weighted SNR statistic
```
titodalcanton, line 269: Typo.

titodalcanton (changes requested):
@a-r-williamson this looks mostly ok, I have a few minor requests above, plus the following one: in the docstrings of the various reweighted SNRs and chi-squared functions, could you add links to the papers giving the definitions of those quantities?

bin/pycbc_multi_inspiral
a-r-williamson: Done in 95caa1cc

pycbc/events/coherent.py
a-r-williamson: Done in f222a2

pycbc/events/coherent.py
a-r-williamson: Done in f222a2

pycbc/events/coherent.py
```diff
@@ -0,0 +1,371 @@
...
+        Network chisq values corresponding to each trigger
+
+    Returns
+    -------
+    rw_snr: numpy.ndarray
+        Array of re-weighted SNRs
+    """
+    denom = ((1 + netwk_chisq) ** idx1) / 2
+    rw_snr = netwk_snr / denom ** idx2
+    return rw_snr
```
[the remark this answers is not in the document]
a-r-williamson: This should be the same re-weighting. I'll switch to use `newsnr`

a-r-williamson:
@titodalcanton (and @pannarale) I think I've addressed all your comments, Tito. I'd love to get this merged as soon as possible, especially as this includes important fixes for bugs that we discovered while tidying up here:

[Serious bug in assigning SNRs and trigger array indices when looping over IFOs](https://github.com/gwastro/pycbc/commit/e532771e412563a51b8686227220c6aa75e62f50)
Each iteration in [this loop](https://github.com/gwastro/pycbc/commit/e532771e412563a51b8686227220c6aa75e62f50#diff-218a84325b9ee909bdcdc6b7e5d423c8b3e634b7176bb0c1a6b5514ea61b6e37L345-R362) was improperly assigning the outputs of `matched_filter_and_cluster`. These need to be copies. 

[Bug in the null SNR re-weighting](https://github.com/gwastro/pycbc/pull/3929/commits/16b6a7b35ff66d8fadcc5225f0cf3a9a1f399b66#diff-4b37abfdbeda70e2b7dbd03499c15377e8b4566d89e7e59523ad2bf0fccb48b9L347-R367)
The re-weighting of the detction statistic was being incorrectly applied with a threshold flat in null SNR. It should have been flat in null SNR up to a threshold in coherent SNR, then increase linearly with coherent SNR * some tunable gradient.

titodalcanton:
@a-r-williamson the diff for `pycbc_multi_inspiral` appears messed up somehow, but it may simply be extremely large. Could you maybe avoid the `if __name__ == '__main__':` block? Since this script is not supposed to be imported, I am not sure it adds anything, and it steals one indentation level across the *entire* script. I hope it was not `black` that did that :)

a-r-williamson:
@titodalcanton Done, thanks.

bin/pycbc_multi_inspiral
```diff
@@ -381,353 +165,359 @@ with ctx:
...
+    # Loop over templates
+    for t_num, template in enumerate(bank):
+        for ifo in args.instruments:
+            if args.cluster_method == "window":
+                cluster_window = int(args.cluster_window * sample_rate)
+            elif args.cluster_method == "template":
                 cluster_window = int(template.chirp_length * sample_rate)
-            elif opt.cluster_window == 0:
+            elif args.cluster_window == 0:
                  cluster_window = int(0)
```
titodalcanton, line 293: Am I missing the point of having this if-else block inside the two for loops? It looks like you could easily move it to just before the loops.

titodalcanton: Ah right, it does depend on the template, but can you at least move it before the loop over the instruments?

titodalcanton: Actually, if you only use `cluster_window` in `cluster_template_network_events()`, then it is probably more readable to move this entire if-else block right before that call.

titodalcanton: Also, `int(0)` is equivalent to just `0`.

titodalcanton:
@a-r-williamson I have one last comment, and you should also fix codeclimate's points about redefining names. These are my last comments I think.

bin/pycbc_multi_inspiral
a-r-williamson: Well spotted. I'll move this down and remove the last check as it's redundant.

titodalcanton (approved):

Issue #3928, opened 2022-02-03
Rethink chirp mass range in `pycbc_optimize_snr`?


titodalcanton:
@spxiwh notes:

```python
minchirp = mchirp * (1 - mchirp / 50.0)
maxchirp = mchirp * (1 + mchirp / 50.0)
```

> It looks like there's a bug here. I think this was supposed to be fractional bounds on mchirp (maybe ones that increase with the mchirp??). But as a specific fail case: if the mchirp is 50, this will search down to a mchirp of 0 and then things will break!
> Although I guess the lines below catch that, but I still think the bounds aren't ideal!

titodalcanton:
I did in fact see this error while playing with `pycbc_optimize_snr`, though I am not sure if this is a consequence of what @spxiwh said.
```
Traceback (most recent call last):
  File "/home/tito/virtualenvs/pycbc/bin/pycbc_optimize_snr", line 350, in <module>
    opt_params = optimize_func(bounds, args, extra_args)
  File "/home/tito/virtualenvs/pycbc/bin/pycbc_optimize_snr", line 135, in optimize_di
    results = differential_evolution(
  File "/home/tito/virtualenvs/pycbc/lib/python3.9/site-packages/scipy/optimize/_differentialevolution.py", line 329, in differential_evolution
    ret = solver.solve()
  File "/home/tito/virtualenvs/pycbc/lib/python3.9/site-packages/scipy/optimize/_differentialevolution.py", line 903, in solve
    result = minimize(self.func,
  File "/home/tito/virtualenvs/pycbc/lib/python3.9/site-packages/scipy/optimize/_minimize.py", line 623, in minimize
    return _minimize_lbfgsb(fun, x0, args, jac, bounds,
  File "/home/tito/virtualenvs/pycbc/lib/python3.9/site-packages/scipy/optimize/lbfgsb.py", line 294, in _minimize_lbfgsb
    raise ValueError("LBFGSB - one of the lower bounds is greater than an upper bound.")
ValueError: LBFGSB - one of the lower bounds is greater than an upper bound.
```

tdent:
Well, maybe an idea behind the formula was that the highest chirp mass in the live bank was less than 50.

tdent:
We could try and fix this mathematically by noting that 1 \pm mchirp/50 is the first order expansion of exp(\pm mchirp/50), and if you use the exponential instead of the linear function the minimum never gets to zero.  Though for mchirp >> 50 it does get very close to 0.

[...]

PR #3931, opened 2022-02-04
Run SNR optimizer in PyCBC Live example/test

titodalcanton:
I found that this does not actually take a lot of time on my test machine, so why not. At the moment the SNR optimizer fails on one of the injections (see #3928) so I did not add any actual check that the SNR optimization is successful. However, if a LIGOLW file is generated, BAYESTAR will run on it, so we can spot cases where `pycbc_optimize_snr` breaks BAYESTAR.

### DIFF
--- a/examples/live/.gitignore
+++ b/examples/live/.gitignore
@@ -3,3 +3,4 @@ template_bank.hdf
 injections.hdf
 stat*.hdf
 output/
+*.fits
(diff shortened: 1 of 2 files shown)


tdent:
I'm not really familiar with how the test suite works here, so maybe I can't give much feedback.

tdent (approved):
It looks reasonable, is the first change to .gitignore related to the snr opt addition though?

titodalcanton:
@tdent the change to `.gitignore` is not specific to the SNR optimizer, but I did not fill like opening a whole PR just for that.

Anyway, it seems that the two SNR optimizer jobs timeout before completing, probably because the machine where the test runs appear to have 2 available cores only. Not much we can do about that I guess.

PR #3933, opened 2022-02-08
Adding test of pycbc_multi_inspiral face-on/-away

a-r-williamson:
This adds an example for `pycbc_multi_inspiral` to test not only the executable, but also the implementation of the face-on/-away signal projection used in PyGRB searches. Relies on gwastro/pycbc-config#76 for inputs.

### DIFF
--- a/examples/multi_inspiral/check_gw170817_trigs.py
+++ b/examples/multi_inspiral/check_gw170817_trigs.py
@@ -25,7 +25,6 @@ mask = (
     & (snrs[2] > 25)
     )
 n = mask.sum()
-status = 0
 if n > 0:
     result = 'PASS'
     status = 0
--- a/examples/multi_inspiral/check_gw170817_trigs.py
+++ b/examples/multi_inspiral/check_gw170817_trigs.py
@@ -1,0 +1,34 @@
+#!/usr/bin/env python
+# Read a pycbc_multi_inspiral HDF5 trigger file and check that it contains
+# triggers compatible with mock GW170817-like injections
+# 2022 Andrew Williamson, Tito Dal Canton
+
+import sys
+import logging
+import h5py
+import numpy as np
+from pycbc import init_logging
+
+init_logging(True)
+gw170817_time = 1187008882.43
+status = 0
+with h5py.File('GW170817_test_output.hdf', 'r') as f:
+    snrs = [
+        f['network/end_time_gc'][:],
+        f['network/coherent_snr'][:],
+        f['network/reweighted_snr'][:],
+        ]
+# search for compatible trigs
+mask = (abs(gw170817_time - snrs[0]) < 0.1)
+n = mask.sum()
+status = 0
+if n > 0:
+    result = 'PASS'
+    status = 0
+else:
+    result = 'FAIL'
+    status = 1
+logging.info(
+    '%s: GW170817 found with coherent SNR = %.2f; reweighted SNR %.2f', result,
+    snrs[1][mask], snrs[2][mask])
+sys.exit(status)
--- a/examples/multi_inspiral/check_gw170817_trigs.py
+++ b/examples/multi_inspiral/check_gw170817_trigs.py
@@ -19,7 +19,11 @@ with h5py.File('GW170817_test_output.hdf', 'r') as f:
         f['network/reweighted_snr'][:],
         ]
 # search for compatible trigs
-mask = (abs(gw170817_time - snrs[0]) < 0.1)
+mask = (
+    (abs(gw170817_time - snrs[0]) < 0.1)
+    & (snrs[1] > 25)
+    & (snrs[2] > 25)
+    )
 n = mask.sum()
 status = 0
 if n > 0:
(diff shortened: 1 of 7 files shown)


examples/multi_inspiral/faceon_faceaway.sh
```diff
@@ -0,0 +1,82 @@
...
+CONFIG_URL=https://github.com/a-r-williamson/pycbc-config/raw/pycbc_multi_inspiral_tests/test/multi_inspiral
+INJ_FILE=faceon_faceaway_injs.hdf
+
+echo -e "\\n\\n>> [`date`] Getting injection file"
+wget -nc ${CONFIG_URL}/${INJ_FILE}
+
+# Generate mock data
+for IFO in 'H1' 'L1' 'V1'; do
+    echo -e "\\n\\n>> [`date`] generating mock data for ${IFO}"
+    `which pycbc_condition_strain` \
```
titodalcanton, line 20: Why is the `which` needed?

examples/multi_inspiral/faceon_faceaway.sh
```diff
@@ -0,0 +1,82 @@
...
+DEC=-0.408407044967
+
+echo -e "\\n\\n>> [`date`] Getting template bank"
+wget -nc ${CONFIG_URL}/${BANK_FILE}
+echo -e "\\n\\n>> [`date`] Bank veto bank"
+wget -nc ${CONFIG_URL}/${BANK_VETO_FILE}
+
+for POL in 'standard' 'left' 'right' 'left+right'; do
+    echo -e "\\n\\n>> [`date`] Running pycbc_multi_inspiral with ${POL} projection"
+    `which pycbc_multi_inspiral` \
```
titodalcanton, line 46: Same comment as above.

titodalcanton:
@a-r-williamson you could also add this example to the test suite. See `/tools/pycbc_test_suite.sh`, I think a good place to run this is just after the PyCBC Live test.

examples/multi_inspiral/faceon_faceaway.sh
a-r-williamson: I had just followed the lead of [examples/inspiral/run.sh](https://github.com/gwastro/pycbc/blob/master/examples/inspiral/run.sh) which evaluates the path with `which` but I guess there's no need.

a-r-williamson:
Hi @titodalcanton, thanks. Yes I was going to push another commit suggesting this, glad you agree.

examples/multi_inspiral/faceon_faceaway.sh
titodalcanton: Ah I see, in that case it was needed because `pycbc_inspiral` was being run through the cProfile module (which has then been commented out). You can just call `pycbc_condition_strain` here.

pycbc/frame/frame.py
```diff
@@ -79,6 +79,7 @@ def _read_channel(channel, stream, start, duration):
     channel_type = lalframe.FrStreamGetTimeSeriesType(channel, stream)
     read_func = _fr_type_map[channel_type][0]
     d_type = _fr_type_map[channel_type][1]
+    print(start, duration)
```
titodalcanton, line 82: I guess this was a debugging print, please remove before merging.

examples/multi_inspiral/check_gw170817_trigs.py
```diff
@@ -0,0 +1,34 @@
...
+        f['network/coherent_snr'][:],
+        f['network/reweighted_snr'][:],
+        ]
+# search for compatible trigs
+mask = (abs(gw170817_time - snrs[0]) < 0.1)
+n = mask.sum()
+status = 0
+if n > 0:
+    result = 'PASS'
+    status = 0
```
pannarale, line 27: This line can be removed

examples/multi_inspiral/faceon_faceaway.sh
```diff
@@ -0,0 +1,82 @@
...
+
+EVENT=1187008882
+STEP=300
+START=$((EVENT - 2 * STEP))
+END=$((EVENT + STEP + 64))
+GPS_START=$((START - 32))
+GPS_END=$((END + 32))
+DUR=$((GPS_END - GPS_START))
+CONFIG_URL=https://github.com/gwastro/pycbc-config/raw/master/test/multi_inspiral
+CONFIG_URL=https://github.com/a-r-williamson/pycbc-config/raw/pycbc_multi_inspiral_tests/test/multi_inspiral
```
pannarale, line 11: This should be removed

examples/multi_inspiral/run.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash -e
+
+CONFIG_URL=https://github.com/gwastro/pycbc-config/raw/master/test/multi_inspiral
+CONFIG_URL=https://github.com/a-r-williamson/pycbc-config/raw/pycbc_multi_inspiral_tests/test/multi_inspiral
```
pannarale, line 4: This line should be removed.

pannarale (changes requested):
Only minore requests.  Nice work!

pycbc/frame/frame.py
a-r-williamson: Removed, thanks.

pannarale (approved):

PR #3935, opened 2022-02-09
fix psd from txt

WuShichao:
@ahnitz Sometimes due to rounding errors, `vals[0]` will fall below the interpolation range, so force `vals[0]` to be `low_freq_cutoff`.

### DIFF
--- a/pycbc/filter/matchedfilter.py
+++ b/pycbc/filter/matchedfilter.py
@@ -1148,7 +1148,7 @@ def get_cutoff_indices(flow, fhigh, df, N):
     kmax: int
     """
     if flow:
-        kmin = int(numpy.ceil(flow / df))
+        kmin = int(flow / df)
         if kmin < 0:
             err_msg = "Start frequency cannot be negative. "
             err_msg += "Supplied value and kmin {} and {}".format(flow, kmin)
--- a/pycbc/filter/matchedfilter.py
+++ b/pycbc/filter/matchedfilter.py
@@ -1322,13 +1322,27 @@ def matched_filter(template, data, psd=None, low_frequency_cutoff=None,
     return snr * norm
 
 _snr = None
-def match(vec1, vec2, psd=None, low_frequency_cutoff=None,
-          high_frequency_cutoff=None, v1_norm=None, v2_norm=None,
-          subsample_interpolation=False, return_phase=False):
-    """ Return the match between the two TimeSeries or FrequencySeries.
+def match(
+    vec1,
+    vec2,
+    psd=None,
+    low_frequency_cutoff=None,
+    high_frequency_cutoff=None,
+    v1_norm=None,
+    v2_norm=None,
+    subsample_interpolation=False,
+    return_phase=False,
+):
+    """Return the match between the two TimeSeries or FrequencySeries.
 
     Return the match between two waveforms. This is equivalent to the overlap
     maximized over time and phase.
+
+    The maximization is only performed with discrete time-shifts,
+    or a quadratic interpolation of them if the subsample_interpolation
+    option is turned on; for a more precise computation
+    of the match between two waveforms, use the optimized_match function.
+    The accuracy of this function is guaranteed up to the fourth decimal place.
 
     Parameters
     ----------
@@ -1368,19 +1382,23 @@ def match(vec1, vec2, psd=None, low_frequency_cutoff=None,
     htilde = make_frequency_series(vec1)
     stilde = make_frequency_series(vec2)
 
-    N = (len(htilde)-1) * 2
+    N = (len(htilde) - 1) * 2
 
     global _snr
     if _snr is None or _snr.dtype != htilde.dtype or len(_snr) != N:
-        _snr = zeros(N,dtype=complex_same_precision_as(vec1))
-    snr, _, snr_norm = matched_filter_core(htilde, stilde, psd,
-                                           low_frequency_cutoff,
-                                           high_frequency_cutoff,
-                                           v1_norm, out=_snr)
+        _snr = zeros(N, dtype=complex_same_precision_as(vec1))
+    snr, _, snr_norm = matched_filter_core(
+        htilde,
+        stilde,
+        psd,
+        low_frequency_cutoff,
+        high_frequency_cutoff,
+        v1_norm,
+        out=_snr,
+    )
     maxsnr, max_id = snr.abs_max_loc()
     if v2_norm is None:
-        v2_norm = sigmasq(stilde, psd, low_frequency_cutoff,
-                          high_frequency_cutoff)
+        v2_norm = sigmasq(stilde, psd, low_frequency_cutoff, high_frequency_cutoff)
 
     if subsample_interpolation:
         # This uses the implementation coded up in sbank. Thanks Nick!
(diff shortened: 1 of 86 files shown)


tdent:
Unrelated to the change, but I don't understand why the line `kmin = int(low_freq_cutoff / delta_f)` occurs twice in this function.  I think one of these should be removed ...

tdent:
I guess it's also not clear why the interpolation has to be done over log frequency .. maybe if there was a comment on that aspect, the extra complexity would be more justifiable.

ahnitz:
@WuShichao Can you rebase this PR and make sure the unittests all pass?

ahnitz:
@MarlinSchaefer Can you verify if this resolves the issue you also saw?

tdent (changes requested):
Can the unnecessary repeat of `kmin = int(low_freq_cutoff / delta_f)` be removed (ie remove line 68) please?

MarlinSchaefer:
@WuShichao I've tried this PR on the minimal example I posted in #3952 and it did not resolve that issue. Can you try if the solution I suggested there solves your problem?

WuShichao:
@MarlinSchaefer I have updated the code with a simpler solution, which can solve both of our issues. Please try it.

MarlinSchaefer:
@WuShichao I've tried your solution and it works for me.

I still have one concern though. Does forcing the frequency values in the way this PR does not violate the underlying assumption of the FrequencySeries that all data-points are equally spaced. This solution forces a differing delta_f between the first and the second sample and the second to last and the last sample, right? I guess this happens on a sub-sample basis. Could this become an issue when only a few output samples are requested?

tdent (approved):
I'm happy with removing the duplicated line - other experts can comment on the non-equally-spaced frequency series aspect

MarlinSchaefer (changes requested):
@WuShichao I've checked against the issue (#3952) I posted. Due to my own error, that issue is ill-posed, as it requests a maximum frequency larger than what the initial data can provide. The code to check should actually be
```python
from pycbc.psd import aLIGOZeroDetHighPower, from_numpy_arrays

psd = aLIGOZeroDetHighPower(1024, 1, low_freq_cutoff=0)
psd = from_numpy_arrays(psd.sample_frequencies, psd, 2047, 0.5, 15)
```
This would be resolved by setting `kmin = int(numpy.ceil(low_freq_cutoff / delta_f))`.

However, in the current version of the PR this error is silently ignored and the last two frequency values in the `vals` array are identical, i.e. `vals[-2]-vals[-1]==0`.

I would suggest not setting `vals[-1]` and to add an explicit check for both the validity of the low frequency cutoff as well as the requested highest frequency. The first is already in the code. The second could look something like
```python
if (length - 1) * delta_f > freq_data[-1]:
    raise ValueError('Requested number of samples exceeds highest '
                     'available frequency in input data. '
                     f'(requested {(length - 1) * delta_f}, available '
                     f'{freq_data[-1]})')
```
In general I think setting the `vals` directly may be dangerous in certain situations and can ignore errors silently.

I'm also not sure why `data_start` is calculated in such a complicated way. Shouldn't `data_start = numpy.searchsorted(freq_data, low_freq_cutoff)` be sufficient? @ahnitz maybe you have more insight here?

WuShichao:
@MarlinSchaefer Thanks for your suggestions, I have updated my code, please check it.

MarlinSchaefer (approved):
This solved my problem and I don't see any immediate problems for other usecases with the code as it is.

WuShichao:
@MarlinSchaefer I also update the `kmin` in the [`get_cutoff_indices`](https://github.com/WuShichao/pycbc/blob/154f975be77a724a25c92687a887e72f3ae3b5f0/pycbc/filter/matchedfilter.py#L1151), unless the `flow` will be different in the optimal SNR calculation (will be divided by 0).

pycbc/filter/matchedfilter.py
```diff
@@ -1148,15 +1148,15 @@ def get_cutoff_indices(flow, fhigh, df, N):
     kmax: int
     """
     if flow:
-        kmin = int(flow / df)
+        kmin = int(numpy.ceil(flow / df))
```
ahnitz, line 1151: @WuShichao I think changing this could be somewhat dangerous in a lot of places. I'm not sure of all the impacts, so I'm a bit reticent to change this here.

pycbc/psd/read.py
```diff
@@ -48,7 +48,7 @@ def from_numpy_arrays(freq_data, noise_data, length, delta_f, low_freq_cutoff):
         raise ValueError('Lowest frequency in input data '
           ' is higher than requested low-frequency cutoff ' + str(low_freq_cutoff))
 
-    kmin = int(low_freq_cutoff / delta_f)
+    kmin = int(numpy.ceil(low_freq_cutoff / delta_f))
```
ahnitz, line 51: Thinking about this a bit more, perhaps there is an easier solution that would be in general nicer anyway. Would it make sense to allow the code to extrapolate up to a sample away from the data set at the boundaries?

WuShichao: @ahnitz I have considered this, I will test just fill `np.inf` first.

pycbc/filter/matchedfilter.py
WuShichao: @ahnitz Yes, this may have some impacts on other parts of PyCBC, but I think we need to use the same `kmin` in both places, otherwise, it will be divided by 0 in the `sigma` function (because `get_cutoff_indices` use a smaller `kmin`, and `read.py` uses `np.zeros` to initialize the psd.)

ahnitz: @WuShichao I agree it should be the same, but the original convention is used everywhere, so I don't think we want to change that.

WuShichao: @ahnitz Maybe we can initialize the PSD with `np.inf` instead of `np.zeros`.

WuShichao:
@ahnitz @MarlinSchaefer I've updated the code, now it allows to extrapolate one sample below or above the boundaries, and also allows using the max available frequency when the requested one is beyond the max available one (this is useful for very low mass CBC).

pycbc/psd/read.py
```diff
@@ -60,12 +60,19 @@ def from_numpy_arrays(freq_data, noise_data, length, delta_f, low_freq_cutoff):
     freq_data = freq_data[data_start:]
     noise_data = noise_data[data_start:]
 
+    if (length - 1) * delta_f > freq_data[-1]:
+        print('Requested number of samples exceeds the highest '
```
MarlinSchaefer, line 64: Not sure if this should be a print statement. How is this usually handled in PyCBC @ahnitz?

MarlinSchaefer (approved):
I think this solution is good and handles most cases.

pycbc/psd/read.py
titodalcanton: If this is a failure condition that should result in the termination of the program, raise a ValueError with this message. Otherwise, if it is just to provide information to the user, call `logging.info()`, `logging.warn()` or `logging.error()`.

ahnitz (approved):

PR #3936, opened 2022-02-10
PyCBC Live: stop uploading PSDs to GraceDB as separate files

titodalcanton:
In the past, the low-latency alert system required the noise PSD estimates associated with a candidate event to be uploaded to GraceDB as a separate LIGOLW file. As described in https://lscsoft.docs.ligo.org/ligo.skymap/interface.html and https://git.ligo.org/emfollow/gwcelery/-/merge_requests/850, we are now transitioning to the assumption that the noise PSDs are part of the original LIGOLW upload, and the second upload is no longer necessary.

This PR removes the code that does the separate PSD upload. It also reorganizes the operations that happen when a candidate is uploaded, so that the first operation is the upload, and the diagnostic plots are done as a second step.

Needs testing.

### DIFF
--- a/pycbc/io/live.py
+++ b/pycbc/io/live.py
@@ -322,6 +322,35 @@ class SingleCoincForGraceDB(object):
         # as GraceDB operations can fail later
         self.save(fname)
 
+        gid = None
+        try:
+            # try connecting to GraceDB
+            gracedb = GraceDb(gracedb_server) \
+                    if gracedb_server is not None else GraceDb()
+
+            # create GraceDB event
+            group = 'Test' if testing else 'CBC'
+            r = gracedb.createEvent(group, "pycbc", fname, search).json()
+            gid = r["graceid"]
+            logging.info("Uploaded event %s", gid)
+
+            if self.is_hardware_injection:
+                gracedb.writeLabel(gid, 'INJ')
+                logging.info("Tagging event %s as an injection", gid)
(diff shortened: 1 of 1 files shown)


ahnitz:
@titodalcanton Any push to move away from ligolw? The format itself ilso non-ideal for this kind of thing.

titodalcanton:
@ahnitz yeah, a while ago I started thinking about a possible HDF5-based file format to replace the LIGOLW document. I have to learn what I have to change on the GraceDB side though. Hopefully I will get to it in the next few months.

ahnitz:
@titodalcanton FYI, the tutorial test should work on a relaunch, sorry.

titodalcanton:
This is now tested on tens of uploads and appears to do the right thing.

titodalcanton:
Some before/after latency measurements here: https://ldas-jobs.ligo.caltech.edu/~tito.canton/pycbc/live/test_pr3936/plot.png

I am not understanding those results yet, so putting back the WIP for now.

titodalcanton:
The latency measurement above is perhaps not as encouraging as I was hoping for (I suspect the "before" curve may not be very well sampled) but at least this PR produces essentially the same result in terms of latency, and at least reduces the storage size associated with each event.

spxiwh (approved):
This all looks good to me. Thanks!

PR #3940, opened 2022-02-21
Un-merged PyGRB post-processing development

pannarale:
This PR refreshes work that had been carried out for the PyGRB post-processing and results presentation, but was never brought to master.

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -684,7 +684,7 @@ def distance_from_chirp_distance_mchirp(chirp_distance, mchirp, ref_mass=1.4):
 
 
 _detector_cache = {}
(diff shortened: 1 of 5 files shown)


pannarale:
Yeah, @ahnitz, I went over to @millsjc's fork and tried to bring to master with this PR some work we had done about a year ago.  I'm not sure what he had done with the c files, but those should not have been touched.  My plan is to clean things up: that should definitely reduce the volume of files.  It really should boil down to PyGRB plotting/post-processing scripts.

pannarale:
So, it looks like a script to fix indentation had been applied to the whole repo, therefore modifying basically all lines of C files as well.  I have restarted this PR anew.

a-r-williamson:
#3947

Issue #3942, opened 2022-02-21
markupsafe breaks jinja2


ahnitz:
We use jinja2 for some results page templating. It requires markupsafe which recently upgraded and is not longer compatible. 

see the discussion here https://github.com/pallets/jinja/issues/1585

ahnitz:
Ok, followup on this, it appears the issue is resolved in later releases of jinja2, we *don't* pin to an old release, however, it appears one of our dependencies is pulling in *flask* which does require an old version (hence causing everything to break).

ahnitz:
Ok, and the issue is Pegasus-wms pinning to an old flask version. 

https://pypi.org/project/pegasus-wms/5.0.1/#files

ahnitz:
I've suggested a PR to the Pegasus team here https://github.com/pegasus-isi/pegasus/pull/36 that would address this issue upstream, since we don't depend on flask (and hence pull in old jinja2 directly).

ahnitz:
I'm closing this is really a Pegasus platform requirement problem. There is no further change we can make on our end to avoid this.

PR #3943, opened 2022-02-21
Switch to minimal pegasus api requirement

ahnitz:
This is an attempt to address the immediate issue of Pegasus having a hard requirement on flask which no longer works due to #3942. In general, it's probably a good idea to keep our dependencies as lean as possible as well.

Assuming this passes tests, is there any reason we don't *just* depend on the API? Doe we actually need anything else for the PyCBC library directly?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -18,7 +18,12 @@ lscsoft-glue>=1.59.3
 ligo-segments
 tqdm
 gwdatafind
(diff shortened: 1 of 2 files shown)


spxiwh (approved):
Looks good! I think when I was testing this, it wasn't possible to just pull the API, but that is all we need, so if this works, then great!

PR #3944, opened 2022-02-21
revert to standard dqsegdb

ahnitz:
This is an attempt to go back to the standard dqsegdb which I had to fork temporarily to hack in python3 support. My understanding is the last release should support python3, so :crossed_fingers:


spxiwh (approved):
I'll just approve this, so it can be merged as soon as this works.

PR #3946, opened 2022-02-23
add in some alternative smoothing methods

GarethCabournDavies:
I wrote this up a while ago, but never actually made the PR in the end.

I found that sometimes the smoothing can be a bit over-aggressive, so I've added a couple of alternative methods to fit_sngls_over_multiparam:
 - smooth_area, which is the same as the current implementation and is the default
 - smooth_distance_weighted, which finds the templates within 3 of the unit area defined as normal, and the smooths them, weighted according to a normal distribution of width 1 (distance defined by --smoothing-width)
 - n_loudest, which counts the triggers in the nearest templates to the template of interest and stops when you reach a threshold (currently hardcoded to 500). This basically takes into account the fact that we are doing this because of small number statistics, and says that we only want to smooth templates if we really need to. This should reduce the cases of templates getting lots of triggers from a specific glitch morphology having their shallow fit parameter smoothed away.

### DIFF
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -95,7 +95,8 @@ def smooth_n_closest(nabove, invalphan, ntotal, dists, total_trigs=500):
     n_triggers = 0
     # Count number of templates required to gather total_trigs templates,
     # start at closest
-    while n_triggers < total_trigs:
+    # total_trigs, if supplied on command line will be a str so convert to int
+    while n_triggers < int(total_trigs):
         n_triggers += nabove[dist_sort[n_triggers]]
         templates_required += 1
     logging.debug("%d templates required to obtain %d triggers",
@@ -119,7 +120,7 @@ _smooth_dist_func = {
     'distance_weighted': smooth_distance_weighted
 }
 
-def smooth(nabove, invalphan, ntotal, dists, smoothing_method):
+def smooth(nabove, invalphan, ntotal, dists, smoothing_method, **kwargs):
     """
     Wrapper for smoothing according to a function defined by smoothing_method
 
@@ -130,7 +131,7 @@ def smooth(nabove, invalphan, ntotal, dists, smoothing_method):
     template of interest
     """
     return _smooth_dist_func[smoothing_method](nabove, invalphan,
-                                               ntotal, dists)
+                                               ntotal, dists, **kwargs)
 
 parser = argparse.ArgumentParser(usage="",
     description="Smooth (regress) the dependence of coefficients describing "
@@ -190,7 +191,7 @@ parser.add_argument("--smoothing-method", default="smooth_tophat",
                          "trucated at three smoothing-widths.")
 parser.add_argument("--smoothing-keywords", nargs='*',
                     help="Keywords for the smoothing function, supplied "
-                         "as key:value pairs, e.g. n_closest:500 to define "
+                         "as key:value pairs, e.g. total_trigs:500 to define "
                          "the number of templates in the n_closest smoothing "
                          "method")
 args = parser.parse_args()
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -1,6 +1,6 @@
 #!/usr/bin/python
 
-# Copyright 2016 Thomas Dent, Alex Nitz
+# Copyright 2016 Thomas Dent, Alex Nitz, Gareth Cabourn Davies
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by the
@@ -18,16 +18,18 @@ import sys, h5py, argparse, logging, pycbc.version, numpy
 from scipy.stats import norm
 from pycbc.events import triggers
 
+
 def dist(i1, i2, parvals, smoothing_width):
     """
-    computes the vector of parameter values at index i1 and
-    set of indices i2, and gives the Euclidean distance between
+    Computes the vector of parameter values at index/indices i1 and
+    index/indices i2, and gives the Euclidean distance between
     the two with a metric of 1/(smoothing width^2)
     """
     dsq = 0
     for v, s in zip(parvals, smoothing_width):
         dsq += (v[i2] - v[i1]) ** 2.0 / s ** 2.0
     return dsq ** 0.5
+
 
 def smooth_templates(nabove, invalphan, ntotal, template_idx,
                      weights=None):
@@ -74,6 +76,7 @@ def smooth_templates(nabove, invalphan, ntotal, template_idx,
                     ntotal_t_smoothed)
     return return_tuple
 
+
 def smooth_tophat(nabove, invalphan, ntotal, dists):
     """
     Smooth templates using a tophat function with templates within unit
@@ -84,6 +87,7 @@ def smooth_tophat(nabove, invalphan, ntotal, dists):
                             invalphan,
                             ntotal,
                             idx_within_area)
+
 
 def smooth_n_closest(nabove, invalphan, ntotal, dists, total_trigs=500):
     """
@@ -104,6 +108,7 @@ def smooth_n_closest(nabove, invalphan, ntotal, dists, total_trigs=500):
     idx_to_smooth = dist_sort[:templates_required]
     return smooth_templates(nabove, invalphan, ntotal, idx_to_smooth)
 
+
 def smooth_distance_weighted(nabove, invalphan, ntotal, dists):
     """
     Smooth templates weighted according to dists in a unit-width normal
@@ -120,6 +125,7 @@ _smooth_dist_func = {
     'distance_weighted': smooth_distance_weighted
 }
 
+
 def smooth(nabove, invalphan, ntotal, dists, smoothing_method, **kwargs):
     """
     Wrapper for smoothing according to a function defined by smoothing_method
@@ -132,6 +138,7 @@ def smooth(nabove, invalphan, ntotal, dists, smoothing_method, **kwargs):
     """
     return _smooth_dist_func[smoothing_method](nabove, invalphan,
                                                ntotal, dists, **kwargs)
+
 
 parser = argparse.ArgumentParser(usage="",
     description="Smooth (regress) the dependence of coefficients describing "
@@ -241,11 +248,10 @@ for param, slog in zip(args.fit_param, args.log_param):
     else:
         raise ValueError("invalid log param argument, use 'true', or 'false'")
 
-# for an exponential fit 1/alpha is linear in the trigger statistic values
-# so calculating weighted sums or averages of 1/alpha is appropriate
 nabove = fits['count_above_thresh'][:]
 ntotal = fits['count_in_template'][:]
-
+# For an exponential fit 1/alpha is linear in the trigger statistic values
+# so calculating weighted sums or averages of 1/alpha is appropriate
 invalpha = 1. / fits['fit_coeff'][:]
 invalphan = invalpha * nabove
 
@@ -255,6 +261,7 @@ alpha_smoothed = []
 rang = numpy.arange(0, len(nabove))
 
 logging.info("Smoothing ...")
+
 # Handle the one-dimensional case of tophat smoothing separately
 # as it is easier to optimize computational performance.
 if len(parvals) == 1 and args.smoothing_method == 'smooth_tophat':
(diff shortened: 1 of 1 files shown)


bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
 from __future__ import division
 
 import sys, h5py, argparse, logging, pycbc.version, numpy
+from scipy.stats import norm
 from pycbc.events import triggers
 
+def dist(i1, i2, parvals, smoothing_width):
```
GarethCabournDavies, line 21: This is just a move, as it was previously defined in-code closer to its use

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+    dsq = 0
+    for v, s in zip(parvals, smoothing_width):
+        dsq += (v[i2] - v[i1]) ** 2.0 / s ** 2.0
+    return dsq ** 0.5
+
+def smooth_templates(nabove, invalphan, ntotal, template_idx,
```
GarethCabournDavies, line 27: This is the function which others use to actually do the smoothing once the indices of the templates to be smoothed are found

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+    invalphan_mean = numpy.average(invalphan[template_idx], weights=weights)
+
+    # return nabove_smoothed, alpha_smoothed, ntotal_smoothed
+    return_tuple = (nabove_t_smoothed,
+                    nabove_t_smoothed / invalphan_mean,
+                    ntotal_t_smoothed)
+    return return_tuple
+
+def smooth_area(nabove, invalphan, ntotal, dists):
+    idx_within_area = numpy.flatnonzero(dists < 1.)
```
GarethCabournDavies, line 41: Hard cut on templates which are more than one unit of `--smoothing-width`-defined distance

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+                            ntotal,
+                            idx_within_area)
+
+def smooth_n_closest(nabove, invalphan, ntotal, dists):
+    dist_sort = numpy.argsort(dists)
+    templates_required = 0
+    n_triggers = 0
+    # Count number of templates required to gather 1000 templates,
+    # start at closest
+    while n_triggers < 500:
```
GarethCabournDavies, line 53: This should probably be un-hardcoded at some point (maybe using a smoothing-keywords option or something?), but I think 500 is probably around where it should be anyway

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -93,27 +157,16 @@ for param, slog in zip(args.fit_param, args.log_param):
     else:
         raise ValueError("invalid log param argument, use 'true', or 'false'")
 
-if 'count_in_template' in fits.keys(): # older files may not have this dataset
```
GarethCabournDavies, line 96: As we don't really need the backwards compatibility to before this was put in, these checks for whether the `count_in_template` is set isn't really needed anymore

ahnitz:
@GarethCabournDavies Is this ready to be reviewed? If so, I think there is not a discrepency between the 1-d and n-d case. It seems the 1-d is as before. If that's intended, then it should probably only work if the smoothing option is set to what it implements (i.e. the old default). Otherwise, I think it would need to be updated to account for some additional weighting functions (not clear if this can be done trivially).

GarethCabournDavies:
> @GarethCabournDavies Is this ready to be reviewed? If so, I think there is not a discrepency between the 1-d and n-d case. It seems the 1-d is as before. If that's intended, then it should probably only work if the smoothing option is set to what it implements (i.e. the old default). Otherwise, I think it would need to be updated to account for some additional weighting functions (not clear if this can be done trivially).

Ah, thanks. I'd missed 1-D out before as was only using it for N-D, but that makes sense.

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
tdent: Might be good to have a line or two of comment.  It looks like it computes the vector of parameter values at two indices (or two sets of indices?) i2, i1 and gives the Euclidean distance between the two with a metric of 1/(smoothing width^2).

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+    for v, s in zip(parvals, smoothing_width):
+        dsq += (v[i2] - v[i1]) ** 2.0 / s ** 2.0
+    return dsq ** 0.5
+
+def smooth_templates(nabove, invalphan, ntotal, template_idx,
+                      weights=None):
+    if weights is None: weights = numpy.ones_like(template_idx)
+    nabove_t_smoothed = numpy.average(nabove[template_idx], weights=weights)
+    ntotal_t_smoothed = numpy.average(ntotal[template_idx], weights=weights)
+    invalphan_mean = numpy.average(invalphan[template_idx], weights=weights)
```
tdent, line 32: maybe add a comment here to explain that the max likelihood fit for 1/alpha is linear in the trigger statistic values, so we perform an average of (n_above / alpha) over templates and then invert this and multiply by (smooth) nabove to obtain smoothed alpha

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
tdent: If you need to explain here, add a line of comment to the code. (Though I'd prefer not to mention how the distances were defined, as that is not relevant to what this function does - it just takes whatever 'dist' values it is given.)

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+    return return_tuple
+
+def smooth_area(nabove, invalphan, ntotal, dists):
+    idx_within_area = numpy.flatnonzero(dists < 1.)
+    return smooth_templates(nabove,
+                            invalphan,
+                            ntotal,
+                            idx_within_area)
+
+def smooth_n_closest(nabove, invalphan, ntotal, dists):
```
tdent, line 47: as above, prefer to have 1 sentence or so of docstring

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
tdent: yes, simplest way to enable unhardcoding would be to have eg `total_trig=500` in the function definition, I would prefer that ..

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+    dist_sort = numpy.argsort(dists)
+    templates_required = 0
+    n_triggers = 0
+    # Count number of templates required to gather 1000 templates,
```
tdent, line 51: this should be 'required to gather at least [500/total_trig] triggers' I guess

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+    # start at closest
+    while n_triggers < 500:
+        n_triggers += nabove[dist_sort[n_triggers]]
+        templates_required += 1
+    logging.debug("%d templates required to obtain %d triggers", templates_required, n_triggers)
+    idx_to_smooth = dist_sort[:templates_required]
+    return smooth_templates(nabove, invalphan, ntotal, idx_to_smooth)
+
+def smooth_distance_weighted(nabove, invalphan, ntotal, dists):
```
tdent, line 60: as above, can we get a sentence or two of docstring?

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+    idx_within_area = numpy.flatnonzero(dists < 3.)
+    weights = norm.pdf(dists[idx_within_area])
```
tdent, line 62: This looks like it applies a one-dimensional standard normal distribution to the distances (truncated to '3 sigma'), correct?

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+    ntotal_t_smoothed = numpy.average(ntotal[template_idx], weights=weights)
+    invalphan_mean = numpy.average(invalphan[template_idx], weights=weights)
+
+    # return nabove_smoothed, alpha_smoothed, ntotal_smoothed
+    return_tuple = (nabove_t_smoothed,
+                    nabove_t_smoothed / invalphan_mean,
+                    ntotal_t_smoothed)
+    return return_tuple
+
+def smooth_area(nabove, invalphan, ntotal, dists):
```
tdent, line 40: Maybe the function name is a bit vague .. how about 'smooth_tophat' as this is a top hat window in some coordinates. 
Would be good to have 1 sentence or so of docstring explaining the format of the inputs - I think the first 3 are arrays covering the full set of templates and the last is an array - also over the full set of templates? - giving the distances of each to the (implicit) one template for which smoothed values are being made.

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -62,6 +118,14 @@ parser.add_argument("--smoothing-width", type=float, nargs='+', required=True,
                          "logs of them) to smooth over. Required. "
                          "This must be a list corresponding to the smoothing "
                          "parameters.")
+parser.add_argument("--smoothing-method", default="smooth_area",
+                    choices = _smooth_dist_func.keys(),
+                    help="Method used to smooth the fit parameters; "
+                         "'smooth_area' (default) finds all templates within "
+                         "unity distance from the template of interest "
```
tdent, line 125: -> 'unit' distance I think - you also should mention the 'weighted' method in this help string

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -15,8 +15,64 @@
...
+    return dsq ** 0.5
+
+def smooth_templates(nabove, invalphan, ntotal, template_idx,
+                      weights=None):
+    if weights is None: weights = numpy.ones_like(template_idx)
+    nabove_t_smoothed = numpy.average(nabove[template_idx], weights=weights)
+    ntotal_t_smoothed = numpy.average(ntotal[template_idx], weights=weights)
+    invalphan_mean = numpy.average(invalphan[template_idx], weights=weights)
+
+    # return nabove_smoothed, alpha_smoothed, ntotal_smoothed
```
tdent, line 34: I think this comment won't be needed if the 'inverse alpha' thing is explained in a separate comment

tdent (changes requested):

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
GarethCabournDavies: Done

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
GarethCabournDavies: Correct, and clarified in the docstring

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
GarethCabournDavies: Changed as suggested, docstring added

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
GarethCabournDavies: Outputs are now explained in docstring

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
GarethCabournDavies: added

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
GarethCabournDavies: Added in docstring explanation of the function

GarethCabournDavies:
> @GarethCabournDavies Is this ready to be reviewed? If so, I think there is not a discrepency between the 1-d and n-d case. It seems the 1-d is as before. If that's intended, then it should probably only work if the smoothing option is set to what it implements (i.e. the old default). Otherwise, I think it would need to be updated to account for some additional weighting functions (not clear if this can be done trivially).

I tried to look at implementing the other options, but then got frustrated so I haven't done that, and so the efficient case is only used if using one dimension *and* tophat smoothing

tdent (approved):
Seems OK, have added a code author and done minor codeclimate 'improvements'

PR #3949, opened 2022-02-25
Support for generating time varying noise 

sum33it:
A support for generating a time varying noise is added. With the help of an additional argument `scale` (default = 1.0), one can generate a time varying noise where total time can be broken into smaller equal chunks (time bins) and an array of scale values for each bin can be provided which act as the time varying function. Then time series generated from all the bins can be stacked together. The seed should be same for all the bins.

While generating the coloured noise, white noise is multiplied with the ASD. The `scale` argument is multiplicative factor to ASD. A more general frequency dependent function can be added in future.

### DIFF
--- a/pycbc/noise/reproduceable.py
+++ b/pycbc/noise/reproduceable.py
@@ -86,7 +86,8 @@ def normal(start, end, sample_rate=16384, seed=0):
 def colored_noise(psd, start_time, end_time,
                   seed=0, sample_rate=16384,
                   low_frequency_cutoff=1.0,
-                  filter_duration=128):
+                  filter_duration=128,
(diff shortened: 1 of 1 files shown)


titodalcanton:
How related is this to #3825?

sum33it:
@titodalcanton This is a parallel approach to #3825 which requires minimal changes in current infrastructure to produce a simple time varying PSD.

sum33it:
@ahnitz Poke.

ahnitz (approved):

ahnitz:
@sum33it please reabse and fix the history

ahnitz (approved):

PR #3954, opened 2022-03-02
Update distribution.yml

ahnitz:


ahnitz:
This updates our wheel construction to include python3.10 and arm64 builds for MacOSX. The latter should *hopefully* at least catch if we break the m1 builds for conda for example before it gets to release point. I don't know if we can do any real testing though since I believe the builds are done by cross-compilation from an x86 test runner. 

not included here, but maybe for the future would be to enable arm64 builds for linux as well.

spxiwh (approved):
The python 3.9 wheel you sent seems to work on my ARM venv, so approving. Thanks!

PR #3957, opened 2022-03-04
Fix for py3 dict keys

tdent:
Minor fix to deal with dict keys() no longer being a sequence.

### DIFF
--- a/bin/pycbc_merge_inj_hdf
+++ b/bin/pycbc_merge_inj_hdf
@@ -28,6 +28,7 @@ import pycbc
 import pycbc.inject
 import pycbc.version
 
+
 def get_gc_end_time(injection):
     """Return the geocenter end time of an injection. Required for seamless
     compatibility with LIGOLW and HDF injection objects, which use different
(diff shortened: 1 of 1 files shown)


bin/pycbc_merge_inj_hdf
```diff
@@ -65,7 +65,7 @@ if __name__ == '__main__':
         injections = pycbc.inject.InjectionSet(inj_file)
 
         inj_table = injections.table
-        inj_table = inj_table[np.sort(inj_table.dtype.fields.keys())]
+        inj_table = inj_table[np.sort(list(inj_table.dtype.fields.keys()))]
```
tdent, line 68: It's probably a good idea to put in a comment saying why this sort is needed (which I am not quite sure of).

tdent:
(Tested by @PRAVEEN-mnl in a workflow )

spxiwh (approved):
Thanks! I'm happy if this fixes the problem. Extra comments are always welcome, but not a barrier to approval. Merge when ready.

PR #3971, opened 2022-03-16
Offline singles BIG changes

GarethCabournDavies:
Changes for the including singles into the offline search as described in the paper by myself and Ian

In the same vein as the way that the multiifo search was kept separate until it was more stable, I've tried to keep everything as a different search, but obviously there is a lot which is reused

I've tried to add a small search test as well, as a result, I think we need to have the PR up, though this is draft at the moment until I can convince the tests to work

PR #3975, opened 2022-03-22
Cuts to triggers/templates for use in findtrigs codes

GarethCabournDavies:
During the offline singles FAR search, we found that we wanted to apply cuts to the triggers and templates in order to remove obvious noise / hopeless for recovery events from the sngls_findtrigs code

This change allows a standard method of cutting triggers or templates from the code in a semi-efficient manner

Triggers can be cut on:
 - Any single-detector ranking
 - Any chisq from the chisq_choices in io/hdf.py (followed by `_chisq`)
 - `psd_var_val`, `end_time` or `sigmasq` read from the trigger_merge file

Templates can be cut on:
 - A subset of conversions from the template bank, as newly defined in `pycbc/bank/conversions.py`
 - Smoothed or unsmoothed template fit parameters (N.B. unsmoothed requires a new option to fit_over_multiparam)

To use this, one would run, eg. 
```
pycbc_coinc_findtrigs \
  $other_arguments \
  --template-cuts \
    fit_by_fit_coeff:4.5:lower \
    chirp_mass:9:upper \
  --trigger-cuts \
    snr:5.5:lower \
    traditional_chisq:10:upper
```

This would set e.g. an upper limit of 9 on the chirp mass of the template


GarethCabournDavies:
I havent yet put the changes needed into the sngls_findtrigs code, as that will be a part of the changes in #3971

bin/all_sky_search/pycbc_fit_sngls_by_template
```diff
@@ -176,7 +176,11 @@ abovethresh, stat = get_stat(args, trigf[args.ifo], args.stat_threshold)
 tid = trigf[args.ifo + '/template_id'][abovethresh]
 time = trigf[args.ifo + '/end_time'][abovethresh]
 if args.save_trig_param:
-    tparam = trigf[args.ifo + '/' + args.save_trig_param][abovethresh]
+    tparam_region = trigf[args.ifo][args.save_trig_param + '_template'][:]
+    # Note: this breaks if there are no triggers in the template, but that
```
GarethCabournDavies, line 180: Though I have pointed out a case where this can break, it is better than before, as it was breaking if there were no triggers _above threshold_ in the template, rather than no triggers at all

GarethCabournDavies:
Sorry for all of the extra commits (disentangling from other work was less obvious than expected), but I think this is ready to be reviewed now

GarethCabournDavies:
Poke on this - codeclimate things look irrelevant

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -224,12 +236,21 @@ def process_template(tnum):
...
-        tids_full[i] = sngl.set_template(tnum)
-        times_full[i] = sngl['end_time']
-        logging.info('%s:%s' % (i, len(tids_full[i])))
+        # Apply single-trigger cuts as defined on the command line
+        tids_uncut = sngl.set_template(tnum)
+        trigger_keep_ids = cuts.apply_trigger_cuts(sngl, trigger_cut_dict)
+
+        tids_full[i] = tids_uncut[trigger_keep_ids]
+        times_full[i] = sngl['end_time'][trigger_keep_ids]
+        logging.info('%s:%s' % (i, len(tids_uncut)))
```
titodalcanton, line 245:
```suggestion
        logging.info('%s:%s', i, len(tids_uncut))
```

pycbc/bank/__init__.py
```diff
@@ -0,0 +1,6 @@
+"""
+This packages contains modules for dealing with the template bank
+"""
+
+
+from .conversions import *
```
titodalcanton, line 6: What is the point of this import?

GarethCabournDavies: I think this means that you can do "from pycbc import bank" and then use "bank.conversions" - however I've never fully got the knack of this kind of thing

GarethCabournDavies:
Poke on this

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+"""
+import logging
+import copy
+import numpy as np
+from pycbc.events import ranking
+from pycbc.io import hdf
+from pycbc.bank import conversions as bank_conv
+from pycbc.io import get_chisq_from_file_choice
+
+sngl_rank_keys = ranking.sngls_ranking_function_dict.keys()
```
tdent, line 37: maybe a line of comment here saying what sngl_rank_keys are going to represent?

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+
+template_fit_param_choices = ['fit_by_fit_coeff', 'smoothed_fit_coeff',
+                              'fit_by_count_above_thresh',
+                              'smoothed_fit_count_above_thresh',
+                              'fit_by_count_in_template',
+                              'smoothed_fit_count_in_template']
+template_param_choices = bank_conv.conversion_options + \
+                             template_fit_param_choices
+
+ineq_choices = ['upper', 'lower', 'upper_inc', 'lower_inc']
```
tdent, line 51: this seems redundant with ineq_functions.keys, no?  Would avoid some duplication to define the dictionary first IMO.

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+
+def insert_cuts_option_group(parser):
+    """
+    Add options to the parser for cuts to the templates/triggers
+    """
+    parser.add_argument('--trigger-cuts', nargs='+',
+                        help="Cuts to apply to the triggers, supplied as "
+                             "PARAMETER:VALUE:LIMIT, where, PARAMETER is the "
+                             "parameter to be cut, VALUE is the value at "
+                             "which it is cut, which must be convertible to "
```
tdent, line 62: Given that there is an error message for 'not convertible to float', you could IMO leave out this part of the help. 
(I doubt there is a case where someone might want to supply a value that was not float-able anyway ..)

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+                             "a float, and LIMIT is one of '"
+                             + "', '".join(ineq_choices) +
+                             "' to indicate the inequality needed. "
+                             "PARAMETER is one of:'"
+                             + "', '".join(trigger_param_choices) +
+                             "'. For example snr:6:LOWER places the "
+                             "requirement that matched filter SNR must be > 6")
```
tdent, line 69: It's not completely clear if triggers with SNR>6 are retained, or are cut.
I suppose it means they are kept, so I suggest '.. removes triggers if matched filter SNR < 6'

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+                        help="Cuts to apply to the triggers, supplied as "
+                             "PARAMETER:VALUE:LIMIT. Format is the same as in "
+                             "--trigger-cuts. PARAMETER can be one of '"
+                             + "', '".join(template_param_choices) + "'.")
+
+
+# What are the inequalities associated with the cuts?
+# 'upper' means upper limit, and so requires value < the threshold
+ineq_functions = {
+    'upper': np.less,
```
tdent, line 80: Again, it's not totally clear if triggers for which the inequality is true are kept or cut.  It looks like they are kept.  So 'upper' means apply a cut at the top end of the range, etc. 
Then I'd suggest '.. requires value < threshold to keep a trigger'

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+    'lower_inc': np.greater_equal
+}
+
+
+def convert_inputstr(inputstr, choices):
+    """
+    Convert the inputstr into a dictionary keyed on parameter
+    with a tuple of the function to be used in the cut, and
+    the float to compare to.
+    Do some checks as to whether the inputstr fits input requirements
```
tdent, line 92: 'Do input checks' is conciser

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+    Check that the inputs given to options in insert_cuts_option_group
+    are sensible, and return the objects used to handle the cuts.
+    """
+    # Deal with the case where no cuts are supplied:
+    if not args.trigger_cuts and not args.template_cuts:
+        return {}, {}
+
+    # So that we can deal with the case where one set of cuts is supplied
+    # but not the other, assign an empty list if no argument given
+    trigger_cut_strs = args.trigger_cuts if args.trigger_cuts else []
```
tdent, line 132: would `args.trigger_cuts or []` also work?  Think I'd prefer this as less repetitive

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+        raise value_e
+
+    return {cut_param: (ineq_functions[cut_limit],
+                        float(cut_value_str))}
+
+
+def ingest_cuts_option_group(args):
+    """
+    Check that the inputs given to options in insert_cuts_option_group
+    are sensible, and return the objects used to handle the cuts.
```
tdent, line 124: I don't think this function does input checking, it just does some housekeeping between trigger and template cut options. So remove comment about check sensible inputs & say eg 'return dicts for trigger and template cuts' ?

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+    """
+    # Deal with the case where no cuts are supplied:
+    if not args.trigger_cuts and not args.template_cuts:
+        return {}, {}
+
+    # So that we can deal with the case where one set of cuts is supplied
```
tdent, line 130: It's self-evident that an empty list is being assigned, so just eg `Deal with the case where one set of cuts is supplied but not the other`

pycbc/events/cuts.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+def apply_trigger_cuts(triggers, trigger_cut_dict):
+    """
+    Fetch/Calculate the parameter for triggers, and then
+    apply the cuts defined in template_cut_dict
+
+    Parameters
+    ----------
+    triggers:
```
tdent, line 156: missing an explanation here ?

pycbc/events/cuts.py
GarethCabournDavies: That's it! I knew there was an easier/more obvious way to do this but couldn't remember it - thanks

pycbc/events/cuts.py
GarethCabournDavies: Thanks - missed before and updated

pycbc/events/cuts.py
```diff
@@ -0,0 +1,347 @@
...
+def apply_trigger_cuts(triggers, trigger_cut_dict):
+    """
+    Fetch/Calculate the parameter for triggers, and then
+    apply the cuts defined in template_cut_dict
+
+    Parameters
+    ----------
+    triggers: ReadByTemplate object
+        The triggers in this particular template. This
+        must have the correct datasets required tio calculate
```
tdent, line 156: typo 'tio'. I'm not sure if the comment about 'okay in standard searches' is going to be informative .. if the function works people won't need the comment, if it doesn't work then it probably won't help with debugging

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -184,6 +187,15 @@ else:
 if tids_with_trigs is not None:
     template_ids = numpy.intersect1d(tids_with_trigs, template_ids)
 
+# Apply cuts as defined on the command line
```
tdent, line 190: not sure why 'as defined on the command line' .. suggest just 'Apply cuts to templates'

pycbc/bank/conversions.py
```diff
@@ -0,0 +1,114 @@
...
+            # This statement should be the reached only if 'duration'
+            # is given, but 'template_duration' is in the bank
+            values = bank['template_duration'][:][template_ids]
+        else:
+            values = pnutils.get_imr_duration(bank['mass1'][:][template_ids],
+                                              bank['mass2'][:][template_ids],
+                                              bank['spin1z'][:][template_ids],
+                                              bank['spin2z'][:][template_ids],
+                                              bank['f_lower'][:][template_ids],
+                                              approximant="SEOBNRv4")
```
tdent, line 97: Not sure if SEOBNRv4 ought to be hard-coded here - in principle I think it should be allowed to be a kwarg, eg `def bank_conversion(parameter, bank, template_ids, duration_approximant='SEOBNRv4'):`

bin/all_sky_search/pycbc_fit_sngls_by_template
tdent: OK ... but how about putting a try-except to catch the no triggers case?

pycbc/bank/conversions.py
```diff
@@ -0,0 +1,114 @@
...
+
+spin_conversions = {
+    'chieff': conv.chi_eff,
+    'chi_eff': conv.chi_eff,
+    'effective_spin': conv.chi_eff,
+    'chi_a': conv.chi_a
+}
+
+
+def bank_conversion(parameter, bank, template_ids):
```
tdent, line 58: Think I'd prefer to call this something less similar to the name of the package and module. Eg `def get_bank_property()` ?

tdent:
Do we want two modules called conversions.py, both at the pycbc/ level and in pycbc/bank/ ?  It seems potentially confusing unless there is a good reason for them to be the same.  Could call this one bank_conversions.py for instance ...

bin/all_sky_search/pycbc_fit_sngls_by_template
GarethCabournDavies: Added now

tdent (approved):

tdent:
Not sure why you are getting failed checks here, but if that is understood/fixed this seems OK to go from my side

PR #3979, opened 2022-03-24
banish the six library

ahnitz:


ahnitz:
@spxiwh There is one new FIXME because I found a function that no longer actually could be reached. If we want that re-implemented, it would need to be done in a separate PR.

spxiwh (commented):
Thanks for pulling out all the six imports!

I would not like to disable the "check that we didn't get redirected to LIGO log in page" check ... The check actually still works as far as I know, but can [possibly] do weird things if a binary file is downloaded, and the bit where we checked if it was binary or not was broken in python3. Given that we've had this working in python3 for a bit of time now, I'd be happier to just remove the "istext" check and always run the lines below, unless we have an example where downloading a binary file causes issues?

ahnitz:
@spxiwh Ok, good point. I've reverted that, but left in the message just for the 'istext' check in case that is wanted later.

spxiwh (approved):
Thanks, this is ready to merge now from my perspective. Codeclimate is reporting one weird thing, which appears unrelated to this, but might be worth checking out to make sure something didn't get broken at some previous point(?)

ahnitz:
@spxiwh This line should actually be tested in the tutorial notebooks (one of the one's Collin wrote). I think what is happening here is a misidentification by CC given that the base type is a dict, but it has setter / getter methods for the property.

PR #3980, opened 2022-03-28
Results page artefact location

GarethCabournDavies:
The artefact for the results page wasn't being uploaded from the test search, but it also looked like it wasn't being put into the folder where it was expected to be.

This fixes the issue, but I'm not sure where the results page was being made originally


GarethCabournDavies:
Output here: https://github.com/gwastro/pycbc/actions/runs/2052024010, where the results page can be downloaded

PR #3985, opened 2022-03-31
fix for dynesty reflect rename

ahnitz:


sum33it (approved):

ahnitz:
This should address two issues in our codebase related to the updated version 1.2.0 of dynesty. For additional details see https://github.com/joshspeagle/dynesty/issues/362 and https://github.com/joshspeagle/dynesty/issues/361. Some changes outside those discussions were also required due to the dropping of the "kwargs" argument from the sampler. 

Changes to our codebase should enable compatibility for <1.1 and 1.2 versions simultaneously so I don't believe we need to add a hard minimum for dynesty yet at 1.2.

PR #3986, opened 2022-03-31
Sngls findtrigs update

GarethCabournDavies:
Separating out the updates to sngls_findtrigs from the offline singles PR, as it makes things a little easier to digest

- Update the sngls_findtrigs code to be more in line with coinc_findtrigs, and therefore more obvious to implement similar things to both codes
- The clustering was previously performed over all triggers, this is now done only on a per-template basis.
- The output files are then the same except for a couple of things:
   - The order of triggers changes, as we move from trigger_id order to template_id order
   - trigger_id is now a float64 rather than int64: no real reason why or why not to do this
   - The attribute timeslide_interval is now given (always zero)

### DIFF
--- a/bin/all_sky_search/pycbc_sngls_findtrigs
+++ b/bin/all_sky_search/pycbc_sngls_findtrigs
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-import argparse, logging, numpy as np
+import argparse, logging, h5py, numpy as np
 from ligo.segments import infinity
 from numpy.random import seed, shuffle
 from pycbc.events import veto, coinc, stat
@@ -94,17 +94,11 @@ trigs = ReadByTemplate(trigger_file,
                        args.gating_veto_windows)
 logging.info("%d triggers in file", trigf[ifo + '/snr'].size)
 
-data_init = {}
-#FIXME - this feels like it should be easier to define
-all_dsets = ['sigmasq', 'chisq', 'chisq_dof', 'coa_phase', 'end_time',
-             'snr', 'template_id', 'sg_chisq', 'psd_var_val']
-
-for ds in all_dsets:
-    data_init[ds] = []
-data_init['trigger_id'] = []
-
-trigs_out = io.DictArray(data=data_init)
 stat_all = []
+trigger_ids_all = []
+template_ids_all = []
+trigger_times_all = []
+
 
 rank_method = stat.get_statistic_from_opts(args, [ifo])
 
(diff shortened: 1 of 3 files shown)


GarethCabournDavies:
Having changed from the dictarray method (which was useful but had to have hard-coded datasets), the statistic values are now different, but all match to within numpy.isclose with default values

GarethCabournDavies:
Poke

bin/all_sky_search/pycbc_sngls_findtrigs
```diff
@@ -1,152 +1,152 @@
...
 args = parser.parse_args()
 
-if (args.veto_files and not args.segment_names) or \
-    (args.segment_names and not args.veto_files):
-    raise RuntimeError('--veto-files and --segment-names are mutually required')
+trigger_file = args.trigger_files[0]
+
+if (args.veto_files and not args.segment_name) or \
+    (args.segment_name and not args.veto_files):
+    raise RuntimeError('--veto-files and --segment-name are mutually required')
```
titodalcanton, line 55:
```suggestion
    parser.error('--veto-files and --segment-name are mutually required')
```

bin/all_sky_search/pycbc_sngls_findtrigs
```diff
@@ -1,152 +1,152 @@
...
+
+if not len(args.veto_files) == len(args.segment_name):
+    raise RuntimeError('--segment-name optionss are required for each --veto-files')
```
titodalcanton, line 58:
```suggestion
if len(args.veto_files) != len(args.segment_name):
    parser.error('--segment-name options are required for each --veto-files')
```

bin/all_sky_search/pycbc_sngls_findtrigs
```diff
@@ -1,152 +1,152 @@
...
-logging.info("Clustering")
-if args.cluster_window:
-    cid = coinc.cluster_over_time(stat, trigs.data['end_time'],
-                                  args.cluster_window)
-    trigs = trigs.select(cid)
-    stat = stat[cid]
-    logging.info("%d triggers after clustering", stat.size)
+num_templates = io.HFile(args.template_bank, "r")['template_hash'].size
+tmin, tmax = parse_template_range(num_templates, args.template_fraction_range)
+logging.info('Analyzing template %s - %s' % (tmin, tmax-1))
```
titodalcanton, line 80:
```suggestion
logging.info('Analyzing template %s - %s', tmin, tmax-1)
```

GarethCabournDavies:
Poke

GarethCabournDavies:
Thanks!

PR #3988, opened 2022-03-31
fix distributions import error

soumide1102:
Currently `pycbc_inference` throws an error:
`AttributeError: module 'pycbc' has no attribute 'distributions'`

Fixing how distributions is imported in distributions/utils.py.

### DIFF
--- a/pycbc/distributions/utils.py
+++ b/pycbc/distributions/utils.py
@@ -31,6 +31,7 @@ from pycbc.types.config import InterpolatingConfigParser
 from pycbc import transforms
 from pycbc import distributions
 
+
 def draw_samples_from_config(path, num=1, seed=150914):
     r""" Generate sampling points from a standalone .ini file.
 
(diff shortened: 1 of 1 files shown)


ahnitz (commented):
Why would this change do anything?

soumide1102:
@ahnitz I produced that error when running pycbc_inference by hand. I'm mystified as well as to why `import pycbc.distributions as distributions` throws an error. But changing it to `from pycbc import distributions` worked and I could complete a pycbc_inference test with that.

soumide1102:
@ahnitz I reverted the change and the error reappeared. 
(Someone else working with me got that error as well on the NERSC cluster.)

ahnitz:
@soumide1102 Can we follow this up in the slack? I'd like to understand what is happening here

soumide1102:
sure.

ahnitz (approved):
Ok, I was able to reproduce this by hacking pycbc to install into python3.6 (not a supported python version any longer). In this scenario, we hit he issue only resolved in python3.7 (see https://bugs.python.org/issue30024) related to how circular imports are resolved. 

The change is innocuous for later python versions so I'm happy to approve this.

PR #3992, opened 2022-04-05
Generate fake frames

maxtrevor:
Created an executable for generating fake frames with state and dq vectors, and idq time series. I intend to use this executable in a future update of the pycbc live example.


titodalcanton:
@maxtrevor I am not sure adding this script to the main repo is the best way forward. At the very least, this duplicates a lot of functionality already existing in `pycbc_condition_strain`. Since we are very early in the integration of iDQ into PyCBC Live, would you mind keeping this script away from the repo for now? I would like to better understand the issue you are having with the other script before deciding that having this in `/bin` is the best way. Apologies but I do not have a lot of time x energy to think about this at the moment.

maxtrevor:
To reproduce that error, run the pycbc_live example somewhere. Take the frame file `H1-SIMULATED_STRAIN-1272790000-32.gwf` it produces. Then run the add_state_dq_vectors.py (the master version) like this:
```
python add_state_dq_vectors.py \
    --input-file H1-SIMULATED_STRAIN-1272790000-32.gwf \
    --strain-channel H1:SIMULATED_STRAIN \
    --output-file ./H1_strain.gwf
```
It gives an error:
```
  File "/home/max.trevor/pycbc_live_dev/pycbc-live-utils/add_state_dq_vectors.py", line 96, in <module>
    write_frame(args.output_file, out_channel_names, out_timeseries)
  File "/home/max.trevor/.conda/envs/pycbc-live-mpich/lib/python3.9/site-packages/pycbc/frame/frame.py", line 404, in write_frame
    raise ValueError("Start and end times of TimeSeries must be integer seconds.")
ValueError: Start and end times of TimeSeries must be integer seconds.
```
Adding a few print statements, the reason is that the strain time series after being read from the gwf now has an ending time of 1272790031.999938965 instead of 1272790032

maxtrevor:
> @maxtrevor I am not sure adding this script to the main repo is the best way forward. At the very least, this duplicates a lot of functionality already existing in `pycbc_condition_strain`. Since we are very early in the integration of iDQ into PyCBC Live, would you mind keeping this script away from the repo for now? I would like to better understand the issue you are having with the other script before deciding that having this in `/bin` is the best way. Apologies but I do not have a lot of time x energy to think about this at the moment.

I have the frame files I currently need for testing the pycbc_live implementation, so no action is needed in the short term. But long term I think we need some functionality in the main repo for generating test frames with more than one channel. After the idq implementation in live is more mature, I plan to add at least an idq channel into the live example, and I don't think we would want to be forced to use a script from pycbc-live-utils for that.

PR #3993, opened 2022-04-05
make rounding more precise in a few places, allow read_frame be less …

ahnitz:
This updates a few locations to produce more dt / df preservation. They weren't wrong per, but if used in codes that don't allow eps rounding then this allow higher interoperability. In particular the lalframe frame reading would ignore the last sample when reading. This caused downstream issues for @maxtrevor (see https://github.com/gwastro/pycbc/pull/3992#issuecomment-1088999391)


ahnitz:
@titodalcanton Poke?

titodalcanton (approved):

PR #3998, opened 2022-04-08
Add function to reconstruct marginalized distance or polarization, feature added to script

ahnitz:
Depends on #3996 

This adds a function to reconstruct the marginalized distance or polarization value of a model. The functionality is added to pycbc_inference_model_stats with an additional argument '--reconstruct-parameters'. 

A few fixes also in place to enable measure the phase for single / relative models.

### DIFF
--- a/bin/inference/pycbc_inference_model_stats
+++ b/bin/inference/pycbc_inference_model_stats
@@ -53,6 +53,8 @@ parser.add_argument("--config-file", nargs="+", type=str, default=None,
                           "with the given file(s).")
 parser.add_argument("--verbose", action="store_true", default=False,
                     help="Print logging messages.")
+parser.add_argument('--reconstruct-parameters', action="store_true",
+                    help="Reconstruct marginalized parameters")
 
 
 # parse command line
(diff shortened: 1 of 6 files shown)


ahnitz:
@cdcapano At the moment, I've added the 'reconstruction' functionality to the pycbc_inference_model_stats script since that was the most similar to what was needed. That may or may not be the most ideal place. 

This PR leaves this as an optional postprecessing step (so nothing automatic within pycbc_inspiral does the reconstruction for you). The latter would need some modification of the sample base class to better handle what the 'finalize' method should do so that maybe it can essentially do what is in this script itself. That might require some restructuring.

bin/inference/pycbc_inference_model_stats
```diff
@@ -92,20 +94,24 @@ model.sampling_transforms = None
...
-# these help for parallelization
+    model.update(**{p: paramvals[p] for p in model.variable_params})
+    _ = model.logposterior    
+    stats = model.get_current_stats()
+    
+    rec = {}
+    if opts.reconstruct_parameters:
+        model.update(**{p: paramvals[p] for p in model.variable_params})
+        rec = model.reconstruct()
+        print(rec)
```
cdcapano, line 106: Did you mean for this print statement to stay here?

cdcapano (approved):
Ok, looks fine to me. Just check that you don't have print statement left behind from debugging. I'm ok with doing this in model stats.

bin/inference/pycbc_inference_model_stats
ahnitz: Woops, no I can remove that. BTW, this PR builds on https://github.com/gwastro/pycbc/pull/3996, which needs to be approved / merged first.

PR #3999, opened 2022-04-11
fix checkpointing after recent dynesty API change

ahnitz:
The recent PR to update pycbc to work with the latest Dynesty (1.2.0) API also turned off checkpointing. This was due to how the options were passed being ignored with the new method. This Patch should address the problem reported by @cnsetzer.


cnsetzer (approved):
I can confirm this is now working and checkpointing as expected. I have verified the checkpoint file for my inference and it is being populated as expected with subsequent checkpoints. 

I will mark this as approved.

PR #4000, opened 2022-04-19
Live singles in ci test

GarethCabournDavies:
Part of testing for https://github.com/gwastro/pycbc/pull/3984 is that we wanted to get the single-detector trigger into the live test

In this PR, we:
- Set the H1 and V1 strain generator to end 32s early
- Add another injection to inject an EM-bright signal into this (L1-only) time
- Make a single fits file (currently just manually filling the appropriate datasets to get something sensible)
- Add the singles to the test `run.sh` script
- Update the `check_results.py` script to know that there may be single-detector events (and call them 'found' rather than 'coinc')

### DIFF
--- a/examples/live/make_singles_fits_file.py
+++ b/examples/live/make_singles_fits_file.py
@@ -1,6 +1,8 @@
-# This script will make a valid singles-fits file for use in the
-# pycbc_live CI tests. It doesn't have much physical meaning,
-# but will give broadly representative numbers for singles
+"""
+This script will make a valid singles-fits file for use in the
+pycbc_live CI tests. It doesn't have much physical meaning,
+but will give broadly representative numbers for singles.
+"""
 import h5py
 import numpy as np
 
--- a/examples/live/make_singles_fits_file.py
+++ b/examples/live/make_singles_fits_file.py
@@ -7,7 +7,7 @@ import numpy as np
 f = h5py.File('single_trigger_fits.hdf','w')
 
 # Some numbers to design the output
-# These are loosley based on the O3a trigger fits file
+# These are loosely based on the O3a trigger fits file
 n_days = 30
 n_bins = 5
 max_duration = 150.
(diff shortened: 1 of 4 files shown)


GarethCabournDavies:
I don't seem to be able to request @maxtrevor to do a review?

examples/live/make_singles_fits_file.py
```diff
@@ -0,0 +1,61 @@
+# This script will make a valid singles-fits file for use in the
+# pycbc_live CI tests. It doesn't have much physical meaning,
+# but will give broadly representative numbers for singles
```
titodalcanton, line 3:
```suggestion
"""
This script will make a valid singles-fits file for use in the
pycbc_live CI tests. It doesn't have much physical meaning,
but will give broadly representative numbers for singles.
"""
```

titodalcanton:
I suggest changing `--frame-read-timeout 100` to something like 10 to avoid having the example stall for 100 s waiting for the final bits of the frame files.

maxtrevor:
> I don't seem to be able to request @maxtrevor to do a review?

Hi Gareth, I was traveling before but I can take a look at this now

GarethCabournDavies:
> > I don't seem to be able to request @maxtrevor to do a review?
> 
> Hi Gareth, I was traveling before but I can take a look at this now

Thanks!

examples/live/make_singles_fits_file.py
```diff
@@ -0,0 +1,63 @@
+"""
```
maxtrevor, line 1: Just making sure I understand, this file is used to calculate the significance of single detector triggers? It represents the results of fitting an exponential to the background distribution of noise triggers?

maxtrevor (commented):
I think the changes to check_results.py , generate_injections.py and run.sh all look reasonable. I just want to make sure I understand what the make_single_fits_file.py is for.

examples/live/make_singles_fits_file.py
GarethCabournDavies: I think @titodalcanton has the correct rights

GarethCabournDavies:
@titodalcanton - I think we just need someone to click 'approve' on this now

titodalcanton (approved):
Thanks @GarethCabournDavies and sorry for the delay. This looks ok.

PR #4003, opened 2022-04-21
Selecting the offsource trials randomly, rather than sequentially

pannarale:
Hey @duncanmmacleod this change should ensure that the off-trials are selected randomly, rather than sequentially from the analysis start time.


duncanmmacleod:
@pannarale, I'm not entirely sure why I've been asked to review this...

pannarale:
@duncanmmacleod, it's your code :-)  Feel free to delegate to others (@a-r-williamson, @titodalcanton)

pannarale:
And I quote: https://github.com/gwastro/pycbc/blob/9fb3fe1c6a0bb4f1f71af2ee2025a77218308449/bin/pygrb/pycbc_grb_trig_combiner#L4

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -371,6 +371,13 @@ parser.add_argument(
     default=6,
     help="The number of off source trials, default: %(default)d",
 )
+parser.add_argument(
+    "-e",
+    "--seed",
+    type=int,
+    default=1234,
```
a-r-williamson, line 378: Do we want a integer default or `None`? For example, this will always give us the same set of slide IDs for every GRB with a maximal offsource.

a-r-williamson (approved):
Looks alright, just the one query regarding the seed default. In production I imagine that for reproducibility we might randomly generate a seed at workflow generation time, or even pass in via config files, so it's not very important.

bin/pygrb/pycbc_grb_trig_combiner
pannarale: Good point.  I will change the default `None`.

PR #4005, opened 2022-04-25
allow assuming science segments are valid, no need to go to server

ahnitz:
This allows one to tell the search workflow that the vetos and or science segments are null and or always OK, respectively. For analyzing mock data, we often have no veto segments and the valid data should be assumed to be the same as the generated frame files.

### DIFF
--- a/pycbc/workflow/injection.py
+++ b/pycbc/workflow/injection.py
@@ -91,7 +91,7 @@ def compute_inj_optimal_snr(workflow, inj_file, precalc_psd_files, out_dir,
                                               'parallelization-factor',
                                               tags))
     except Exception as e:
-        print(e)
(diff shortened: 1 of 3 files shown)


ahnitz:
Here is an example of what the corresponding workflow section might look like if you provide the frame files in a cache format. 

```
[workflow-datafind]
datafind-method = FROM_PREGENERATED_LCF_FILES
datafind-pregenerated-cache-file-h1 = ${resolve:./h1.lcf} 
datafind-pregenerated-cache-file-l1 = ${resolve:./l1.lcf}
datafind-check-segment-gaps = update_times
datafind-check-frames-exist = raise_error

[workflow-segments]
segments-science = on
segments-vetoes = off
```

PR #4007, opened 2022-04-27
Generalizing ifos tag in pycbc_make_bank_verifier_workflow

pannarale:
Use the actual IFOs of interest in naming files, rather than 'HL'.


spxiwh (approved):
Looks good.

PR #4010, opened 2022-04-29
add method to convert pycbc ts to astropy

ahnitz:


titodalcanton (approved):
Looks good to me.

The `format='gps', scale='utc'` syntax looks super confusing to me (Astropy's [documentation](https://docs.astropy.org/en/stable/api/astropy.time.Time.html#astropy.time.Time) and [Wikipedia page](https://en.wikipedia.org/wiki/Time_standard#Current_time_standards) make it even more confusing) but I did a couple of tests and it leads to the correct result.

PR #4012, opened 2022-05-02
similar to PR #3922 and PR#3994 but for pycbc_dark_vs_bright_injections

pannarale:
This is to fix errors with pycbc_dark_vs_bright_injections when dealing with lsctables.
Also, it avoids hitting an error when they symmetric mass ratio is not set in the input xml table (by simply calculating it).

### DIFF
--- a/bin/pycbc_dark_vs_bright_injections
+++ b/bin/pycbc_dark_vs_bright_injections
@@ -84,8 +84,19 @@ output_bright = opts.output_bright
 if opts.write_compress:
     if not output_bright.endswith('gz'):
         output_bright = output_bright+'.gz'
+# make a list of columns that are present in the input table.
+# The : split is needed for columns like `process:process_id`,
+# which must be listed as `process:process_id` in `lsctables.New()`,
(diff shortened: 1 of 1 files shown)


bin/pycbc_dark_vs_bright_injections
```diff
@@ -105,7 +116,7 @@ for i, inj in enumerate(injections.table):
         out_sim_inspiral_bright.append(inj)
     # NSBH systems
     elif numpy.logical_not(numpy.logical_and(m1 >= opts.ns_bh_boundary, m2 >= opts.ns_bh_boundary)):
-        eta = inj.eta
+        eta = inj.eta if inj.eta is not None else m1*m2/(m1+m2)**2
```
titodalcanton, line 119: Please use `pycbc.conversions.eta_from_mass1_mass2()` instead of the explicit equation.

pannarale: Done.

titodalcanton (approved):

PR #4014, opened 2022-05-02
fix pool properties upon resume for dynesty 1.2

ahnitz:
@cnsetzer Reported that upon resuming from a checkpoint Dynesty 1.2 would limit to 1 core. I confirmed this was the case and it seems to be due to addition pool metadata that by default would be restored in 1.2. We need to add these to the "do not restore" list so that it uses the scratch made properties on restart.


ahnitz:
@cnsetzer Please confirm if it is easy that this resolves your issue.

PR #4020, opened 2022-05-06
Add PSD to the initial XML document

GarethCabournDavies:
In the offline event XML upload script, we want to include the PSD in the standard XML file to pass on to gracedb

Additional change needed to fix what looks like a change in the CreateREAL8FrequencySeries function in lal

~~I'm currently hitting an error with uploading, but think that's authentication rather than any change here~~ Fixed

### DIFF
--- a/bin/pycbc_upload_xml_to_gracedb
+++ b/bin/pycbc_upload_xml_to_gracedb
@@ -40,7 +40,7 @@ from pycbc.io.live import gracedb_tag_with_version
 from pycbc.io.ligolw import LIGOLWContentHandler
 from pycbc.psd import interpolate
 from pycbc.types import FrequencySeries
-from pycbc.results import ifo_color
+from pycbc.results import generate_asd_plot
 
 def check_gracedb_for_event(gdb_handle, query, far):
     """
@@ -247,26 +247,15 @@ for event in coinc_table:
                              tag_name=['analyst_comments'])
 
     if args.generate_plots:
-        psd_fig, psd_ax = plt.subplots(1)
-        for ifo in sorted(psddict.keys()):
-            curr_psd = psddict[ifo]
-            psd_ax.loglog(curr_psd.sample_frequencies,
-                          curr_psd ** 0.5 / pycbc.DYN_RANGE_FAC,
-                          c=ifo_color(ifo), label=ifo)
-        psd_ax.legend()
-        psd_ax.set_xlim([10, 1300])
-        psd_ax.set_ylim([3E-24, 1E-20])
-        psd_ax.set_xlabel('Frequency (Hz)')
-        psd_ax.set_ylabel('ASD')
-        psd_png_filename = f'{ifos_str}-PYCBC_{id_str}_PSD-{time:d}-1.png'
-        fullpath_psd = os.path.join(args.output_directory, psd_png_filename)
-        psd_fig.savefig(fullpath_psd)
-        logging.info("Saved PSD plot %s", psd_png_filename)
+        asd_png_filename = f'{ifos_str}-PYCBC_{id_str}_ASD-{time:d}-1.png'
+        fullpath_asd = os.path.join(args.output_directory, asd_png_filename)
+        generate_asd_plot(psddict, fullpath_asd)
+        logging.info("Saved ASD plot %s", asd_png_filename)
         if not args.no_upload:
             gracedb.writeLog(
-            r["graceid"], "PyCBC PSD estimate from the time of event",
-            filename=psd_png_filename,
-            tag_name=["psd"], displayName=['PSDs'])
+            r["graceid"], "PyCBC ASD estimate from the time of event",
+            filename=fullpath_psd,
+            tag_name=["asd"], displayName=['ASDs'])
             logging.info("Uploaded file %s to event %s.", psd_png_filename,
                          r["graceid"])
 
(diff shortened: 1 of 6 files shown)


GarethCabournDavies:
I added in the ability to plot/upload the psd as well while I was messing about with this

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -181,70 +192,91 @@ for event in coinc_table:
...
+            psd_ax.loglog(curr_psd.sample_frequencies,
+                          curr_psd ** 0.5 / pycbc.DYN_RANGE_FAC,
+                          c=ifo_color(ifo), label=ifo)
+        psd_ax.legend()
+        psd_ax.set_xlim([10, 1300])
+        psd_ax.set_ylim([3E-24, 1E-20])
+        psd_ax.set_xlabel('Frequency (Hz)')
+        psd_ax.set_ylabel('ASD')
+        psd_png_filename = '{}-PYCBC_{}-{:d}-psd.png'.format(ifos_str, id_str,
+                                                             time)
```
titodalcanton, line 262:
```suggestion
        psd_png_filename = f'{ifos_str}-PYCBC_f{id_str}_PSD-{time:d}-1.png'
```
Note that I made the file name consistent with the usual convention (duration at the end, even if bogus) and moved the "PSD" label into the arbitrary-string part. Maybe you did it the other way because you want the file names to naturally sort lexicographycally and group event files and their PSD plots together?

titodalcanton:
@GarethCabournDavies this looks ok to me, with that minor comment, and a more general one: ideally we would use a single code to make both the online and offline PSD plots on GraceDB. Have you checked if that is easy to do? Maybe there could be a function in `pycbc.results` to take care of that plot?

GarethCabournDavies:
> @GarethCabournDavies this looks ok to me, with that minor comment, and a more general one: ideally we would use a single code to make both the online and offline PSD plots on GraceDB. Have you checked if that is easy to do? Maybe there could be a function in `pycbc.results` to take care of that plot?

That makes a lot of sense. I'll see what I can do with that

bin/pycbc_upload_xml_to_gracedb
GarethCabournDavies: Looking at it now, we plot the ASD, not the PSD, so maybe that should be updated in the naming for both plots?

titodalcanton: Good point. Some of us may be used to just calling it "PSD" anyway, and would be confused by not finding a file with "PSD" in the name, but this does not seem a good reason to use the wrong terminology.

GarethCabournDavies:
Added docstring/preamble (due to CC request) - I tried to convey that some of this is moved/copied from earlier code. From what I could see this was a combination of @spxiwh and @titodalcanton (https://github.com/gwastro/pycbc/pull/2628), so have added names in alphabetical order

pycbc/results/psd.py
```diff
@@ -0,0 +1,69 @@
...
+                      curr_psd[1:] ** 0.5 / DYN_RANGE_FAC,
+                      c=ifo_color(ifo), label=ifo)
+
+    asd_ax.legend()
+    asd_ax.set_xlim([10, 1300])
+    asd_ax.set_ylim([3E-24, 1E-20])
+    asd_ax.set_xlabel('Frequency (Hz)')
+    asd_ax.set_ylabel('ASD')
+    asd_fig.savefig(output_filename)
+
```
titodalcanton, line 69:
```suggestion

__all__ = ["generate_asd_plot"]
```

pycbc/io/live.py
```diff
@@ -426,9 +419,9 @@ def upload(self, fname, gracedb_server=None, testing=True,
                     displayName=['SNR timeseries']
                 )
                 gracedb.writeLog(
-                    gid, 'PSD plot upload',
-                    filename=psd_series_plot_fname,
-                    tag_name=['psd'], displayName=['PSDs']
+                    gid, 'ASD plot upload',
+                    filename=asd_series_plot_fname,
+                    tag_name=['asd'], displayName=['ASDs']
```
titodalcanton, line 424: I suggest keeping the `tag_name` to `psd`. I am not sure how GraceDB would react to the change.

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -181,70 +192,80 @@ for event in coinc_table:
...
+    if args.generate_plots:
+        asd_png_filename = f'{ifos_str}-PYCBC_{id_str}_ASD-{time:d}-1.png'
+        fullpath_asd = os.path.join(args.output_directory, asd_png_filename)
+        generate_asd_plot(psddict, fullpath_asd)
+        logging.info("Saved ASD plot %s", asd_png_filename)
+        if not args.no_upload:
+            gracedb.writeLog(
+            r["graceid"], "PyCBC ASD estimate from the time of event",
+            filename=fullpath_psd,
+            tag_name=["asd"], displayName=['ASDs'])
```
titodalcanton, line 258: See below, I suggest sticking to `psd` as the `tag_name`.

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -26,6 +26,9 @@ import logging
 from ligo.gracedb.rest import GraceDb
 import numpy as np
 import h5py
+import matplotlib
+matplotlib.use('agg')
+from matplotlib import pyplot as plt
```
titodalcanton, line 31: Are these imports still needed?

pycbc/results/psd.py
```diff
@@ -0,0 +1,69 @@
...
+#
+#                                   Preamble
+#
+# =============================================================================
+#
+"""
+Module to generate PSD figures
+"""
+import matplotlib
+matplotlib.use('agg')
```
titodalcanton, line 30: Do not set matplotlib's backend in a module, it will mess up some use cases as soon as somebody imports (a module that imports) `pycbc.results`.

GarethCabournDavies:
Updated to @titodalcanton's suggestions

Issue #4019, opened 2022-05-05
PyCBC incompatible with Astropy 5.1rc1


duncanmmacleod:
The following error occurs when attempting to import `pycbc.cosmology` in an environment with Astropy 5.1rc1 [[ref](https://github.com/gwpy/gwpy/runs/6299808407?check_suite_focus=true)]:

```pytb
/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pycbc/conversions.py:37: in <module>
    import pycbc.cosmology
/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/pycbc/cosmology.py:297: in <module>
    for _c in astropy.cosmology.parameters.available}
/opt/hostedtoolcache/Python/3.9.12/x64/lib/python3.9/site-packages/astropy/cosmology/__init__.py:38: in __getattr__
    raise AttributeError(f"module {__name__!r} has no attribute {name!r}.")
E   AttributeError: module 'astropy.cosmology' has no attribute 'parameters'.
```

The error occurs here:

https://github.com/gwastro/pycbc/blob/5a7e3f478c9d1a16c7a1658d806637514277a219/pycbc/cosmology.py#L295-L297

I can't pin down exactly which change in Astropy results in this error (I think this used to work entirely by accident), but it should be fixed by directly importing the necessary parameters module (`from astropy.cosmology import parameters`).

ahnitz:
This is now fixed by #4022

PR #4021, opened 2022-05-06
check astropy 5.1rc1

ahnitz:


ahnitz:
A failure has been reported by @duncanmmacleod in #4019, but we should run the tests and see if any other errors pop up ahead of time.

ahnitz:
The initial error reported in #4019 should now be addressed in #4022

ahnitz:
All unittests passed with 5.1rc1, so now closing PR.

PR #4023, opened 2022-05-08
Add a Hierarchical model to inference

cdcapano:
Adds a class for hierarchical models. Work in progress.

### DIFF
--- a/docs/inference.rst
+++ b/docs/inference.rst
@@ -487,6 +487,7 @@ Examples
     inference/examples/gw150914.rst
     inference/examples/single.rst
     inference/examples/relative.rst
+    inference/examples/hierarchical.rst
 
 ===============================================
 Visualizing the Posteriors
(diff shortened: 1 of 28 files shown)


ahnitz:
I think the multi signal, shared data model can subclass from this with just an update of the loglikelihood function. I'll try to add that in combination with changes needed to the data models. Out of curiosity can this model contain also another heirachical model? If no, would that be difficult bookkeeping wise?

cdcapano:
Depends on #4029.

ahnitz:
@cdcapano Why have you added a new intermediate class? Is that planning something for the future? If not, it seems like it's adding a shell of complexity for no reason, no?

ahnitz:
@cdcapano I'll defer to you on that one, but thanks for clarifying the problem.

ahnitz (approved):
Please merge once the tests pass

cdcapano:
@ahnitz I reverted the intermediate class back (option 2 above). I'm just waiting for the `run.sh` to finish on atlas to make sure it completes ok, then I'll merge.

ahnitz:
@cdcapano Did this work out?

cdcapano:
It's still running, but I'm just going to squash and merge since there doesn't appear to be anything wrong. I can add a plot of the results to the doc page in a followup PR.

PR #4025, opened 2022-05-11
multi-signal relative model 

ahnitz:
This is a work in progress of adding in the features needed for the multi-signal relative model.

### DIFF
--- a/examples/inference/multisignal/rel.ini
+++ b/examples/inference/multisignal/rel.ini
@@ -1,0 +1,82 @@
+[model]
+name = multi_signal
+submodels = event1 event2
+
+[event1__model]
+name = relative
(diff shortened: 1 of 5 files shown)


ahnitz:
@sum33it @cdcapano Ok, I think this works now. There's a couple things I need to improve for some more advanced uses, (1) the grids atm, should be the same between models, but I can change that in a follow-up PR. Also in a follow-up, I will add in the ability choose the method for estimating the t_of_f function (i.e. to map frequency to a specific time to handle earth rotation effects). I need the latter to properly handle higher modes for very long NSBH for example.

cdcapano (approved):
Sorry, missed this. Looks good.

PR #4026, opened 2022-05-13
allow setting the universe in the config file

ahnitz:


examples/workflow/generic/simple_subworkflow_data/run.sh
```diff
@@ -1 +1 @@
-python simple.py --workflow-name test --config-files simple.ini --submit-now
```
spxiwh, line 1: Why change this?

ahnitz: Accidental merge due to testing

pycbc/workflow/core.py
```diff
@@ -264,8 +256,8 @@ def __init__(self, cp, name,
             super(Executable, self).__init__(self.pegasus_name,
                                              installed=self.installed)
 
-        self._set_pegasus_profile_options()
-        self.set_universe(self.universe)
+        if self.universe:
+            self.set_universe(self.universe)
```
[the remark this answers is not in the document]
ahnitz: @spxiwh It's handling the case that universe was set to the class *explicitly* through its init arguments. Otherwise, that would always be ignored.

sum33it (approved):
Ok. I have checked this on a parallel universe and it is working as intended. I set `condor|universe = parallel ` in the `[pegasus_profile-condorpool_shared]` section as suggested by @spxiwh in slack channel.

pycbc/workflow/core.py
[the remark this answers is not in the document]
ahnitz: @spxiwh Ah, I'm not sure. I think it might make sense to just take this out from the class. I can check if we aren't using this anyway.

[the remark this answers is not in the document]
ahnitz: @spxiwh I've gone ahead and remove the option and the uses elsewhere in the repo (just jobsetup.py AFAIK)

pycbc/workflow/jobsetup.py
```diff
@@ -908,10 +906,7 @@ class PycbcSplitInspinjExecutable(Executable):
     The class responsible for running the pycbc_split_inspinj executable
     """
     current_retention_level = Executable.INTERMEDIATE_PRODUCT
-    def __init__(self, cp, exe_name, num_splits, universe=None, ifo=None,
-                 out_dir=None):
-        super().__init__(cp, exe_name, universe, ifo, out_dir, tags=[])
-        self.num_splits = int(num_splits)
```
spxiwh, line 914: I think this will break this one. num_splits is needed

ahnitz: You are right.

PR #4030, opened 2022-05-15
add .ini support in pycbc_brute_bank

WuShichao:
@ahnitz Allow `pycbc_brute_bank` to use the `.ini` file, I have run bank generation using this PR for 10 mins, no problem, when this PR is in good shape, I will test it by entire template bank generation. I will also update/clean the read config function in this PR soon.

### DIFF
--- a/bin/bank/pycbc_brute_bank
+++ b/bin/bank/pycbc_brute_bank
@@ -385,6 +385,8 @@ def draw(rtype):
             l &= mc > args.min_mchirp
 
     else:
+        # Apply boundary conditions.
+        params = dists_joint.apply_boundary_conditions(**params)
         # Apply constraints.
         l = dists_joint.contains(params)
         # Apply boundaries.
@@ -465,6 +467,7 @@ while tau0s < args.tau0_end:
                          region, kloop, r, len(bank), kconv, len(bank) - blen)
             if uconv:
                 logging.info('Ratio of convergences: %2.3f' % (kconv / (uconv)))
+                logging.info('Progress: {:.0%} completed'.format(tau0s/args.tau0_end))
 
             if kloop == 1:
                 okconv = kconv
--- a/bin/bank/pycbc_brute_bank
+++ b/bin/bank/pycbc_brute_bank
@@ -1,8 +1,29 @@
 #!/usr/bin/env python
+
+# Copyright (C) 2017-2020 Alex Nitz, Duncan Macleod
+# Copyright (C) 2022 Shichao Wu, Alex Nitz
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 3 of the License, or (at your
+# option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
 """Generate a bank of templates using a brute force stochastic method.
 """
 import numpy, h5py, logging, argparse, numpy.random, sys
 import pycbc.waveform, pycbc.filter, pycbc.types, pycbc.psd, pycbc.fft, pycbc.conversions
+from pycbc.distributions import read_params_from_config
+from pycbc.distributions.utils import draw_samples_from_config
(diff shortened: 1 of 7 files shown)


bin/bank/pycbc_brute_bank
```diff
@@ -1,8 +1,29 @@
 #!/usr/bin/env python
+
+# Copyright (C) 2017-2020 Alex Nitz, Duncan Macleod
```
ahnitz, line 3: This part doesn't look right?

bin/bank/pycbc_brute_bank
```diff
@@ -278,29 +312,54 @@ def draw(rtype):
...
+                params[name] = samples[name]
+
     elif rtype == 'kde':
         trail = 300
         if trail > len(bank):
             trail = len(bank)
         p = bank.keys()
         p = [k for k in p if k not in fdict]
         p.remove('approximant')
+        if args.input_config is not None and 'q' in static_args:
```
ahnitz, line 332: Let's avoid hardcoding any parameters. I think what we want is if you provide a config file, the kde should be drawn in the variable parameters you choose (only) and then you make you add in the static args and apply the same conversions / transforms you get from the config file. That way it is completely generic, make sense?

ahnitz, line 332: Let's avoid hardcoding any parameters. I think what we want is if you provide a config file, the kde should be drawn in the variable parameters you choose (only) and then you make you add in the static args and apply the same conversions / transforms you get from the config file. That way it is completely generic, make sense?

bin/bank/pycbc_brute_bank
```diff
@@ -278,29 +312,54 @@ def draw(rtype):
...
             params[k] = v
 
+        if args.input_config is not None and 'q' in static_args:
+            params['mass1'] = pycbc.conversions.mass1_from_mtotal_q(params['total_mass'], static_args['q'])
+            params['mass2'] = pycbc.conversions.mass2_from_mtotal_q(params['total_mass'], static_args['q'])
+
     params['approximant'] = numpy.array([args.approximant]*size)
 
-    # Filter out stuff
+    # Filter out stuff (kde method may also generate samples outside boundaries).
```
ahnitz, line 350: If you have the config file, this isn't the right way to way to apply the filter. There is a much easier method that already handles both boundaries *and* constraints. Just use the contains operator of the JointDistribution. . 


Look at https://github.com/gwastro/pycbc/blob/master/pycbc/distributions/joint.py#L220

ahnitz, line 350: If you have the config file, this isn't the right way to way to apply the filter. There is a much easier method that already handles both boundaries *and* constraints. Just use the contains operator of the JointDistribution. . 


Look at https://github.com/gwastro/pycbc/blob/master/pycbc/distributions/joint.py#L220

bin/bank/pycbc_brute_bank
ahnitz: This also avoids passing anything to the kde that is singular since the variable params by definition *must* vary.

ahnitz: This also avoids passing anything to the kde that is singular since the variable params by definition *must* vary.

pycbc/distributions/joint.py
```diff
@@ -242,6 +242,39 @@ def contains(self, params):
             result = result.item()
         return result
 
+    def __contains__(self, params):
```
ahnitz, line 245: @WuShichao Thank you for putting this together. I think it is nearly ready to go. Given that as you've shown you can do array input with the 'in' operator (this appears to be a limitation of the __contains__ operator itself !, I think perhaps it would be good to rename this method from __contains__ to something else. 

I think the current 'contains' method is only used in a few places. I propose we swap this out as that name is confusing. Perhaps it can be renamed to 'within_constraints' and your __constraints__ should be renamed to simply 'constrains'. That would mean renaming any existing uses 'of constrains' to point to 'within_constraints' instead.

pycbc/distributions/joint.py
```diff
@@ -242,6 +242,39 @@ def contains(self, params):
...
+            If params was an array, or if params a dictionary and one or more
+            of the parameters are arrays, will return an array of booleans.
+            Otherwise, a boolean.
+        """
+        params = self.apply_boundary_conditions(**params)
+        result = True
+        for dist in self.distributions:
+            param_name = dist._params[0]
+            contain_array = numpy.ones(len(params[param_name]), dtype=bool)
+            # `__contains__` in `pycbc.distributions.bounded`
```
ahnitz, line 269: Maybe add a note that we should fix this in the future. I think we don't need to hold up this PR anymore for this particular change, but we should have it so there is method that of the distributions that can except a vector form.

ahnitz (approved):
@WuShichao This is looking nice. Please see my last two comments which I hope are easy to address. Once that's done and the tests pass, I believe this should be merged.

PR #4032, opened 2022-05-24
add multi signal class + support to gaussian_noise model

ahnitz:
This PR depends on #4023 

This adds a new model meant to contain models which both (1) use shared data (2) the signals might overlap enough to worry about cross terms in the likelihood function. If (2) doesn't apply, one can simply use the hierarchical model directly as this model necessarily has more limited support in terms of which submodels can be included. 

At least one submodel must implement the following two methods

```
def multi_signal_support(self):

def multi_loglikelihood(self, models):
```

The first returns the set of *other* models which are supported for this classes loglikelihood function. At the moment, I plan to only implement loglikelihood functions that support other instances of the same class, but it might be useful in the future to allow cases where mixed models are used. However, in that case, one of the models must explicitly say it supports those models. This is necessarily the case, as the logl function has to be aware of the internals each model to actually perform the needed calculations.

This PR also adds these functions to the 'gaussian_noise' model. Future PRs will add support to other models.


WuShichao:
@ahnitz I think this will also be useful to "CBC + glitch" PE if one has the glitch model (maybe some wavelets summation) through the custom waveform plugin.

ahnitz:
@WuShichao Yes, I agree.

pycbc/inference/models/gaussian_noise.py
```diff
@@ -852,7 +853,46 @@ def _nowaveform_loglr(self):
...
+
+                if slc1 != slc2:
+                    raise RuntimeError('gaussian noise implementation of multi'
+                                       'signal model requires all frequency'
+                                       'cutoffs to be the same!')
+
+                h1h2 = h1[slc1].inner(h2[slc1]).real  # < h1, h2>
+                loglr += - h1h2  # This is -0.5 * (<h1|h2> + <h2|h1>)
+        return loglr + self.lognl
+
```
cdcapano, line 894: @ahnitz Wouldn't it be easier to just sum up the waveforms, then pass that through the normal loglr or loglikelihood function?

pycbc/inference/models/gaussian_noise.py
```diff
@@ -883,6 +923,10 @@ def _loglr(self):
                 return self._nowaveform_loglr()
             else:
                 raise e
+
+        if save_waveforms:
+            self.wfs = {}
+
```
cdcapano, line 929: Rather than add this, I think it'd be better to move [this function](https://github.com/gwastro/pycbc/blob/master/pycbc/inference/models/gated_gaussian_noise.py#L197-L223) from `BaseGatedGaussianNoise` over to `BaseGaussianNoise`, then modify the `loglr` function currently in `GaussianNoise` to call `get_waveforms` rather than the waveform generator directly. This would also be a useful step toward having a script that plots the maxL waveforms. Also could be useful for my comment above: then the `multi_loglikelihood` could just modify what `get_waveforms` does (maybe through an attribute), then call the loglr function.

pycbc/inference/models/gaussian_noise.py
ahnitz: Ok, done.

pycbc/inference/models/gaussian_noise.py
ahnitz: @cdcapano The function in gated gaussian needs to remain as it all the things in the inner loop aren't appropriate for the regular model. I think I need to leave the function in gated gaussian there, but possibly inherit from the one in gaussian noise (added).

ahnitz:
@cdcapano Can you please recheck?

pycbc/inference/models/hierarchical.py
```diff
@@ -498,3 +498,62 @@ def map_params(params):
...
+
+    def _loglikelihood(self):
+        for lbl, model in self.submodels.items():
+            # Update the parameters of each
+            model.update(**{p.subname: self.current_params[p.fullname]
+                            for p in self.param_map[lbl]})
+
+        # Calculate the combined loglikelihood
+        p = self.primary_model
+        logl = self.submodels[p].multi_loglikelihood(self.submodels.values())
```
cdcapano, line 552: I think you'll be calling the primary model's loglikelihood twice here, won't you? The submodels dictionary will contain the primary model.

pycbc/inference/models/gaussian_noise.py
```diff
@@ -852,7 +856,59 @@ def _nowaveform_loglr(self):
...
+            wfs.append(m.get_waveforms())
+
+        # combine into a single waveform
+        combine = {}
+        for det in self.data:
+            mlen = max([len(x[det]) for x in wfs])
+            [x[det].resize(mlen) for x in wfs]
+            combine[det] = sum([x[det] for x in wfs])
+
+        loglr = self._loglr(waveforms=combine)
```
cdcapano, line 887: I'd rather not change the API from the base models for the core likelihood functions. You don't need to do that anyway. Call self's get_waveforms, then add to the `current_wfs` the other models' waveforms. Inside `_loglr`, `get_waveforms` will be called again. Since `_current_wfs` will have been set, it will return the summed waveforms. You'll need to reset `_current_wfs` to None after, so that subsquent calls to the normal `loglr` (if someone were to do that) regenerates just this model's waveforms.

pycbc/inference/models/gaussian_noise.py
```diff
@@ -852,7 +856,59 @@ def _nowaveform_loglr(self):
...
+        """
+        return [type(self)]
+
+    def multi_loglikelihood(self, models):
+        """ Calculate a multi-model (signal) likelihood
+        """
+        # Generate the waveforms for each submodel
+        wfs = []
+        for m in models:
+            wfs.append(m.get_waveforms())
```
cdcapano, line 878: It's odd that you're not calling `self`'s `get_waveforms` here. You're relying on the user passing in `self` in the list of models, which isn't evident from the API. Naively, I would think that calling `model.multi_loglikelihood(models)` would include that model in the loglikelihood, otherwise, why would this be an instance method and not a stand-alone function?

cdcapano (changes requested):
See requests below. Also, it looks like a bunch of the tests are failing.

cdcapano (approved):
Looks good, thanks for addressing my comments.

PR #4033, opened 2022-05-24
add multi signal support to the single_template model

ahnitz:
This depends on #4032 and #4023 and adds multi signal support to the single template model. Similar to the gaussian noise model, at this time, it only supports the case of multiple instances of this model and not any mixed model cases.


ahnitz:
To prove this working, I added a simple example where we assume there were two BNS with exactly identical parameters in the data around GW170817. As expected, we see that the distance recovery is essentially just multiplied by two (this would not be the case if cross terms were not accounted for)


single

PR #4038, opened 2022-05-27
Some typos & clarify function name

tdent:


bin/pycbc_optimize_snr
```diff
@@ -112,14 +112,14 @@ def compute_network_snr_core(v, *argv, raise_err=False):
     return network_snrsq ** 0.5, snr_series_dict
 
 
-def compute_network_snr(v, *argv):
+def compute_minus_network_snr(v, *argv):
```
GarethCabournDavies, line 115: It looks like this has introduced an error (which isn't failing on the CI for some reason), as the function name hasnt been update where it is used in the optimize_shgo or optimize_di - and the same thing in the specific _pso version

This hasn't shown up as pycbc_optimize_snr is run through a python subprocess, but I'm not sure how to pass through the error

tdent: See #4072 ..

PR #4039, opened 2022-05-27
Initial/ minimal code for live pastro, other cleanups

tdent:


tdent:
Mostly placeholders at the moment, next tasks are to set up at least one working calculation method

titodalcanton:
@tdent I did not realize how much this PR grew recently. As is, this will conflict with #4086, which I hope to merge in the next 24 hours, and with a few other changes that I would like to propose in the next ~week. I see this PR does a few separate things:

1. Implement p_astro for PyCBC Live.
2. Rename the SingleCoincForGraceDB class.
3. Fix some issues with matplotlib imports/backend.
4. Whitespace/comment edits.

Would you be able to split off 2 to a separate PR? I agree with the change, but since we are intentionally breaking the API, I would like that to happen in a fairly simple PR with a clear title for reference. I would also ask to split 3-4 to one or two separate PRs to make reviewing 1 easier, if possible.

Do you need one last rebase on master? I think I already saw some of the proposed edits previously.

bin/pycbc_live
```diff
@@ -877,7 +883,8 @@ with ctx:
         if evnt.rank > 0:
             evnt.commit_results((results, data_end()))
         else:
-            psds = {ifo: data_reader[ifo].psd for ifo in data_reader if data_reader[ifo].psd is not None}
+            psds = {ifo: data_reader[ifo].psd for ifo in data_reader \
+                    if data_reader[ifo].psd is not None}
```
titodalcanton, line 887: The backslash is not needed, and you can also indent the `for` if you like.

tdent:
Well, 3) seemed to be necessary for testing, as otherwise I was running into all sorts of errors when just trying to do pycbc_live --help.
Since whitespace / comment edits and matplotlib import fixes would otherwise require PRs of their own which would also need to be reviewed I don't think the total burden is reduced by splitting them off. 

I see the case for putting the SingleCoinc rename in its own PR as otherwise it might be 'hidden' here.

titodalcanton:
Just saying that it takes me exponentially longer to review a longer PR. Anyway, my only strict request at the moment is splitting off point 2, so let's start with that.

bin/pycbc_live
```diff
@@ -702,6 +706,7 @@ pycbc.init_logging(args.verbose, format=log_format)
 
 ctx = scheme.from_cli(args)
 fft.from_cli(args)
+padata = livepau.PAstroData(args, args.bank_file)
```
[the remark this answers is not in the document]
tdent: Hmm, I suppose I had some reason for splitting off the bank as a separate argument. It is a logically separate input from the other stuff provided via the args (ie the method and json spec file).

bin/pycbc_live
```diff
@@ -342,12 +343,13 @@ class LiveEventManager(object):
...
-            event = SingleCoincForGraceDB(live_ifos, coinc_results, bank=bank,
+            event = SingleCoincForGraceDB(live_ifos, coinc_results,
+                                          padata=padata, bank=bank,
                                           psds=psds, followup_data=fud,
                                           low_frequency_cutoff=f_low,
                                           channel_names=args.channel_name,
-                                          mc_area_args=self.mc_area_args,
-                                          )
+                                          mc_area_args=self.mc_area_args)
+                                          #p_astro_data=self.padata)
```
tdent, line 352: whoops, missed this .. will remove asap

bin/pycbc_live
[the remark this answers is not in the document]
tdent: I changed this to explicitly pass only the method, spec file and bank file as arguments not the full args object as I think this is easier to debug/maintain.

pycbc/io/live.py
```diff
@@ -266,36 +271,77 @@ def __init__(self, ifos, coinc_results, **kwargs):
...
+        if 'padata' in kwargs:
+            padata = kwargs['padata']
+            trigger_data = {
+                'mass1': sngl_populated.mass1,
+                'mass2': sngl_populated.mass2,
+                'spin1z': sngl_populated.spin1z,
+                'spin2z': sngl_populated.spin2z,
+                'network_snr': network_snrsq ** 0.5,
+                'far': far}
+            horizons = {ifo: self.psds[ifo].dist}
```
titodalcanton, line 284: `for ifo in …` missing?

tdent: interesting, I wonder what this did given that it didn't produce an error

titodalcanton: It probably created a one-item dict with a random ifo from the code above.

titodalcanton:
I also suggest merging `live_pastro.py` and `live_pastro_utils.py` since they are not too long and their classes and functions are fairly self-explanatory.

Also please add the `__all__ = […]` boilerplate at the end of those/that file(s) as done in other modules.

tdent:
Conflict now in pycbc live .. as expected I guess

pycbc/population/live_pastro.py
```diff
@@ -0,0 +1,152 @@
...
+    with open(spec_file) as specf:
+        pa_spec = json.load(specf)
+    # check that the file has the right contents
+    assert 'param' in pa_spec
+    assert 'bin_edges' in pa_spec  # should be a list of floats
+    assert 'sig_per_yr_binned' in pa_spec  # signal rate per bin (per year)
+    # do the lengths of bin arrays match?
+    assert len(pa_spec['bin_edges']) == len(pa_spec['sig_per_yr_binned']) + 1
+    assert 'ref_bns_horizon' in pa_spec  # float
+    assert 'netsnr_thresh' in pa_spec  # float
```
titodalcanton, line 32: One could be a bit more aggressive here and enforce even more assumptions, e.g.
```python
    assert pa_spec['ref_bns_horizon'] > 0
    assert pa_spec['netsnr_thresh'] > 0
```

pycbc/population/live_pastro.py
```diff
@@ -0,0 +1,152 @@
...
+        Prerequisite data for p astro calc
+    bankf: string
+        Path to HDF5 template bank file
+
+    Returns
+    -------
+    bank_data: dictionary
+        Template counts binned over specified param
+    """
+    bank = h5py.File(bankf, 'r')
```
titodalcanton, line 51: Use a `with` block to access the bank file.

pycbc/population/live_pastro_utils.py
```diff
@@ -0,0 +1,103 @@
...
+    """
+
+    live_pastro_group = parser.add_argument_group('Options for live p_astro '
+                                                  'calculation')
+
+    # Only one choice so far, allow for more later
+    live_pastro_group.add_argument('--p-astro-method',
+                                   choices=['template_param_bins'])
+    live_pastro_group.add_argument('--p-astro-spec-data',
+        help='File containing information to set up p_astro calculation')
```
titodalcanton, line 26: Have you considered having a single `--p-astro-definition` option taking the JSON file name as argument, and defining the method with an additional field in the JSON file? That seems a bit more compact to me, and it allows us to drop the verify checks below.

pycbc/population/live_pastro_utils.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Returns
+    -------
+    args : argparse Namespace object
+    """
+    if args.p_astro_method is not None and args.p_astro_spec_data is None:
+        raise RuntimeError('Need a p astro data spec file for method %s! ' %
+                           args.p_astro_method)
+    if args.p_astro_method is None and args.p_astro_spec_data is not None:
+        raise RuntimeError('Need to specify p_astro method to do a p_astro '
```
titodalcanton, line 47: If we do want to keep this code as is, these errors should not be exceptions (and certainly not RuntimeErrors), but calls to `parser.error()`. So you would need to pass the parser object as an extra arg to this function (yes, I know we do not do that in other `verify_*` calls, but we should!).

bin/pycbc_live
```diff
@@ -756,6 +761,8 @@ pycbc.init_logging(args.verbose, format=log_format)
 
 ctx = scheme.from_cli(args)
 fft.from_cli(args)
+padata = livepau.PAstroData(args.p_astro_method, args.p_astro_spec_data,
```
titodalcanton, line 764: In one of the open PRs, I am trying to reduce the number of global variables we use in `pycbc_live`. Could you make `padata` an attribute of the LiveEventManager object instead, initialized in its constructor?

titodalcanton:
@tdent this looks almost ready to go, apart from a few coding comments. If you prefer to merge this functionality this week, though, we can keep those changes for a later PR.

pycbc/population/live_pastro_utils.py
tdent: Simplified option structure means these all go away ..

bin/pycbc_live
```diff
@@ -375,6 +377,7 @@ class LiveEventManager(object):
         event = SingleCoincForGraceDB(
             live_ifos,
             coinc_results,
+            padata=padata,
```
titodalcanton, line 380: Should this be `self.padata`?

bin/pycbc_live
```diff
@@ -444,13 +446,14 @@ class LiveEventManager(object):
...
-            # apply a trials factor of the number of active detectors
+            # Apply a trials factor of the number of active detectors
             ifar /= len(active)
             single['foreground/ifar'] = ifar
 
             live_ifos = [i for i in fud if 'snr_series' in fud[i]]
 
-            event = SingleCoincForGraceDB(live_ifos, single, bank=bank,
+            event = SingleCoincForGraceDB(live_ifos, single,
+                                          padata=padata, bank=bank,
```
titodalcanton, line 456: Probably `self.padata` as well?

bin/pycbc_live
tdent: ah, this only works because 'padata' has already been assigned a value .. so yes, probably should be fixed

titodalcanton (approved):
Looks good to go for now. Further refinements in future PRs.

PR #4040, opened 2022-05-27
Add ability to use plugin models

cdcapano:
This adds the ability for users to define their own models that can be used in pycbc inference, without needing to modify the source code. Borrows the same techniques as used for the plugin waveforms.

### DIFF
--- a/pycbc/inference/models/__init__.py
+++ b/pycbc/inference/models/__init__.py
@@ -20,6 +20,8 @@ assuming various noise models.
 """
 
 
+from .base import BaseModel
+from .base_data import BaseDataModel
 from .analytic import (TestEggbox, TestNormal, TestRosenbrock, TestVolcano,
                        TestPrior, TestPosterior)
 from .gaussian_noise import GaussianNoise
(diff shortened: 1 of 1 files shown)


ahnitz (approved):
Looks right.

ahnitz:
@cdcapano Proposed change to your PR to fix the circular import issue https://github.com/cdcapano/pycbc/pull/5

cdcapano:
I was having trouble with circular imports. Basic problem was custom model classes need to inherit from `BaseModel`, but this setup a circular import since the plugins are resolved when the models module is initialized. To fix, I've delayed the plugin resolving to when the models are first accessed, rather than when they're imported. I did this by replacing the `inference.models.models` dict with a function, `inference.models.get_models()`. This will break backward compatibility for anyone that previously accessed the `models` dict directly (I think this is done in a couple of the tutorials). It doesn't affect the `pycbc_inference` executable though, so should probably be ok.

cdcapano:
@ahnitz To avoid breaking backward compatibility, I decided to duck type the `models` dictionary instead so that it automatically adds the plugins the first time its accessed. I think I covered all possible ways it could be accessed. I left in the new functions, with the idea that these should be used anyway. If the checks pass I'll merge, unless you object.

I'll add documentation and a tutorial in another PR.

ahnitz:
@cdcapano Ok, sounds good.

ahnitz (approved):

PR #4047, opened 2022-06-02
added pycbc_plot_kde_vals

PRAVEEN-mnl:


bin/all_sky_search/pycbc_plot_kde_vals
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python
+
+import h5py, argparse
+import matplotlib.pyplot as plt
+from matplotlib.colors import LogNorm
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument('--signal-file')
+parser.add_argument('--template-file', required=True)
+parser.add_argument('--param', nargs='+', required=True, help='mchirp, eta, chi_eff')
```
tdent, line 10: I think the code should be able to handle the case of parameters with different names, as long as the input files have datasets with these names.  The help should say that you can specify one parameter to plot vs the kde, if the plot type is kde_vs_param, or two parameters separated by a space if the plot type is param_vs_param.  If the user specifies a third parameter it looks like the code doesn't do anything with it ...

bin/all_sky_search/pycbc_plot_kde_vals
```diff
@@ -0,0 +1,64 @@
...
+parser.add_argument('--plot-type', choices=['kde_vs_param', 'param_vs_param'])
+parser.add_argument('--which-kde', choices=['signal_kde','template_kde', 'ratio_kde'])
+parser.add_argument('--plot-dir', required=True)
+parser.add_argument('--verbose', action='count')
+args = parser.parse_args()
+
+signal_dataset = h5py.File(args.signal_file, 'r')
```
tdent, line 17: If the user doesn't want to plot the signal kde or ratio, it should not be necessary to read the signal file.  Also, it's a bit confusing to create a hdf file object and call it 'dataset'

bin/all_sky_search/pycbc_plot_kde_vals
```diff
@@ -0,0 +1,64 @@
...
+template_dataset = h5py.File(args.template_file, 'r')
+param0 = template_dataset[args.param[0]][:]    #mchirp,eta,chi_eff
+param1 = template_dataset[args.param[1]][:]
+param2 = template_dataset[args.param[2]][:]
+kde_template = template_dataset['kde_template'][:]
+kde_signal = signal_dataset['kde_sig'][:]
```
tdent, line 23: This should be inside an if statement together with the line that creates the hdf file object for the signal file.

bin/all_sky_search/pycbc_plot_kde_vals
```diff
@@ -0,0 +1,64 @@
...
+parser.add_argument('--which-kde', choices=['signal_kde','template_kde', 'ratio_kde'])
+parser.add_argument('--plot-dir', required=True)
+parser.add_argument('--verbose', action='count')
+args = parser.parse_args()
+
+signal_dataset = h5py.File(args.signal_file, 'r')
+template_dataset = h5py.File(args.template_file, 'r')
+param0 = template_dataset[args.param[0]][:]    #mchirp,eta,chi_eff
+param1 = template_dataset[args.param[1]][:]
+param2 = template_dataset[args.param[2]][:]
```
tdent, line 21: I don't think the code uses param2 at all!  Also, if the user wants to plot kde vs param it should not be necessary to make 2 inputs to the param option.

bin/all_sky_search/pycbc_plot_kde_vals
```diff
@@ -0,0 +1,64 @@
...
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument('--signal-file')
+parser.add_argument('--template-file', required=True)
+parser.add_argument('--param', nargs='+', required=True, help='mchirp, eta, chi_eff')
+parser.add_argument('--plot-type', choices=['kde_vs_param', 'param_vs_param'])
+parser.add_argument('--which-kde', choices=['signal_kde','template_kde', 'ratio_kde'])
+parser.add_argument('--plot-dir', required=True)
+parser.add_argument('--verbose', action='count')
+args = parser.parse_args()
```
tdent, line 15: Can you add some input checking so that if the user chooses kde_vs_param the length of args.param should be constrained to be exactly 1, if they choose param_vs_param the length should be exactly 2.
The basic point is that there should not be any options given to the code that have no effect. So e.g. in the kde_vs_param case any --param options beyond the first one are not used and would be meaningless.

tdent:
@PRAVEEN-mnl can you paste two example plots here that could be produced with the kde_vs_param or with the param_vs_param options?

tdent:
@PRAVEEN-mnl I think I asked previously for you to paste some sample plot output files into a comment here, can you do that now with the current version of the code?

PRAVEEN-mnl:
mchirp_vs_eta_at_signal_kde
mchirp_vs_eta_at_ratio_kde
template_kde_vs_mchirp
template_kde_vs_chi_eff

tdent:
@PRAVEEN-mnl These look OK, but there are a few things to be fixed. The tick labels are too small - probably this is because of your choice of figure size.  

Primarily, the 'mchirp' axis is still a linear axis and is labelled 'mchirp' although I suppose what is plotted there is ln(mchirp).  So the axis label is not correct.  In any case, it is better to use a log axis rather than take the log of a quantity and plot it on a linear axis. 

In addition the user does not have a choice to plot the parameters on log axes or not.  It would be better to have an option like `--log-param` which takes the same number of inputs as `--param` and indicates whether the parameter(s) should be plotted on a log axis (True) or a linear axis (False). See the options in https://github.com/gwastro/pycbc/blob/master/bin/all_sky_search/pycbc_fit_sngls_over_multiparam .

bin/all_sky_search/pycbc_plot_kde_vals
```diff
@@ -0,0 +1,85 @@
...
+
+import numpy, h5py, argparse, logging
+import matplotlib.pyplot as plt
+from matplotlib.colors import LogNorm
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument('--signal-file')
+parser.add_argument('--template-file', required=True)
+parser.add_argument('--param', nargs='+', required=True, help='specify one-param and two-param in case of kde_vs_param and param_vs_param respectively')
+parser.add_argument('--log-axis', nargs='+', choices=['True', 'False'], required=True)
```
tdent, line 11: This looks OK, it would be nice to write a short help message for it

tdent:
@PRAVEEN-mnl I removed some plt.show() statements as these do not work when logged in on a cluster and we want the code to just save a plot png file. 
Have you checked whether the code can be run on a cluster to produce a png file? There may be some slight change we need to the setup for matplotlib.

tdent (approved):

PR #4053, opened 2022-06-14
Add statmap hierarchical removal against exclusive

GarethCabournDavies:
Adding in the possibility to do hierarchical removal against exclusive IFAR in the add_statmap code

While fixing this, a couple of errors in the implementation of the significance module for the add_statmap code were found (Not flagged due to not using HR in the tests), these have been fixed

### DIFF
--- a/bin/all_sky_search/pycbc_add_statmap
+++ b/bin/all_sky_search/pycbc_add_statmap
@@ -83,9 +83,7 @@ if args.max_hierarchical_removal == 0:
                      "These are conflicting options. Use with --help for more "
                      "information.")
 else :
-    is_bkg_inc = (args.hierarchical_removal_against == 'inclusive')
-    is_bkg_exc = (args.hierarchical_removal_against == 'exclusive')
-    if not(is_bkg_inc or is_bkg_exc):
+    if args.hierarchical_removal_against == 'none':
         parser.error("--max-hierarchical-removal requires a choice of which "
                      "background to remove foreground triggers against, "
                      "inclusive or exclusive. Use with --help for more "
@@ -455,7 +453,7 @@ bg_time_ct = {f_in.attrs['ifos'].replace(' ',''): f_in.attrs['background_time']
 # Counter for number of removals
 h_iterations = 0
 
-if is_bkg_inc:
+if args.hierarchical_removal_against == 'inclusive':
     ifar_key = 'ifar'
 else:
     ifar_key = 'ifar_exc'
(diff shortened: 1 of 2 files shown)


bin/all_sky_search/pycbc_add_statmap
```diff
@@ -532,42 +580,42 @@ while True:
...
     for key in all_ifo_combos:
         bnlouder, fnlouder = significance.get_n_louder(
-            significance_dict[ifo_combo_key],
             sep_bg_data[key].data['stat'],
             sep_fg_data[key].data['stat'],
-            sep_bg_data[key].data['decimation_factor'])
+            sep_bg_data[key].data['decimation_factor'],
+            **significance_dict[ifo_combo_key])
         # In principle, bg time should be adjusted, but is expected to be a
         # negligible corection
```
tdent, line 603: can you fix 'corection' ?

bin/all_sky_search/pycbc_add_statmap
```diff
@@ -68,6 +74,23 @@ if args.max_hierarchical_removal and injection_style:
...
+# Check that the user chose inclusive or exclusive background to perform
+# hierarchical removals of foreground triggers against.
+if args.max_hierarchical_removal == 0:
+    if args.hierarchical_removal_against != 'none':
+        parser.error("User Error: 0 maximum hierarchical removals chosen but "
+                     "option for --hierarchical-removal-against was given. "
+                     "These are conflicting options. Use with --help for more "
+                     "information.")
+else :
+    is_bkg_inc = (args.hierarchical_removal_against == 'inclusive')
```
tdent, line 86: I think you could save a couple of variables and a few lines by doing the check on 
`if args.hierarchical_removal_against is 'none'` and later also switching on the args attribute .. these two Boolean variables are basically not used for anything beyond repackaging the option value

tdent: i.e. if you can get by without adding variables that more or less reproduce the options, do so

tdent (changes requested):
In general looks OK, though I haven't checked every detail - see minor suggested changes in individual comments

bin/all_sky_search/pycbc_add_statmap
GarethCabournDavies: That's fair - this logic was copied over from other files which use HR, so I've edited it in the coinc_statmap file as well

tdent: Looks OK.  (Coinc statmap HR has very chatty comments doesn't it!)

tdent (approved):

PR #4054, opened 2022-06-14
MKL causes FFTW to segfault if loaded first

ahnitz:
This changes FFTW to use RTLD_DEEPBIND when loading the shared libraries as opposed to RTLD_GLOBAL. The latter is ineffective if MKL is first loaded (which can happen due to innocuous reasons like someone loading numpy before pycbc) because the symbols will still point to the MKL version. DEEPBIND enforces that the symbol lookup first goes to the loaded library rather than the global table so we can ensure FFTW functions are actually called when they are intended.

### DIFF
--- a/pycbc/libutils.py
+++ b/pycbc/libutils.py
@@ -160,7 +160,7 @@ def get_libpath_from_dirlist(libname, dirs):
     # If we get here, we didn't find it...
     return None
 
-def get_ctypes_library(libname, packages, mode=RTLD_DEEPBIND):
+def get_ctypes_library(libname, packages, mode=DEFAULT_RTLD_MODE):
     """
     This function takes a library name, specified in architecture-independent fashion (i.e.
     omitting any prefix such as 'lib' or suffix such as 'so' or 'dylib' or version number) and


ahnitz:
This fixes the test case segfault reported in #4052

pycbc/libutils.py
```diff
@@ -26,6 +26,14 @@
 from collections import deque
 from subprocess import getoutput
 
+
+# DLOPEN mode that requies symbols to first look up locally to this
+# library rather than first going to the global symbol table
+# This should be used when multiple libraries may export incompatible
+# symbols
+RTLD_DEEPBIND = 0x00008
```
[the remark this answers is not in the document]
spxiwh: I would prefer using this (thanks Josh!), but it seems like mac does not have this. If we are right that this flag is not needed in mac anyway, then perhaps just wrap this in a "is linux" block and use default flags for mac.

pycbc/fft/fftw.py
```diff
@@ -30,8 +30,8 @@
 # We need to construct them directly with CDLL so
 # we can give the RTLD_GLOBAL mode, which we must do
 # in order to use the threaded libraries as well.
-double_lib = get_ctypes_library('fftw3',['fftw3'],mode=ctypes.RTLD_GLOBAL)
-float_lib = get_ctypes_library('fftw3f',['fftw3f'],mode=ctypes.RTLD_GLOBAL)
```
spxiwh, line 34: Should the mode be set deeper, ie. in the `get_ctypes_library` function itself, so that it is used *all the time* when using ctypes??

spxiwh: ... Or make this flag the default??

pycbc/libutils.py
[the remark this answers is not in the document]
josh-willis: At the least it should probably really be:
```
import os
if hasattr(os, RTLD_DEEPBIND):
    RTLD_DEEPBIND = os.RTLD_DEEPBIND
else
    RTLD_DEEPBIND = 0x00008 # Or maybe we set it to RTLD_GLOBAL if that flag
                   # just doesn't exist on a given platform?
```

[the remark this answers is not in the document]
ahnitz: They are the same value, so yes, in testing they both work for linux at least.

pycbc/fft/fftw.py
josh-willis: We should check, but I think this would make sense.  Certainly if we are putting logic to decide whether or not a given installation has (or should use) `RTLD_DEEPCOPY` versus `RTLD_GLOBAL`, then I think it makes sense to do that in only one place.  However I think I prefer the route of making it a default, so that we retain the ability to do something different on a library by library basis, in case we find we need to do something different elsewhere.

pycbc/libutils.py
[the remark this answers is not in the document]
ahnitz: @josh-willis Ok, looks like that this doesn't work on macos unfortunately as the variable doesn't exist on that platform. Darn.

spxiwh:
I'm happy to have this be the default *and* be set in fftw.py.

But I think we do need to use `os. RTLD_DEEPBIND`. If this isn't defined in mac it implies that this would do nothing in mac anyway, and then it's not clear what setting this would do on macs at all.

I would prefer not changing the previous behaviour *at all* on mac, and using `os. RTLD_DEEPBIND` on linux.

I tried running the initial test at the top of https://github.com/gwastro/pycbc/issues/4052 on my mac and it ran fine anyway!!

pycbc/libutils.py
```diff
@@ -31,7 +31,12 @@
 # library rather than first going to the global symbol table
 # This should be used when multiple libraries may export incompatible
 # symbols
-RTLD_DEEPBIND = 0x00008
+#
+# This is only defined for linux systems, on macosx, the default behavior
+# is similar
+RTLD_DEEPBIND = 0
+if hasattr(os, 'RTLD_DEEPBIND'):
+    RTLD_DEEPBIND = os.RTLD_DEEPBIND
```
josh-willis, line 40: so maybe taking @spxiwh 's comment into account, what we want here is:
```
if hasattr(os, 'RTLD_DEEPBIND'):
    DEFAULT_RTLD_MODE = os.RTLD_DEEPBIND
else:
   DEFAULT_RTLD_MODE = ctypes.DEFAULT_MODE
```
and then change below to:
```
def get_ctypes_library(libname, packages, mode=DEFAULT_RTLD_MODE)
```
which then also allows override in individual libraries.  Or would you rather export something like `LIBUTILS_RTLD_DEEPCOPY` and have that the default, so that we are then also explicitly setting that in `fftw` rather than relying on the default?  Although it is arguably also confusing to have something that implies you're doing `DEEPCOPY` but in fact on some os's it is a noop.

pycbc/libutils.py
```diff
@@ -26,6 +26,19 @@
...
 
+
+# DLOPEN mode that requies symbols to first look up locally to this
+# library rather than first going to the global symbol table
+# This should be used when multiple libraries may export incompatible
+# symbols
+#
+# This is only defined for linux systems, on macosx, the default behavior
+# is similar
+RTLD_DEEPBIND = 0
```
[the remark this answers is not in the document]
ahnitz: Gosh darn, it looks like the bit definitions aren't even compatible, for comparison with glibc https://code.woboq.org/userspace/glibc/bits/dlfcn.h.html

pycbc/libutils.py
ahnitz: @josh-willis Yeah, either way, non-perfect, but I think I'll opt for just setting up the default as you've proposed and fftw won't set the mode at all.

ahnitz:
@josh-willis Do you recall why we were setting the mode at all in FFTW to RTLD_GLOBAL? That's the ony niggle I'm not sure about.

ahnitz:
Test failures atm are due to git.ligo.org being down.

josh-willis:
> @josh-willis Do you recall why we were setting the mode at all in FFTW to RTLD_GLOBAL? That's the ony niggle I'm not sure about.

I'm afraid I don't.  The things I would want to check are that (for instance) we can load wisdom from a file and then that wisdom gets used, since that's a variable internal inside of FFTW that we can't inspect.  I may have just added that "to be sure."

Also, reading over the `ctypes` docs, I see this:
```

ctypes.RTLD_GLOBAL

    Flag to use as mode parameter. On platforms where this flag is not available, it is defined as the integer zero.

ctypes.RTLD_LOCAL

    Flag to use as mode parameter. On platforms where this is not available, it is the same as RTLD_GLOBAL.

ctypes.DEFAULT_MODE

    The default mode which is used to load shared libraries. On OSX 10.3, this is RTLD_GLOBAL, otherwise it is the same as RTLD_LOCAL.
```
So before, when in FFTW we were explicitly passing `RTLD_GLOBAL`, on new enough macs, we may in fact have been passing zero, since we were *not* opening it with the default for that OS.

ahnitz:
@josh-willis I tested this fixed the initial issue reported in #4052 in conda. 

@josh-willis @spxiwh Is there anything specific we need to run that isn't covered in the unittests, or should we merge this now? 

I think after we merge we should also do a release.

pycbc/libutils.py
josh-willis: LGTM assuming all unit tests pass. And I agree we should do a release after this.

josh-willis (approved):

josh-willis:
As none of the codebeat issues even seem related to these changes, I am fine with merging this whenever you are ready @ahnitz . I'll leave that to you, and also to wait and see if @spxiwh has additional feedback.

spxiwh (approved):
Also looks good to me. For anyone referring to this PR in the future, we should be clear that this failure won't *always* happen if you have MKL and FFTW installed on conda. For e.g. our basic tests use conda, and install both MKL and FFTW, and work fine (though they may use `pip install` to bring in at least MKL). The IGWN conda environments also seem to work fine (and I assume there conda is used to bring in MKL). I think the common factor is some form of accelerated libraries (BLAS??, numpy??) that are brought in when you want MKL for more than just doing FFTs.

Anyway, happy to have this merged and let's cut a new release.

PR #4057, opened 2022-06-16
examples: make slice indices integers

nogoegst:
Python 3 now divides integers producing floats, so this to go back to integer division.


ahnitz:
@nogoegst Thank you for this PR! It seems quite straightforward.

PR #4061, opened 2022-06-16
PyGRB Workflow Generation Updates

jakeb245:
This contains the changes made needed to make the PyGRB workflow generator work with `pycbc_multi_inspiral`. Currently, a simple workflow containing splitbank and inspiral jobs is able to be generated and successfully ran with the right configuration options. 

Some of the code was already a bit rough, and I haven't done a lot to fix that yet. I may have even made it worse at times, but it does work. Therefore I am marking this PR as work in progress as I'd like to clean some things up before anything is merged. Advice on how to do that would be appreciated! I tried to make it work with the original coh_ptf code for now, but dropping that would make things simpler. 

There are also some fixes to `pycbc_multi_inspiral` within this PR. One of them was a fault of myself not noticing a check in #3798. My fix to that here works (I think), but I intend to come up with a better way of doing it. I also changed the call to the auto chi-squared test so that it uses the SNR timeseries like it's supposed to. Before, it was using the list of trigger SNRs and sometimes causing an index error. I'm considering moving these fixes to a separate pull request.


pycbc/workflow/matched_filter.py
```diff
@@ -240,9 +240,9 @@ def setup_matchedfltr_dax_generated_multi(workflow, science_segs, datafind_outs,
 
     if match_fltr_exe == 'pycbc_multi_inspiral':
         exe_class = select_matchedfilter_class(match_fltr_exe)
-        cp.set('inspiral', 'longitude',\
+        cp.set('inspiral', 'ra',\
```
titodalcanton, line 243:
```suggestion
        cp.set('inspiral', 'ra',
```

pycbc/workflow/matched_filter.py
```diff
@@ -240,9 +240,9 @@ def setup_matchedfltr_dax_generated_multi(workflow, science_segs, datafind_outs,
...
                str(radians(float(cp.get('workflow', 'ra')))))
-        cp.set('inspiral', 'latitude',\
+        cp.set('inspiral', 'dec',\
```
titodalcanton, line 245:
```suggestion
        cp.set('inspiral', 'dec',
```

pycbc/workflow/segment.py
```diff
@@ -305,15 +305,16 @@ def generate_triggered_segment(workflow, out_dir, sciencesegs):
     # How many IFOs meet minimum data requirements?
     min_seg = segments.segment(triggertime - onbefore - minbefore - padding,
                                triggertime + onafter + minafter + padding)
+    # Work in progess below
     scisegs = segments.segmentlistdict({ifo: sciencesegs[ifo]
             for ifo in sciencesegs.keys() if min_seg in sciencesegs[ifo]
             and abs(sciencesegs[ifo]) >= minduration})
-
+
```
titodalcanton, line 312:
```suggestion

```

pycbc/workflow/jobsetup.py
```diff
@@ -698,18 +702,19 @@ def create_node(self, data_seg, valid_seg, parent=None, inj_file=None,
...
-        if self.cp.has_option('inspiral', 'do-bank-veto'):
+        if self.cp.has_option('inspiral', 'do-bank-veto') or self.cp.has_option('workflow-inspiral', 'bank-veto-bank-file'):
             if not bankVetoBank:
                 raise ValueError("%s must be given a bank veto file if the "
                                  "argument 'do-bank-veto' is given"
                                  % self.name)
-            node.add_input_opt('--bank-veto-templates', bankVetoBank)
-
+            node.add_input_opt('--bank-veto-bank-file', bankVetoBank)
+
```
titodalcanton, line 711:
```suggestion

```

jakeb245:
Since creating this PR I've made some progress. Currently the code is able to generate a PyGRB workflow running the following jobs in order.
1. Splitbank
2. Inspiral (no injections)
3. Trig combiner
4. Trig cluster

The rest of the PyGRB workflow is still under development by others. I think this PR is in a state where it can be merged (after review) with more changes coming in new PRs as other parts of the workflow are finished.

jakeb245:
I'm going to be unable to work on this for several weeks, so I'll give an update on this PR and then respond to reviews until it can be merged.

This PR brings the PyGRB workflow generation code to a state where it can run the new executables until post-processing plots/efficiency. See [this wiki](https://github.com/gwastro/pycbc/wiki/PyGRB) for more information. Some general issues I've found are:
- `pycbc_grb_inj_combiner` does not work with the output of `pycbc_grb_inj_finder`. The file structures do not seem to be compatible as written. The workflow generates and runs both executables without failure, but the combiner does not do anything.
- Currently, the config files given to `pycbc_create_injections` must be hardcoded for the event-specific parameters. In the future, we likely want the workflow generator to dynamically generate config files with those settings. Getting EM bright and jitter skyloc functionality merged into master should likely happen before that.
- CodeClimate is failing, but the issues don't seem worth fixing to me. I don't quite understand the issues, so I'm happy to fix them if requested.

Some of these issues are outside the scope of the PyGRB workflow generator and can be fixed in seperate PRs. As more updates to the PyGRB workflow are finalized, more updates to the workflow generator will be required. These updates should cover all work that has been finished and merged into master. There are also some small fixes in the executables required for them to run.

bin/pycbc_hdf_splitinj
```diff
@@ -0,0 +1,70 @@
...
+
+import argparse
+from pycbc.inject import InjectionSet
+import h5py
+import numpy as np
+import pycbc.version
+
+"""
+Splits injections created using pycbc_create_injections into smaller sets
+"""
```
titodalcanton, line 11: Move the docstring between the shebang and the imports.

bin/pycbc_hdf_splitinj
```diff
@@ -0,0 +1,70 @@
...
+
+# Parse command line
+parser = argparse.ArgumentParser()
+parser.add_argument("--version", action="version",
+                    version=pycbc.version.git_verbose_msg)
+parser.add_argument("-f", "--output-files", nargs='*', default=None,
+                   help="Names of output files")
+
+parser.add_argument("-i", "--input-file", help="Injection file to be split")
```
titodalcanton, line 20:
```suggestion
parser.add_argument("-f", "--output-files", nargs='*', required=True,
                    help="Names of output files")
parser.add_argument("-i", "--input-file", required=True,
                    help="Injection file to be split")
```

bin/pycbc_hdf_splitinj
```diff
@@ -0,0 +1,70 @@
...
+
+args = parser.parse_args()
+
+# Read in input file as both an hdf file and an InjectionSet object
+inj_file = h5py.File(args.input_file)
```
titodalcanton, line 25:
```suggestion
inj_file = h5py.File(args.input_file, 'r')
```

bin/pycbc_hdf_splitinj
```diff
@@ -0,0 +1,70 @@
...
+
+# InjectionSet.write() requires static params as a dictionary,
+# so get that from file object.
+# Ignore the "static_params" copy.
+static_params = {key : inj_file.attrs[key] for key in inj_file.attrs
+                 if key != 'static_args'}
+
+# Also get the names of variable params as write_args
+write_args = [arg for arg in inj_table.fieldnames
+              if arg not in static_params.keys()]
```
titodalcanton, line 39: You can probably drop the `.keys()` here.

bin/pycbc_hdf_splitinj
```diff
@@ -0,0 +1,70 @@
...
+injs_per_split = np.zeros(num_splits) + ideal_split
+
+# Handle the remainder if it exists
+if remainder > 0:
+    for i in range(remainder):
+        injs_per_split[i] += 1
+
+# Sanity check: did we get account for all injs?
+if sum(injs_per_split) != num_injs:
+    raise ValueError("Not all injections were accounted for!")
```
titodalcanton, line 57: An assert is probably better than a ValueError here:
```suggestion
assert sum(injs_per_split) == num_injs, "Not all injections were accounted for!"
```

bin/pycbc_hdf_splitinj
```diff
@@ -0,0 +1,70 @@
...
+              if arg not in static_params.keys()]
+
+num_injs = len(inj_table)
+num_splits = len(args.output_files)
+
+# Ideally, the number of injections is divsible by number of splits
+# with no remainder, but that is not always true
+ideal_split = num_injs // num_splits
+remainder = num_injs % num_splits
+injs_per_split = np.zeros(num_splits) + ideal_split
```
titodalcanton, line 48:
```suggestion
injs_per_split = np.zeros(num_splits, dtype=int) + ideal_split
```

bin/pycbc_hdf_splitinj
```diff
@@ -0,0 +1,70 @@
...
+if sum(injs_per_split) != num_injs:
+    raise ValueError("Not all injections were accounted for!")
+
+# Sort injections by time
+inj_table.sort(order='tc')
+
+# Split injections into a list of smaller sets
+for i in range(num_splits):
+    # Number of injections in this split
+    injs_in_split = int(injs_per_split[i])
```
titodalcanton, line 65:
```suggestion
    injs_in_split = injs_per_split[i]
```

bin/pycbc_hdf_splitinj
titodalcanton: After seeing the comment towards the end of the script, I would also add a sentence to the docstring explaining that the split injections are organized so as to maximize the time between injections in the output files. Otherwise, the user may simply assume that the first output file will get the first N injections in the input file, etc.

bin/pycbc_multi_inspiral
```diff
@@ -349,10 +349,8 @@ with ctx:
             # Only do this is idx doesn't get time shifted out of the time
             # we are looking at.
             idx_dict = {
-                ifo: idx[ifo][np.logical_and(
-                    idx[ifo] > time_delay_idx[ifo],
-                    idx[ifo] - time_delay_idx[ifo] < len(snr_dict[ifo])
-                    )
+                ifo: idx[ifo][
+                    idx[ifo] + time_delay_idx[ifo] < len(snr_dict[ifo]) - 1
```
titodalcanton, line 353: Since I am not familiar with the code around this change, what is the reason for the change? Was the old logic incorrect? Should the code also check that `idx[ifo] + time_delay_idx[ifo] >= 0`?

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -476,9 +478,9 @@ numpy.random.seed(args.seed)
 # Randomly pick num_trials offsource trials
 offtrials = numpy.random.choice(range(sum(offsource_ntrials)), args.num_trials, replace=False)
 offtrials.sort()
-for i in offtrials:
-    _ts = bins["OFFSOURCE"][0][0] if i < offsource_ntrials[0] else bins["OFFSOURCE"][1][0]
-    _ts += i*trialtime
+for i, j in enumerate(offtrials):
+    _ts = bins["OFFSOURCE"][0][0] if j < offsource_ntrials[0] else bins["OFFSOURCE"][1][0]
+    _ts += j*trialtime
```
titodalcanton, line 483:
```suggestion
    _ts += j * trialtime
```

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -77,8 +78,11 @@ wflow.cp = _workflow.set_grb_start_end(wflow.cp, start, end)
 # Retrieve science segments
 currDir = os.getcwd()
 segDir = os.path.join(currDir, "segments")
-sciSegs, segsFileList = _workflow.setup_segment_generation(wflow, segDir)
+sciSegsFile = _workflow.get_segments_file(wflow, "science", 'segments-science', segDir)
```
titodalcanton, line 81:
```suggestion
sciSegsFile = _workflow.get_segments_file(wflow, 'science', 'segments-science', segDir)
```

pycbc/workflow/grb_utils.py
```diff
@@ -36,10 +36,45 @@
...
+        'pycbc_grb_trig_combiner': PycbcGrbTrigCombinerExecutable,
+        'pycbc_grb_trig_cluster': PycbcGrbTrigClusterExecutable,
+        'pycbc_grb_inj_finder': PycbcGrbInjFinderExecutable,
+        'pycbc_grb_inj_combiner': PycbcGrbInjCombinerExecutable
+    }
+    try:
+        return exe_to_class_map[exe_name]
+    except KeyError:
+        raise NotImplementedError(
+            "No job class exists for executable %s, exiting" % curr_exe)
```
titodalcanton, line 75:
```suggestion
    if exe_name not in exe_to_class_map:
        raise ValueError(f"No job class exists for executable {curr_exe}")
```

pycbc/workflow/grb_utils.py
```diff
@@ -365,3 +400,177 @@ def get_sky_grid_scale(
...
+
+    # Find injections from triggers
+    exe_class = select_grb_pp_class(wf, "inj_finder")
+    job_instance = exe_class(wf.cp, "inj_finder")
+    inj_find_files = FileList([])
+    for inj_tag in inj_tags:
+        tag_inj_files = FileList([f for f in inj_files
+                                  if inj_tag in f.tags])
+        tag_insp_files = FileList([f for f in inj_insp_files
+                                   if inj_tag in f.tags[1]])
```
titodalcanton, line 438: Is the `[1]` index intended on the last line (and not intended a couple of lines before)? If so, consider adding a comment explaining why you only look at the second element.

pycbc/workflow/grb_utils.py
```diff
@@ -365,3 +400,177 @@ def get_sky_grid_scale(
...
+        node.add_opt('--verbose')
+        node.add_opt("--ifo-tag", ifo_tag)
+        node.add_opt("--grb-name", self.trigger_name)
+        node.add_opt("--trig-start-time", self.trig_start_time)
+        node.add_opt("--segment-dir", seg_dir)
+        node.add_input_list_opt("--input-files", insp_files)
+        node.add_opt("--user-tag", "PYGRB")
+        node.add_opt("--num-trials", self.num_trials)
+        # Prepare output file tag
+        user_tag = "PYGRB_GRB{}".format(self.trigger_name)
```
titodalcanton, line 488: F-strings are great, use them:
```suggestion
        user_tag = f"PYGRB_GRB{self.trigger_name}"
```

pycbc/workflow/grb_utils.py
```diff
@@ -365,3 +400,177 @@ def get_sky_grid_scale(
...
+    def create_node(self, input_file, out_dir, ifo_tag, segment, tags=None):
+        if tags is None:
+            tags = []
+        node = Node(self)
+        node.add_input_opt('--input-files', input_file)
+        out_name = input_file.name[:-3] + '-FILTERED.h5'
+        out_file = File(ifo_tag, 'inj_combiner', segment,
+                        os.path.join(out_dir, out_name), tags=tags)
+        node.add_output_opt('--output-file', out_file)
+        node.add_opt('--max-inclination', 30)
```
titodalcanton, line 575: This sounds like something that we absolutely want in a config file and not hardcoded here, maybe add a FIXME comment.

pycbc/workflow/injection.py
```diff
@@ -226,7 +227,13 @@ def setup_injection_workflow(workflow, output_dir=None,
             inj_job = exe(workflow.cp, inj_section_name,
                           out_dir=output_dir, ifos='HL',
                           tags=curr_tags)
-            node = inj_job.create_node(full_segment)
+            if exe is PycbcCreateInjectionsExecutable:
```
titodalcanton, line 230: Does this actually work, or did you intend to use `isinstance()`?

pycbc/workflow/jobsetup.py
```diff
@@ -290,8 +290,12 @@ def multi_ifo_coherent_job_setup(workflow, out_files, curr_exe_job,
...
     else:
         ipn_sky_points = None
-        bank_veto = datafind_outs[-1]
-        frame_files = datafind_outs[:-1]
+        if 'bank_veto_bank' in datafind_outs[-1].name:
+            bank_veto = datafind_outs[-1]
+            frame_files = datafind_outs[:-1]
+        else:
+            bank_veto = None
+            frame_files = datafind_outs
```
titodalcanton, line 298: This function seems particularly fragile (and the comment just above the change appears to agree). Looping over `datafind_outs` until you find one with the right name seems a better way, would that not work?

pycbc/workflow/jobsetup.py
```diff
@@ -696,35 +700,27 @@ def create_node(self, data_seg, valid_seg, parent=None, inj_file=None,
...
         if dfParents is not None:
-            node.add_arg('--frame-cache %s' % \
-                         " ".join([":".join([frameCache.ifo, frameCache.name])\
-                                   for frameCache in dfParents]))
-            for frameCache in dfParents:
-                node.add_input(frameCache)
-            #node.add_input_list_opt('--frame-cache', dfParents)
+            frame_arg = '--frame-files'
+            for frame_file in dfParents:
+                frame_arg += ' ' + ':'.join([frame_file.ifo, frame_file.name])
```
titodalcanton, line 721: Much simpler with an f-string:
```suggestion
                frame_arg += f' {frame_file.ifo}:{frame_file.name}'
```

pycbc/workflow/matched_filter.py
```diff
@@ -240,10 +239,10 @@ def setup_matchedfltr_dax_generated_multi(workflow, science_segs, datafind_outs,
...
     if match_fltr_exe == 'pycbc_multi_inspiral':
         exe_class = select_matchedfilter_class(match_fltr_exe)
-        cp.set('inspiral', 'longitude',\
-               str(radians(float(cp.get('workflow', 'ra')))))
-        cp.set('inspiral', 'latitude',\
-               str(radians(float(cp.get('workflow', 'dec')))))
+        cp.set('inspiral', 'ra',
+               str(cp.get('workflow', 'ra')))
+        cp.set('inspiral', 'dec',
+               str(cp.get('workflow', 'dec')))
```
titodalcanton, line 245: Does this change imply that the config files will now have to specify RA and declination in radians, or that the inspiral executable now takes degrees?

bin/pycbc_multi_inspiral
jakeb245: The time delay was previously applied using a numpy operation. It was changed so that time delay is now added as necessary. The logic check here was overlooked when that change was made, so when I was running workflow with many triggers it would try and run chi-squared tests on triggers with indices outside of the data. There certainly could be cases where more logic checks are necessary.

bin/pygrb/pycbc_grb_inj_finder
```diff
@@ -322,10 +313,13 @@ with h5py.File(outfilename, "w") as h5out:
...
-        injs = injarr[injlist]
+        injs = allinjections[injlist]
         for field in injs.fieldnames:
-            grp[field] = injs[field]
-
+            # h5py does not support unicode strings
+            if 'U' not in str(injs[field].dtype):
+                grp[field] = injs[field]
+            else:
+                grp[field] = str(injs[field])
```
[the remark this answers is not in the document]
jakeb245: I had to do this because h5py does not seem to support Unicode data types. It works, but there may be a better way to work around it.

pycbc/workflow/injection.py
jakeb245: I probably did intend to use `isinstance()` but this does work

pycbc/workflow/jobsetup.py
jakeb245: This code is definitely questionable. I can look into rewriting it.

pycbc/workflow/matched_filter.py
jakeb245: I believe I changed this because `pycbc_multi_inspiral` was changed to use RA and declination in radians.

pycbc/workflow/injection.py
titodalcanton: I am not convinced:
```
In [1]: class Test:
   ...:     def __init__(self):
   ...:         self.foo = 12
   ...: 

In [2]: x = Test()

In [3]: isinstance(x, Test)
Out[3]: True

In [4]: x is Test
Out[4]: False
```

titodalcanton: Ah sorry, `select_generic_executable()` returns the actual class, not an instance. So yeah, this would work.

bin/pygrb/pycbc_grb_inj_finder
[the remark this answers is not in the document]
titodalcanton: I get the h5py and Unicode issue. I think my issue with the last line is the following. `injs[field]` would be a Numpy array of strings here. What you want to do is convert the Numpy array of strings to a list of Python strings, which h5py will support. However, the last line will not do that. It will create a *single* string, containing just how Numpy would print the array. This should do what you want:
```suggestion
                grp[field] = [str(s) for s in injs[field]]
```

pycbc/workflow/splittable.py
```diff
@@ -147,8 +150,8 @@ def setup_splittable_dax_generated(workflow, input_tables, out_dir, tags):
             num_injs = int(cp.get_opt_tags("em_bright_filter", "max-keep",
                                            tags))
         else:
-            num_injs = int(cp.get_opt_tags("workflow-injections", "num-injs",
-                                           tags))
+            num_injs = int(cp.get_opt_tags("workflow-injections",
+                                           "ninjections", tags))
```
titodalcanton, line 154: This seems to require a syntax change in the configuration file. It would be good if the new workflow could run with old configuration files as much as possible. Given that this is a trivial change, can we avoid it for now?

pycbc/workflow/matched_filter.py
jakeb245: I believe I was thinking that with `pycbc_multi_inspiral` now taking RA and dec, it could be directly provided from the config file without any conversion. Looking at it again, I'm not sure if that's correct. I'll have to look into this further.

pycbc/workflow/grb_utils.py
jakeb245: I think this was because the first file in the inspiral file list is (but is not always) the bank veto bank. Looking at it now, I can probably come up with a better way to get around it.

titodalcanton: Any update on this?

pycbc/workflow/matched_filter.py
titodalcanton: Has this been looked into?

titodalcanton: On the call, we decided to keep the conversion from degrees (easy for whoever sets up the .ini file from a GRB) to radians (which is what `pycbc_multi_inspiral` understands).

pycbc/workflow/grb_utils.py
titodalcanton: Please check this or at least add a FIXME comment so we do not forget.

pycbc/workflow/splittable.py
jakeb245: I remembered why I did this... it was because `pycbc_create_injections` uses this variable name. I've reverted it and a comment clarifying this reasoning. I can also add a check for the injections executable, or there could be a better way to handle it.

titodalcanton: Ok.

titodalcanton (approved):
Thanks @jakeb245, I think it is now time to get this merged so development can continue.

PR #4063, opened 2022-06-17
Awkde setup fix

PRAVEEN-mnl:


tdent (approved):
Seems OK, we will switch this back to mennthor master branch again when that has been updated with the fix

tdent:
FWIW, I put this on 'draft' status when I did not know if the awkde set change had been merged to master yet. But that can be reversed

PR #4066, opened 2022-06-24
Don't import bank output utils at package level

tdent:
Output utils is abandonware used only for ligolw compatibility, & would like to avoid mandatorily importing io code with tmpltbank


tdent:
OK .. these are not tested but I can try git grep to see what uses the sngl_inspiral or ethinca functions.

titodalcanton (approved):
Looks good to me.

PR #4050, opened 2022-06-13
add statmap files for use with single-detector triggers

GarethCabournDavies:
Adding two file which convert findtrigs output --> IFAR

Exclude_zerolag and add_statmap can be used with these as inputs without modification to those codes

PR #4070, opened 2022-06-27
Sngls results plotting

GarethCabournDavies:
Some changes were required to allow search including single-detector events to have sensible results pages

These changes are compatible with #4050 

- hdfinjfind: add if statements as the pivot/fixed ifo doesn't make sense without timeshifts
- foundmissed plots: add maximum SNRs and minimum distances to the coloring schemes, as these are decisive in searches which include single-detector triggers
- ifar_catalog: events now may have _really_ low ifar, so using minimum ifar as a limit to the x-axis no longer makes sense
- page_foreground: changes are basically to remove the requirement to say coincs from the code
- page_ifar: change gets rid of the legacy files support
- page_snrifar: change fixes an issue where open-box values could be inferred from closed-box plot limits

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_hdfinjfind
+++ b/bin/all_sky_search/pycbc_coinc_hdfinjfind
@@ -51,7 +51,7 @@ def xml_to_hdf(table, hdf_file, hdf_key, columns):
         else:
             new_col = col
         key = os.path.join(hdf_key, new_col)
-        hdf_append(hdf_file, key, numpy.array(table.get_column(col),
+        hdf_append(hdf_file, key, numpy.array(table.getColumnByName(col),
                                         dtype=numpy.float32))
 
 parser = argparse.ArgumentParser(description=__doc__)
@@ -142,8 +142,8 @@ for trigger_file, injection_file in zip(args.trigger_files,
         indoc = ligolw_utils.load_filename(injection_file, False,
                                            contenthandler=LIGOLWContentHandler)
         sim_table = lsctables.SimInspiralTable.get_table(indoc)
-        inj_time = numpy.array(sim_table.get_column('geocent_end_time') +
-                               1e-9 * sim_table.get_column('geocent_end_time_ns'),
+        inj_time = numpy.array(sim_table.getColumnByName('geocent_end_time').asarray() +
+                               1e-9 * sim_table.getColumnByName('geocent_end_time_ns').asarray(),
                                dtype=numpy.float64)
     else:
         inj_file = CBCHDFInjectionSet(injection_file)
@@ -175,33 +175,38 @@ for trigger_file, injection_file in zip(args.trigger_files,
     logging.info('Found: %s, Missed: %s' %
                  (len(found_within_time), len(missed_within_time)))
 
-    logging.info('Removing injections in vetoed time')
+    if args.veto_file:
+        logging.info('Removing injections in vetoed time')
 
-    # Put individual detector vetoes into list of all vetoed indices, if an
-    # injection is vetoed in N ifos then its index will appear N times
-    vetoid = numpy.array([])
-    veto_dict = {}
-    for ifo in ifo_list:
-        vi, _ = indices_within_segments(inj_time, [args.veto_file], ifo=ifo,
-                                        segment_name=args.segment_name)
-        vetoid = numpy.append(vetoid, vi)
-        veto_dict[ifo] = vi
+        # Put individual detector vetoes into list of all vetoed indices, if an
+        # injection is vetoed in N ifos then its index will appear N times
+        vetoid = numpy.array([])
+        veto_dict = {}
+        for ifo in ifo_list:
+            vi, _ = indices_within_segments(inj_time, [args.veto_file], ifo=ifo,
+                                            segment_name=args.segment_name)
+            vetoid = numpy.append(vetoid, vi)
+            veto_dict[ifo] = vi
 
-    # Find out which of these indices are unique, and how many times each
-    # vetoed index occurs - this is how many detectors it is vetoed in
-    vetoid_unique, count_vetoid = numpy.unique(vetoid, return_counts=True)
+        # Find out which of these indices are unique, and how many times each
+        # vetoed index occurs - this is how many detectors it is vetoed in
+        vetoid_unique, count_vetoid = numpy.unique(vetoid, return_counts=True)
 
-    # remove injections where the number of unvetoed ifos is less than the
-    # minimum specified by the user
-    vetoed_all = vetoid_unique[(len(ifo_list) - count_vetoid)
-                               < args.min_required_ifos]
-    found_after_vetoes = numpy.array([i for i in found_within_time
-                                      if i not in vetoed_all])
-    missed_after_vetoes = numpy.array([i for i in missed_within_time
-                                      if i not in vetoed_all]).astype(int)
+        # remove injections where the number of unvetoed ifos is less than the
+        # minimum specified by the user
+        vetoed_all = vetoid_unique[(len(ifo_list) - count_vetoid)
+                                   < args.min_required_ifos]
+        found_after_vetoes = numpy.array([i for i in found_within_time
+                                          if i not in vetoed_all])
+        missed_after_vetoes = numpy.array([i for i in missed_within_time
+                                          if i not in vetoed_all]).astype(int)
+        logging.info('Found: %s, Missed: %s' %
+                     (len(found_after_vetoes), len(missed_after_vetoes)))
+    else:
+        veto_dict = {ifo: [] for ifo in ifo_list}
+        found_after_vetoes = found_within_time
+        missed_after_vetoes = missed_within_time.astype(int)
 
-    logging.info('Found: %s, Missed: %s' %
-                 (len(found_after_vetoes), len(missed_after_vetoes)))
 
     found_fore = numpy.arange(0, len(stat), 1)[left[found]]
     found_fore_v = numpy.arange(0, len(stat), 1)[left[found_after_vetoes]]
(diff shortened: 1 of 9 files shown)


GarethCabournDavies:
I've only just realised this was left open for so long. I've added a couple of needed fixes as my tests were failing, but things had got past the CI tests somehow.

I've also made a single-source-of-truth fix in page_foreground and a (significant) efficiency saving if no veto file is given in the hdfinjfind code.

GarethCabournDavies:
Fixes
 - ligolw doesnt seem to have get_column any more, and so using getColumnByName seems to work
 - use logging rather than print statements in pycbc_ifar_catalog
 - hdf io file was not doing input sanitation on the template_ids, which should be ints but weren't, causing an indexing problem

bin/all_sky_search/pycbc_coinc_hdfinjfind
```diff
@@ -175,33 +175,38 @@ for trigger_file, injection_file in zip(args.trigger_files,
...
-
-    # remove injections where the number of unvetoed ifos is less than the
-    # minimum specified by the user
-    vetoed_all = vetoid_unique[(len(ifo_list) - count_vetoid)
-                               < args.min_required_ifos]
-    found_after_vetoes = numpy.array([i for i in found_within_time
-                                      if i not in vetoed_all])
-    missed_after_vetoes = numpy.array([i for i in missed_within_time
-                                      if i not in vetoed_all]).astype(int)
+    if args.veto_file:
```
GarethCabournDavies, line 178: This is _really_ slow even if you don't have the veto file - save time if not

bin/plotting/pycbc_ifar_catalog
```diff
@@ -236,9 +241,10 @@ if opts.open_box:
 if opts.open_box:
     # scale the plot around the foreground
     pylab.ylim(0.7, 0.05 * len(fore_cumnum))
-    pylab.xlim(30 * min(fore_ifar), 3 * max_ifar)
+    pylab.xlim(min(expected_ifar) / 30, 3 * max_ifar)
```
tdent, line 244: why is this an improvement?

bin/plotting/pycbc_page_foundmissed
```diff
@@ -9,16 +9,30 @@ import pycbc.pnutils
...
 parser.add_argument('--injection-file',
                     help="The hdf injection file to plot", required=True)
 parser.add_argument('--axis-type', default='mchirp', choices=['mchirp',
                     'effective_spin', 'time', 'mtotal', 'mass_ratio'])
 parser.add_argument('--log-x', action='store_true', default=False)
 parser.add_argument('--distance-type', default='decisive_optimal_snr',
-                    choices=['decisive_distance', 'dec_chirp_distance',
-                             'chirp_distance', 'comb_optimal_snr',
-                             'decisive_optimal_snr', 'redshift'],
+                    choices=list(labels.keys()),
```
tdent, line 35: I think you should just be able to do `list(labels)` in python3

bin/plotting/pycbc_page_foundmissed
```diff
@@ -9,16 +9,30 @@ import pycbc.pnutils
...
+        'dec_chirp_distance': 'Injected Decisive Chirp Distance (Mpc)',
+        'min_eff_distance': 'Minimum Effective Injected Distance (Mpc)',
+        'min_eff_chirp_distance': 'Minimum Effective Injected Chirp Distance (Mpc)',
+        'chirp_distance': 'Injected Chirp Distance (Mpc)',
+        'comb_optimal_snr': 'Combined Optimal SNR',
+        'decisive_optimal_snr': 'Decisive Optimal SNR',
+        'max_optimal_snr': 'Maximum Optimal SNR',
+        'redshift': 'Redshift',
+        'time': 'Time (s)',
+        'effective_spin': 'Weighted Aligned Spin',
```
tdent, line 25: I think the usual term for this is now 'Effective Inspiral Spin'

bin/plotting/pycbc_page_foundmissed
```diff
@@ -59,6 +75,10 @@ elif args.far_type == 'exclusive':
     ifar_found = f['found_after_vetoes/ifar_exc'][:]
     far_title = 'Exclusive'
 
+if args.limit_ifar:
+    ifar_above_thresh = ifar_found > args.limit_ifar
```
tdent, line 79: could you replace these two lines with a single call to eg numpy.minimum (element-wise minimum)?

bin/plotting/pycbc_page_foundmissed
```diff
@@ -200,7 +209,10 @@ if args.gradient_far:
         # Set up tick labels - there will always be 5
         min_tick = numpy.ceil(min(numpy.log10(color)))
         max_tick = numpy.floor(max(numpy.log10(color)))
-        tick_step = numpy.floor((max_tick - min_tick) / 5)
+        if (min_tick - max_tick) < 5:
```
tdent, line 212: put a comment to explain what this does? Also does it invalidate the previous comment about 'there will always be 5' ?

tdent (changes requested):
See comments. Generally looks OK but a few changes are a bit mysterious, and/or could probably be done with fewer lines/operations.

bin/plotting/pycbc_page_foundmissed
GarethCabournDavies: This was fixed in #4153, so was changed by a rebase anyway

bin/plotting/pycbc_page_foundmissed
```diff
@@ -95,11 +115,18 @@ try:
...
 
-    # "Decisive" distance is the second smallest effective distance
+    # "Decisive" distance for coincidences is the second smallest
+    # effective distance
     dvals['decisive_distance'] = numpy.sort(eff_dists)[:,1]
     dvals['dec_chirp_distance'] = \
         pycbc.pnutils.chirp_distance(dvals['decisive_distance'],
                                      vals['mchirp'])
+    # If we include single-detector triggers in the analysis, the
+    # "decisive" distance would be the minimum distance
```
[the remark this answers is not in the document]
GarethCabournDavies: It does look confusing on its own; it was meant to be a part of the larger comment with the one above, so have move the two to be together. So yes the minimum are the proper substitute for decisive distances when included

I have looked for a good description to change "decisive" - we could go for `second_minimum`?

bin/plotting/pycbc_ifar_catalog
GarethCabournDavies: This is a leftover from before I did the switch to counting louder events below threshold instead of extrapolating the fit backwards.

Basically the minimum fore_ifar could have been _really_ low, and so anything below the minimum expected_ifar was essentially rubbish.

However stuff at really low IFAR is now telling us that something has gone wrong, so I will revert it

bin/plotting/pycbc_page_foundmissed
```diff
@@ -95,11 +113,18 @@ try:
...
 
-    # "Decisive" distance is the second smallest effective distance
+    # "Decisive" distance for coincidences is the second smallest
+    # effective distance
     dvals['decisive_distance'] = numpy.sort(eff_dists)[:,1]
     dvals['dec_chirp_distance'] = \
         pycbc.pnutils.chirp_distance(dvals['decisive_distance'],
                                      vals['mchirp'])
+    # When including single-detector triggers in the analysis, the
+    # "decisive" distance is be the minimum distance
```
tdent, line 124: So .. I think the use of 'decisive' only makes sense in the coinc context as it was thought of originally as the distance that controls whether a coinc is made or not. Here I would just write eg "When single-detector triggers are included in the analysis, minimum distance is the relevant quantity for injection efficiency"

bin/plotting/pycbc_ifar_catalog
```diff
@@ -239,6 +244,7 @@ if opts.open_box:
     pylab.xlim(30 * min(fore_ifar), 3 * max_ifar)
 else:
     pylab.ylim(0.7, max(expected_cumnum))
+    pylab.xlim(min(expected_ifar) / 30, 3 * max(upper[4]))
```
tdent, line 247: I still haven't quite worked out why this setting for xmin is what we want. I guess 'expected_ifar' is just a guess anyway in terms of starting at 1e-4 (about 1/3 of a day) so if we want to allow for even less significant events we might want a smaller number .. ?

bin/plotting/pycbc_page_foundmissed
```diff
@@ -9,16 +9,30 @@ import pycbc.pnutils
...
 from pycbc.detector import Detector
 
+labels={'mchirp': 'Chirp Mass',
+        'mtotal': 'Total Mass',
+        'mass_ratio': 'Mass Ratio',
+        'decisive_distance': 'Decisive Distance (Mpc)',
+        'dec_chirp_distance': 'Decisive Chirp Distance (Mpc)',
+        'min_eff_distance': 'Minimum Effective Distance (Mpc)',
+        'min_eff_chirp_distance': 'Minimum Effective Chirp Distance (Mpc)',
+        'chirp_distance': 'Injected Chirp Distance (Mpc)',
```
tdent, line 19: I would still remove the 'Injected' here as the other labels seem to do OK without it

bin/plotting/pycbc_ifar_catalog
GarethCabournDavies: Removed - I don't know why but I thought Id reverted to the original already

bin/plotting/pycbc_page_foundmissed
GarethCabournDavies: Fixed now

bin/plotting/pycbc_page_foundmissed
GarethCabournDavies: Done

GarethCabournDavies:
N.B. The CC failure is due to previous changes, and makes sense to not "fix"

tdent (approved):

PR #4071, opened 2022-06-28
Make combine statmap options more like actual usage

tdent:
Previous default behaviour (before #4053) was to remove against inclusive FAR at the combine statmap stage, so add this option in order to test HR in this stage going forwards

### DIFF
--- a/examples/search/analysis.ini
+++ b/examples/search/analysis.ini
@@ -197,9 +197,11 @@ veto-window = 0.100
 cluster-window = 10.0
 
 [combine_statmap]
+cluster-window = ${statmap|cluster-window}
+
+[combine_statmap-full_data]
 max-hierarchical-removal = ${workflow-results|max-hierarchical-removal}
 hierarchical-removal-against = inclusive
-cluster-window = ${statmap|cluster-window}
 
 [foreground_censor]
 
--- a/examples/search/analysis.ini
+++ b/examples/search/analysis.ini
@@ -197,6 +197,8 @@ veto-window = 0.100
 cluster-window = 10.0
 
 [combine_statmap]
+max-hierarchical-removal = ${workflow-results|max-hierarchical-removal}
+hierarchical-removal-against = inclusive
 cluster-window = ${statmap|cluster-window}
 
 [foreground_censor]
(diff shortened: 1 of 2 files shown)


GarethCabournDavies:
That looks sensible to me

tdent:
@ahnitz any idea how to diagnose the small workflow failures?

tdent:
> I had a quick look in the logs, and it is failing on the add_statmap injection jobs, which makes sense as the injection files dont have inclusive ifars.
> 
That makes sense .. but where do you find the logs?

GarethCabournDavies:
> > I had a quick look in the logs, and it is failing on the add_statmap injection jobs, which makes sense as the injection files dont have inclusive ifars.
> 
> That makes sense .. but where do you find the logs?

Click details next to the failed jobs, then under "summary" on the left hand side, it has the logs under artifacts
image

tdent:
> > > I had a quick look in the logs, and it is failing on the add_statmap injection jobs, which makes sense as the injection files dont have inclusive ifars.
> >  
> > That makes sense .. but where do you find the logs?
> 
> Click details next to the failed jobs, then under "summary" on the left hand side, it has the logs under artifacts 

Oh, I was looking in the 'build' section which has a bit at the end about 'storing log files', however that in the end turned out to only be the dagman log file for running the workflow, not the individual job logs.  As those are 160MB I'm impressed that you managed to look quickly.

tdent:
Well, still failing ..

tdent:
The error was in a foreground minifollowup job .. 
```
          File "/opt/hostedtoolcache/Python/3.8.13/x64/bin/pycbc_foreground_minifollowup", line 107, in <module>
            raise KeyError('Sort variable {0} not in {1}: sort choices in '
        KeyError: 'Sort variable ifar not in background_exc: sort choices in background_exc are H1/time, H1/trigger_id, L1/time, L1/trigger_id, V1/time, V1/trigger_id, decimation_factor, stat, template_id, timeslide_id'
```

tdent:
OK, I think I know what is happening here .. production plotting ini files say to sort on 'stat' not ifar, and ifar doesn't exist for time slide events.

PR #4073, opened 2022-06-29
Live singles fits and snr optimization updates

GarethCabournDavies:
- Update to allow the pycbc live singles fits file to update during an analysis - i.e. move the file opening to the "calculate_ifar" test, rather than opening at the start of the analysis and keeping the fit parameters static
 - During this I noticed that pycbc_live was getting rather bloated, so moved the LiveEventManager into its own module, live_event_manager.py under pycbc/events. Some global variables were being used in the class/methods rather than local, so this is updated
 - There was previously no way to optimize the SNR of single-detector events, so this has been added
    - These will be run, but not uploaded to gracedb if there is a nearby coincident event being uploaded (this might want to be changed?)

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -15,22 +15,19 @@ import os.path
 import itertools
 import platform
 import subprocess
-from shutil import which
 from mpi4py import MPI as mpi
 from pycbc.pool import BroadcastPool
-from pycbc import fft, version, waveform, scheme, makedir
+from pycbc import fft, version, waveform, scheme
 from pycbc.types import MultiDetOptionAction, MultiDetMultiColonOptionAction
-from pycbc.filter import LiveBatchMatchedFilter, compute_followup_snr_series
-from pycbc.filter import followup_event_significance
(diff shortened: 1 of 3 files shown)


titodalcanton:
I suggest doing a preliminary PR that *just* moves the `LiveEventManager` to a module, without functional changes.

GarethCabournDavies:
Yeah, that's my thinking (but will be doing it backwards, i.e. functional changes first, then move to a module)

GarethCabournDavies:
Plus it gets my merge number up :chart_with_upwards_trend: :wink:

titodalcanton:
Could you name it pycbc.events.live instead of pycbc.events.live_event_manager? We may want to use the module for more than just the event manager class.

GarethCabournDavies:
I thought of that, but wanted to avoid the name clash with pycbc/io/live.py - maybe live_events.py?

titodalcanton:
I do not think there would be any clash, for example we already have events.distributions.gaussian and events.noise.gaussian. pycbc.events.live_events seems repetitive to me.

PR #4074, opened 2022-06-29
Update live single fits to reload file each time

GarethCabournDavies:
Move fits file reading into the check_singles function, so that the fit parameters are updated rather than loaded once and static


titodalcanton (approved):
LGTM, we will probably need some targeted testing soon.

PR #4075, opened 2022-06-29
Live optimize singles

GarethCabournDavies:
- Move the optimize_snr setup to its own function
- allow singles to call it
   - It is only called when no coincs have been uploaded

N.B. There's a bit of an oddity in the test case where the SNR is optimised for the single found at the same time as the coinc, however if this had been uploaded, the optimisation would not have been done. But I think this is okay to leave in

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -215,6 +215,105 @@ class LiveEventManager(object):
                     triggers['foreground/ifar'], pvalue, self.pvalue_livetime)
             triggers['foreground/pvalue_{}'.format(ifo)] = pvalue
 
+
+    def setup_optimize_snr(self, results, ifos, bank, fname, data_readers, gid):
+        template_id = \
+            results['foreground/{}/template_id'.format(ifos[0])]
+        p = props(bank.table[template_id])
+        apr = p.pop('approximant')
+        min_buffer = bank.minimum_buffer + 0.5
+        buff_size = \
+            pycbc.waveform.get_waveform_filter_length_in_time(apr, **p)
+
+        tlen = bank.round_up((buff_size + min_buffer) \
+            * bank.sample_rate)
+        flen = int(tlen / 2 + 1)
+        delta_f = bank.sample_rate / float(tlen)
+        cmd = 'timeout {} '.format(args.snr_opt_timeout)
+        exepath = which('pycbc_optimize_snr')
+        cmd += exepath + ' '
+        data_fils_str = '--data-files '
+        psd_fils_str = '--psd-files '
+        for ifo in ifos:
+            curr_fname = \
+                fname.replace('.xml.gz',
+                              '_{}_data_overwhitened.hdf'.format(ifo))
+            curr_data = data_readers[ifo].overwhitened_data(delta_f)
+            curr_data.save(curr_fname)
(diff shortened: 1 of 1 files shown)


titodalcanton (approved):
LGTM

bin/pycbc_live
```diff
@@ -280,106 +379,14 @@ class LiveEventManager(object):
...
-                log_fname = fname.replace('.xml.gz', '_optimize_snr.log')
-
-                logging.info('Running %s with log to %s', cmd, log_fname)
-                with open(log_fname, "w") as logfile:
-                    subprocess.Popen(
-                        cmd, shell=True, stdout=logfile, stderr=logfile
-                    )
+            if self.run_snr_optimization \
+                    and self.ifar_upload_threshold < ifar:
+                self.setup_optimize_snr(coinc_results, coinc_ifos, bank,
```
tdent, line 384: should the change from `live_ifos` to `coinc_ifos` have any effect?

tdent:
At first glance there could be issues related to the differences between `ifos` / `live_ifos` / `coinc_ifos` which are treated slightly differently in the old code.  (Alas, there isn't much in the way of comments to explain why.)

bin/pycbc_live
GarethCabournDavies: I had tried to keep the original live/coinc ifos the same as before (as I wasn't sure on the differences), but it looks like I didn't succeed

tdent: TBH it's not clear if this could be causing an error, but I think worth checking anyway. Maybe Ian can give you some hints on what the different ifo lists are all doing

ahnitz:
@GarethCabournDavies @tdent See the logic here and on line 330/331. 
https://github.com/GarethCabournDavies/pycbc/blob/188ad38bf53f2dc080e24464c69162c8eb478352/bin/pycbc_live#L343

bin/pycbc_live
tdent: OK, if this is a problem is someone planning to open an issue or PR?

GarethCabournDavies: Yes, I am at a conference this week, and busy for the next couple, but hope to get to this in the coming weeks

bin/pycbc_live
```diff
@@ -215,6 +215,105 @@ class LiveEventManager(object):
...
             triggers['foreground/pvalue_{}'.format(ifo)] = pvalue
 
+
+    def setup_optimize_snr(self, results, ifos, bank, fname, data_readers, gid):
+        template_id = \
+            results['foreground/{}/template_id'.format(ifos[0])]
+        p = props(bank.table[template_id])
+        apr = p.pop('approximant')
+        min_buffer = bank.minimum_buffer + 0.5
+        buff_size = \
```
ahnitz, line 225: @titodalcanton @GarethCabournDavies So I think this should pass 

```
bank.approximant(template_id)
```

to the function and then it should work fine.

bin/pycbc_live
titodalcanton: I am looking into this.

titodalcanton: See #4086, closing this conversation.

PR #4077, opened 2022-06-30
adding request-disk argument for dag generation

soumenroy:
The request-disk argument is added in the `pycbc_make_banksim` script to avoid the `ERROR: Failed to commit job submission`.  Therefore, we need to supply the 'request-disk' in the [workflow] section of the [configuration file](http://pycbc.org/pycbc/latest/html/banksim.html?highlight=effectualness), e.g., `request-disk=1024`.

### DIFF
--- a/bin/pycbc_make_banksim
+++ b/bin/pycbc_make_banksim
@@ -12,7 +12,7 @@ from glue.pipeline import CondorDAGJob, CondorDAGNode, CondorDAG, CondorJob
 
 class BaseJob(CondorDAGJob, CondorJob):
     def __init__(self, log_dir, executable, cp, section, gpu=False,
-                 accounting_group=None):
+                 accounting_group=None, request_disk=None):
         CondorDAGJob.__init__(self, "vanilla", executable)
 
         if gpu:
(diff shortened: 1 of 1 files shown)


titodalcanton:
@soumenroy thanks! I was thinking of something even simpler actually, just giving it a reasonable hardcoded setting, without the need to get it from the config file:
```
self.add_condor_cmd('request_disk', '1024')
```
Also, could you make the same change to `pycbc_make_faithsim`? The code should be essentially identical to banksim.

soumenroy:
Hi @titodalcanton,

Your opinion is indeed more straightforward than what I have done so far. I am not sure whether 1024 is a robust choice. What do you think if I make this argument's default value be 1024 instead of hardcoded? The user can either supply a value or ignore it in this case. Let me know your opinion.

I will look into the `pycbc_make_faithsim` and let you know.

soumenroy:
Hi @spxiwh,

Thanks for the information. 

I just made it hard coded and also committed the required changes in `pycbc_make_faithsim`.

titodalcanton:
Can `add_condor_cmd()` take an integer as second argument, or does it have to be a string? Have you tested your change @soumenroy?

soumenroy:
Hi @titodalcanton,

I have checked my changes and was able to submit the jobs successfully. However, I had to add `.gz` at the end of injection and bank file names because of changes in `pycbc_splitbank` and `pycbc_split_inspinj`. Please check my last commit.

Issue #4081, opened 2022-07-11
SNR optimization failing in PyCBC Live


maxtrevor:
I am currently trying to run PyCBC Live on some frames on CIT that I have copied over from the MDC. 
When I launch PyCBC Live using the script `/home/max.trevor/CBC/preO4/pycbc_idq/mdc_test/run_saved_frames.sh`  it starts successfully and runs until the first time it finds a coincident event. Then it hits the following error:

```
2022-07-11 10:00:07,894 node1390 0 Coincs 0: H1-L1: 131626 in cbuffer
2022-07-11 10:00:07,895 node1390 0 Found H1-L1 coinc with ifar 0.0024896405191148154
2022-07-11 10:00:07,895 node1390 0 computing followup data for coinc
2022-07-11 10:00:07,896 node1390 0 Generating IMRPhenomD, 16.0s, 168, starting from 20.0 Hz
2022-07-11 10:00:07,995 node1390 0 Coincident candidate! Saving as /local/max.trevor/idq_mdc_test/triggers/coinc-1262302864.3544922-dlxdjv.xml.gz
2022-07-11 10:00:08,044 node1390 0 Source probabilities file saved as /local/max.trevor/idq_mdc_test/triggers/coinc-1262302864.3544922-dlxdjv_probs.json
Traceback (most recent call last):
  File "/home/max.trevor/.conda/envs/pycbc-live-mpich/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/home/max.trevor/.conda/envs/pycbc-live-mpich/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/home/max.trevor/.conda/envs/pycbc-live-mpich/lib/python3.9/site-packages/mpi4py/__main__.py", line 7, in <module>
    main()
  File "/home/max.trevor/.conda/envs/pycbc-live-mpich/lib/python3.9/site-packages/mpi4py/run.py", line 198, in main
    run_command_line(args)
  File "/home/max.trevor/.conda/envs/pycbc-live-mpich/lib/python3.9/site-packages/mpi4py/run.py", line 47, in run_command_line
    run_path(sys.argv[0], run_name='__main__')
  File "/home/max.trevor/.conda/envs/pycbc-live-mpich/lib/python3.9/runpy.py", line 268, in run_path
    return _run_module_code(code, init_globals, run_name,
[...]

PR #4083, opened 2022-07-15
make python3 compatible for approximant names

ahnitz:
@titodalcanton @GarethCabournDavies I believe this should be the fix needed for #4081


titodalcanton (approved):
I confirm this fixes the problem. I briefly tried to find an "upstream" solution in the `TemplateBank` constructor in order to avoid this change, as I find the previous single-line solution easier to understand, but I could not get it to work. So this goes in.

PR #4084, opened 2022-07-18
add reconstruction of phase if it has been marginalized

ahnitz:
This patch allows one to demarginalize the coa_phase if it has been analytically marginalized over. I also add an option to 
'at_time' to allow for simple linear interpolation instead of nearest point. This smooths over some issues in the sample rate of the single template model (there are some obvious future improvements here as well).

### DIFF
--- a/pycbc/inference/models/tools.py
+++ b/pycbc/inference/models/tools.py
@@ -259,7 +259,7 @@ class DistMarg():
             elif self.marginalize_vector:
                 s, h = s[xl2], h[xl2]
             elif self.distance_marginalization:
-                s, h = s[xl1], h[xl1]
+                s, h = s[xl], h[xl]
 
             phasev = numpy.linspace(0, numpy.pi*2.0, int(1e4))
             phasel = (numpy.exp(2.0j * phasev) * s).real + h
--- a/pycbc/inference/models/tools.py
+++ b/pycbc/inference/models/tools.py
@@ -197,7 +197,7 @@ class DistMarg():
             skip_vector = True
             interpolator = None
             if self.reconstruct_phase:
-                return_complex=True
+                return_complex = True
 
         return marginalize_likelihood(sh_total, hh_total,
                                       phase=self.marginalize_phase,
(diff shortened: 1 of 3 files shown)


ahnitz:
Comparison of posterior with / without marginalization of the coa_phase, polarization, and distance. 
com

pycbc/inference/models/tools.py
```diff
@@ -378,7 +405,9 @@ def marginalize_likelihood(sh, hh,
             else:
                 vweights = dist_weights
 
-        if phase:
+        if return_complex:
+            return sh, -0.5 * hh
+        elif phase:
```
cdcapano, line 410: A little off topic, could you add a comment here explaining why you need to add `sh` to the `log(i0e(sh))`? Julian was asking me the other day, and it took me awhile to recall (it's because we want log(I0(sh)), but the scipy `i0e` function returns `exp(-|sh|)I0(sh))`.

Actually, now that I think of it... there might be a bug here. I think that should be `+ abs(sh)`. According to the [scipy docs](https://docs.scipy.org/doc/scipy/reference/generated/scipy.special.i0e.html), `i0e` returns `exp(-abs(x))*i0(x)`. That means if `sh` is negative, then we'll end up returning `log(i0e(sh)) - 2sh`. I'm not sure why this hasn't bitten us before; maybe because sh is positive for things we care about? Still, that would affect evidence calculations.

cdcapano (changes requested):
See my comments below. Mostly documentation requests. Unrelated to this patch, but I think there's a bug in the phase marginalization; see my comment. Can you verify?

Also, the unit tests are failing, I think because you changed the default return value to interpolation rather than nearest. You'll need to update the unit test to account for it.

pycbc/inference/models/tools.py
ahnitz: @cdcapano If you calculate the marginalization, you have to take |sh| as the argument, so it is never negative. Look earlier in the code, this is already done at line  https://github.com/gwastro/pycbc/blob/c809928a2cec48afdbed7a841990462ea8926eae/pycbc/inference/models/tools.py#L382. I guess 412/413 is redundant though, which may have caused the confusion here.

cdcapano: Oh, I missed that. Ok, it's fine then. Man, the logic in this function is complicated.

ahnitz:
@cdcapano Can this now be merged? The final CC warning has to be ignored.

cdcapano (approved):
Looks good

PR #4086, opened 2022-07-19
PyCBC Live: fix SNR opt regression + docstrings + code cleanup

titodalcanton:
PR #4075 introduced a regression in PyCBC Live's SNR optimization: it only uses the interferometers that originally produced the candidate, instead of all the interferometers with usable data. This reverts to the original behavior.

This PR also adds a docstring for `setup_optimize_snr()`, inverts the logic in `check_coincs()` (allowing us to drop one indentation level for most of the function) and includes some minor reformatting and f-string-izing.

I have an old branch with more aggressive refactoring of the SNR optimization, including changes to the files involved in the process. After this is merged, I will try to get that rebased and turned into a PR.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -219,6 +219,29 @@ class LiveEventManager(object):
     def setup_optimize_snr(
         self, results, live_ifos, coinc_ifos, bank, fname, data_readers, gid
     ):
+        """Setup and start the network SNR optimization process for a
+        candidate event. See arXiv:2008.07494 for details.
+
+        Parameters
+        ----------
+        results : dict
+            Information about the candidate.
+        live_ifos : list
+            List of detectors with usable data.
+        coinc_ifos : list
+            List of detectors that provided a reliabile estimate of the merger
+            time for this candidate. Must not be a superset of `live_ifos`.
+        bank :
+            Template bank.
+        fname : str
+            File name where the candidate information has been saved.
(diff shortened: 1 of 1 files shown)


titodalcanton:
I am thinking that the LiveEventManager instance should carry the bank and data reader as attributes. This would allow us to drop a few arguments from a few methods of LiveEventManager (including `setup_optimize_snr()`). @ahnitz do you expect any problem with that?

bin/pycbc_live
```diff
@@ -448,8 +492,15 @@ class LiveEventManager(object):
             if self.ifar_upload_threshold < ifar \
                     and not any(nearby_coincs) \
                     and self.run_snr_optimization:
-                self.setup_optimize_snr(single, [ifo], bank, fname,
-                                        data_reader, gid)
+                self.setup_optimize_snr(
+                    single,
+                    [ifo],
+                    [ifo],
```
GarethCabournDavies, line 498: I think this will still do the optimisation only on the single-detector that produced the event - I think we get all of them if we use `active`?

titodalcanton: The regression that this PR intends to fix is strictly related to coincident candidates, but yup, I noticed this too. I guess I should have asked in #4075, but what is the intention here? Do we want to use all available data for the SNR opt, or just the data with a detectable signal?

PR #4091, opened 2022-07-21
Calc kde calc

PRAVEEN-mnl:


bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -11,6 +11,9 @@ parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument('--signal-file')
 parser.add_argument('--template-file', required=True)
 parser.add_argument('--min-mass', type=float, default=None, required=True)
+parser.add_argument('--fit-param', nargs='+', required=True,
+                    help='Parameters over which KDE is calculated')
```
tdent, line 15: As discussed, the input param names should be expected to be just commonly used strings like eg 'mchirp', 'eta', 'q', 'chi_eff', 'mtotal' etc, which can be evaluated by feeding the template bank and the parameter name to the bank conversions function or by feeding the signal masses and spins and the parameter name to the triggers.get_param function.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -65,8 +68,21 @@ chi_eff_n = convert.chi_eff(mass1_n, mass2_n, spin1z_n, spin2z_n)
 mchirp_n = convert.mchirp_from_mass1_mass2(mass1_n, mass2_n)
 eta_n = convert.eta_from_mass1_mass2(mass1_n, mass2_n)
 
+assert len(args.fit_param) == len(args.log_param)
+
 if args.make_template_kde:
-    sample_n = numpy.vstack((numpy.log(mchirp_n), eta_n, chi_eff_n)).T
+    param0 = eval(args.fit_param[0][:])
```
tdent, line 74: As we discussed, these lines can be replaced by a loop over the list of fit param names from the command line. 

Also I think the template parameters have to be evaluated for all cases, or I think it's a good idea to do so, so the evaluation can occur outside the 'if' statement

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -65,8 +68,21 @@ chi_eff_n = convert.chi_eff(mass1_n, mass2_n, spin1z_n, spin2z_n)
...
+    param1 = eval(args.fit_param[1][:])
+    param2 = eval(args.fit_param[2][:])
+    if args.log_param == ['False', 'False', 'False']:
+        sample_n = numpy.vstack((param0, param1, param2)).T
```
tdent, line 78: So here, even if we do not know the number of parameters, we can still make a list or tuple of arrays of parameter values (via looping over the fit param names given on the command line) and feed that to vstack.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -105,7 +121,18 @@ if args.make_signal_kde:
...
+    param2 = eval(args.fit_param[2][:])
+    if args.log_param == ['False', 'False', 'False']:
+        sample_s = numpy.vstack((param0, param1, param2)).T
+    elif args.log_param == ['True', 'False', 'False']:
+        sample_s = numpy.vstack((numpy.log(param0), param1, param2)).T
+    elif args.log_param == ['True', 'True', 'False']:
+        sample_s = numpy.vstack((numpy.log(param0), numpy.log(param1), param2)).T
+    else:
+        raise ValueError("invalid log_param argument, use 'True', or 'False'")
+
```
tdent, line 135: When outputting values to the kde file, I would make sure the output file has at least all the parameters present in the template bank file - usually this comprises mass1, mass2, spin1, spin2z, but you can read the list of such parameters from the bank file.  Also, I would include the parameters used for the KDE as datasets in the output file.  (I would save the parameter values themselves not their logs.)
Then we need to also record whether the KDE was made over the parameter values or over their log: I think a nice way to record the parameter choice etc is to make an attr to record the fit param option, the value can be eg a string consisting of the param names separated by commas, and then another attr recording whether the log was used, the value can be a string with the True or False choices specified by the user.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -4,16 +4,19 @@ import numpy, h5py, operator, argparse, logging
...
 from pycbc import libutils
+from pycbc.events import triggers
 akde = libutils.import_optional('awkde')
 kf = libutils.import_optional('sklearn.model_selection')
 
 parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument('--signal-file')
 parser.add_argument('--template-file', required=True)
-parser.add_argument('--min-mass', type=float, default=None, required=True)
+parser.add_argument('--min-mass', type=float, default=None)
```
tdent, line 14: Can you include a help message for this option please. Note that at present we are applying the minimum mass to the signal masses but not to the templates, this should be clear from the help

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
 if args.make_signal_kde and args.make_template_kde:
     raise Exception("Choose only one option out of --make-signal-kde and --make-template-kde")
```
tdent, line 62: According to what Tito usually comments, I think using parser.error() would be preferable to the raise Exception statement.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
-mass2_n = template_file['mass2'][:]
-spin1z_n = template_file['spin1z'][:]
-spin2z_n = template_file['spin2z'][:]
-chi_eff_n = convert.chi_eff(mass1_n, mass2_n, spin1z_n, spin2z_n)
-mchirp_n = convert.mchirp_from_mass1_mass2(mass1_n, mass2_n)
-eta_n = convert.eta_from_mass1_mass2(mass1_n, mass2_n)
+#calling template data and constructing template_kde_hdf_file over same template params
+temp_file = h5py.File(args.template_file, 'r')
+approximant = temp_file['approximant'][:]
+f_lower = temp_file['f_lower'][:]
```
tdent, line 67: I am not sure why you are using f_lower for getting the number of templates. Not all template files necessarily have a f_lower dataset and we do not use the f_lower values in the KDE, I would prefer to use mass1 here.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
+tid = numpy.arange(len(f_lower))
+mass1, mass2, spin1z, spin2z = triggers.get_mass_spin(temp_file, tid)
+chi_eff = convert.chi_eff(mass1, mass2, spin1z, spin2z)
```
tdent, line 70: I don't know why you are still calculating these values here (template chi_eff, mchirp, eta).  If they are needed to make the kde then they should be calculated as a parameter in the loop over args.fit_param later. If they are not used in making the kde they do not have to be calculated.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
+mchirp = convert.mchirp_from_mass1_mass2(mass1, mass2)
+eta = convert.eta_from_mass1_mass2(mass1, mass2)
 
 assert len(args.fit_param) == len(args.log_param)
```
tdent, line 74: This should be checked before doing any calculations.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
-    param2 = eval(args.fit_param[2][:])
-    if args.log_param == ['False', 'False', 'False']:
-        sample_n = numpy.vstack((param0, param1, param2)).T
-    elif args.log_param == ['True', 'False', 'False']:
-        sample_n = numpy.vstack((numpy.log(param0), param1, param2)).T
-    elif args.log_param == ['True', 'True', 'False']:
-        sample_n = numpy.vstack((numpy.log(param0), numpy.log(param1), param2)).T
+for param, slog in zip(args.fit_param, args.log_param):
+    data = triggers.get_param(param, args, mass1, mass2, spin1z, spin2z)
+    if slog in ['false', 'False', 'FALSE']:
```
tdent, line 79: I am not sure why you are allowing all these different values when the only two allowed choices for the option as you have written it in the parser are 'True' and 'False'. You only need to check if the value is 'True' or 'False' and everything else should be an error.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
+        logging.info('Using param: %s', param)
+        parvals.append(data)
+    elif slog in ['true', 'True', 'TRUE']:
+        logging.info('Using log param: %s', param)
+        parvals.append(numpy.log(data))
     else:
-        raise ValueError("invalid log_param argument, use 'True', or 'False'")    
+        raise ValueError("invalid log param argument, use 'true', or 'false'")
```
tdent, line 86: this should be True and False, exactly as the choices you allow in the argparse setup

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
+sample1 = numpy.vstack((parvals)).T
```
tdent, line 87: this variable name is not very informative, I suggest eg 'template_pars'

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
+grid_pts = sample1
```
tdent, line 88: is it really necessary to create two variables with exactly the same value?

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
-    hdf.create_dataset('mass1', data=mass1_n)
-    hdf.create_dataset('mass2', data=mass2_n)
-    hdf.create_dataset('spin1z', data=spin1z_n)
-    hdf.create_dataset('spin2z', data=spin2z_n)
-    hdf.create_dataset('mchirp', data=mchirp_n)
-    hdf.create_dataset('eta', data=eta_n)
-    hdf.create_dataset('chi_eff', data=chi_eff_n)
-    hdf.create_dataset('grid_pts', data=grid_pts_n)
+    hdf.create_dataset('kde_temp', data=kde_data)
+    hdf.attrs['kde_tempplate'] = 'temp_file'
```
tdent, line 98: I don't see what information this attribute is providing, can it be removed?

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
+    hdf.attrs['fit_param'] = 'args.fit_param'
```
tdent, line 99: It looks like this creates an attribute with the name 'fit_param' and the value is the string 'args.fit_param'.  I don't see why recording this string is useful.  I think I suggested to make an attribute which records the fit params that the user has specified on the command line.  So that would require the _value_ of args.fit_param to be converted to a string.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
+    hdf.attrs['log_param'] = 'args.log_param'
```
tdent, line 100: Same comment as for the attribute 'fit_param'. I don't think this actually records what the user specified for the log option.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -58,51 +61,57 @@ def optimizedparam(sampleval, nfold=2):
...
-mass1_n = template_file['mass1'][:]
-mass2_n = template_file['mass2'][:]
-spin1z_n = template_file['spin1z'][:]
-spin2z_n = template_file['spin2z'][:]
-chi_eff_n = convert.chi_eff(mass1_n, mass2_n, spin1z_n, spin2z_n)
-mchirp_n = convert.mchirp_from_mass1_mass2(mass1_n, mass2_n)
-eta_n = convert.eta_from_mass1_mass2(mass1_n, mass2_n)
+#calling template data and constructing template_kde_hdf_file over same template params
+temp_file = h5py.File(args.template_file, 'r')
+approximant = temp_file['approximant'][:]
```
tdent, line 66: Rather than assuming the template bank file always has 'approximant' and 'f_lower' datasets and no others except the masses and spins, it would be better to get a list of all the datasets in the bank file (whatever they are) and then loop over them getting their values from the file. Then the same datasets can be written into the output file.

bin/all_sky_search/pycbc_template_kde_calc
tdent: Also, if you set a default value of 0 for this option you can simplify the code used to implement the minimum mass.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -114,38 +123,38 @@ if args.make_signal_kde:
...
         idx = numpy.full(N_original, True)
-    mass2_s = mass2
-    mass1_s = data_signal['mass1'][idx]
-    assert min(mass1_s - mass2_s) > 0
-    eta_s = data_signal['eta'][idx]
-    chi_eff_s  = data_signal['chi_eff'][idx]
-    mchirp_s = data_signal['mchirp'][idx]
+    mass1 = data_signal['mass1'][idx]
+    assert min(mass1 - mass2) > 0
+    eta = data_signal['eta'][idx]
```
tdent, line 128: Similarly to the template parameters, you shouldn't have to evaluate eta, chi_eff, mchirp explicitly here, they should only be used if they are specified as fit_param options

bin/all_sky_search/pycbc_template_kde_calc
tdent: Also, since these datasets from the original bank file are always going to be present in the output file, both in the signal KDE and template KDE cases, it would be preferable to open the output file and write the datasets into it before doing any other computation.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -114,38 +123,38 @@ if args.make_signal_kde:
...
-        sample_s = numpy.vstack((param0, param1, param2)).T
-    elif args.log_param == ['True', 'False', 'False']:
-        sample_s = numpy.vstack((numpy.log(param0), param1, param2)).T
-    elif args.log_param == ['True', 'True', 'False']:
-        sample_s = numpy.vstack((numpy.log(param0), numpy.log(param1), param2)).T
-    else:
-        raise ValueError("invalid log_param argument, use 'True', or 'False'")
+    for param, slog in zip(args.fit_param, args.log_param):
+        data = data_signal[param]
+        if slog in ['false', 'False', 'FALSE']:
```
tdent, line 134: Same comments here as for the template param case, you are only allowing the strings 'True' and 'False' as options.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -114,38 +123,38 @@ if args.make_signal_kde:
...
+            logging.info('Using param: %s', param)
+            paramvals.append(data)
+        elif slog in ['true', 'True', 'TRUE']:
+            logging.info('Using log param: %s', param)
+            paramvals.append(numpy.log(data))
+        else:
+            raise ValueError("invalid log param argument, use 'true', or 'false'")
 
-    grid_pts_s = numpy.array((numpy.log(mchirp_n), eta_n, chi_eff_n)).T
+    sample2 = numpy.vstack((paramvals)).T
```
tdent, line 143: Suggest to call this eg 'signal_pars'

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -114,38 +123,38 @@ if args.make_signal_kde:
...
-    hdf.create_dataset('kde_sig', data=kde_data_s)
-    hdf.attrs['signal_kde'] = 'kde_sig'
-    hdf.create_dataset('mass1', data=mass1_s)
-    hdf.create_dataset('mass2', data=mass2_s)
-    hdf.create_dataset('mchirp', data=mchirp_s)
-    hdf.create_dataset('eta', data=eta_s)
-    hdf.create_dataset('chi_eff', data=chi_eff_s)
-    hdf.create_dataset('grid_pts', data=grid_pts_s)
+    hdf.create_dataset('kde_sig', data=kde_data)
+    hdf.attrs['kde_signal'] = 'data_signal'
```
tdent, line 151: as for the template case, I don't see the point of making this attribute.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -114,38 +123,38 @@ if args.make_signal_kde:
...
+    hdf.attrs['fit_param'] = 'args.fit_param'
+    hdf.attrs['log_param'] = 'args.log_param'
+    hdf.create_dataset('mass1', data=mass1)
```
tdent, line 154: We want the output file to have the same set of datasets in both the signal and template KDE cases, ie the datasets mass1, mass2, spin1z, spin2z, approximant, f_lower etc (whatever was in the original bank file) should be present in the output file. As suggested in another comment the way to do this without repeating lines of code is to open the output file earlier on and immediately write the original template file datasets into it.

bin/all_sky_search/pycbc_template_kde_calc
tdent: @PRAVEEN-mnl do you need more explanation for this?

bin/all_sky_search/pycbc_template_kde_calc
tdent: @PRAVEEN-mnl can you address this comment/request?

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -53,40 +60,48 @@ def optimizedparam(sampleval, nfold=2):
...
-    hdf.close()
+    with h5py.File(args.output_file, "w") as f_dest:
+        g1 = f_dest.create_group("new_data")
+        g1.create_dataset("template_kde", data=template_kde)
+        g1.create_dataset("grid", data=sample1)
+        for name in args.fit_param:
+            g1.create_dataset(name, data= triggers.get_param(name, args, mass1, mass2, spin1z, spin2z))
+        g1.attrs['fit_param'] = args.fit_param
+        g1.attrs['log_param'] = args.log_param
+        with h5py.File("sbbh2.hdf", "r") as f_src:
```
tdent, line 100: I am not seeing why you have put the file name in the code here. All the files to be used by the code should be given as input options. This code should be able to be run with _any_ bank file, not just one called 'sbbh2.hdf'.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -52,44 +59,50 @@ def optimizedparam(sampleval, nfold=2):
     maxFOM = FOM[(optbw, optalpha)]
     return optbw, optalpha
 
+assert len(args.fit_param) == len(args.log_param)
```
tdent, line 62: Can you move this line to just after `init_logging` to make it clearer that it is input checking

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -52,44 +59,50 @@ def optimizedparam(sampleval, nfold=2):
...
+
 if args.make_signal_kde and args.make_template_kde:
-    raise Exception("Choose only one option out of --make-signal-kde and --make-template-kde")
+    parser.error("Choose only one option out of --make-signal-kde and --make-template-kde")
```
tdent, line 65: Same for these lines, can you move them to just after argument parsing and `init_logging`

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -52,44 +59,50 @@ def optimizedparam(sampleval, nfold=2):
...
+
+#calling template data and constructing template_kde_hdf_file over template params
```
tdent, line 67: I don't think these bits of code are _constructing_ a hdf file - the comment can be just 'Obtaining template parameters'

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -52,44 +59,50 @@ def optimizedparam(sampleval, nfold=2):
...
+temp_file = h5py.File(args.template_file, 'r')
+mass1 = temp_file['mass1']
+tid = numpy.arange(len(mass1))  #tid : integer or array of int for Indices of the entries to be returned
```
tdent, line 70: suggest "Array of template ids" for the comment.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -4,13 +4,20 @@ import numpy, h5py, operator, argparse, logging
...
 from pycbc import libutils
+from pycbc.events import triggers
 akde = libutils.import_optional('awkde')
 kf = libutils.import_optional('sklearn.model_selection')
 
 parser = argparse.ArgumentParser(description=__doc__)
 parser.add_argument('--signal-file')
 parser.add_argument('--template-file', required=True)
-parser.add_argument('--min-mass', type=float, default=None, required=True)
+parser.add_argument('--min-mass', type=float, default=None,
```
tdent, line 14: This help message does not actually describe what the option does - i.e. remove all signal events with mass2 < min_mass.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -52,44 +59,50 @@ def optimizedparam(sampleval, nfold=2):
...
+    data = triggers.get_param(param, args, mass1, mass2, spin1z, spin2z)
+    if slog in ['False']:
+        logging.info('Using param: %s', param)
+        template_pars.append(data)
+    elif slog in ['True']:
+        logging.info('Using log param: %s', param)
+        template_pars.append(numpy.log(data))
+    else:
+        raise ValueError("invalid log param argument, use 'True', or 'False'")
+sample1 = numpy.vstack((template_pars)).T
```
tdent, line 84: can this be called something more informative than 'sample1' ? eg 'temp_samples' to make it clear that it concerns the template parameters?

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -52,73 +64,78 @@ def optimizedparam(sampleval, nfold=2):
...
-    hdf.create_dataset('mchirp', data=mchirp_s)
-    hdf.create_dataset('eta', data=eta_s)
-    hdf.create_dataset('chi_eff', data=chi_eff_s)
-    hdf.create_dataset('grid_pts', data=grid_pts_s)
-    hdf.close()
+    signal_kde = kde_awkde(signal_samples, temp_samples, alp=optalpha, gl_bandwidth=optbw)
+
+    with h5py.File(args.output_file, "w") as f_dest:
+        f_dest.create_dataset("signal_kde", data=signal_kde)
+        for name in args.fit_param:
```
tdent, line 136: Is it possible to not write these pieces of code for outputting the fit param values and original template file values to the output hdf file twice?  Since you are doing the same for the template kde and signal kde cases, can this be done with only one piece of code?

tdent (approved):

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -1,23 +1,38 @@
-#!/usr/bin/env python
+#3dimensional_kde
```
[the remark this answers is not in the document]
tdent: (I edited to put it back)

PR #4092, opened 2022-07-21
use latest macosx version for testing / builds

ahnitz:


spxiwh (approved):
macos-latest makes more sense than a locked, old, version.

PR #4093, opened 2022-07-21
time / sky marginalization  + rework / simplify marginalization scripts

ahnitz:


ahnitz:
This patch simplifies how vector marginalizations are done e.g. those that are really just brute-force "due a bunch of samples and integrate over them". The new interface is that the likelihood model (if intended to handle vector marginalizations of a given parameter) should simply expect a vector in the current_params, rather than some other method to get the vector values. This allows for more generic marginalizations, centralizes the code that needs to handle things like drawing / chossing parameters,  and simplifies writing models that support it.

ahnitz:
@cdcapano I think this is now ready for review. I've tried to make it backwards compatible (but with an added deprecation warning to nudge people for the next couple releases that the interface may change). 

I've updated the single template example, but am leaving a full docs rewrite (i.e. adding an explicit marginalizations dedicated page) until I add support in at least the relative model (future PR, I'm not adding that in here).

ahnitz:
Ok, workflow failure is because it reuses the single template example configuration and without the demarginalization step in the workflow, some of the parameters it was trying to plot are missing.I'll think later how to fix that generally (actually a future PR might be to allow pycbc_inference to also do the demarginalization itself if requested), but for now will have it use the old config file.

pycbc/inference/models/tools.py
```diff
@@ -193,82 +247,293 @@ def marginalize_loglr(self, sh_total, hh_total,
...
+        # Is it worth doing this if some detector has low SNR?
+        sref = None
+        iref = None
+        idx = []
+        dx = []
+        mcweight = None
+        for ifo in ifos:
+            snr = snrs[ifo]
+            start = max(tcmin - .031, snr.start_time)
+            end = min(tcmax + .031, snr.end_time)
```
cdcapano, line 417: Please define a variable (maybe `EARTH_CROSSING_TIME`) for 0.031 and use that here, so we know what this is.

cdcapano (approved):
Looks good. Just one request; see comment. Once you address that, feel free to merge.

PR #4094, opened 2022-07-21
PyCBC Live: fix bug in mechanism that prevents multiple uploads

titodalcanton:
Fixes a couple bugs introduced in #3984. I have tested that uploads work after this. We do need a way to try some kind of fake uploads in our test suite though.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -87,7 +87,7 @@ class LiveEventManager(object):
         self.run_snr_optimization = run_snr_optimization
 
         # Keep track of which events have been uploaded
-        self.last_few_coincs_uploaded = numpy.zeros(10, dtype=float)
+        self.last_few_coincs_uploaded = []
 
         if self.run_snr_optimization and self.rank == 0:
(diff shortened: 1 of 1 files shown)


GarethCabournDavies:
> We do need a way to try some kind of fake uploads in our test suite though.

Changes look good to me - any ideas on how to implement this in the test suite?

GarethCabournDavies (approved):

ahnitz:
@GarethCabournDavies Is the gdb server code public so one can just start up a test instance in the CI?

GarethCabournDavies:
> @GarethCabournDavies Is the gdb server code public so one can just start up a test instance in the CI?

I didn't think it was,but looks like it might be [here](https://git.ligo.org/computing/gracedb/server). I won't pretend to be anywhere near able to understand how to implement it though!

titodalcanton:
@ahnitz I have been wondering the same actually. As far as I know it is a Django setup, so in principle it should even be possible to spin up an instance using an sqlite3 backend. I can have a look in the coming weeks.

PR #4109, opened 2022-08-13
Rationalize upload options

tdent:
"Nothing should be uploaded to gracedb without the --enable-gracedb-upload option." (IW Harry, 13 Aug 2022)

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -262,7 +262,7 @@ class LiveEventManager(object):
             * bank.sample_rate)
         flen = int(tlen / 2 + 1)
         delta_f = bank.sample_rate / float(tlen)
-        cmd = 'timeout {} '.format(snr_opt_timeout)
+        cmd = 'timeout {} '.format(args.snr_opt_timeout)
         exepath = which('pycbc_optimize_snr')
         cmd += exepath + ' '
         data_fils_str = '--data-files '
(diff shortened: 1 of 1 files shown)


bin/pycbc_live
```diff
@@ -757,6 +760,9 @@ fft.verify_fft_options(args, parser)
 
 if args.output_background is not None and len(args.output_background) != 2:
     parser.error('--output-background takes two parameters: period and path')
+if not args.enable_gracedb_upload and args.enable_single_detector_upload:
```
GarethCabournDavies, line 763: This looks good and is a better way to do it than the one I proposed (which would have been enforcing this silently)

bin/pycbc_live
```diff
@@ -702,34 +702,37 @@ parser.add_argument('--enable-profiling', type=int, metavar='RANK',
...
-                         " with frame file production.")
+parser.add_argument('--enable-single-detector-upload', action='store_true', default=False,
+                    help='Upload single ifo events to GraceDB')
 parser.add_argument('--gracedb-server', metavar='URL',
                     help='URL of GraceDB server API for uploading events. '
                          'If not provided, the default URL is used.')
 parser.add_argument('--gracedb-search', type=str, default='AllSky',
                     help='String going into the "search" field of the GraceDB '
                          'events')
+parser.add_argument('--ifar-upload-threshold', type=float, required=True,
```
GarethCabournDavies, line 727: The changes in the command line options look to just be a re-ordering. It would be good to revert them to keep the git history looking sensible

GarethCabournDavies (approved):
Thanks for sorting this! Minor comment on changes outwith the idea of the PR, but looks good to me

bin/pycbc_live
GarethCabournDavies: Actually, the changes are good as they are rationalising the options as in the PR title - ignore previous comment

tdent: Yes, the intention is to group the FAR calculation vs upload-related options separately.

PR #4110, opened 2022-08-16
Single events to require both upload options

GarethCabournDavies:
@spxiwh found an issue where single-detector events were attempted to upload to gracedb even when `--enable-gracedb-upload` was not in place, this is as `--enable-single-detector-upload` was still enabled.

This patch means both are required, for extra safety


GarethCabournDavies:
Sorry, didnt see #4109 before starting this

PR #4112, opened 2022-08-17
Reorganize APIs related to GraceDB uploads

titodalcanton:
In a previous PR, @tdent proposed a better name for the `SingleCoincForGraceDB` class. I asked for that change to be made in a dedicated PR, so here it is.

This PR also moves some strictly LIGOLW-specific code (which is used by code that is more general than PyCBC Live) to `pycbc.io.ligolw`, fixes some docstrings, and does some minor code cleanup.

I am also thinking of renaming the `pycbc.io.live` module to `pycbc.io.gracedb`, as there is nothing really specific to PyCBC Live in it (in principle it could be used by `pycbc_upload_xml_to_gracedb` too). Feel free to object though.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -28,7 +28,7 @@ from pycbc.strain import StrainBuffer
 from pycbc.events.ranking import newsnr
 from pycbc.events.coinc import LiveCoincTimeslideBackgroundEstimator as Coincer
 from pycbc.events.single import LiveSingle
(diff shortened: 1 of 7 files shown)


tdent (approved):
Looks fine. Format 'f' strings seem to be the future...

titodalcanton:
I am letting this run for a while with uploads enabled to test it.

PR #4119, opened 2022-08-26
Correct the way PyCBC Live inits p_astro

titodalcanton:
Only initialize p_astro (and source classification, and a bunch of other attributes) for the root process, and do print p_astro details at initialization when `--verbose` is used.


tdent:
Seems fine, if it has been minimally tested and prints out the counts OK

tdent (approved):

titodalcanton:
Yes, this has been tested and prints the same numbers as it did before (and only does to on the root process).

PR #4121, opened 2022-09-06
Make no-mass-gap default option for online source classification 

veronica-villa:
Given the removal of MassGap as a source category in O4 we set the no-mass-gap option as the default behaviour of the code, resulting in a {p(BNS), p(NSBH), p(BBH)} output.

### DIFF
--- a/pycbc/mchirp_area.py
+++ b/pycbc/mchirp_area.py
@@ -31,8 +31,8 @@ def insert_args(parser):
                                    "to different CBC sources.")
     mchirp_group.add_argument('--src-class-mass-gap-max', type=float,
                               metavar=('MAX_GAP'),
-                              help="Upper limit of the mass gap, that corresponds "
-                                   "to the minimum mass of a black hole. "
+                              help="Upper limit of the mass gap, corresponding"
+                                   " to the minimum mass of a black hole. "
                                    "Used as limit of integration of the "
                                    "different CBC regions when considering "
                                    "the MassGap category.")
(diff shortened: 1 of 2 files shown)


tdent:
It looks like the test suite is failing with an error in the calc probabilities - 

> File "/home/runner/work/pycbc/pycbc/.tox/py-search/lib/python3.8/site-packages/pycbc/mchirp_area.py", line 280, in calc_probabilities
>     areas = calc_areas(trig_mc_det, mass_limits, mass_bdary, z, mass_gap)
> TypeError: calc_areas() missing 1 required positional argument: 'mass_gap_separate'

pycbc/mchirp_area.py
```diff
@@ -21,20 +21,21 @@ def insert_args(parser):
                                              "source probabilities of a "
                                              "candidate event using the snr, "
                                              "mchirp, and effective distance.")
-    mchirp_group.add_argument('--src-class-mass-range', type=float, nargs=2,
-                              metavar=('MIN_M2', 'MAX_M1'),
-                              default=[1.0, 45.0],
+    mchirp_group.add_argument('--src-class-mass-ranges', type=float, nargs=3,
```
tdent, line 24: Suggest 'src-class-mass-limits' instead

pycbc/mchirp_area.py
```diff
@@ -21,20 +21,21 @@ def insert_args(parser):
...
+                              metavar=('MIN_M2', 'MAX_M1', 'MAX_NS'),
+                              default=[1.0, 3.0, 45.0],
```
tdent, line 26: Should it be clear which order these must be supplied in? Here you have max ns last in the list of metavars but the default has it second. Maybe it would be better to not try and rely on the numbers being in any order and instead sort them automatically and then assign the middle value as max ns?

pycbc/mchirp_area.py
```diff
@@ -64,16 +65,30 @@ def insert_args(parser):
 
 
 def from_cli(args):
-    return {'mass_limits': {'max_m1': args.src_class_mass_range[1],
-                            'min_m2': args.src_class_mass_range[0]},
-            'mass_bdary': {'ns_max': args.src_class_mass_gap[0],
-                           'gap_max': args.src_class_mass_gap[1]},
+    if args.src_class_mass_gap_max:
```
tdent, line 68: I think you want to put in a check that max gap is not lower than max ns here, otherwise weird things might happen.

pycbc/mchirp_area.py
```diff
@@ -234,22 +256,28 @@ def calc_probabilities(mchirp, snr, eff_distance, src_args):
...
         else:
-            probabilities = {"BNS": 1.0, "NSBH": 0.0, "BBH": 0.0,
-                             "Mass Gap": 0.0}
+            probabilities = {"BNS": 1.0, "NSBH": 0.0, "BBH": 0.0}
 
     else:
-        areas = calc_areas(trig_mc_det, mass_limits, mass_bdary, z, mass_gap)
+        areas = calc_areas(trig_mc_det, mass_limits, mass_bdary, z, mass_gap, mass_gap_separate)
         total_area = sum(areas.values())
         probabilities = {key: areas[key]/total_area for key in areas}
```
tdent, line 282: Should probably put spaces around ' / ' as an arithmetic operator here.

pycbc/mchirp_area.py
```diff
@@ -64,16 +65,30 @@ def insert_args(parser):
...
 
 def from_cli(args):
-    return {'mass_limits': {'max_m1': args.src_class_mass_range[1],
-                            'min_m2': args.src_class_mass_range[0]},
-            'mass_bdary': {'ns_max': args.src_class_mass_gap[0],
-                           'gap_max': args.src_class_mass_gap[1]},
+    if args.src_class_mass_gap_max:
+        return {'mass_limits': {'max_m1': args.src_class_mass_ranges[1],
+                                'min_m2': args.src_class_mass_ranges[0]},
+                'mass_bdary': {'ns_max': args.src_class_mass_ranges[2],
```
tdent, line 71: Here you are assuming that ns max is the 3rd argument to the mass ranges option, but the default has it as 2nd argument ..

pycbc/mchirp_area.py
veronica-villa: I put the metavars in that order because of the help text, but I think your option of sort them before it's better

pycbc/mchirp_area.py
veronica-villa: should it be a warning, a full error or just a check that max_gap > max_ns and in case is it not use max_gap = max_ns?

tdent: Tito always suggests a parser error in these cases .. see eg https://github.com/gwastro/pycbc/blob/master/bin/pycbc_live#L779

tdent (approved):
Looks good now, you can make codeclimate happier by reformatting the 'too long' lines

pycbc/mchirp_area.py
veronica-villa: thanks!

pycbc/mchirp_area.py
```diff
@@ -67,6 +67,8 @@ def insert_args(parser):
 def from_cli(args):
     mass_limits_sorted = sorted(args.src_class_mass_limits)
     if args.src_class_mass_gap_max:
+        if args.src_class_mass_gap_max < mass_limits_sorted[1]:
+            parser.error('MAX_GAP value cannot be lower than MAX_NS limit')
```
tdent, line 71: Hmm, I don't know if it will work here as you need the parser object, which 'args' by itself does not have access to. You may have to put this check into wherever there is parser.parse_args() instead .. @titodalcanton any advice here?

veronica-villa: ah, you're right! it should be here then: https://github.com/gwastro/pycbc/blob/master/bin/pycbc_live#L777

titodalcanton: Or, perhaps cleaner, pass the parser object to `from_cli()` as well i.e. `from_cli(parser, args)`.

tdent (approved):
Looks OK, have you done minimal testing?

veronica-villa:
> Looks OK, have you done minimal testing?

no, I didn't have time :/

PR #4126, opened 2022-09-08
Generate skymaps without sampling / enable pycbc inference to run model with only reconstructed params

ahnitz:
Depends on #4093 

This adds a dummy sampler that enables pycbc inference to generate results without requiring the model to have any params that need to be sampled over. This is useful for most uses of the single_template model as all parameters can be effectively marginalized over. One can simply set the sampler to 'dummy' and give a 'num_samples' option. 

I've added an example config file + run script which demonstrates generating the GW170817 results for all parameters the single template model supports.

### DIFF
--- a/bin/inference/pycbc_inference
+++ b/bin/inference/pycbc_inference
@@ -146,7 +146,7 @@ with ctx:
 
     # Run the sampler
     sampler.run()
-    
+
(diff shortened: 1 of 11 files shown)


sum33it (approved):
Looks good to me.

PR #4129, opened 2022-09-13
add time marginalization [+ra+dec+pol+phase] to full waveform model

ahnitz:
Depends on #4126 and #4093.

This adds a new model 'marginalized_time' that uses ffts to generate the polarization SNR timeseries. It can use this to set up time + sky + phase + distance marginalization or any combination thereof.  Inclination cannot be marginalized and phase is only valid for dominant mode waveforms as normal.


ahnitz:
WIP is only for the included example as it should be properly separated and perhaps included in documentation. The library code can be reviewed.

ahnitz:
Added example page

https://www.atlas.aei.uni-hannover.de/work/ahnitz/doctest2/html/inference/examples/margtime.html

ahnitz:
@sum33it @cdcapano Can one of the two of you review this PR? I've posted the associated documentation example above.

docs/inference/examples/margtime.rst
```diff
@@ -0,0 +1,53 @@
...
+
+Now run:
+
+.. literalinclude:: ../../../examples/inference/margtime/run.sh
+   :language: bash
+
+:download:`Download <../../../examples/inference/margtime/run.sh>`
+
+This will run the ``dynesty`` sampler. When it is done, you will have a file called
+``demarg_150914.hdf`` which contains the results. It should take just a few minutes to run.
```
cdcapano, line 45: Add something here about how you are reconstructing the marginalized parameters after the inference is done. When I first looked at this, I missed that in the script, so was confused why the marginalized parameters were showing up in the corner plot.

pycbc/inference/models/tools.py
```diff
@@ -421,7 +421,7 @@ def draw_sky_times(self, snrs):
             end = min(tcmax + EARTH_RADIUS, snr.end_time)
             snr = snr.time_slice(start, end, mode='nearest')
 
-            w = snr.squared_norm().numpy()
+            w = snr.squared_norm().numpy() / 2.0
```
cdcapano, line 424: Wait, is that a bug? Should the divide by 2 have been there before? If not, does this break anything?

cdcapano (commented):
Few comments; see below. Also, I take it this means that when the time is reconstructed, the timing resolution is not sub-sample, correct? Might be worth pointing that out in the documentation for the class.

pycbc/inference/models/tools.py
ahnitz: It should have been a 2, but it didn't affect the testing so far or results. It previously would only affect the choice of monte-carlo samples to be slightly less efficient than otherwise. The weights were self-consistent otherwise, hence why no observed issue in the results themselves before.

docs/inference/examples/margtime.rst
ahnitz: @cdcapano Ok,  I've updated the text to point this out.

ahnitz:
@cdcapano The time reconstruction is sub-sample, though it uses a locally quadratic fit to the data to do so.

pycbc/inference/models/marginalized_gaussian_noise.py
```diff
@@ -194,6 +194,157 @@ def _loglr(self):
...
+            fp, fc = self.dets[det].antenna_pattern(
+                                    params['ra'],
+                                    params['dec'],
+                                    params['polarization'],
+                                    params['tc'])
+            dt = self.dets[det].time_delay_from_earth_center(params['ra'],
+                                                             params['dec'],
+                                                             params['tc'])
+            dtc = params['tc'] + dt
+            cplx_hd = fp * cplx_hpd[det].at_time(dtc,
```
ahnitz, line 333: @cdcapano This is where the sub-sample interpolation comes into play. The monte-carlo points for the time are not limited to the sample locations. If you recall, the times are choices by their sample, but then we dither the time to +- half a sample, so all times are covered.

cdcapano (approved):
Ok, thanks for the explanation and for adding the additional docs.

PR #4130, opened 2022-09-14
LISA SMBHB PE example and doc

WuShichao:
This is the PR for the LISA SMBHB parameter estimation example and corresponding documentation.


companion.txt
```diff
@@ -27,3 +27,11 @@ sympy>=1.9
 # Needed for KDE trigger statistics
 git+https://github.com/mennthor/awkde.git@master
 scikit-learn
+
+# Needed for `BBHx` package to work with PyCBC
+gcc_linux-64
+gxx_linux-64
+gsl
+lapack=3.6.1
```
ahnitz: @WuShichao These aren't python packages. Remember you need to put this in tox.ini. I also doubt you need the gcc / gxx parts, but perhaps you might. Add one at a time so we don't do anything unnecessarily. 

Second, the error you posted is because it can't even process the line, this isn't quite the syntax for setting a version requirement even it were a python package. See the requirements.txt file for how versions are done for future reference.

ahnitz:
@WuShichao Can you post a link to the generated docs?

tox.ini
```diff
@@ -46,6 +46,8 @@ commands = bash tools/pycbc_test_suite.sh
 [testenv:py-inference]
 deps =
     {[base]deps}
+    gsl
```
ahnitz, line 49: @WuShichao This would try to install with pip, see at the top where conda_deps are given, not that uses a different variable.

ahnitz: You probably should just add the items to tbe global conda_deps section anyway, otherwise bbhx will try to be installed for tests that don't have the prereqs. One alternative is to add the bbhx pip commands here rather than companion.txt.

examples/inference/lisa_smbhb/plot.py
```diff
@@ -0,0 +1,67 @@
+import subprocess
+import pickle
+import numpy as np
+
```
ahnitz, line 4: @WuShichao I think we don't need this file at all. Can't you replace this with a direct call to pycbc_inference_plot_posterior? Just put in the expected values directly.

docs/inference/examples/lisa_smbhb.rst
```diff
@@ -0,0 +1,67 @@
...
+----------------------------------------------
+
+This example shows how to use PyCBC for parameter estimation of supermassive black hole binaries (SMBHB) 
+in LISA mock data. The `data <https://zenodo.org/record/7078835>`_ are generated from 
+`LISA Data Challenge 2a: Sangria <https://lisa-ldc.lal.in2p3.fr/challenge2a>`_, 
+and `BBHx <https://github.com/mikekatz04/BBHx>`_ package is used to generate the ``IMRPhenomD`` template and calculate 
+the corresponding TDI response for LISA. Relative binning (heterodyned likelihood) 
+is used during sampling to speed up the computation of likelihood functions. Before doing parameter estimation, 
+you need to install BBHx and `the corresponding PyCBC waveform plugin <https://github.com/ConWea/BBHX-waveform-model>`_
+(in the future, this waveform plugin will be a part of BBHx).
```
ahnitz, line 12: @WuShichao Let's not promise this. That's the BBHx devs choice.

ahnitz (approved):
@WuShichao Please remove the one statement I commented on, otherwise, I think this is good enough to merge. Any other comments to improve the text could be addressed in a future PR.

PR #4138, opened 2022-09-22
iDQ strain and frame support in PyCBC Live

maxtrevor:
This PR contains code needed for the implementation of iDQ in PyCBC Live. It contains only the changes that are common to all of  my proposed implementations of iDQ in PyCBC. Since these changes are common to all of the implementations, I think it makes sense to get a head start on merging them in, especially so @spxiwh can look at any optimization needed. 
In particular this PR contains:
A new iDQBuffer class in frame.py which subclasses the DataBuffer class
Support for creating iDQ buffers as instance attributes of the StrainBuffer class (plural because two buffers are used, one for the actual iDQ data channel and one for the IDQ_OK channel )
A change to the timerseries class's at_time method to support looking up multiple times with one method call

This PR shows multiple commits but all but one relate to me rebasing my personal fork of PyCBC with updates from gwastro/master . Please squash the commits when the merge is approved.


maxtrevor:
Requesting review from @titodalcanton and @spxiwh

pycbc/strain/strain.py
```diff
@@ -1547,6 +1555,17 @@ def __init__(self, frame_src, channel_name, start_time,
             self.dq = pycbc.frame.StatusBuffer(frame_src, data_quality_channel,
                                                start_time, **sb_kwargs)
 
+        if idq_channel is not None:
+            assert idq_state_channel is not None, 'Each ifo with an idq channel requires an idq state channel'
```
titodalcanton, line 1559:
```suggestion
            if idq_state_channel is None:
                raise ValueError('Each detector with an iDQ channel requires an iDQ state channel as well')
```

pycbc/types/timeseries.py
```diff
@@ -273,8 +273,13 @@ def at_time(self, time, nearest_sample=False, interpolate=None):
...
             time += self.delta_t / 2.0
-
-        return self[int((time-self.start_time)*self.sample_rate)]
+        deltat = time-self.start_time
+        if hasattr(deltat,'__len__'):
+            if len(deltat) > 0:
+                return self[(deltat*self.sample_rate).astype('int')]
+            else:
+                return _numpy.array([])
+        return self[int(deltat*self.sample_rate)]
```
titodalcanton, line 282: Some more formatting recommendations:
```suggestion
        deltat = time - self.start_time
        if hasattr(deltat, '__len__'):
            if len(deltat) > 0:
                return self[(deltat * self.sample_rate).astype('int')]
            return _numpy.array([])
        return self[int(deltat * self.sample_rate)]
```

maxtrevor:
Is this good to go?

titodalcanton:
@maxtrevor I see a genuine failure in the test suite. Please have a look.

titodalcanton:
Ah, it looks like you just need to add the relevant iDQ command-line arguments to `pycbc_live`.

maxtrevor:
@titodalcanton the only failure I can see in the test suite is this
```
2022-09-22 17:00:12,409 PASS: GW170817 found with coherent SNR = 27.16; reweighted SNR 26.64
    Pass.
/home/runner/work/pycbc/pycbc
ERROR: InvocationError for command /usr/bin/bash tools/pycbc_test_suite.sh (exited with code 1)
___________________________________ summary ____________________________________
ERROR:   py-search: commands failed
Error: Process completed with exit code 1.
```
I have no clue what this error is or how to fix it. I don't see anything to do with iDQ

titodalcanton:
`args` will not have an `idq_channel` attribute unless you add the appropriate `parser.add_argument()` call to `pycbc_live`.

maxtrevor:
Ah now I understand. I have those lines in my full branches for running the test analysis, but didn't include the add argument lines in this PR

maxtrevor:
The only check failing now is codeclimate. I am not sure how to address the 3 issues it has raised.

pycbc/types/timeseries.py
```diff
@@ -273,8 +273,12 @@ def at_time(self, time, nearest_sample=False, interpolate=None):
 
         if nearest_sample:
             time += self.delta_t / 2.0
-
-        return self[int((time-self.start_time)*self.sample_rate)]
+        deltat = time - self.start_time
```
spxiwh, line 276: I do agree with CodeClimate about the lines here. It's not clear why we need 3 return statements on successive lines, but it does become quite confusing what is going on here.

If this complexity *is* needed, can it be well commented (ie. why check for `__len__` on delta_t, that feels hacky! ... If you are asking, is deltat a numpy array there's clearer ways to do that) I'm guessing that you are wanting this to run on an array of times, rather than just a single value? If that's true the documentation does need updating, and it should be checked if the other code paths work in that way.

maxtrevor: So yes the reason to check for __len__ is to check if the list is a numpy array or something similar. The whole point of this change is to allow for checking multiple times with one function call. 
I am open to finding a way to reduce the complexity of this but not sure how to do so. Basically the current logic is:
Is the input time an array or a single value?
If Its an array, I had to add a check for an array of length 0, because there was an if it get an array of length 0 otherwise. 
If it is not an array, keep the original functionality.
I have been running PyCBC Live with this change in effect and it does not seem to have broken anything else.

spxiwh: Does the time_slice method (https://github.com/gwastro/pycbc/blob/dc5ff851280c65879ecda96a4faee99d3431f1a0/pycbc/types/timeseries.py#L178), not do what you want? ... It's better to use (or adapt) existing methods, rather than trying to force behaviour into another, especially in the core PyCBC Classes.

spxiwh: Perhaps not, if your list of times is a set of unconnected times.

maxtrevor: Also I would like to mention that checking if `__len__` exists is exctly what is done in the definitions of all the single detector stats in ranking.py

maxtrevor: Indeed, time_slice does not do what I need. The araay that gets passed in is an error of single detector trigger times which will not generally be a continuous stretch

pycbc/frame/frame.py
```diff
@@ -792,3 +792,112 @@ def advance(self, blocksize):
...
+        Parameters
+        ----------
+        times: float
+            The times whose idq values are needed
+
+        Returns
+        -------
+        values: float
+            The values of the idq buffer at the given times
+        """
```
spxiwh, line 844: So it looks like the call to `at_time` is here. But this method never seems to be called? 

.... Also, I think you've said that both `times` and `values` are not floats, but numpy arrays (of floats?). Please update docstrings here (and have a quick check that others are accurate too.

pycbc/types/timeseries.py
spxiwh: So I'm not opposed to having this work on a set of times, but we may need to consider whether this should still be called `at_time`; `at_times` would make more sense. It also then needs to work in *all* case if multiple values are given, and the code here will fail if given an array and `interpolate == 'linear'`. The doc strings must also accurately reflect what the code can do.

spxiwh: (Or something along the lines of, we change this to `at_times` and make it work only on arrays. Then write an additional `at_time` method, which takes one value, converts to a length one array, and just calls on to `at_times` ... This should all also work with length 0 arrays, so I'm not sure yet why that needs to be a special case).

pycbc/frame/frame.py
maxtrevor: So the actual call to at_time is not in this PR. My logic was:
I am testing 3 seperate ways of implementing iDQ. They each have a seperate branch. Only of these branches will eventually be merged in. There is code that is common to all three branches. I decided to have two seperate PRs, one for the code common to all branches (this PR) and another for the statistic I decide on and code specific to supporting that statistic.

How at_time will be used is in the main pycbc_live loop, after collecting all the triggers for an ifo and removing triggers that get flagged, each trigger needs to know the idq vale at its time for that can be passed along to the single detector stat. I am doing this by pulling the array of times of the surviving triggers, passing it the iDQBuffer and asking it to give back an array with the values of idq at the time of each trigger. 

I did not know that I need to make a distinction in the docstring between a float and an array of floats. I can correct that, but I think these methods will actually work whether they are passed an array or a single value. Should the docstring say 'float or array of floats' ?

pycbc/types/timeseries.py
maxtrevor: Fundamentally I can not think of a way to write the at_time function that meets all of the following criteria:
Preserves original functionality when passed a single number (i.e. returns a single number and not a length 1 array)
Has the functionality I need of being able to check an array of times
Doesn't error when given an array of length 0 (this happens if no single detector triggers occur)
Has less than three return statements

I can make it so the three return statements aren't on three successive lines but I think the problem is larger than that.

maxtrevor: I like the idea of having seperate at_time and at_times methods. I can code that. 
The reason for the length 0 special case is because of the need to make indices into ints. The indices of the times to check are calculated using floats and then cast to to an int. The numpy ` astype(int) ` method gives an error if called by an array of length 0.

maxtrevor: I was getting an error if a length 0 array got passed in, which occurs if no single detector triggers happen in an analysis block. But I think that was actually happening with a previous version of this method, the check for 0 may not be necessary anymore.

Is it acceptable if I just leave the at_time method exactly as it was, and then add an at_times method that takes an array and has my desired functionality?

maxtrevor: I made a new commit doing what I suggested above. I returned at_time to its previous version, and made a seperate at_times method.

pycbc/types/timeseries.py
```diff
@@ -273,9 +273,23 @@ def at_time(self, time, nearest_sample=False, interpolate=None):
...
         if nearest_sample:
             time += self.delta_t / 2.0
-
         return self[int((time-self.start_time)*self.sample_rate)]
 
+    def at_times(self, times, nearest_sample = False):
+        """ Return an array of values at the specified gps times
+
+        Parameters
+        ----------
```
spxiwh, line 282: Don't forget to add `times` here as well.

pycbc/types/timeseries.py
spxiwh: Okay this looks good. I think the two could (should) be consolidated later, but I think it's now clear what to expect.

maxtrevor:
Tests are failing, it seems like they couldn't get files they needed for some reason? Its ntot something connected to my code. I previously passed all the non codeclimate checks until I updated a docstring.

titodalcanton:
There is a button "re-run jobs" on the top right of the Actions page. I just used it.

maxtrevor:
> There is a button "re-run jobs" on the top right of the Actions page (you may not be able to see it if you are not a maintainer or whatever). I just used it.

Thanks!

bin/pycbc_live
```diff
@@ -606,6 +606,12 @@ parser.add_argument('--state-channel', action=MultiDetMultiColonOptionAction,
                          "corresponds to CAT1 information")
 parser.add_argument('--analyze-flags', action=MultiDetOptionAction, nargs='+',
                     help='The flags that must be in the "good" state to analyze data')
+parser.add_argument('--idq-channel', action=MultiDetMultiColonOptionAction,
```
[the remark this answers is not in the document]
titodalcanton: What I mean is, do you need `MultiDetMultiColonOptionAction` as opposed to the simpler `MultiDetOptionAction`?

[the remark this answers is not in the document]
titodalcanton: (the docstrings of those classes should describe the difference clearly, if you want to know more).

titodalcanton:
We can ignore those last Codeclimate errors. See the last open comment and then I think we are done.

bin/pycbc_live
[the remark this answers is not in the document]
maxtrevor: @spxiwh I am willing to change this to ` MultiDetOptionActSpecial ` . But that is not what is used for ` --channel-name ` currently. Looking at the docstrings it does seem like ` MultiDetOptionActSpecial ` is the correct one to use and the option for ` --channel-name ` should probably be changed.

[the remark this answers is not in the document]
titodalcanton: Nevermind @maxtrevor, I found https://github.com/gwastro/pycbc/pull/3695 and I think what you did initially is sensible. Sorry for the amnesia.

maxtrevor:
I think this is ready to merge?

titodalcanton (approved):
Thanks @maxtrevor.

PR #4148, opened 2022-10-04
set consistent phase convention for inference models / maybe bugfix? 

ahnitz:
I might consider this PR a bugfix. In running some simulations on the marginalized models and testing the accuracy of parameter reconstruction I noticed an inconsistency in how we were handling the phase. 

 We were inconsistent in our inner products about the use of 'coa_phase' in various models. Some of these were less obvious than others as they occur in marginalized models where this would be unobservable except when trying to use the partial terms to do phase reconstruction (i.e the complex inner products). Sometimes we would conjugate the data and other times the signal when doing the complex inner products. Either is fine, but if we want to reconstruct the phase, this introduces a sign inconsistency that I was not accounting for. I think the right solution is to use a consistent convention. I've decided to move everything (where it wasn't already) to use the convention of conjugating the template. This was the convention already established in most of the literature and also what we do in our matched filtering code. 

Some models (only relbin atm) may choose to internally use another convention for convenience, but should return the complex inner products with the right convention (you can invert one to another by conjugation). 

@cdcapano Note, the conjugation order for the 'inner' product method is that 'self' is conjugated so the instance the method is called on, so this involved switching the order in a couple places.

### DIFF
--- a/pycbc/inference/models/gaussian_noise.py
+++ b/pycbc/inference/models/gaussian_noise.py
@@ -950,7 +950,7 @@ class GaussianNoise(BaseGaussianNoise):
                 h[self._kmin[det]:kmax] *= self._weight[det][slc]
 
                 # the inner products
-                cplx_hd = self._whitened_data[det][slc].inner(h[slc])  # <h, d>
+                cplx_hd = h[slc].inner(self._whitened_data[det][slc])  # <h, d>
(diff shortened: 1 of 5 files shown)


cdcapano (approved):
Thanks for finally fixing this!

PR #4150, opened 2022-10-04
Reduce wget verbosity in tests

titodalcanton:
Going through the log generated by the test suite is annoying because thousands and thousands of lines are just wget showing the progress of files being downloaded. This adds `-nv` to those wget calls, so they do not show the progress in the logs.


spxiwh (approved):
Makes sense, and looks good to me. (I've also struggled with really long log files!)

PR #4151, opened 2022-10-05
small update for LISA PE doc

WuShichao:
@ahnitz In this PR, I add a second link for BBHx to make it more visible.


lauranuttall:
Thanks, I completely missed this when I ran your example so this will really help

lauranuttall (approved):
Thank you, adding this link is really useful

PR #4153, opened 2022-10-06
Change foundmissed tick step

dethodav:
This MR addresses the case where the range of recovered injection FARs covers less than 5 orders of magnitude. This scenario can arise when a small number of injections are recovered (likely due to analyzing a small amount of time).  Rather than return an error, tick marks are now placed on every order of magnitude.

### DIFF
--- a/bin/plotting/pycbc_page_foundmissed
+++ b/bin/plotting/pycbc_page_foundmissed
@@ -200,7 +200,7 @@ if args.gradient_far:
         # Set up tick labels - there will be 5 if possible
         min_tick = numpy.ceil(min(numpy.log10(color)))
         max_tick = numpy.floor(max(numpy.log10(color)))
-        tick_step = max(1,numpy.floor((max_tick - min_tick) / 5))
+        tick_step = max(1, numpy.floor((max_tick - min_tick) / 5))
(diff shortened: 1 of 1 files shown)


bin/plotting/pycbc_page_foundmissed
```diff
@@ -197,10 +197,10 @@ if args.gradient_far:
         c = plot.colorbar()
         c.set_label('False Alarm Rate $(yr^{-1})$, %s' % far_title)
 
-        # Set up tick labels - there will always be 5
+        # Set up tick labels - there will be 5 if possible
         min_tick = numpy.ceil(min(numpy.log10(color)))
         max_tick = numpy.floor(max(numpy.log10(color)))
-        tick_step = numpy.floor((max_tick - min_tick) / 5)
+        tick_step = max(1,numpy.floor((max_tick - min_tick) / 5))
```
titodalcanton, line 203:
```suggestion
        tick_step = max(1, numpy.floor((max_tick - min_tick) / 5))
```

titodalcanton (approved):
Looks good with just a trivial whitespace correction.

GarethCabournDavies:
Any reason not to merge this?

PR #4154, opened 2022-10-08
Template bank project : changes in faithsim codes 

MPillas:
Changes in `pycbc_make_faithsim` & `pycbc_faithsim` to use them with python3

### DIFF
--- a/bin/pycbc_faithsim
+++ b/bin/pycbc_faithsim
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#!/home/marion.pillas/template_bank_venv/bin/python3
 ## Copyright (C) 2012  Alex Nitz
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by the
@@ -26,6 +26,7 @@
 import logging
 from numpy import complex64
 import argparse
+import sys
 from ligo.lw import utils as ligolw_utils
 from ligo.lw import lsctables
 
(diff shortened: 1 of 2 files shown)


ahnitz (approved):

ahnitz:
@MPillas Thank you for your contribution!

MPillas:
In the `pycbc_make_faithsim` code, is there a way to know or control the number of xml bank files the` inj.xml` file will be split into ? 

I'm referring to this part of the code : 

```
logging.info("Creating injection file")
inj_str = "lalapps_inspinj " + get_ini_opts(confs, "inspinj") + "--output inj.xml"
os.system(inj_str)

logging.info("Splitting template bank")
subprocess.call(['pycbc_splitbank',
                 '--templates-per-bank', str(templates_per_job),
                 '-t', 'inj.xml',
                 '-o', 'bank/bank'])

num_banks = len(glob.glob("bank/bank*"))
```

MPillas:
Problem in `pycbc_faithsim_collect_results` (`part of pycbc_make_faithsim`)

`Traceback (most recent call last):
  File "scripts/pycbc_faithsim_collect_results", line 50, in <module>
    pdata[field] = bt.get_column(field)
AttributeError: 'SimInspiralTable' object has no attribute 'get_column'`

But in the documentation I can clearly see the attribute `get_column` in the `glue.ligolw.lsctables.SimInspiralTable` class. The problem comes from the `mass1` column otherwise it's working.

spxiwh:
Hi Marion. We don't use glue.ligolw any more, we now use ligo.lw. The
syntax there is a little different (not sure offhand what the replacement
is, but I think it would only be a small change).

PR #4157, opened 2022-10-12
cleanup of pycbc live single trigger fits codes

GarethCabournDavies:
given updated stat and cuts modules, updates the live singles fitting code to actually use these modules rather than do its own thing the same

Some minor (one-line) changes to other codes, as the cuts module is called multiple times in this code, but once in others meaning that it spams the logging here without the change

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_findtrigs
+++ b/bin/all_sky_search/pycbc_coinc_findtrigs
@@ -196,7 +196,7 @@ template_ids = cuts.apply_template_cuts(
     template_ids=template_ids)
 
 logging.info("%d out of %d templates kept after applying template cuts",
-             len(template_ids), trigs.singles[0].bank['mass1'].size)
+             len(template_ids), len(trigs.singles[0].bank))
 
 # 'data' will store output of coinc finding
 # in addition to these lists of coinc info, will also store trigger times and
--- a/bin/all_sky_search/pycbc_coinc_findtrigs
+++ b/bin/all_sky_search/pycbc_coinc_findtrigs
@@ -196,7 +196,7 @@ template_ids = cuts.apply_template_cuts(
     template_ids=template_ids)
 
 logging.info("%d out of %d templates kept after applying template cuts",
-             len(template_ids), trigs.singles.bank['mass1'].size)
+             len(template_ids), trigs.singles[0].bank['mass1'].size)
 
 # 'data' will store output of coinc finding
 # in addition to these lists of coinc info, will also store trigger times and
--- a/bin/all_sky_search/pycbc_coinc_findtrigs
+++ b/bin/all_sky_search/pycbc_coinc_findtrigs
@@ -195,6 +195,8 @@ template_ids = cuts.apply_template_cuts(
     ifos=trigs.ifos,
     template_ids=template_ids)
 
+logging.info("%d out of %d templates kept after applying template cuts",
+             len(template_ids), trigs.singles.bank['mass1'].size)
 
 # 'data' will store output of coinc finding
 # in addition to these lists of coinc info, will also store trigger times and
--- a/bin/all_sky_search/pycbc_sngls_findtrigs
+++ b/bin/all_sky_search/pycbc_sngls_findtrigs
@@ -113,6 +113,9 @@ template_ids = cuts.apply_template_cuts(
     ifos=[ifo],
     template_ids=template_ids)
 
+logging.info("%d out of %d templates kept after applying template cuts",
+             len(template_ids), trigs.bank['mass1'].size)
+
 for tnum in template_ids:
     tids_uncut = trigs.set_template(tnum)
 
--- a/bin/all_sky_search/pycbc_sngls_findtrigs
+++ b/bin/all_sky_search/pycbc_sngls_findtrigs
@@ -114,7 +114,7 @@ template_ids = cuts.apply_template_cuts(
     template_ids=template_ids)
 
 logging.info("%d out of %d templates kept after applying template cuts",
-             len(template_ids), trigs.bank['mass1'].size)
+             len(template_ids), len(trigs.bank))
 
 for tnum in template_ids:
     tids_uncut = trigs.set_template(tnum)
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -26,7 +26,7 @@ parser.add_argument("--conservative-percentile", type=int, default=95,
                     help="What percentile to use for the conservative "
                          "combined fit. Integer in range 50-99. Default=95")
 parser.add_argument("--output", required=True,
-                    help="Output directory for final results")
+                    help="Output file for combined fit parameters")
 parser.add_argument("--ifos", required=True, nargs="+",
                     help="list of ifos fo collect info for")
 
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -244,10 +244,15 @@ for ifo in args.ifos:
     fout_ifo.attrs['total_counts'] = counts_out[ifo].sum()
 
     if args.output_plot_file_format:
-        alpha_lines += ax_alpha.plot(ad, alpha_all,
+        # If no bins had any triggers on that day, ignore
+        valid_days = np.logical_and(alpha_all > 0, np.isfinite(alpha_all))
+        dates_out = ad[valid_days]
+        count_all = count_all[valid_days]
+        alpha_all = alpha_all[valid_days]
+        alpha_lines += ax_alpha.plot(dates_out, alpha_all,
                                      c='k', linestyle='-', linewidth=2,
                                      label="daily overall alpha")
-        count_lines += ax_count.plot(ad, count_all,
+        count_lines += ax_count.plot(dates_out, count_all,
                                      c='k', linestyle='-', linewidth=2,
                                      label="overall count")
         alpha_lines += [ax_alpha.axhline(overall_meanalpha,
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -13,15 +13,11 @@
 # Public License for more details.
 
 import h5py, numpy as np, argparse
-import matplotlib
-matplotlib.use('agg')
-from matplotlib import pyplot as plt
-import re, datetime, logging
+import logging
 import pycbc
-import os
 
 parser = argparse.ArgumentParser(usage="",
-    description="Plot histograms of triggers split over various parameters")
+    description="Combine fitting parameters from several different files")
 parser.add_argument("--verbose", action="store_true",
                     help="Print extra debugging information", default=False)
 parser.add_argument("--trfits-files", nargs="+", required=True,
@@ -29,14 +25,8 @@ parser.add_argument("--trfits-files", nargs="+", required=True,
 parser.add_argument("--conservative-percentile", type=int, default=95,
                     help="What percentile to use for the conservative "
                          "combined fit. Integer in range 50-99. Default=95")
-parser.add_argument("--output-directory", required=True,
+parser.add_argument("--output", required=True,
                     help="Output directory for final results")
-parser.add_argument("--output-id-string",
-                    help="Optional ID string to include in the output "
-                         "filename")
-parser.add_argument("--output-plot", action='store_true',
-                    help="Flag to plot fit coefficient vs time and counts "
-                         "vs time plots")
 parser.add_argument("--ifos", required=True, nargs="+",
                     help="list of ifos fo collect info for")
 
@@ -51,11 +41,6 @@ if args.conservative_percentile < 50 or \
     parser.error("--conservative-percentile must be between 50 and 99, "
                  "otherwise it is either not a percentile, or not "
                  "conservative.")
-
-colours = ['r', (1.0, 0.6, 0), 'y', 'g', 'c', 'b', 'm',
-           (0.25,0.9,0)]
-# If the above colours are exhausted, revert to the default colour cycle
-colours += plt.rcParams['axes.prop_cycle'].by_key()['color']
 
 counts_all = {ifo:[] for ifo in args.ifos}
 alphas_all = {ifo:[] for ifo in args.ifos}
@@ -75,12 +60,13 @@ with h5py.File(args.trfits_files[0], 'r') as fit_f0:
 
 live_times = {ifo : [] for ifo in args.ifos}
 
-logging.info("Checking through %d files", len(args.trfits_files))
+trigger_file_times = []
 
-epoch = datetime.date(2000,1,1)
+n_files = len(args.trfits_files)
+logging.info("Checking through %d files", n_files)
+
 for f in args.trfits_files:
 
-    # Get the date information, there are a couple of ways to do this:
     fits_f = h5py.File(f, 'r')
     # Check that the file uses the same setup as file 0, to make sure
     # all coefficients are comparable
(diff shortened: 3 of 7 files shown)


bin/live/pycbc_live_single_trigger_fits
```diff
@@ -198,110 +177,193 @@ for filename in files:
...
         continue
+
+    # Triggers for this file only:
+    triggers = {}
     with h5py.File(f, 'r') as fin:
         # Open the file: does it have the ifo group and snr dataset?
         for ifo in args.ifos:
             if not (ifo in fin and 'snr' in fin[ifo]):
                 continue
             live_time[ifo] += 8
```
GarethCabournDavies, line 188: Note: I (still) don't like that this is hard-coded, but don't see a way to get this from the files themselves

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -198,110 +177,193 @@ for filename in files:
...
+    fout.attrs['fit_function'] = args.fit_function
+    fout.attrs['fit_threshold'] = args.fit_threshold
+
+if not args.output_fit_plot:
+    logging.info("Done")
+    exit(0)
 
-fout.close()
-logging.info("Done")
+import matplotlib
```
GarethCabournDavies, line 315: Here onwards is some code I had around from doing the plots of the fits. The output is, for example, like this:

image

pycbc/events/cuts.py
```diff
@@ -388,7 +388,4 @@ def apply_template_cuts(bank, template_cut_dict, template_ids=None,
             raise ValueError("Cut parameter " + parameter + " not recognised."
                              " This shouldn't happen with input sanitisation")
 
-    logging.info("%d out of %d templates kept after applying template cuts",
```
GarethCabournDavies, line 391: This logging is applied once in the offline codes (pycbc_sngls_findtrigs and pycbc_coinc_findtrigs), but many times (once per file) in pycbc_live_single_trigger_fits, and spams the logging.

I have moved the logging to outside the function for this to be clearer

bin/live/pycbc_live_combine_single_fits
```diff
@@ -48,8 +48,7 @@ if args.conservative_percentile < 50 or \
                  "otherwise it is either not a percentile, or not "
                  "conservative.")
 
-colours = ['r', (1.0, 0.6, 0), 'y', 'g', 'c', 'b', 'm',
-           (0.8, 0.25, 0), (0.25, 0.8, 0)]
+colours = ['r',(1.0,0.6,0),'y','g','c','b','m',(0.8,0.25,0),(0.25,0.8,0)]
```
GarethCabournDavies, line 51: Adding more colours as we are looking at extending the number of bins

tdent: is this adding more or just removing the spaces in the existing list?

GarethCabournDavies:
poke @titodalcanton

bin/live/pycbc_live_combine_single_fits
```diff
@@ -48,24 +52,45 @@ if args.conservative_percentile < 50 or \
                  "otherwise it is either not a percentile, or not "
                  "conservative.")
 
-colours = ['r', (1.0, 0.6, 0), 'y', 'g', 'c', 'b', 'm',
-           (0.8, 0.25, 0), (0.25, 0.8, 0)]
+colours = ['r', (1.0,0.6,0), 'y', 'g', 'c', 'b', 'm',
```
GarethCabournDavies, line 55:
```suggestion
colours = ['r', (1.0, 0.6, 0), 'y', 'g', 'c', 'b', 'm',
```

bin/live/pycbc_live_single_trigger_fits
titodalcanton: We can put this in in `pycbc_live` if that helps, should be a trivial change.

bin/live/pycbc_live_combine_single_fits
titodalcanton: Note that you can actually use the colormaps to color lines. Maybe worth considering if you have many curves that are "indexed" by a continuous variable?

bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -195,6 +195,8 @@ template_ids = cuts.apply_template_cuts(
     ifos=trigs.ifos,
     template_ids=template_ids)
 
+logging.info("%d out of %d templates kept after applying template cuts",
+             len(template_ids), trigs.singles[0].bank['mass1'].size)
```
titodalcanton, line 199: Does `len(trigs.singles[0].bank)` work by any chance? Clearer IMO.

bin/all_sky_search/pycbc_sngls_findtrigs
```diff
@@ -113,6 +113,9 @@ template_ids = cuts.apply_template_cuts(
     ifos=[ifo],
     template_ids=template_ids)
 
+logging.info("%d out of %d templates kept after applying template cuts",
+             len(template_ids), trigs.bank['mass1'].size)
```
titodalcanton, line 117: Same as above.

bin/live/pycbc_live_combine_single_fits
```diff
@@ -49,23 +53,44 @@ if args.conservative_percentile < 50 or \
...
 
 live_times = {ifo : [] for ifo in args.ifos}
 
-for f in files:
+logging.info("Checking through %d files", len(args.trfits_files))
+
+epoch = datetime.date(2000,1,1)
+for f in args.trfits_files:
+
+    # Get the date information, there are a couple of ways to do this:
```
titodalcanton, line 83: This comment seems more useful just before line 94.

titodalcanton: Also, I would add in the comment a short reminder of why we need to do all that date logic - why can we not simply look at the GPS times of the triggers?

bin/live/pycbc_live_combine_single_fits
```diff
@@ -89,14 +116,15 @@ for f in files:
         analysis_day = int(m.group(3))
         n_since_epoch = (datetime.date(analysis_year, analysis_month,
                                        analysis_day)
-                         - datetime.date(2000,1,1)).days
+                         - epoch).days
         analysis_dates += [n_since_epoch]
+
     for ifo in args.ifos:
         if ifo not in fits_f:
+            logging.info(f + " has no " + ifo + " triggers")
```
titodalcanton, line 124: This condition is expected fairly often, so I would not spam the stderr so much.

bin/live/pycbc_live_combine_single_fits
```diff
@@ -243,7 +289,7 @@ for ifo in args.ifos:
         ax_alpha.legend(alpha_lines, alpha_labels, loc='lower center',
                         ncol=3, bbox_to_anchor=(0.5, 1.01))
         fig_alpha.tight_layout()
-        fig_alpha.savefig(args.output_plot_file_format.format(ifo + "-alphas"))
+        fig_alpha.savefig(outname + f'_{ifo}_fit_coeffs.png')
```
titodalcanton, line 292:
```suggestion
        fig_alpha.savefig(f'{outname}_{ifo}_fit_coeffs.png')
```

bin/live/pycbc_live_combine_single_fits
```diff
@@ -252,6 +298,6 @@ for ifo in args.ifos:
                         ncol=3, bbox_to_anchor=(0.5, 1.01))
         ax_count.grid(zorder=-30)
         fig_count.tight_layout()
-        fig_count.savefig(args.output_plot_file_format.format(ifo + "-counts"))
+        fig_count.savefig(outname + f'_{ifo}_counts.png')
```
titodalcanton, line 301:
```suggestion
        fig_count.savefig(f'{outname}_{ifo}_counts.png')
```

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -56,140 +41,148 @@ parser.add_argument("--fit-function", default="exponential",
...
 pycbc.init_logging(args.verbose)
 
+# Check the bin inputs to see if they are sensible
+if args.duration_bin_edges and (args.duration_bin_start or
+                                args.duration_bin_end or
+                                args.num_duration_bins):
+    # duration bin edges specified as well as the linear/logarithmi
+    raise parser.error("Cannot use --duration-bin-edges with "
+                       "--duration-bin-start, --duration-bin-end or "
+                       "--num-duration-bins.")
```
titodalcanton, line 97:
```suggestion
    parser.error("Cannot use --duration-bin-edges with "
                 "--duration-bin-start, --duration-bin-end or "
                 "--num-duration-bins.")
```

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -56,140 +41,148 @@ parser.add_argument("--fit-function", default="exponential",
...
+
+if not args.duration_bin_edges and not (args.duration_bin_start and
+                                        args.duration_bin_end and
+                                        args.num_duration_bins):
+    raise parser.error("--duration-bin-start, --duration-bin-end and "
+                       "--num-duration-bins must be set if not using "
+                       "--duration-bin-edges.")
```
titodalcanton, line 104:
```suggestion
    parser.error("--duration-bin-start, --duration-bin-end and "
                 "--num-duration-bins must be set if not using "
                 "--duration-bin-edges.")
```

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -56,140 +41,148 @@ parser.add_argument("--fit-function", default="exponential",
...
+
+if args.duration_bin_end and \
+        args.duration_bin_end <= args.duration_bin_start:
+    raise parser.error("--duration-bin-end must be greater than "
+                       "--duration-bin-start, got %.2f and %.2f",
+                       args.duration_bin_end, args.duration_bin_start)
```
titodalcanton, line 110:
```suggestion
    parser.error("--duration-bin-end must be greater than "
                 "--duration-bin-start, got "
                 f"{args.duration_bin_end} and {args.duration_bin_start}")
```

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -56,140 +41,148 @@ parser.add_argument("--fit-function", default="exponential",
...
 
 # Loop through files - add events which meet the immediately gettable
 # criteria
 date_directory = os.path.join(args.top_directory, args.analysis_date)
-events = {i: DictArray(data={rp: np.array([])
-                       for rp in required_datasets}) for i in args.ifos}
+
+if not os.path.exists(date_directory):
+    logging.warning("The directory %s does not exist.", date_directory)
+    raise OSError
```
titodalcanton, line 165:
```suggestion
    raise FileNotFoundError(f"The directory {date_directory} does not exist.")
```

bin/live/pycbc_live_single_trigger_fits
titodalcanton: In any case, taking it from the file name is probably better than hardcoding it. The current value will not work for the early-warning search, for example.

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -198,110 +191,219 @@ for filename in files:
...
-if cutting_newsnr:
-    logging.info("Keeping events with new_snr %s %.3f" % (nsineq[0], nsthresh))
-    ineq_fn = inequality_string_to_function(nsineq)
     for ifo in args.ifos:
-        idx_keep = np.nonzero(ineq_fn(events[ifo].data['new_snr'], nsthresh))[0]
-        if len(idx_keep):
-            events[ifo] = events[ifo].select(idx_keep)
+        # Continue if not triggers found in file
+        if ifo not in triggers:
+            continue
```
titodalcanton, line 227: If you simply iterate over `triggers` instead of over `args.ifos`, you can skip this check, right?

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -198,110 +191,219 @@ for filename in files:
...
+                          triggers_cut, args.sngl_ranking)
+
+        triggers_cut[args.sngl_ranking] = sngls_value
+
+        triggers_da = DictArray(data=triggers_cut)
+
+        # If we are clustering, take the max sngl_ranking value
+        if args.cluster:
+            # Use sngls_ranking parameter for clustering
+            max_idx = sngls_value.argmax()
```
titodalcanton, line 256: Unless `sngls_value` is a more complicated object than I am imagining, it seems like this is taking the argmax over the entire day, is this the desired behavior?

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -198,110 +191,219 @@ for filename in files:
...
+            triggers_cut = triggers_da.select(max_idx)
+
+        if no_triggers_yet[ifo]:
```
titodalcanton, line 259: Do you really need the auxiliary `no_triggers_yet` dict? Can you not do this instead:
```python
if ifo in events:
    events[ifo] += triggers_da
else:
    events[ifo] = triggers_da
```

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -198,110 +191,219 @@ for filename in files:
...
+alphas = {i: np.zeros(n_bins, dtype=np.float32) for i in args.ifos}
+counts = {i: np.zeros(n_bins, dtype=np.float32) for i in args.ifos}
+binned_sngl_stats = {ifo: {} for ifo in args.ifos}
+
 for ifo in args.ifos:
-    for key in events[ifo].data:
-        fout['{}/triggers/{}'.format(ifo, key)] = events[ifo].data[key]
-    event_bin = np.array([tbins[d] for d in events[ifo].data['template_duration']])
-    for bin_num in range(args.num_duration_bins):
+    if ifo not in events: continue
```
titodalcanton, line 291: Can you avoid this check if you simply iterate over `events` instead?

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -198,110 +191,219 @@ for filename in files:
...
-fout['bins_lower'] = tbins.lower()
+with h5py.File(oput_file, 'w') as fout:
 
-fout.attrs['analysis_date'] = args.analysis_date
-fout.attrs['input'] = sys.argv
-fout.attrs['cuts'] = ineq_str
-fout.attrs['fit_function'] = args.fit_function
-fout.attrs['fit_threshold'] = args.fit_threshold
+    for ifo in args.ifos:
+        if ifo not in events: continue
```
titodalcanton, line 317: Can you skip this check if you iterate over `events` instead?

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -198,110 +191,219 @@ for filename in files:
...
-                                      events[ifo].data[args.fit_param][inbin],
+                                      binned_sngl_stats[ifo][bin_num],
                                       args.fit_threshold)
-    fout['{}/fit_coeff'.format(ifo)] = alphas[ifo]
-    fout['{}/counts'.format(ifo)] = counts[ifo]
-    fout[ifo].attrs['live_time'] = live_time[ifo]
 
-fout['bins_upper'] = tbins.upper()
-fout['bins_lower'] = tbins.lower()
+with h5py.File(oput_file, 'w') as fout:
```
titodalcanton, line 314: Add a `logging.info()` just before this indicating that you are writing the results. We want to know when the fitting has finished and we are about to touch the filesystem.

bin/live/pycbc_live_single_trigger_fits
titodalcanton: We should definitely make these nice plots, but I suggest splitting the plotting code to a separate script. We do not necessarily want to put the plots in the same place as the fit results anyway, they should probably go together with the daily lag and single trigger plots.

titodalcanton:
@GarethCabournDavies sorry for the slowness, this was a massive one. See comments above, some are required fixes.

bin/live/pycbc_live_single_trigger_fits
GarethCabournDavies: This is the max over the trigger file only

bin/live/pycbc_live_combine_single_fits
GarethCabournDavies: Wait, that's a much better way to do this. How did I not think of that before.
I suppose we may not always decide to output the time of the triggers which we used in the fitting from the other code, but we probably will do, especially now the plotting has been split off

bin/live/pycbc_live_combine_single_fits
GarethCabournDavies: That's a great idea - they are now indexed based on how far they are through the "rainbow_r" colormap (with an option to choose colormap)

bin/live/pycbc_live_single_trigger_fits
GarethCabournDavies: For info - updated plot using the reversed rainbow
image

GarethCabournDavies:
Okay, I think this is ready for re-review now

GarethCabournDavies:
I'm going to be away now until the new year, but I am happy for edits to be made and this to be merged when ready

titodalcanton (approved):
Given that Gareth is going to be away for several weeks, I propose merging this now and starting to test it, as it does not affect any currently used analysis.

PR #4159, opened 2022-10-19
Adding per-event-type live pastro

tdent:
To get a better estimate of per-event-type noise and signal densities, remove the current FAR trials factor and account for the triggered detectors in the horizon based sensitivity scaling.

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -340,32 +340,32 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
 
         network_snr2 += snr ** 2
 
-    ifos = sorted(set(ifos) - set(subthreshold_ifos))
-    if not ifos:
-        raise RuntimeError('all interferometers have SNR below threshold.'
-                           ' Is this really a candidate event?')
+    trig_ifos = sorted(set(ifos) - set(subthreshold_ifos))
+    if not trig_ifos:
+        raise RuntimeError(f'All interferometers have SNR below threshold '
+                           '{thresh_SNR}. Is this really a candidate event?')
 
     coinc_results['foreground/stat'] = network_snr2 ** 0.5
     coinc_results['foreground/ifar'] = ifar
 
     # BAYESTAR's convention for end time of subthreshold detectors
-    subthreshold_sngl_time = np.mean(
-            [coinc_results['foreground/%s/end_time' % ifo] for ifo in ifos])
+    subthreshold_sngl_time = np.mean([
+        coinc_results['foreground/%s/end_time' % ifo] for ifo in trig_ifos])
 
     window_bins = int(snr_series_duration * sample_rate)
-    followup_data = {}
-    for ifo in ifos + subthreshold_ifos:
+    skyloc_data = {}
+    for ifo in ifos:
         key = 'foreground/{}/'.format(ifo)
 
         if ifo in subthreshold_ifos:
             coinc_results[key + 'end_time'] = subthreshold_sngl_time
 
-        followup_data[ifo] = {
+        skyloc_data[ifo] = {
             'psd': interpolate(coinc_results[key + 'psd_series'], 0.25)
         }
 
-        # now go back to the SNR time series and cut out
-        # a shorter piece centered on the end time
+        # Go back to the SNR time series and select a piece centered on the
+        # end time
         time = coinc_results[key + 'end_time']
         snr_series = coinc_results[key + 'snr_series']
         peak_bin = int((time - snr_series.start_time) / snr_series.delta_t)
@@ -381,22 +381,26 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
             max_bin = peak_bin + window_bins + 1
             min_bin = peak_bin - window_bins
 
-        followup_data[ifo]['snr_series'] = \
+        skyloc_data[ifo]['snr_series'] = \
                 snr_series[min_bin:max_bin].astype(np.complex64)
 
-    kwargs = {'psds': {ifo: followup_data[ifo]['psd']
-                       for ifo in ifos + subthreshold_ifos},
+    kwargs = {'psds': {ifo: skyloc_data[ifo]['psd'] for ifo in ifos},
               'low_frequency_cutoff': f_low,
               'high_frequency_cutoff': f_upper,
-              'followup_data': followup_data,
+              'skyloc_data': skyloc_data,
               'channel_names': channel_names}
 
-    doc = live.CandidateForGraceDB(ifos, coinc_results, **kwargs)
+    doc = live.CandidateForGraceDB(
+        trig_ifos,
+        trig_ifos,
+        coinc_results,
+        **kwargs
+    )
 
     ligolw_file_path = os.path.join(tmpdir, file_name_tag + '.xml')
 
     if gracedb_server:
-        comments = ['Manual followup up from PyCBC']
+        comments = ['Manual followup from PyCBC']
         gid = doc.upload(ligolw_file_path,
                          gracedb_server=gracedb_server,
                          testing=test_event,
@@ -406,13 +410,11 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
         doc.save(ligolw_file_path)
 
     # Defining the approximant to feed into BAYESTAR
-
     row = WaveformArray.from_kwargs(
         mass1=mass1,
         mass2=mass2,
         spin1z=spin1z,
         spin2z=spin2z)
-
     bs_approximant = waveform.bank.parse_approximant_arg(approximant, row)[0]
 
     # BAYESTAR uses TaylorF2 instead of SPAtmplt
--- a/bin/pycbc_optimize_snr
+++ b/bin/pycbc_optimize_snr
@@ -367,7 +367,7 @@ spin2z = opt_params[3]
 
 # Prepare for GraceDB upload
 coinc_results = {}
-followup_data = {}
+skyloc_data = {}
 
 loudest_ifo = None
 loudest_snr_allifos = 0
@@ -437,9 +437,9 @@ for idx, ifo in enumerate(ifos):
     coinc_results['foreground/'+ifo+'/chisq_dof'] = 1
     coinc_results['foreground/'+ifo+'/template_duration'] = \
         fp['template_duration'][()]
-    followup_data[ifo] = {}
-    followup_data[ifo]['psd'] = interpolate(data[ifo].psd, 0.25)
-    followup_data[ifo]['snr_series'] = snr_series_dict[ifo]
+    skyloc_data[ifo] = {}
+    skyloc_data[ifo]['psd'] = interpolate(data[ifo].psd, 0.25)
+    skyloc_data[ifo]['snr_series'] = snr_series_dict[ifo]
 
     # Find loudest SNR within coincidence window of the largest peak
     snr_peak = abs(loudest_snrs[ifo])
@@ -448,8 +448,8 @@ for idx, ifo in enumerate(ifos):
     # Add small buffer
     lttbd += 0.005
 
-    time_window = [loudest_snr_time_allifos-lttbd,
-                   loudest_snr_time_allifos+lttbd]
+    time_window = [loudest_snr_time_allifos - lttbd,
+                   loudest_snr_time_allifos + lttbd]
 
     snr_slice = snr_series_dict[ifo].time_slice(time_window[0], time_window[1])
     max_snr_idx = numpy.argmax(abs(snr_slice))
@@ -476,13 +476,16 @@ for ifo in fp['channel_names'].keys():
 
 mc_area_args = load_hdf5_to_dict(fp, 'mc_area_args/')
 
-kwargs = {'psds': {ifo: followup_data[ifo]['psd'] for ifo in ifos},
+kwargs = {'psds': {ifo: skyloc_data[ifo]['psd'] for ifo in ifos},
           'low_frequency_cutoff': flow,
-          'followup_data': followup_data,
+          'skyloc_data': skyloc_data,
           'channel_names': channel_names,
           'mc_area_args': mc_area_args}
 
+# Treat all ifos as having triggers
+# Do not recalculate p_astro
 doc = live.CandidateForGraceDB(
+    ifos,
     ifos,
     coinc_results,
     upload_snr_series=True,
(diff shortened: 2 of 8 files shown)


titodalcanton:
Under test at IJCLab.

tdent:
@ahnitz maybe you could look at the changes to the test module?  I think they ought to have been made back in 2019 ..

pycbc/io/live.py
```diff
@@ -166,30 +170,16 @@ def __init__(self, ifos, coinc_results, **kwargs):
...
-                     for ifo in ifos]) + self.time_offset
-
-        # For subthreshold detectors, respect BAYESTAR's assumptions and checks
-        bayestar_check_fields = ('mass1 mass2 mtotal mchirp eta spin1x '
-                                 'spin1y spin1z spin2x spin2y spin2z').split()
-        for sngl in sngl_inspiral_table:
-            if sngl.ifo in followup_ifos:
-                for bcf in bayestar_check_fields:
-                    setattr(sngl, bcf, getattr(sngl_populated, bcf))
-                sngl.end = lal.LIGOTimeGPS(self.merger_time)
```
titodalcanton, line 181: Are you intending to remove this? I do not see the same functionality being reimplemented elsewhere.

tdent: Yes, because with the O3 behaviour followup_ifos was by design an empty list and all the sngl inspiral fields are populated in the SNR followup code.

pycbc/io/live.py
```diff
@@ -44,8 +44,8 @@ def __init__(self, ifos, coinc_results, **kwargs):
             Dictionary providing SNR time series for each detector, to be used
             in sky localization with BAYESTAR. The format should be
             `followup_data['H1']['snr_series']`. More detectors can be present
-            than given in `ifos`. If so, the extra detectors will only be used
-            for sky localization.
+            than in `ifos`. If so, extra detectors will only be used for sky
```
titodalcanton, line 47:
```suggestion
            than in `coinc_ifos`. If so, extra detectors will only be used for sky
```

pycbc/io/live.py
```diff
@@ -56,35 +56,39 @@ def __init__(self, ifos, coinc_results, **kwargs):
...
         else:
             self.snr_series = None
-            usable_ifos = ifos
-            followup_ifos = []
+            snr_ifos = coinc_ifos
+
+        # Set merger time to the mean of ifo peaks - to be used for file naming
+        self.merger_time = numpy.mean(
+                    [coinc_results[f'foreground/{i}/end_time']
+                     for i in snr_ifos]) + self.time_offset
```
titodalcanton, line 91: I know this is probably not what the old code was doing, but the average should be taken over `coinc_ifos` instead, no? Otherwise we are including potentially unreliable time estimates in the average.

pycbc/io/live.py
titodalcanton: Yeah you are right that this code had no effect when the class is used by PyCBC Live. I am not sure about `pycbc_make_skymap` though, that should be checked. We may also need this code again once we implement "KAGRA mode", depending on how that is implemented.

pycbc/io/live.py
```diff
@@ -368,13 +359,10 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
-                    base = 'foreground/{}/'.format(ifo)
-                    snr = self.coinc_results[base + 'snr']
-                    mt = (self.coinc_results[base + 'end_time']
-                          + self.time_offset)
-                    pl.plot([mt - ref_time], [snr], c=ifo_color(ifo),
-                            marker='x')
+                base = 'foreground/{}/'.format(ifo)
+                snr = self.coinc_results[base + 'snr']
+                mt = (self.coinc_results[base + 'end_time'] + self.time_offset)
+                pl.plot([mt - ref_time], [snr], c=ifo_color(ifo), marker='x')
```
titodalcanton, line 365: I think here you are again assuming that each `snr_ifo` will have a corresponding entry in `coinc_results`. This is the case right now for PyCBC Live, but may not be in the future (e.g. KAGRA used only for BAYESTAR but not for improving the FAR) or for other applications.

I would not make this change.

test/test_io_live.py
```diff
@@ -81,16 +84,15 @@ def do_test(self, n_ifos, n_ifos_followup):
...
+            # fill in the various fields - all ifos now get end_time, etc
+            base = 'foreground/' + ifo + '/'
+            results[base + 'end_time'] = t_peak + offset
+            results[base + 'snr'] = amplitude
+            results[base + 'sigmasq'] = np.random.uniform(1e6, 2e6)
             followup_data[ifo] = {'snr_series': snr_series,
                                   'psd': psd}
 
-        for ifo, k in itertools.product(trig_ifos, self.template):
+        for ifo, k in itertools.product(all_ifos, self.template):
```
titodalcanton, line 95: Please revert this change. I do want the `CandidateForGraceDB` class to support the case where we do have some detectors below threshold in addition to the ones that generated the candidate, even if this situation is not met in the current implementation of PyCBC Live.

titodalcanton:
Most of this PR is fine. However, there seems to be some removal of functionality from the `CandidateForGraceDB` class, and I want to keep that functionality even if not *presently* used by PyCBC Live. It may be used by `pycbc_make_skymap` (I do not remember right now) and we may want to use it for KAGRA in a few months.

pycbc/io/live.py
tdent: I would be OK with not removing this if the set of ifos it potentially concerns can be clearer. 'Followup' could refer either to significance followup or EM followup ie sky map.
In either case the ifo requires a SNR time series to be calculated, though, and I don't know where that would happen that doesn't already populate this information elsewhere (?)

test/test_io_live.py
tdent: The current situation does, if I understood it, allow for subthreshold ifos, but these are fed to CandidateFGDB as if they were triggering ifos, ie with the same structure of SNR time series and peak 'trigger' values.
What is the alternative if ifos are to be treated differently but still used for sky localization? Their SNR series will have to be evaluated, where is that supposed to happen?

tdent:
@titodalcanton See individual comment discussions - I can't work out what is the functionality that is supposedly being removed. I understand that we may want to make a distinction in the future between significance/SNR followup and sky loc use, but I don't see how this would be done based on recycling bits of now-inactive O2 code side by side with the O3 behaviour. It seems to me better to throw away the O2 structure which currently does nothing and cannot be meaningfully tested or evaluated, and get to a simpler and more comprehensible version of the O3 state, which might then be a basis for putting 'back' whatever bits of the O2 code will actually be needed, in a hopefully better overall structure. 
I am not much of a fan of keeping old lines of code around that currently do nothing (and so confuse developers and maintainers) but might in future be reactivated in some not yet clear way.
If we must have them, comment them out so it is absolutely clear they do nothing atm.

pycbc/io/live.py
tdent: If you mean the O3 code, indeed it took all ifos with a SNR time series, and those all had an identified peak/trigger time. Is the point here that the peak time was 'unreliable' if the SNR was too low (eg GW170817 in V1)?
We can discuss whether this is enough of a problem that we want to make a change wrt O3 behaviour, but that would be beyond the scope of this PR. I have tried to only reproduce (and rationalise!) O3 behaviour here concerning all candidate info except pastro. I think it might confuse reviewers if we choose to change the definition of the event time in the same PR as the pastro change.

test/test_io_live.py
tdent: More broadly, what should be written in the CandidateFGDB docstring to better describe what it ought to do with the different sets of ifos?
Unless we have a technical description of the desired behaviour it's hard to write a meaningful test

tdent:
Many changes removing the ability for followup_data to index different ifos from coinc_results have been removed from the branch

tdent:
@titodalcanton I think it's ready to start testing / checking over again now

titodalcanton:
Running at IJCLab.

bin/pycbc_make_skymap
```diff
@@ -381,22 +381,26 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
...
               'low_frequency_cutoff': f_low,
               'high_frequency_cutoff': f_upper,
-              'followup_data': followup_data,
+              'skyloc_data': skyloc_data,
               'channel_names': channel_names}
 
-    doc = live.CandidateForGraceDB(ifos, coinc_results, **kwargs)
+    doc = live.CandidateForGraceDB(
+        trig_ifos,
+        trig_ifos,
```
titodalcanton, line 395: Why is the second argument not `ifos`?

bin/pycbc_optimize_snr
```diff
@@ -493,7 +496,7 @@ if args.enable_gracedb_upload:
     comment = ('Automatic PyCBC followup of trigger '
                '<a href="/events/{0}/view">{0}</a> to find the template '
                'parameters that maximize the SNR. The FAR of this trigger is '
-               'copied from {0} and does not reflect this triggers\' template '
+               'copied from {0} and does not reflect this trigger\'s template '
```
titodalcanton, line 499: Should the comment say the same for p_astro?

pycbc/events/coinc.py
```diff
@@ -1212,7 +1209,7 @@ def _find_coincs(self, results, valid_ifos):
             offsets = numpy.concatenate(offsets)
             ctime0 = numpy.concatenate(ctimes[self.ifos[0]]).astype(numpy.float64)
             ctime1 = numpy.concatenate(ctimes[self.ifos[1]]).astype(numpy.float64)
-
+            logging.info("Clustering %s-%s coincs", self.ifos[0], self.ifos[1])
```
titodalcanton, line 1212: Import and use the `ppdets()` function here.

bin/pycbc_make_skymap
titodalcanton: Ah, I guess because they do not contribute to the "significance", by looking at the updated docstring… Strictly speaking, none of the detectors contribute to the significance in this case.

pycbc/io/live.py
```diff
@@ -22,30 +22,34 @@ class CandidateForGraceDB(object):
...
             the hdf file for this time.
         psds: dict of FrequencySeries
-            Dictionary providing PSD estimates for all involved detectors.
+            Dictionary providing PSD estimates for all detectors observing.
         low_frequency_cutoff: float
             Minimum valid frequency for the PSD estimates.
         high_frequency_cutoff: float, optional
             Maximum frequency considered for the PSD estimates. Default None.
-        followup_data: dict of dicts, optional
+        skyloc_data: dict of dicts, optional
```
titodalcanton, line 47: "skyloc" is not a particularly better name than "followup" here, because BAYESTAR does more than *sky* localization (full 3D spatial localization + Bayes factors). Maybe I should have just chosen "bayestar_data" in the beginning :)

titodalcanton: And yes, the docstring below is also not completely correct because of the same point.

pycbc/io/live.py
```diff
@@ -59,32 +63,37 @@ def __init__(self, ifos, coinc_results, **kwargs):
         self.template_id = coinc_results[f'foreground/{ifos[0]}/template_id']
         self.coinc_results = coinc_results
         self.ifos = ifos
+        assert len(coinc_ifos) < 3, f"I can't handle {coinc_ifos} coinc ifos!"
+        self.psds = kwargs['psds']
```
titodalcanton, line 67: Whan happens with singles then?

pycbc/io/live.py
```diff
@@ -216,17 +226,22 @@ def __init__(self, ifos, coinc_results, **kwargs):
 
         # P astro calculation
         if 'padata' in kwargs:
-            padata = kwargs['padata']
+            assert len(coinc_ifos) < 3, \
+                f"p_astro can't handle {coinc_ifos} coinc ifos!"
```
titodalcanton, line 230: Same comment as above here, this fails for single-detector candidates?

pycbc/io/live.py
```diff
@@ -216,17 +226,22 @@ def __init__(self, ifos, coinc_results, **kwargs):
...
                 'mass2': sngl_populated.mass2,
                 'spin1z': sngl_populated.spin1z,
                 'spin2z': sngl_populated.spin2z,
                 'network_snr': network_snrsq ** 0.5,
-                'far': far}
-            horizons = {ifo: self.psds[ifo].dist for ifo in self.psds}
+                'far': far,
+                'triggered': coinc_ifos,
+                # Active = potentially contributing to significance:
+                # ignore ifos that only contribute to sky loc
```
titodalcanton, line 240: I would expect "active" to mean "with usable data", and I think in other contexts we do indeed use "active" to mean "in observing mode", so I am slightly concerned about this nomenclature here. Why not "trig_ifos" / "sig_ifos" / "ifos"?

titodalcanton:
@tdent a few more comments, but I am less concerned now.

bin/pycbc_optimize_snr
tdent: Dealing with pastro for the SNR max upload is beyond the scope of this PR

pycbc/io/live.py
tdent: The intention here is not to give a complete description of bayestar, it is to resolve the ambiguity of 'followup' now that we have significance followup as well

bin/pycbc_make_skymap
tdent: Any request for changes ?

pycbc/events/coinc.py
tdent: from detector.py?

pycbc/io/live.py
tdent: I expect coinc_ifos to be length 1 .. ??

tdent: Actually I meant to remove this assert and only have the one for the pastro calculation, since there is no fundamental reason to exclude 3 coinc ifos here

pycbc/io/live.py
tdent: singles should have coinc_ifos of length 1 ?

pycbc/io/live.py
tdent: You mean the use of 'active' in the trigger_data dict that is fed to pastro? The thinking there is, the ifos active from the point of view of possibly influencing p astro.  The p astro module does not need to know about or think about other ifos ;)

tdent: Has been swapped out for 'sensitive'

bin/pycbc_make_skymap
titodalcanton: I suppose not.

titodalcanton (approved):
I think we can approve this now. @tdent could you please fix the remaining Codeclimate issues, which look trivial enough?

PR #4161, opened 2022-10-21
add time marginalized heterodyne likelikelihood

ahnitz:
This adds a new model to handle heterodyne likelihoods that allow time marginalized (and related such as sky) optimizations.

There is also a bugfix to the sky normalization included here. I made a mistake in the normalization which I found during testing the detailed results in a pp test.


ahnitz:
I've updated and split apart PRs so closing this now in favor of those

PR #4162, opened 2022-10-21
Fix a waveform plugin issue that missed the waveform length estimator

yi-fan-wang:
When a user provides a frequency domain (fd) waveform plugin and a waveform length estimator, PyCBC doesn't make a time domain (td) version for the waveform correctly. It seems because PyCBC calls the waveform plugin later than making the fd/td transformation for a waveform model.

This PR encapsulates the fd/td transformation in a function, and explicitly calls this function in the waveform plugin function when the user provides a waveform length estimation.

### DIFF
--- a/pycbc/waveform/plugin.py
+++ b/pycbc/waveform/plugin.py
@@ -69,6 +69,7 @@ def add_length_estimator(approximant, function):
     from pycbc.waveform.waveform import td_fd_waveform_transform
     td_fd_waveform_transform(approximant)
 
+
 def retrieve_waveform_plugins():
     """ Process external waveform plugins
     """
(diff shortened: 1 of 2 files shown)


ahnitz (commented):
@yi-fan-wang Yeah, this looks right. Can you double check if it works though as expected by testing with a time length plugin? If it works (and address the CC messages) then I think this can be merged.

yi-fan-wang:
@ahnitz I tested this PR with this waveform plugin: https://github.com/yi-fan-wang/TestingGR_with_Gravwaves. It's a frequency domain waveform with a waveform length estimator. The following codes work well for me:

```
import pycbc.waveform
hp, hc = pycbc.waveform.get_fd_waveform(approximant="birefringence",baseapprox='IMRPhenomD',
                                        mass1=30,mass2=30,delta_f=1/32,distance=1000,f_lower=20,parity_mpvinverse=100)
hp_td, hc_td = pycbc.waveform.get_td_waveform(approximant="birefringence",baseapprox='IMRPhenomD',
                                        mass1=30,mass2=30,delta_t=1./1024,f_lower=20,distance=1000,
                                        parity_mpvinverse=100)
```

ahnitz (approved):

PR #4166, opened 2022-10-28
Kde coinc stat

PRAVEEN-mnl:


pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+            A list containing the filenames of hdf format files used to help
+            construct the coincident statistics. The files must have a 'stat'
+            attribute which is used to associate them with the appropriate
+            statistic class.
+        ifos: list of strs, not used here
+            The list of detector names
+        """
+        ExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
+                                         ifos=ifos, **kwargs)
+        parsed_attrs = [f.split('-') for f in self.files.keys()]
```
tdent, line 1787: perhaps add a comment here to say what are the exact names of the attributes you are looking for?

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        self.data_name = [at[0] for at in parsed_attrs if
+                       (len(at) == 2 and at[1] == 'kde_file')]
+        if not len(self.data_name):
+            raise RuntimeError("None of the statistic files has the required "
+                               "attribute called {dname}-kde_file !")
```
tdent, line 1792: what is 'dname' ? I don't think it has been defined yet

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        self.kde_by_tid = {}
+        for dname in self.data_name:
+            self.kde_by_tid[dname] = self.assign_kdes(dname)
+            #print('self.kde_by_tid=', self.kde_by_tid[dname])
+        self.curr_tnum = None       
+ 
+    def assign_kdes(self, dname):
+        """
+        Extract fits from fit files
```
tdent, line 1801: I guess you want to say 'extract values from a KDE file' or something similar?

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+
+        Parameters
+        -----------
+        dname: str
+            signal or template.
+
+        Returns
+        -------
+        rate_dict: dict
```
tdent, line 1810: The return value is not called 'rate_dict' it is called 'kde_by_tid_dict', please edit the comment here

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+                                         ifos=ifos, **kwargs)
+        parsed_attrs = [f.split('-') for f in self.files.keys()]
+        self.data_name = [at[0] for at in parsed_attrs if
+                       (len(at) == 2 and at[1] == 'kde_file')]
+        if not len(self.data_name):
+            raise RuntimeError("None of the statistic files has the required "
+                               "attribute called {dname}-kde_file !")
+        self.kde_by_tid = {}
+        for dname in self.data_name:
+            self.kde_by_tid[dname] = self.assign_kdes(dname)
```
tdent, line 1795: This seems too complicated, I don't think you need to make an assignment here, you can put the assignment into the assign_kdes() method

pycbc/events/stat.py
tdent: Or you might not want to use a return value at all ..

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+
+        Parameters
+        -----------
+        dname: str
+            signal or template.
+
+        Returns
+        -------
+        rate_dict: dict
+            A dictionary containing kde values for signal and template files
```
tdent, line 1811: does the function return a dictionary containing both signal and template kdes, or does it just return one of them?

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+
+        """
+        kde_file = self.files[dname+'-kde_file']
+        template_id = kde_file['template_id'][:]
```
tdent, line 1815: Do you expect the values will need sorting?  I thought the values in the file are always going to be in the same order as the templates in the template file.

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        tid_sort = numpy.argsort(template_id)
+        kde_by_tid_dict = {}
+        kde_by_tid_dict[dname+'_data_kde'] = \
+            kde_file['data_kde'][:][tid_sort]
+        print('self.kde_by_tid_dict', kde_by_tid_dict)
```
tdent, line 1820: We will not want print statements when the code is used in production

pycbc/events/stat.py
tdent: Or you might want to return just the kde values, not a dictionary, because it's not necessary to have such a complicated structure for the KDEs as it was for the background fits that depend on both the detector and the template

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        -------
+        rate_dict: dict
+            A dictionary containing kde values for signal and template files
+
+        """
+        kde_file = self.files[dname+'-kde_file']
+        template_id = kde_file['template_id'][:]
+        tid_sort = numpy.argsort(template_id)
+        kde_by_tid_dict = {}
+        kde_by_tid_dict[dname+'_data_kde'] = \
```
tdent, line 1818: I don't know if it is necessary to create another dictionary here. Would it work if the function just returned the array of KDE values?

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+            as self.ifos
+
+        Returns
+        -------
+        value: log of coinc signal rate density for the given single-ifo
+            triggers and time shifts
+        """
+        logr_s = ExpFitFgBgNormStatistic.logsignalrate(self, stats,
+                                                       shift, to_shift)
+        signal_kde = self.kde_by_tid["signal"]
```
tdent, line 1846: This will fail if the kde_by_tid dictionary doesn't have entries called "signal" and "template". So, you should put a check in the __init__ method to see if those entries exist.

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        template_kde = self.kde_by_tid["template"]
+        logr_s += [numpy.log(i/j) for i, j in zip(signal_kde, template_kde)]
```
tdent, line 1848: This looks like a very strange way of doing a calculation over an array. But I think to get the correct values of the KDEs, you need to know which template or templates is being analyzed. How does this function know which template the triggers in 'stats' come from? You need to get the information about which template we are looking at and use that to index the signal and template kde arrays.

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        print('logr_s=', logr_s)
+        return logr_s
+
+    def single(self, trigs):
```
tdent, line 1852: Can you put this method before the rank_stat_coinc method in the code, as it logically comes before?

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        -------
+        numpy.ndarray
+            The array of single detector values
+        """
+        try:
+            tnum = trigs.template_num
+        except AttributeError:
+            tnum = trigs['template_id']
+        self.curr_tnum = self.kde_by_tid[tnum]
+        print(self.curr_tnum)
```
tdent, line 1875: Again we will not want to have print statements in the code when used in production.

pycbc/events/stat.py
```diff
@@ -1758,6 +1757,155 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        -------
+        value: log of coinc signal rate density for the given single-ifo
+            triggers and time shifts
+        """
+        logr_s = ExpFitFgBgNormStatistic.logsignalrate(self, stats,
+                                                       shift, to_shift)
+        signal_kde = self.kde_by_tid["signal"]
+        template_kde = self.kde_by_tid["template"]
+        logr_s += [numpy.log(i/j) for i, j in zip(signal_kde, template_kde)]
+        print('logr_s=', logr_s)
```
tdent, line 1849: Can you remove print statements from code in the PR please

pycbc/events/stat.py
tdent: What do you want to do with the curr_tnum variable? You are not using it for anything.

pycbc/events/stat.py
```diff
@@ -1758,6 +1758,145 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+            statistic class.
+        ifos: list of strs, not used here
+            The list of detector names
+        """
+        ExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
+                                         ifos=ifos, **kwargs)
+        #looking for 'signal' and 'template' attributes in kde files
+        # and is represented as dname
+        parsed_attrs = [f.split('-') for f in self.files.keys()]
+        self.data_name = [at[0] for at in parsed_attrs if
```
tdent, line 1791: suggest to call this 'kde_names'

pycbc/events/stat.py
tdent: Instead of this check I think you want to check that self.kde_names has exactly 2 entries and they are "signal" and "template"

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+            construct the coincident statistics. The files must have a 'stat'
+            attribute which is used to associate them with the appropriate
+            statistic class.
+        ifos: list of strs, not used here
+            The list of detector names
+        """
+        ExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
+                                         ifos=ifos, **kwargs)
+        #looking for 'signal' and 'template' attributes in kde files
+        #and is represented as kname
```
tdent, line 1841: This comment doesn't really help, later in the code you use 'kname' as a dummy variable but that is already clear from just reading the code.

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        parsed_attrs = [f.split('-') for f in self.files.keys()]
+        self.kde_names = [at[0] for at in parsed_attrs if
+                       (len(at) == 2 and at[1] == 'kde_file')]
+        if not len(self.kde_names):
```
tdent, line 1845: I already said that this check is not sufficient. You need to code a check that there are exactly two elements in self.kde_names and that the elements are "signal" and "template". The code here only checks for `len(self.kde_names)` which is not enough for the statistic to be calculated

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+            raise RuntimeError("None of the statistic files has the required "
+                               "attribute called {signal}-kde_file or " 
+                               "{template}-kde_file !")
+        self.kde_by_tid = {}
+        for kname in self.kde_names:
+            self.assign_kdes(kname)
+        self.curr_tnum = None
```
tdent, line 1852: Add a comment saying that this variable will hold the template ids of events for the statistic calculation

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+                               "attribute called {signal}-kde_file or " 
+                               "{template}-kde_file !")
+        self.kde_by_tid = {}
+        for kname in self.kde_names:
+            self.assign_kdes(kname)
+        self.curr_tnum = None       
+ 
+    def assign_kdes(self, kname):
+        """
+        Extract values from from KDE files
```
tdent, line 1856: Typo 'from from'

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+            signal or template.
+        """
+        with h5py.File(self.files[kname+'-kde_file'], 'r') as kde_file:
+            self.kde_by_tid[kname+'_kdevals'] = kde_file['data_kde'][:]
+
+    def single(self, trigs):
+        """
+        Calculate the necessary single detector information
+
+        In this case the ranking rescaled (see the lognoiserate method here)
```
tdent, line 1870: This docstring is not useful for the code here. Remove it and write another one

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        """
+        try:
+            tnum = trigs.template_num  # exists if accessed via coinc_findtrigs
+            ifo = trigs.ifo
+        except AttributeError:
+            tnum = trigs['template_id']  # exists for SingleDetTriggers
+            assert len(self.ifos) == 1
+            # Should be exactly one ifo provided
+            ifo = self.ifos[0]
+        if self.curr_tnum is None:
```
tdent, line 1892: This is OK, but it would be better to put this 'if' before the try/except block. 
The code you have at the moment always gets the tnum value from the trigs object, but it is not always used or needed because self.curr_tnum might not be None. You should only get the tnum values if they are needed, which is only the case if self.curr_tnum is None.

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+            Dictionary-like object holding single detector trigger information.
+
+        Returns
+        -------
+        numpy.ndarray
+            The array of single detector values
+        """
+        try:
+            tnum = trigs.template_num  # exists if accessed via coinc_findtrigs
+            ifo = trigs.ifo
```
tdent, line 1886: Is `ifo` needed for anything in this code?  If not, why are you getting the value here?

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        except AttributeError:
+            tnum = trigs['template_id']  # exists for SingleDetTriggers
+            assert len(self.ifos) == 1
```
tdent, line 1889: Why do you have this code about self.ifos and assigning the variable `ifo`? Is it needed for anything?

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+
+        Returns
+        -------
+        value: log of coinc signal rate density for the given single-ifo
+            triggers and time shifts
+        """
+        logr_s = ExpFitFgBgNormStatistic.logsignalrate(self, stats, shift, to_shift)
+        signal_kde = self.kde_by_tid["signal_kdevals"][self.curr_tnum]
+        template_kde = self.kde_by_tid["template_kdevals"][self.curr_tnum]
+        logr_s += numpy.log(signal_kde/template_kde)
```
tdent, line 1920: looks OK, but for formatting you should have spaces around the division operator.

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,147 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        -------
+        numpy.ndarray
+            Array of limits on the limifo single statistic to
+            exceed thresh.
+        """
+        loglr = ExpFitFgBgNormStatistic.coinc_lim_for_thresh(
+                    self, s, thresh, limifo, **kwargs)
+        signal_kde = self.kde_by_tid["signal_kdevals"][self.curr_tnum]
+        template_kde = self.kde_by_tid["template_kdevals"][self.curr_tnum]
+        loglr += numpy.log(signal_kde/template_kde)
```
tdent, line 1950: same comment - spaces around ` / `

tdent:
It looks like this is nearly finished, but there are some lines of code that are not necessary that seem to have been copied from other stat classes and there are some formatting issues that codeclimate is unhappy with, see here 
https://codeclimate.com/github/gwastro/pycbc/pull/4166 
these should be basically easy (but boring) to fix

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,139 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        ifos: list of strs, not used here
+            The list of detector names
+        """
+        ExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
+                                         ifos=ifos, **kwargs)
+        # the stat file attributes are hard-coded as 'signal-kde_file' & 'template-kde_file' 
+        parsed_attrs = [f.split('-') for f in self.files.keys()]
+        self.kde_names = [at[0] for at in parsed_attrs if
+                       (len(at) == 2 and at[1] == 'kde_file')]
+        assert self.kde_names == ["signal", "template"] and len(self.kde_names) == 2, \
```
tdent, line 1844: This is not quite correct, this will fail if the list is `["template", "signal"]` but that would be acceptable. Also if you specify exactly what list kde_names should be and the list has 2 elements, then obviously len() will give 2, so there is no point testing also for the length. 
I suggest using `sorted()` to get the two strings in a specific order for testing.

tdent:
Done apart from the test on the kde_names and addressing codeclimate.

PRAVEEN-mnl:
mchirp_vs_eta
mchirp_vs_chi_eff
eta_vs_chi_eff

These are the plots between mchirp vs eta, mchirp vs chi_eff & eta vs chi_eff. The color scale is the difference between KDE and BBH statistics values.

pycbc/events/stat.py
```diff
@@ -1810,6 +1810,142 @@ def coinc_lim_for_thresh(self, s, thresh, limifo,
...
+        trigs:dict of numpy.ndarrays, h5py group (or similar dict-like object)
+            Dictionary-like object holding single detector trigger information
+
+        Returns
+        -------
+        numpy.ndarray
+            The array of single detector values
+        """
+        try:
+            tnum = trigs.template_num # exists if accessed via coinc_findtrigs
```
spxiwh, line 1884: This doesn't look like it is safe, and I think may not work as intended in either the offline or online modes.

In the online modes the "single" method is called when triggers are added (I think) and the rank_stat_coinc called later with a larger set of triggers. Therefore in that case self.curr_tnum will not correspond to the "stats" sent to logsignalrate. 

In the offline mode, I think that the "single" method is called for all of the single detector triggers, and then logsignalrate is called on all potential coincs. Again, I don't think that self.curr_tnum here then aligns with the "stats" sent to logsignalrate below.

I may have misunderstood this, but I think the template number will need to be added to the dtype of "trigs" and then in here we call ExpFitFgBgNormStatistic.single to get other parameters and then add template_id into this.

tdent: @spxiwh The intention is that this works (and only offline!) in a similar way to the BBH statistic with the curr_mchirp variable. In coinc_findtrigs the triggers and coincs should be processed strictly one template at a time, so we are thinking that it should be sufficient to assign tnum when the singles are read for a given template prior to finding coincs.

PRAVEEN-mnl:
mchirp_vs_eta
mchirp_vs_chi_eff
eta_vs_chi_eff
kde_diff_vs_stat_diff

The first 3 plots are the same as above between _mchirp vs eta_, _mchirp vs chi_eff_ & _eta vs chi_eff__. The color scale is the difference between KDE and BBH statistics values in all these. The 4th plot is between _ln(density ratio) vs stat difference_ and the color scale is ln(mchirp)

tdent (commented):
There are a few changes to formatting and comments that I will make via the web browser before approving, otherwise this looks good

tdent (approved):

PR #4168, opened 2022-11-03
Use solid sphere moment of inertia

dfinstad:
We probably want to approximate the NS as a solid sphere (2/5), and not a spherical shell (2/3)


ahnitz (approved):
@dfinstad You may need to rebase from master

PR #4172, opened 2022-11-07
generate LISA TDI-1.5/2.0 noise in PyCBC

WuShichao:
@ahnitz In this PR, I reimplement LISA TDI PSD for (X,Y,Z) and (A,E,T) channels, and sky/polarization averaged sensitivity curve from the LDC manual. This PR will allow PyCBC to simulate LISA noise. The signal injection is in the next PR.

Currently, I need to do 6 more things:

- [x] test this module with `pycbc.noise.noise_from_psd`
- [x] test this module with `pycbc_condition_strain`
- [x] replace `averaged_lisa_fplus_sq_approx` with a more accurate numerical fit
- [x] add analytical fit for Galatic confusion noise PSD
- [x] add sensitivity curve transformed TDI PSD (X,Y,Z)
- [x] add support for multiple channels' noise generation when using `pycbc_inference`

### DIFF
--- a/bin/pycbc_condition_strain
+++ b/bin/pycbc_condition_strain
@@ -75,6 +75,7 @@ parser.add_argument('--frame-duration', metavar='SECONDS', type=int,
                          'in seconds')
 
 pycbc.strain.insert_strain_option_group(parser)
+pycbc.psd.insert_psd_option_group(parser)
 args = parser.parse_args()
 
 if args.frame_duration is not None and args.frame_duration <= 0:
--- a/bin/pycbc_condition_strain
+++ b/bin/pycbc_condition_strain
@@ -75,7 +75,6 @@ parser.add_argument('--frame-duration', metavar='SECONDS', type=int,
                          'in seconds')
 
 pycbc.strain.insert_strain_option_group(parser)
-pycbc.psd.insert_psd_option_group(parser)
 args = parser.parse_args()
 
 if args.frame_duration is not None and args.frame_duration <= 0:
--- a/pycbc/psd/__init__.py
+++ b/pycbc/psd/__init__.py
@@ -203,7 +203,8 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
     psd_options.add_argument("--fake-strain-extra-args",
                              nargs='+', action=DictOptionAction,
                              metavar='PARAM:VALUE', default={}, type=dict,
-                             help="(optional) Extra arguments passed to the PSD models.")
+                             help="(optional) Extra arguments passed to the "
+                                  "PSD models.")
     # Options specific to XML PSD files
     psd_options.add_argument("--psd-file-xml-ifo-string",
                              help="If using an XML PSD file, use the PSD in "
--- a/pycbc/psd/__init__.py
+++ b/pycbc/psd/__init__.py
@@ -84,7 +84,7 @@ def from_cli(opt, length, delta_f, low_frequency_cutoff,
         # PSD from lalsimulation or file
         if opt.psd_model:
             psd = from_string(opt.psd_model, length, delta_f, f_low,
-                              **opt.extra_args)
+                              **opt.psd_extra_args)
         elif opt.psd_file or opt.asd_file:
             if opt.asd_file:
                 psd_file_name = opt.asd_file
--- a/pycbc/psd/__init__.py
+++ b/pycbc/psd/__init__.py
@@ -187,6 +187,11 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
     psd_options.add_argument("--psd-model",
                              help="Get PSD from given analytical model. ",
                              choices=get_psd_model_list())
+    psd_options.add_argument("--fake-strain-extra-args",
+                             nargs='+', action=DictOptionAction,
+                             metavar='PARAM:VALUE', default={}, type=float,
+                             help="(optional) Extra arguments passed to "
+                             "the PSD models.")
     psd_options.add_argument("--psd-file",
                              help="Get PSD using given PSD ASCII file")
     psd_options.add_argument("--asd-file",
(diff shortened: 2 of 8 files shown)


ahnitz:
(1) hook up your functions so they are accesible form pycbc.psd.from_string

ahnitz:
(2) your functions should follow the standard argument layout of psd functions (see any other psd)

ahnitz:
@WuShichao Address the CC issues and the prior questions and then I think this nearly ready to merge.

WuShichao:
@ahnitz Thanks! I will update this PR during this week's workshop.

pycbc/psd/__init__.py
```diff
@@ -27,7 +28,7 @@
 from pycbc.types import ensure_one_opt, ensure_one_opt_multi_ifo
 
 def from_cli(opt, length, delta_f, low_frequency_cutoff,
-             strain=None, dyn_range_factor=1, precision=None):
+             strain=None, dyn_range_factor=1, precision=None, **kwargs):
```
ahnitz, line 31: Should this just get the option from opt like other things do?  I think in general it is better to be explicit, but in this case, we've somewhat set a convention and it would be better to not mix conventions.

ahnitz: I think the difference here with things like length, df, and flow is that those are required for any psd. The options and extra kwargs are more like the choice of psd type which we have already set up to take from the command line. Since you are adding the option to the command line anyway, it would probably be most consistent to use the options directly.

pycbc/types/optparse.py
```diff
@@ -214,6 +215,64 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+            required=required,
+            help=help,
+            metavar=metavar)
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        # Again this is modified from the standard argparse 'append' action
+        err_msg = "Issue with option: %s \n" %(self.dest,)
+        err_msg += "Received value: %s \n" %(' '.join(values),)
+        if getattr(namespace, self.dest, None) is None:
+            setattr(namespace, self.dest, DictWithDefaultReturn())
```
ahnitz, line 260: Do we want to use DictWithDefaultReturn here? or just populate a regular dictionary?

ahnitz (changes requested):
See, comments. I think besides a few minor style things, this looks like it might be the right thing. I'd like confirmation with an example / plot posted here to demonstrate that the PR works after addressing the comments. Also, please fix the cc issues.

pycbc/psd/__init__.py
WuShichao: @ahnitz Thanks for this suggestion! I just realized that our extra options are included in `opt` if I understand it correctly.

pycbc/types/optparse.py
```diff
@@ -214,6 +215,64 @@ def __call__(self, parser, namespace, values, option_string=None):
                 raise ValueError(err_msg)
         setattr(namespace, self.dest, items)
 
+class ExtraArgsOptionAction(argparse.Action):
```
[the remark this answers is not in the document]
WuShichao: @ahnitz Makes sense. Let me change the name.

WuShichao:
@ahnitz Poke :)

pycbc/psd/__init__.py
```diff
@@ -82,7 +82,11 @@ def from_cli(opt, length, delta_f, low_frequency_cutoff,
     if (opt.psd_model or opt.psd_file or opt.asd_file):
         # PSD from lalsimulation or file
         if opt.psd_model:
-            psd = from_string(opt.psd_model, length, delta_f, f_low)
+            if hasattr(opt, 'fake_strain_extra_args'):
+                psd = from_string(opt.psd_model, length, delta_f, f_low,
+                                  opt.fake_strain_extra_args)
```
ahnitz, line 87: This should be
(1) Why do we need an if statement here? Can you set a default of an empty dictionary?

(2)
```
**opt.fake_strain_extra_args)
```

pycbc/strain/strain.py
```diff
@@ -527,6 +535,10 @@ def insert_strain_option_group(parser, gps_times=True):
     data_reading_group.add_argument("--fake-strain-sample-rate",
                 default=16384, type=float,
                 help="Sample rate of the fake data generation")
+    data_reading_group.add_argument("--fake-strain-extra-args",
+                nargs='+', action=DictOptionAction,
+                metavar='PARAM:VALUE', type=str,
```
ahnitz, line 540: Can you add empty dict as a default here?

ahnitz:
Also, don't forget to add the option to https://github.com/gwastro/pycbc/blob/f1a8c6ea0034b6adcaba728da58f827670eb6df7/pycbc/strain/strain.py#L631

ahnitz:
Add a multi-detector version. This may require adding a multi-detector option parser that can handle a dictionary as an element.

pycbc/psd/__init__.py
```diff
@@ -82,7 +83,8 @@ def from_cli(opt, length, delta_f, low_frequency_cutoff,
     if (opt.psd_model or opt.psd_file or opt.asd_file):
         # PSD from lalsimulation or file
         if opt.psd_model:
-            psd = from_string(opt.psd_model, length, delta_f, f_low)
+            psd = from_string(opt.psd_model, length, delta_f, f_low,
+                              **opt.fake_strain_extra_args)
```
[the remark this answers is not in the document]
WuShichao: @ahnitz Let me just name it `extra_args`.

bin/pycbc_condition_strain
```diff
@@ -75,6 +75,7 @@ parser.add_argument('--frame-duration', metavar='SECONDS', type=int,
                          'in seconds')
 
 pycbc.strain.insert_strain_option_group(parser)
+pycbc.psd.insert_psd_option_group(parser)
```
ahnitz, line 78: Why do you have this? You don't even use the corresponding 'from_cli' function that would use the options in this group. Is this a leftover from some test?

WuShichao: I remember the build test will call `pycbc_condition_strain`, if I don't add this, then `opt` in this file doesn't have `extra-args`.

ahnitz: @WuShichao Take another pass through reasoning that through. Why did you get an error? What is the purpose of these option groups? If you never call its corresponding function that uses the options. How can it fail?

ahnitz (approved):

PR #4174, opened 2022-11-08
pycbc_banksim : fix the wrong help message

MPillas:
Change in the help message about the template-file : the script can take an hdf template bank file as input, not only xml file


titodalcanton (approved):
Thanks @MPillas.

PR #4175, opened 2022-11-09
iDQ based veto in PyCBC Live

maxtrevor:
This PR adds the option to create vetoes by thresholding on a data quality timeseries.


maxtrevor:
@titodalcanton

pycbc/frame/frame.py
```diff
@@ -702,10 +723,12 @@ def check_valid(self, values, flag=None):
         """
         if self.valid_on_zero:
             valid = values.numpy() == 0
+        elif self.threshold is not None:
+            valid = values.numpy() > self.threshold
         else:
             if flag is None:
                 flag = self.valid_mask
-            valid = numpy.bitwise_and(values.numpy(), flag) == flag
```
titodalcanton, line 708: Why is this change needed?

maxtrevor: The idq_ok channel should be a binary on/off timeseries of whether to use iDQ's results. It unfortunately gets read in as a float, which causes an error here if not cast to an int.

bin/pycbc_live
```diff
@@ -985,20 +988,45 @@ with ctx:
...
+                if data_reader[ifo].idq is not None:
+                    logging.info("Checking %s's iDQ information", ifo)
+                    start = data_reader[ifo].start_time
+                    times = results[ifo]['end_time']
+                    # if idq_states = 0, idq info was invalid. Don't let
+                    # those triggers get vetoed
+                    idq_states = data_reader[ifo].idq_state.at_times(times)
+                    idq_valid = idq_states == 1
+                    idq_invalid = numpy.logical_not(idq_valid)
+                    invalid_idx = numpy.where(idq_invalid)[0]
```
titodalcanton, line 1014: I think this is preferred than `where()`:
```suggestion
                    invalid_idx = numpy.flatnonzero(idq_invalid)
```

bin/pycbc_live
```diff
@@ -985,20 +988,45 @@ with ctx:
...
+
+                    # for triggers with valid idq info, only keep those with
+                    # idq FAP > threshold
+                    valid_idx = numpy.where(idq_valid)[0]
```
titodalcanton, line 1018: Same as above
```suggestion
                    valid_idx = numpy.flatnonzero(idq_valid)
```

bin/pycbc_live
```diff
@@ -985,20 +988,45 @@ with ctx:
...
-                    if len(results[ifo][key]):
-                        results[ifo][key] = results[ifo][key][idx]
+                if data_reader[ifo].dq is not None:
+                    logging.info("Checking %s's DQ vector", ifo)
+                    start = data_reader[ifo].start_time
+                    times = results[ifo]['end_time']
+                    idx = data_reader[ifo].dq.indices_of_flag(
+                            start, valid_pad, times,
+                            padding=data_reader[ifo].dq_padding)
+                    logging.info('Keeping %d/%d %s triggers',
```
titodalcanton, line 1000:
```suggestion
                    logging.info('Keeping %d/%d %s triggers after DQ flags',
```

bin/pycbc_live
```diff
@@ -985,20 +988,45 @@ with ctx:
...
+                    # for triggers with valid idq info, only keep those with
+                    # idq FAP > threshold
+                    valid_idx = numpy.where(idq_valid)[0]
+                    valid_keep = data_reader[ifo].idq.indices_of_flag(
+                            start, valid_pad, times[valid_idx],
+                            padding=data_reader[ifo].dq_padding)
+                    valid_idx = valid_idx[valid_keep]
+                    idx = numpy.append(invalid_idx, valid_idx)
+
+                    logging.info('Keeping %d/%d %s triggers',
```
titodalcanton, line 1025:
```suggestion
                    logging.info('Keeping %d/%d %s triggers after iDQ',
```

bin/pycbc_live
```diff
@@ -985,20 +988,45 @@ with ctx:
...
+                    # those triggers get vetoed
+                    idq_states = data_reader[ifo].idq_state.at_times(times)
+                    idq_valid = idq_states == 1
+                    idq_invalid = numpy.logical_not(idq_valid)
+                    invalid_idx = numpy.where(idq_invalid)[0]
+
+                    # for triggers with valid idq info, only keep those with
+                    # idq FAP > threshold
+                    valid_idx = numpy.where(idq_valid)[0]
+                    valid_keep = data_reader[ifo].idq.indices_of_flag(
```
titodalcanton, line 1019: I am a bit confused by what is being done here. What is the reasoning behind using `at_times()` for the iDQ status channel and `indices_of_flag()` for the iDQ FAP channel? Where are you actually checking that the iDQ FAP is above the given threshold?

pycbc/frame/frame.py
titodalcanton: Pretty sure that is a bug (or at least a questionable choice of data formats) on the iDQ side. I suppose it is late/hard to get that changed though?

bin/pycbc_live
titodalcanton: Sorry I overlooked the code that answers the second question, but the first one stands.

pycbc/strain/strain.py
```diff
@@ -1558,14 +1562,19 @@ def __init__(self, frame_src, channel_name, start_time,
...
-                                                   increment_update_cache=increment_update_cache)
+            if idq_threshold is None:
+                raise ValueError('If an iDQ channel is provided, a veto threshold must also be provided')
+            self.idq_threshold = idq_threshold
+            self.idq = pycbc.frame.StatusBuffer(frame_src, idq_channel, start_time,
+                                                max_buffer=max_buffer,
+                                                force_update_cache=force_update_cache,
+                                                increment_update_cache=increment_update_cache,
+                                                threshold=self.idq_threshold)
+            self.idq_state = pycbc.frame.StatusBuffer(frame_src, idq_channel, start_time,
```
titodalcanton, line 1573: Should the second argument here not be `idq_state_channel`?

titodalcanton:
@maxtrevor overall comment: I think there is a genuine bug, apart from that this should do what we want. I would like to propose an alternative implementation though. I do not particularly like the idea of having to carry around *two* `StatusBuffer` objects for iDQ and having to worry about that in the main `pycbc_live` script. Have you considered having a single `iDQBuffer` object that carries its two `StatusBuffer` objects as attributes, and knows how to properly treat them as a pair?

pycbc/frame/frame.py
titodalcanton: @maxtrevor you can now revert this change, right?

maxtrevor: Unfortunately not. I changed the code to use the correct idq ok channel but it turns out that channel is being written as a float. Patrick Godwin said that on the iDQ side it is much easier to write frame files with all channels having the same data type and its relatively trivial to cast the channel to an int after reading it. He would prefer not to change this on the idq side right now as review is already ongoing. So I think we are stuck with this casting to int for now at least.

titodalcanton: I see… *sigh*. Bigger comment coming in the main discussion.

titodalcanton:
In light of @maxtrevor's comment about iDQ status channel type, and after more carefully reviewing the current code in `pycbc.frame`, I am now turning my previous comment into a stronger request, and I would prefer not merging until that change is made.

Note that the DataBuffer class has a dtype parameter. It defaults to float, but StatusBuffer changes it to integer when inheriting from DataBuffer. Given that iDQ channels are all float instead (let's ignore whether that is appropriate or not) then using two StatusBuffer objects to hold the iDQ data does not seem the correct approach to me. I think we really want a single iDQBuffer class that carries two float *DataBuffer* attributes for value and status, but does not inherit from DataBuffer itself. We are then free to implement whatever specific logic iDQ uses (based on floats or ints, one or two channels, etc) inside the iDQ-specific class, and StatusBuffer can continue to retain its current meaning as a simpler integer bitmask status information.

Does this make sense?

maxtrevor:
> In light of @maxtrevor's comment about iDQ status channel type, and after more carefully reviewing the current code in `pycbc.frame`, I am now turning my previous comment into a stronger request, and I would prefer not merging until that change is made.
> 
> Note that the DataBuffer class has a dtype parameter. It defaults to float, but StatusBuffer changes it to integer when inheriting from DataBuffer. Given that iDQ channels are all float instead (let's ignore whether that is appropriate or not) then using two StatusBuffer objects to hold the iDQ data does not seem the correct approach to me. I think we really want a single iDQBuffer class that carries two float _DataBuffer_ attributes for value and status, but does not inherit from DataBuffer itself. We are then free to implement whatever specific logic iDQ uses (based on floats or ints, one or two channels, etc) inside the iDQ-specific class, and StatusBuffer can continue to retain its current meaning as a simpler integer bitmask status information.
> 
> Does this make sense?

It does make sense. I think I am most of the way to having this working already. 

One thing occurred to me though. The StatusBuffer object uses dtype int unless told otherwise. So why was it giving me an error unless I explicitly cast to int? It seems like it used a float anyways even though the default dtype was int. This is mostly irrelevant but it doesnt seem to make any sense.

maxtrevor:
@titodalcanton I think this now does everything we want

maxtrevor:
@titodalcanton The current version should not be merged. I need some help with bugfixing asap. I was very confident in the most recent commit until i tried to run a full analysis with it. It runs, but it is spitting out an error ` XLAL Error - XLALFrStreamReadREAL8TimeSeries (LALFrStreamReadTS_source.c:211): Internal function call failed ` over and over again while running. I have not see this with any previous verison of my code and have no clue what is causing it. Troubleshooting the underlying C code is outside of my current skillset.

maxtrevor:
This should be ready to go at this point, I have an analysis that has been running for the last 24 hours and has vetoes active at expected times

titodalcanton:
This seems ok to be merged now. Just a minor final request: do you need to store the iDQ threshold as an attribute of the StrainBuffer? If not, please remove it.

titodalcanton:
…and same comment for the `at_times()` method of DataBuffer.

titodalcanton (approved):

PR #4178, opened 2022-11-11
Minor GraceDB API change

titodalcanton:
`createEvent()` becomes `create_event()`.

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -448,14 +448,20 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
     shutil.move(skymap_fits_name, final_fits_path)
 
     if gracedb_server:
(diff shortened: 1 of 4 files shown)


GarethCabournDavies (changes requested):
This looks sensible, and I can see from https://ligo-gracedb.readthedocs.io/en/gracedb-2.7.8/api.html#ligo.gracedb.rest.GraceDb.create_event that this is the correct usage.

I went through the gdb client code, and checked that:
1. Create_event is exactly the same as createEvent
1. Why it didn't just stop working
 
Turns out it was retro-aliased, so both things are immediately true

Change requested: It does look like all other functions have been changed from camelCase to snake_case as well, so if you could also update writeLabel --> write_label and writeLog --> write_log, that would be great.

titodalcanton:
Thanks Gareth, yeah I noticed because I ran the tox tests and it reported the deprecation warning. Requested change coming in a moment.

GarethCabournDavies (approved):

PR #4179, opened 2022-11-14
add get_td_det_waveform_from_fd_det()

WuShichao:
@ahnitz This PR is for LISA time-domain injection.

### DIFF
--- a/pycbc/waveform/waveform.py
+++ b/pycbc/waveform/waveform.py
@@ -775,6 +775,65 @@ def get_td_waveform_from_fd(rwrap=0.2, **params):
                                            params['f_lower']))
     return hp, hc
 
+def get_td_det_waveform_from_fd_det(template=None, rwrap=0.2, **params):
+    """ Return time domain version of fourier domain approximant which
+    includes detector response, with padding and tapering at the start
+    of the waveform.
+
+    Parameters
+    ----------
+    rwrap: float
+        Cyclic time shift parameter in seconds. A fudge factor to ensure
+        that the entire time series is contiguous in the array and not
+        wrapped around the end.
+    params: dict
+        The parameters defining the waveform to generator.
+        See `get_fd_det_waveform_sequence`.
+
+    Returns
+    -------
+    dict
+        The detector-frame waveform (with detector response) in time
+        domain. Keys are requested data channels.
+    """
+    # determine the duration to use
+    full_duration = duration = get_waveform_filter_length_in_time(
+                                "IMRPhenomD", **params)
+    nparams = params.copy()
+
+    while full_duration < duration * 1.5:
+        full_duration = get_waveform_filter_length_in_time(
+            "IMRPhenomD", **nparams)
+        nparams['f_lower'] *= 0.99
+
+    if 'f_ref' not in nparams:
+        nparams['f_ref'] = params['f_lower']
+
+    # factor to ensure the vectors are all large enough. We don't need to
+    # completely trust our duration estimator in this case, at a small
+    # increase in computational cost
+    fudge_duration = (max(0, full_duration) + .1 + rwrap) * 1.5
+    fsamples = int(fudge_duration / params['delta_t'])
+    N = pnutils.nearest_larger_binary_number(fsamples)
+    fudge_duration = N * params['delta_t']
+
+    nparams['delta_f'] = 1.0 / fudge_duration
+    wfs = {}
+    for ifo in self.data:
+        wfs = wfs | get_fd_det_waveform_sequence(ifos=ifo, **nparams)
+
+    # Resize to the right sample rate
+    tsize = int(1.0 / params['delta_t'] /  nparams['delta_f'])
+    fsize = tsize // 2 + 1
+    for ifo in self.data:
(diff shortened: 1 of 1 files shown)


pycbc/waveform/waveform.py
```diff
@@ -775,6 +775,73 @@ def get_td_waveform_from_fd(rwrap=0.2, **params):
...
+
+    Returns
+    -------
+    dict
+        The detector-frame waveform (with detector response) in time
+        domain. Keys are requested data channels.
+    """
+    kwds = props(template, **params)
+    # determine the duration to use
+    full_duration = duration = pnutils.get_imr_duration(
```
ahnitz, line 801: We shouldn't hardcode this. What we need is to have a length estimator for the approximant and for it to be advertised using the plugin. @WuShichao Can you add the requisite function (maybe Imrphenomd used here is fine!) in the bbhx plugin and update it's setup.py so that it advertises a length estimator? That way you don't need to hardcode here. That may not be appropriate for other approximants.

pycbc/waveform/waveform.py
```diff
@@ -775,6 +775,73 @@ def get_td_waveform_from_fd(rwrap=0.2, **params):
...
+    # factor to ensure the vectors are all large enough. We don't need to
+    # completely trust our duration estimator in this case, at a small
+    # increase in computational cost
+    fudge_duration = (max(0, full_duration) + .1 + rwrap) * 1.5
+    fsamples = int(fudge_duration / kwds['delta_t'])
+    N = pnutils.nearest_larger_binary_number(fsamples)
+    fudge_duration = N * kwds['delta_t']
+
+    nparams['delta_f'] = 1.0 / fudge_duration
+    wfs = get_fd_det_waveform_sequence(**nparams)
```
ahnitz, line 826: Does it seem right to call a sequency version here? Shouldn't we be calling something like get_fd_det_waveform? If we don't have that, shouldn't we? Should the plugin advertise support for that? (and maybe we need to add that to the list of interfaces if we don't already have it). 

What do you think? Wouldn't that be more consistant?

ahnitz (changes requested):
See comments. 

Otherwise, it would be good to have an example to demonstrate the interface works. Maybe a couple line example and post the plot?

ahnitz:
@WuShichao Rebase this from your previous PR

pycbc/inject/inject.py
```diff
@@ -637,6 +650,7 @@ def supported_approximants():
         for d in [waveform.waveform.td_wav, waveform.waveform.fd_wav]:
             for key in d:
                 all_apprxs.extend(d[key])
+        all_apprxs.extend(waveform.waveform.fd_det)
```
ahnitz, line 653: Why is this a separate line?

ahnitz:
@WuShichao You should probably leave out the multiband changes for a separate PR as you are still testing those, unless you've shown those to work.

pycbc/waveform/waveform.py
```diff
@@ -769,12 +770,120 @@ def get_td_waveform_from_fd(rwrap=0.2, **params):
     hp = hp.cyclic_time_shift(-rwrap)
     hc = hc.cyclic_time_shift(-rwrap)
 
-    hp = wfutils.fd_to_td(hp, left_window=(nparams['f_lower'],
-                                           params['f_lower']))
-    hc = wfutils.fd_to_td(hc, left_window=(nparams['f_lower'],
-                                           params['f_lower']))
+    hp = wfutils.fd_to_td(hp, delta_t=params['delta_t'],
```
ahnitz, line 773: Why are you giving 'delta_t' to these? I don't think these should be required.

pycbc/waveform/waveform.py
```diff
@@ -769,12 +770,120 @@ def get_td_waveform_from_fd(rwrap=0.2, **params):
...
+    else:
+        required = parameters.fd_required
+    check_args(input_params, required)
+    return wav_gen(**input_params)
+
+get_fd_det_waveform.__doc__ = get_fd_det_waveform.__doc__.format(
+    params=parameters.fd_waveform_params.docstr(prefix="    ",
+           include_label=False).lstrip(' '))
+
+def get_td_det_waveform_from_fd_det(template=None, rwrap=0.2, **params):
```
ahnitz, line 814: Does the implementation need to be duplicated from get_td_waveform_from_fd? Can't you share most of the code?

ahnitz:
@WuShichao I think this PR is almost ready to go, but see the remaining three comments.

pycbc/inject/inject.py
WuShichao: @ahnitz This needs to be a separate line. Because I follow `fd_det_squence`, just set `fd_det` as {}, not a `_scheme.ChooseBySchemeDict()` as `fd_wav`

WuShichao:
@ahnitz This PR is ready for your review. I combine `get_td_waveform_from_fd` and `get_td_det_waveform_from_fd_det` by calling the same base function `_base_get_td_waveform_from_fd`, but just expose those two wrapper functions to users. The failed and canceled jobs are caused by url timeout issue.

ahnitz (approved):

PR #4181, opened 2022-11-16
Fix for LIGOLW API change (gz → compress)

titodalcanton:
Fixing some seldom used code for a fairly old API change.


pannarale (changes requested):
Hi @titodalcanton, please check the code climate issue.

pannarale (approved):

PR #4182, opened 2022-11-16
Fix error in `pycbc_optimal_snr`

titodalcanton:
Fixes the following error when using `pycbc_optimal_snr` with input and output both in XML format:
```
  File "[…]/bin/pycbc_optimal_snr", line 299, in <module>
    new_inj_table = lsctables.New(lsctables.SimInspiralTable,
  File "[…]/lib/python3.9/site-packages/ligo/lw/lsctables.py", line 153, in New
    new.appendColumn(name)
  File "[…]/lib/python3.9/site-packages/ligo/lw/table.py", line 719, in appendColumn
    raise ligolw.ElementError("invalid Column '%s' for Table '%s'" % (name, self.Name))
ligo.lw.ligolw.ElementError: invalid Column 'process_id' for Table 'sim_inspiral' 
```

### DIFF
--- a/bin/pycbc_optimal_snr
+++ b/bin/pycbc_optimal_snr
@@ -30,12 +30,13 @@ import h5py
 import pycbc
 import pycbc.inject
 import pycbc.psd
+import pycbc.version
 from ligo.lw import utils as ligolw_utils
+from ligo.lw import lsctables
 from pycbc.filter import sigma, make_frequency_series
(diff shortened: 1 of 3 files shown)


tdent:
I didn't think people would still want this mode of operation ...

titodalcanton:
@tdent I was slightly surprised too, but apparently that is the case…

pannarale (approved):

PR #4184, opened 2022-11-17
Change logging to indicate UTC offset

GarethCabournDavies:
Update logging to be in UTC - not being in UTC causes some weirdness for logs when spanning a daylight savings change.

This should probably be advertised more generally, so I will ask for "any objection" in the pycbc-code slack channel

### DIFF
--- a/pycbc/__init__.py
+++ b/pycbc/__init__.py
@@ -55,7 +55,7 @@ class LogFormatter(logging.Formatter):
     def formatTime(self, record, datefmt=None):
         ct = self.converter(record.created).astimezone()
         t = ct.strftime("%Y-%m-%dT%H:%M:%S")
-        s = "%s.%03d" % (t, record.msecs)
+        s = f"{t}.{record.msecs:03d}"
         s += ct.strftime('%z')
         return s
 
--- a/pycbc/__init__.py
+++ b/pycbc/__init__.py
@@ -46,10 +46,14 @@ except:
 __version__ = pycbc_version
 
 
-_default_format = '%(asctime)s %(message)s'
-
-
 class LogFormatter(logging.Formatter):
+    """
+    Format the logging appropriately
+    This will return the log time in the ISO 6801 standard,
+    but with millisecond precision
+    https://en.wikipedia.org/wiki/ISO_8601
+    e.g. 2022-11-18T09:53:01.554+00:00
+    """
     converter = dt.fromtimestamp
 
     def formatTime(self, record, datefmt=None):
@@ -60,7 +64,7 @@ class LogFormatter(logging.Formatter):
         return s
 
 
-def init_logging(verbose=False, format=_default_format):
+def init_logging(verbose=False, format='%(asctime)s %(message)s'):
     """Common utility for setting up logging in PyCBC.
 
     Installs a signal handler such that verbosity can be activated at
--- a/pycbc/__init__.py
+++ b/pycbc/__init__.py
@@ -32,6 +32,7 @@ warnings.filterwarnings("ignore", message="numpy.ufunc size changed")
 import logging
 import random
 import string
+import time
 
 try:
     # This will fail when pycbc is imported during the build process,
@@ -45,7 +46,7 @@ except:
 __version__ = pycbc_version
 
 
-def init_logging(verbose=False, format='%(asctime)s %(message)s'):
+def init_logging(verbose=False, format='%(asctime)s UTC: %(message)s'):
     """Common utility for setting up logging in PyCBC.
 
     Installs a signal handler such that verbosity can be activated at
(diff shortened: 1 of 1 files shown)


titodalcanton:
I am not opposed to this in principle, in fact it would solve a difficulty with PyCBC Live's lag plots. However, I am not sure forcing the timestamps to UTC this is the most desirable way. Is there a way to add the timezone offset instead, as in e.g. https://en.wikipedia.org/wiki/ISO_8601? Note that we do want to keep the microsecond precision too and I am not sure what the ISO 8601 says about that.

GarethCabournDavies:
I've update to use the `2022-11-17T06:26:53.066-0800` format.

Should we have the logging level included?

pycbc/__init__.py
```diff
@@ -45,7 +46,21 @@
 __version__ = pycbc_version
 
 
-def init_logging(verbose=False, format='%(asctime)s %(message)s'):
+_default_format = '%(asctime)s: %(message)s'
+
+
+class Formatter(logging.Formatter):
```
titodalcanton, line 52: Use a more descriptive name for the class, maybe LogFormatter?

pycbc/__init__.py
```diff
@@ -45,7 +46,21 @@
...
+
+    def formatTime(self, record, datefmt=None):
+        ct = self.converter(record.created).astimezone()
+        t = ct.strftime("%Y-%m-%dT%H:%M:%S")
+        s = "%s.%03d" % (t, record.msecs)
+        s += ct.strftime('%z')
+        return s
+
+
+def init_logging(verbose=False, fmt=_default_format):
```
titodalcanton, line 63: What motivates the change from `format` to `fmt`? This requires other changes in the code base (and is breaking the tests).

pycbc/__init__.py
```diff
@@ -45,7 +46,21 @@
...
+_default_format = '%(asctime)s %(message)s'
+
+
+class LogFormatter(logging.Formatter):
+    converter = dt.fromtimestamp
+
+    def formatTime(self, record, datefmt=None):
+        ct = self.converter(record.created).astimezone()
+        t = ct.strftime("%Y-%m-%dT%H:%M:%S")
+        s = "%s.%03d" % (t, record.msecs)
```
titodalcanton, line 58: F-strings please :)
```suggestion
        s = f"{t}.{record.msecs:03d}"
```
Is `record.msecs` really an integer?

pycbc/__init__.py
GarethCabournDavies: I didn't want to overwrite the format builtin, but I now realise thats not needed

pycbc/__init__.py
GarethCabournDavies: I used this to set it: https://stackoverflow.com/questions/6290739/python-logging-use-milliseconds-in-time-format/7517430#7517430

And it looks like it based on:
https://docs.python.org/3/library/logging.html#logrecord-attributes

But apparently it is a float

pycbc/__init__.py
```diff
@@ -45,7 +46,21 @@
 __version__ = pycbc_version
 
 
-def init_logging(verbose=False, format='%(asctime)s %(message)s'):
+_default_format = '%(asctime)s %(message)s'
+
+
+class LogFormatter(logging.Formatter):
```
titodalcanton, line 52: Could you please add a docstring to describe why we use this custom class, indicating the example timestamp format we discussed?

pycbc/__init__.py
```diff
@@ -45,7 +46,21 @@
 __version__ = pycbc_version
 
 
-def init_logging(verbose=False, format='%(asctime)s %(message)s'):
+_default_format = '%(asctime)s %(message)s'
```
titodalcanton, line 49: I actually suggest keeping the default string in the function signature, because that makes the default show up explicitly in the documentation. I think that would be lost if we do this change.

titodalcanton:
Just running a quick test of this, I see the format is actually
```
2022-11-18T11:20:40.768+0000
```
i.e. no colon to separate the hour and minutes of the offset. Can it be included easily by just switching the timezone formatter?

titodalcanton:
I also get a lot of inconsistency where some code still uses the old format, and some code uses the new one. This happens even within the same executable, for example `pycbc_optimize_snr`. Is there a way to make the custom format default for every module that has been imported or is going to be?

GarethCabournDavies:
> Just running a quick test of this, I see the format is actually
> 
> ```
> 2022-11-18T11:20:40.768+0000
> ```
> 
> i.e. no colon to separate the hour and minutes of the offset. Can it be included easily by just switching the timezone formatter?

Not "easily",  but I put in a hack

titodalcanton:
Ok I think this behaves correctly during a daylight saving change:
```
2022-10-30T02:59:54.912+02:00 Some info
2022-10-30T02:59:57.815+02:00 Some info
2022-10-30T02:00:00.719+01:00 Some info
2022-10-30T02:00:03.622+01:00 Some info
```

titodalcanton (approved):
I have verified that the signal mechanism and the `format` kwarg continue to work and that a daylight saving change is handled correctly, so this looks good to me. My only complaint is to the logging module for having neither this as the default timestamp format, nor an easier way to implement it 😠

PR #4185, opened 2022-11-17
Banksim jobs 

MPillas:
Add a way to read the approximant from an injection file in hdf format in the `pycbc_banksim` script
Fix the .`segs ` issue and write the properties of the input bank like : https://github.com/gwastro/pycbc/blob/07715891b47c77bb126803e94ca7acc1abaad91a/pycbc/workflow/psdfiles.py#L123


spxiwh (approved):
Thanks!

Issue #3526, opened 2020-11-13
Workflow generation command should be better accessible in offline search


GarethCabournDavies:
Currently the way to find the workflow generation command is to look at the workflow generation log and to click 'command line' on the results pages

As this is an important part of being able to see what the setup of the run was at a later point, we should give this its own page or file

PR #4186, opened 2022-11-21
Results page generation command

GarethCabournDavies:
Add the command used to generate the workflow to its own page (rather than remaining in the metadata of other files only)

### DIFF
--- a/bin/workflows/pycbc_make_coinc_search_workflow
+++ b/bin/workflows/pycbc_make_coinc_search_workflow
@@ -737,6 +737,9 @@ create_versioning_page(rdir['workflow/version'], container.cp)
 log_file_html = wf.File(workflow.ifos, 'WORKFLOW-LOG', workflow.analysis_time,
                         extension='.html', directory=rdir['workflow'])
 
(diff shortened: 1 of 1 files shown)


ahnitz:
@GarethCabournDavies You should see the 'results' zip file here 

https://github.com/gwastro/pycbc/actions/runs/3516714213

That should contain the web page so you can just download/extract that and view offline to see if it worked as you wanted.

GarethCabournDavies:
Thanks! I managed to get them displaying and it looked a bit off, so it now looks a bit better

GarethCabournDavies:
This resolves #3526

GarethCabournDavies:
Example of the updated output:
image

ahnitz:
@GarethCabournDavies Should this be a section in the "well" i.e. at the top and always visible or in the B section where it would be shaded by default? The latter you can do by just not including the file in any layout well file.

GarethCabournDavies:
I think it makes sense in the well - the log is straight after it, so to me it makes sense for the workflow section to show Command then Log for the workflow generation

GarethCabournDavies:
I'm going to be away now until the new year, but I am happy for edits to be made and this to be merged when ready

ahnitz (approved):

PR #4187, opened 2022-11-22
Start GraceDB uploads and setup_optimize_snr process in threads

xangma:
These changes allow the Rank0 MPI process of pycbc_live to not wait for the GraceDB event upload process to finish before continuing to process data. The setup_optimize_snr function depends on the gid returned from the GraceDB event upload interation, so should be started after the upload is finished. As such, it has also been moved to be within the new upload_in_thread function.

We should have a conversation about [daemons](https://superfastpython.com/stop-daemon-thread/). Currently if the main thread dies, all spawned threads will also be terminated, meaning a thread could get interrupted mid-upload unless we program the main thread to wait. Is this what we want?


ahnitz:
@xangma Interesting patch. I'm just curious. Is this effective despite the GIL? I guess it could be if most of the time is spent in system / non-python code which. Is that right? Is there a reason no simply to fork?

xangma:
Thanks! Yes, from my initial testing it is effective despite the GIL - however we are trying to solve the quite specific issue of multiple uploads to GraceDB in quick succession. Currently Rank0 spends some time waiting on GraceDB responses and setting up the request session again, so this allows it to push the matched filtering on while it waits. 

One of the next things on my list to investigate is keeping the request session object around so the connections isn't reestablished every upload.

When you ask about the potential of forking - do you mean into a new MPI worker/process group?

titodalcanton:
@xangma looks like you need to rebase on current master to incorporate recent changes to the same code.

@ahnitz we did consider a few ways to do this (threading, forking, using a separate MPI rank etc). We thought threading would be a good compromise between extra complexity and not introducing extra latency, but I am all hears for opinions on this.

ahnitz:
@xangma @titodalcanton Ok, sounds good. 

I was curious if you had tried forking as that in principle should be nearly zero cost on linux and you'd avoid the GIL issue as what you'd do is create a fork and have one path continue on with processing data, while the other does the upload then exits. 

I had considered adding this long time ago, but one issue which I never checked about is if this would cause any state information problems. I think it would be OK, since the upload fork doesn't need to access MPI sockets, but these things can sometimes have subtle bugs.

titodalcanton:
I cannot recall at the moment how threading vs forking was chosen, but yeah, I cannot see why forking should not work as well. We are already using the multiprocessing module for background estimation, so that should not create problems.

bin/pycbc_live
```diff
@@ -359,6 +361,55 @@ class LiveEventManager(object):
...
+    def create_gdb(self):
+        """
+        Function to create GraceDB session.
+        """
+        from ligo.gracedb.rest import GraceDb
+        logging.info('Creating GraceDB session')
+        # Set up dict for args that reload the certificate if it will expire in the reload_buffer time
+        # These args are documented here: https://ligo-gracedb.readthedocs.io/en/latest/api.html#ligo.gracedb.rest.GraceDb
+        # Because we do not change any of the request session values when running the code, it should remain thread safe. 
+        gdbargs = {'reload_certificate': True, 'reload_buffer': 300}
```
titodalcanton, line 373: I am not sure what the impact of reloading the cert is (probably very small) but 300 s seems a bit too often. I propose an hourly reload.

bin/pycbc_live
```diff
@@ -359,6 +361,55 @@ class LiveEventManager(object):
...
+        self.gracedb = GraceDb(args.gracedb_server, **gdbargs) if args.gracedb_server else GraceDb(**gdbargs)
```
titodalcanton, line 374: Avoid using `args` here, as it is not an argument of the method. I think `gracedb_server` is one of the attributes of `self`.

titodalcanton: You can also simplify this line by (optionally) putting the server URL in `gdbargs` instead.

pycbc/io/live.py
```diff
@@ -248,21 +251,31 @@ def __init__(self, coinc_ifos, ifos, coinc_results, **kwargs):
...
                                                     kwargs['mc_area_args'])
+            kwargs['hasmassgap_args'] = copy.deepcopy(kwargs['mc_area_args'])
+            kwargs['hasmassgap_args']['mass_gap'] = True
+            kwargs['hasmassgap_args']['mass_bdary']['ns_max'] = 3.0
+            kwargs['hasmassgap_args']['mass_bdary']['gap_max'] = 5.0
+            self.hasmassgap = calc_probabilities(
+                                  coinc_inspiral_row.mchirp,
+                                  coinc_inspiral_row.snr,
+                                  min(eff_distances),
+                                  kwargs['hasmassgap_args'])['Mass Gap']
```
titodalcanton, line 269: This looks like Veronica's recent change. Rebase on master to clean this up.

bin/pycbc_live
xangma: Ah, the `reload_buffer` function works slightly differently to that. When it's set to 300s it means that whenever the GraceDb client interacts with the server, it checks whether the certificate expiration time is within the next 300s (or if you've passed it), and if so, reloads the certificate from disk before starting your server interaction. I haven't been able to test that the cert reloads yet, but I can do a test asap?

I'm thinking of doing this by:
1) Creating a new cert with a short lifespan.
2) Starting python, creating the connection object, and connecting to GraceDb with it.
3) Waiting until the cert expires, then recreating it on disk.
4) Trying to connect to GraceDb again without recreating the connection object.

I'm pretty sure they have done a very similar test in the development of the GraceDb Client package to ensure the code works :-)

bin/pycbc_live
xangma: I definitely agree, and I was originally going to assign them during the `__init__` of the class e.g. `self.gracedb_server = args.gracedb_server`, however, I did notice that `args` is used in a bunch of other class methods already without being explicitly passed in to those either? I'm unsure which convention we want to adopt?

pycbc/io/live.py
xangma: I am good at many things, it seems git is not one of them.
I think I've fixed it, but I can't seem to update the PR? Is the correct button available to you?

bin/pycbc_live
titodalcanton: Ok. Then it seems we can test by simply letting the analysis run for one week, and making sure no uploads fail.

bin/pycbc_live
titodalcanton: That is right, a lot of the code does pass around `args`, however I would like *new* code to stop doing that.

pycbc/io/live.py
titodalcanton: Let's discuss over Slack about this.

bin/pycbc_live
xangma: Yes I agree. Would it be ok to leave this as it is, then I can do a separate branch fixing it for all of the methods in that class?

titodalcanton: Fair enough.

pycbc/io/live.py
```diff
@@ -527,7 +533,7 @@ def upload(self, fname, gracedb_server=None, testing=True,
             except Exception as exc:
                 logging.error('Failed to upload p_astro file for %s', gid)
                 logging.error(str(exc))
-
+
```
titodalcanton, line 536: Please remove this extra whitespace.

pycbc/io/live.py
```diff
@@ -288,10 +291,14 @@ def save(self, fname):
...
-        ligolw_utils.write_filename(self.outdoc, fname, compress='auto')
+        kwargs = {}
+        if threading.current_thread() is not threading.main_thread():
+            # avoid an error due to no ability to do signal handling in threads
+            kwargs['trap_signals'] = None
+        ligolw_utils.write_filename(self.outdoc, fname, compress='auto', **kwargs)
 
         save_dir = os.path.dirname(fname)
-
+
```
titodalcanton, line 301: Please remove the extra whitespace.

pycbc/io/live.py
xangma: Done :-)

pycbc/io/live.py
xangma: Done :-)

titodalcanton:
@xangma looks like you need a careful rebase on master (possible conflict with #4219).

titodalcanton:
I am going to let this run for the weekend on the IJCLab cluster.

xangma:
> I am going to let this run for the weekend on the IJCLab cluster.

Yep, after all of these rebases etc. I think this is a good idea. I'm also running it on an ldas dev node for a few hours today.

titodalcanton:
This now seems to work at IJCLab. It does upload candidates, and I can see that the search carries on while the uploader thread is uploading things. I am puzzled by the time between the uploads of the various bits though, sometimes tens of seconds. I am going to look at a few of the uploaded events in detail.

titodalcanton (approved):
I have no further comments on the diff, and this seems to work how it should.

PR #4190, opened 2022-11-24
fix corner case in live event followup

tdent:
Another iteration of renaming / reorganizing lists of ifos, addressing https://github.com/gwastro/pycbc/issues/4189


titodalcanton:
Is the `ifos` argument to the constructor of CandidateForGraceDB still used after this change?

titodalcanton:
Ok so the code now filters `ifos` to only those detectors that explicitly have an end time estimate. This should indeed fix the problem.

titodalcanton (approved):

PR #4194, opened 2022-11-29
[WIP] Enable new waveform interface

bhooshan-gadre:
This will allow us to use new LVK waveform interface.

This requires different lalsuite branch as well. 

#https://git.ligo.org/pycbc/action-items/-/issues/11

### DIFF
--- a/pycbc/waveform/waveform.py
+++ b/pycbc/waveform/waveform.py
@@ -42,6 +42,8 @@ from .spa_tmplt import spa_tmplt, spa_tmplt_norm, spa_tmplt_end, \
                       spa_tmplt_precondition, spa_amplitude_factor, \
                       spa_length_in_time
 
+import gwsignal
+
 class NoWaveformError(Exception):
     """This should be raised if generating a waveform would just result in all
     zeros being returned, e.g., if a requested `f_final` is <= `f_lower`.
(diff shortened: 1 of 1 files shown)


pycbc/waveform/waveform.py
```diff
@@ -42,6 +42,8 @@
                       spa_tmplt_precondition, spa_amplitude_factor, \
                       spa_length_in_time
 
+import gwsignal
```
ahnitz, line 45: @bhooshan-gadre Maybe whatever this package is importing should just export the standard pycbc waveform interface? We have a plugin architecture so packages can advertise support for the pycbc waveform interfaces. 

https://pycbc.org/pycbc/latest/html/waveform_plugin.html

bhooshan-gadre: @ahnitz Soon, even the lalsuite waveform infrastructure will be modified. So apart from the new waveform, support for lalsuite-based waveforms may break in upcoming releases. 

I will try to use pycbc waveform interface as well and then we can decide.

bhooshan-gadre: We also need special lalsuite (with hash 391604dba2) for this.

tdent:
It seems you will want to fetch master, rebase wrt master (fixing conflicts etc) and then do git push origin HEAD -f since the rebase is going to rewrite the branch history.

bhooshan-gadre:
closing this in favour of #4295

PR #4195, opened 2022-11-29
changed name of multi p astro file

tdent:
Seems simple.

### DIFF
--- a/pycbc/io/live.py
+++ b/pycbc/io/live.py
@@ -286,7 +286,7 @@ class CandidateForGraceDB(object):
 
         # Save multi-cpt p astro as json
         if self.astro_probs is not None:
-            self.multipa_file = 'pycbc_p_astro.json'
+            self.multipa_file = 'pycbc.p_astro.json'
             with open(self.multipa_file, 'w') as multipaf:
                 json.dump(self.astro_probs, multipaf)
             logging.info('Multi p_astro file saved as %s', self.multipa_file)
--- a/pycbc/io/live.py
+++ b/pycbc/io/live.py
@@ -456,7 +456,7 @@ class CandidateForGraceDB(object):
                 gracedb.write_log(
                     gid, 'Multi-component p_astro JSON file upload',
                     filename=self.multipa_file,
-                    tag_name=['em_follow']
+                    tag_name=['p_astro']
                 )
                 logging.info('Uploaded multi p_astro for %s', gid)
             except Exception as exc:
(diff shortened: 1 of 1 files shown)


titodalcanton:
The requirement document seems to indicate `pycbc.p_astro.json` though, with a period.

titodalcanton:
This change also seems problematic because all candidates will now create a file with the same name, overwriting the p_astro file from previous candidates.

tdent:
> This change also seems problematic because all candidates will now create a file with the same name, overwriting the p_astro file from previous candidates.

I know, this is a generic issue with the pycbc model of 'save all files locally and also upload them to gracedb' - if the main method for storing information is considered to be the gracedb event then what we do with files in the pycbc live account is not important.

Any concrete suggestion?

titodalcanton:
> …what we do with files in the pycbc live account is not important.

In my view, it is essential that we keep copies of the candidate files on a permanent and independent storage, since we do not control what really happens with GraceDB. Uploads can fail, and candidates uploaded to gracedb-playground are periodically deleted.

For a concerete suggestion, this would be solved by #4197.

tdent:
> > …what we do with files in the pycbc live account is not important.
> 
> In my view, it is essential that we keep copies of the candidate files on a permanent and independent storage, since we do not control what really happens with GraceDB. Uploads can fail, and candidates uploaded to gracedb-playground are periodically deleted.
> 
> For a concerete suggestion, this would be solved by #4197.

the first part of the quote says, _if_ we consider gracedb as the main storage, which might be what other pipelines are doing, and I'm guessing is what other bits of the collaboration are expecting, but is not necessarily what we want.

titodalcanton:
In my opinion we put this on hold until #4197 (or equivalent) is addressed.

titodalcanton:
We also need the `p_astro` tag on the file, by the way.

tdent:
@titodalcanton any changes needed here?

titodalcanton:
Probably just a rebase on master.

titodalcanton:
Sorry, you also need to make sure the file ends up in the right directory for the candidate. Use `os.path.basename()` to figure out the path to the directory containing the XML file, then use `os.path.join()` to combine the directory path with the new file name.

tdent:
> Use os.path.basename() to figure out the path to the directory containing the XML file, then use os.path.join() to combine the directory path with the new file name.

I think you mean os.path.dirname() ?

tdent:
Changes here are being rolled into #4205, closing

PR #4201, opened 2022-12-04
Add function to use only time around peaks for sky / time marginalization

ahnitz:
this adds a function to determine what time ranges to use for each detector based on the observed SNR peak of a reference signal's SNR time series. This speeds up marginalization over sky location.

PR #4202, opened 2022-12-04
Add new marginalize time example and fixes for when few monte-carlo points are found

ahnitz:
This adds a fix to the case where no valid sky locations are identified using the SNR based draws. If only a small fraction of guessed points are valid, it reverts to prior drawn values instead of potentially producing *zero* (which of course causes an error which this patch not avoids). 

This also adds a new software injection based example of the marginalized_time model. 

Depends on #4201 

@KoustavChandra This PR should hopefully address the issue you were seeing with the marginalized_time model. Please give it a try and let me know if it helps in a case you were seeing fail.

### DIFF
--- a/examples/inference/margtime/injection.ini
+++ b/examples/inference/margtime/injection.ini
@@ -2,14 +2,14 @@
 
 [static_params]
 tc = 1126259462.420
-mass1 = 37
-mass2 = 62
+mass1 = 80
+mass2 = 20
 ra = 2.2
 dec = -1.25
 inclination = 2.5
 coa_phase = 1.5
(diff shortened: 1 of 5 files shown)


ahnitz:
@sum33it Perhaps you can take a pass through and do a signoff of this. It seems to resolve @KoustavChandra's issue, so I think this works correctly. The main thing was to handle cases where monte carlo points are all invalid, in which case, we should fall back to drawing *valid* points from the prior.

ahnitz:
@sum33it Poke

cdcapano (approved):

PR #4204, opened 2022-12-05
fix ZeroDivisionError at low freq

WuShichao:
@ahnitz This PR fixes the `ZeroDivisionError` at low frequencies if there're zeros at the beginning of the reference waveform in the `Relative` model. I have tested this PR, it works.


WuShichao:
@ahnitz Yes, I have tested this with my modification of the waveform plugin by using LISA SMBHB example in PyCBC doc, and got almost same results: https://www.atlas.aei.uni-hannover.de/work/shichao.wu/lisa_smbhb_mass_tc_0.png

Issue #4196, opened 2022-11-30
Return a mass gap probability independently of BNS, NSBH, BBH probabilities


tdent:
At the moment the mchirp_area calculation only returns mutually exclusive BNS/NSBH/BBH probabilities, but it may be desirable to also return a separate mass gap probability (for an area that partly overlaps the NSBH and BBH ones).

titodalcanton:
Fixed by #4205?

PR #4205, opened 2022-12-05
Use mchirp_area to include a HasMassGap probability

veronica-villa:
Work in progress to upload a separate probability for MassGap to be included in the EM Bright properties file. Related to issue #4196.

### DIFF
--- a/pycbc/io/live.py
+++ b/pycbc/io/live.py
@@ -261,9 +261,10 @@ class CandidateForGraceDB(object):
             kwargs['hasmassgap_args']['mass_bdary']['ns_max'] = 3.0
             kwargs['hasmassgap_args']['mass_bdary']['gap_max'] = 5.0
             self.hasmassgap = calc_probabilities(coinc_inspiral_row.mchirp,
-                                                    coinc_inspiral_row.snr,
-                                                    min(eff_distances),
-                                                    kwargs['hasmassgap_args'])['Mass Gap']
+                                                 coinc_inspiral_row.snr,
+                                                 min(eff_distances),
+                                                 kwargs['hasmassgap_args']
+                                                )['Mass Gap']
         else:
             self.probabilities = None
             self.hasmassgap = None
@@ -295,7 +296,7 @@ class CandidateForGraceDB(object):
             self.basename = fname.replace('.xml.gz', '')
 
         # Save EMBright properties info as json
-        if  self.hasmassgap is not None:
+        if self.hasmassgap is not None:
             self.embright_file = self.basename + '_em_bright.json'
             with open(self.embright_file, 'w') as embrightf:
                 json.dump({'HasMassGap': self.hasmassgap}, embrightf)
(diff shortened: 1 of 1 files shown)


tdent:
@veronica-villa Do you have any test of this, even with a 'minimal' configuration?

pycbc/io/live.py
```diff
@@ -310,6 +319,14 @@ def save(self, fname):
             logging.info('P_astro file saved as %s', self.pastro_file)
         return
 
+        # Save EMBright properties info as json
+        if  self.hasmassgap is not None:
+            self.embright_file = self.basename + '_em_bright.json'
+            with open(self.embright_file, 'w') as embrightf:
+                json.dump({'HasMassGap': self.hasmassgap}, embrightf)
```
tdent, line 326: To be compatible with the schema of the expected 'em bright' file you should also add keys 'HasNS' and 'HasRemnant'.  The values for these are still to be determined (we do not have very clear guidance from low latency yet) so I would set the values to -1 in order to be clear that they are not valid probabilities.

tdent (changes requested):
One comment/request, but otherwise if this has been minimally tested it looks OK

tdent:
Ah, it seems there is one test suite failure, see here https://github.com/gwastro/pycbc/actions/runs/3622373938/jobs/6107006270#step:6:1335

pycbc/io/live.py
```diff
@@ -310,6 +319,14 @@ def save(self, fname):
             logging.info('P_astro file saved as %s', self.pastro_file)
         return
```
tdent, line 320: Ah, it seems you need to move this 'return' otherwise the code will just terminate the method here without reaching the embright saving block.

pycbc/io/live.py
tdent: Here is the reasoning - json does not have a None or NaN or null object, so we would have to use "None" or whatever string. But this would probably break whatever parsing is in gwcelery assuming float values. I think we don't want to make people rewrite gwcelery parsing, but we do want them to think about switching between using pipeline values if valid, vs using the 'standard' calculation if not valid.

titodalcanton:
This should be rebased on master.

pycbc/io/live.py
```diff
@@ -310,6 +319,14 @@ def save(self, fname):
...
 
+        # Save EMBright properties info as json
+        if  self.hasmassgap is not None:
+            self.embright_file = self.basename + '_em_bright.json'
```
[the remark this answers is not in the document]
titodalcanton: And you also need to make sure it ends up in the right directory. See #4195.

pycbc/io/live.py
tdent: Then you want the other fields not to be in the .json, i.e. no change in the current proposed code. In the absence of clear guidance from LL this may be the best short-term course.

pycbc/io/live.py
```diff
@@ -450,6 +468,20 @@ def upload(self, fname, gracedb_server=None, testing=True,
                                gid)
                 logging.error(str(exc))
 
+        # Upload em_bright properties JSON
+        if self.hasmassgap is not None:
+            try:
+                gracedb.write_log(
+                    gid, 'EM Bright properties JSON file upload',
+                    filename=self.embright_file,
+                    tag_name=['em_follow']
```
tdent, line 477: According to the upload spec document the tag name should now be 'em_bright'.

veronica-villa:
@tdent @titodalcanton this is has been tested on a pe mock signal using snr-optimization and runs as expected. An example of this can be checked in CIT `/home/veronica.villa/preO4/pycbclive-test/triggers/MDC/2022_01_01/candidate_1325030418.054_ezlb`.

tdent (approved):

PR #4207, opened 2022-12-07
Update get.sh

WuShichao:


titodalcanton:
@WuShichao please also add the `--show-error` option to the `curl` calls in that script, so `curl` actually prints something when the download fails.

WuShichao:
@titodalcanton Thanks. Let me add this too.

titodalcanton:
…however do please keep the `--silent` option, otherwise `curl` will needlessly spam the log :)

titodalcanton:
Final request, could you do the same fix to the other `get.sh` scripts in the `examples/inference/` directories? At least those which run as part of the test suite. It looks like they have the same problem.

PR #4209, opened 2022-12-08
Add a Fisher distribution for sky locations

Prasia-Pankunni:
1) Have created the class FisherDist in `sky_location.py` of `distributions.`
2) Added the definitions `new_z_to_euler`, `rotate_euler` to `transforms.py`
3) Added Fisher to the list of available distributions to` __init__.py`
4) Added definitions `decra2polaz` and `polaz2radec` to `conversions.py`


pycbc/transforms.py
```diff
@@ -2501,3 +2501,51 @@ def read_transforms_from_config(cp, section="transforms"):
...
     return order_transforms(trans)
+
+
+def new_z_to_euler(new_z):
+    """
+    From the new Z axis expressed in (polar, azimuthal) angles of the
+    initial coordinate system, return the (alpha, beta) Euler angles
+    that rotate the old Z axis (0, 0) to the new Z axis.
+    """
+    return (lal.PI_2 + new_z[1]) % (2 * lal.PI), new_z[0]
```
titodalcanton, line 2512: I suggest using `numpy.pi` rather than `lal.PI` and `lal.PI_2`.

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +33,47 @@ class UniformSky(angular.UniformSolidAngle):
...
+    def __init__(self, mu, kappa, if_radians=True):
+        self.kappa = kappa
+        if if_radians is True:
+            self.mu = (mu[1], mu[0])
+        elif if_radians is False:
+            self.mu = numpy.array(numpy.deg2rad([mu[1], mu[0]]))
+        else:
+            raise ValueError("You should give either 'True' or 'False' "
+                             "(for whether the angles are in radians or "
+                             "degrees respectively)")
```
titodalcanton, line 65: In the spirit of keeping things as simple as necessary, I would name the parameter just `radians`, and implement the code as follows:
```suggestion
        if radians:
            self.mu = (mu[1], mu[0])
        else:
            self.mu = numpy.array(numpy.deg2rad([mu[1], mu[0]]))
```

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +33,47 @@ class UniformSky(angular.UniformSolidAngle):
     _default_polar_angle = 'dec'
     _default_azimuthal_angle = 'ra'
 
+class FisherDist():
+    """A distribution that returns a random (ra, dec) angle drawn from the
+    Fisher distribution. Assume that the concentration parameter (kappa)
+    is large so that we can use a Rayleigh distribution about the north
+    pole and rotate it to be centered at the (ra, dec) coordinate mu.
```
titodalcanton, line 40: If may be worth checking if the assumption on `kappa` is being violated, and raising a warning or maybe even a ValueError.

pycbc/distributions/sky_location.py
titodalcanton: The docstring should also indicate what the `radians` parameter does (see docstrings for other classes for an example).

pycbc/transforms.py
```diff
@@ -2509,7 +2509,7 @@ def new_z_to_euler(new_z):
     initial coordinate system, return the (alpha, beta) Euler angles
     that rotate the old Z axis (0, 0) to the new Z axis.
     """
-    return (lal.PI_2 + new_z[1]) % (2 * lal.PI), new_z[0]
+    return (numpy.pi/2 + new_z[1]) % (2 * numpy.pi), new_z[0]
```
pannarale, line 2512:
```suggestion
    return (0.5*numpy.pi + new_z[1]) % (2 * numpy.pi), new_z[0]
```

pycbc/transforms.py
Prasia-Pankunni: The suggested change is made

pycbc/distributions/sky_location.py
Prasia-Pankunni: The suggested change is made and additional details regarding radians is provided in the docstring

pycbc/distributions/sky_location.py
Prasia-Pankunni: Theoretically it says in the limit kappa --> infinity . So, is it okay if I mention it in the docstring so that it will be understood that the class works for only high values of kappa ?

pycbc/transforms.py
Prasia-Pankunni: The suggested change is made

titodalcanton:
@Prasia-Pankunni could you link or paste a few plots showing that the points generated by the code follow the expected distribution?

Prasia-Pankunni:
> @Prasia-Pankunni could you link or paste a few plots showing that the points generated by the code follow the expected distribution?

Yes.. attached below fisher_old.png (corresponding to the old code) and fisher_new.png (corresponding to the new class created)

fisher_new
fisher_old

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +34,68 @@ class UniformSky(angular.UniformSolidAngle):
...
+        rot_eu = self.rvs_polaz(size)
+        ra_new = []
+        dec_new = []
+        for i in range(size):
+            ra_ = rot_eu[i][1]
+            ra_new.append(ra_)
+            dec_ = rot_eu[i][0]
+            dec_new.append(dec_)
+        ra_a = numpy.array([ra_new])
+        dec_p = numpy.array([dec_new])
```
titodalcanton, line 96: The loop and use of lists may be very slow if one wants to draw millions of points. Since `rot_eu` is a Numpy array, you can avoid the loop:
```suggestion
ra_a = rot_eu[:,1]
dec_p = rot_eu[:,0]
```
Much easier to read too!

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +34,68 @@ class UniformSky(angular.UniformSolidAngle):
...
+        right_ascension, declination = polaz2decra(dec_p, ra_a)
```
titodalcanton, line 97: The function is called `polaz2decra()`, but it returns ra, dec? This is confusing, so either change the name to `polaz2radec()` or swap the order of the returned values.

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +34,68 @@ class UniformSky(angular.UniformSolidAngle):
...
+    As in UniformSky, the declination (dec) varies from pi/2 to-pi/2
+    and right ascension (ra) varies from 0 to 2pi. And the angles
+    should be provided in (ra,dec) format in radians (mu_radians=True),
+    rather than factors of pi, or in degrees (mu_radians=False).
+
+    References:
+      * http://en.wikipedia.org/wiki/Von_Mises-Fisher_distribution
+      * http://arxiv.org/pdf/0902.0737v1 (states the Rayleigh limit)
+    """
+    name = 'fisher_dist'
```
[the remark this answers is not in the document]
titodalcanton: And same goes for the class name, I would use just `Fisher`.

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +34,68 @@ class UniformSky(angular.UniformSolidAngle):
...
+    is large so that we can use a Rayleigh distribution about the north
+    pole and rotate it to be centered at the (ra, dec) coordinate mu.
 
-__all__ = ['UniformSky']
+    Assume kappa = 1 / sigma**2
+
+    As in UniformSky, the declination (dec) varies from pi/2 to-pi/2
+    and right ascension (ra) varies from 0 to 2pi. And the angles
+    should be provided in (ra,dec) format in radians (mu_radians=True),
+    rather than factors of pi, or in degrees (mu_radians=False).
```
titodalcanton, line 48: Which units should `kappa` be given in? Please document this in the docstring as well. Should the option `mu_radians` also affect the units of `kappa`?

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +34,68 @@ class UniformSky(angular.UniformSolidAngle):
...
+        Randomly draw multiple samples from the Fisher distribution
+        and returns (polar, azimuthal) angle values.
+        """
+        arr = numpy.array([
+            numpy.random.rayleigh(scale=1./numpy.sqrt(self.kappa),
+                                  size=size),
+            numpy.random.uniform(low=0,
+                                 high=2*numpy.pi,
+                                 size=size)]).reshape((2, size)).T
+        alpha, beta = new_z_to_euler(self.mu_values)
```
titodalcanton, line 79: `alpha` and `beta` are constants fixed at initialization, so you could call`new_z_to_euler()` in `__init__` and store `alpha` and `beta` instead of `mu_values`.

pycbc/distributions/sky_location.py
Prasia-Pankunni: The suggested change has been made

pycbc/distributions/sky_location.py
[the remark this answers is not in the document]
Prasia-Pankunni: The suggested change has been made now

pycbc/distributions/sky_location.py
Prasia-Pankunni: The unit is now included in the docstring. It should be in steradians as well. Yes, mu_radians should be affecting units of kappa not directly but via rotate_euler.

pycbc/distributions/sky_location.py
Prasia-Pankunni: The suggested change has been made

pycbc/distributions/sky_location.py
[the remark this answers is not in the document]
titodalcanton: Apparently the change has been undone, so please remake it.

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +35,56 @@ class UniformSky(angular.UniformSolidAngle):
     _default_polar_angle = 'dec'
     _default_azimuthal_angle = 'ra'
 
+class Fisher():
```
ahnitz, line 38: Does a more specific name make sense? It seems this applies only to sky locations and some some generic fisher matrix?

pycbc/transforms.py
```diff
@@ -2501,3 +2501,61 @@ def read_transforms_from_config(cp, section="transforms"):
...
+    Take an Nx2 array of N (theta, phi) vectors on the unit sphere
+    (that is, (polar, azimuthal) angles in radians) and apply
+    rotations through the Euler angles alpha, beta, and gamma
+    in radians, using the ZXZ convention.
+    """
+    c = numpy.cos
+    s = numpy.sin
+
+    # Define rotation matrix
+    R = numpy.array(
```
ahnitz, line 2526: I think scipy can quickly produce rotation matrices. Is this providing functionality not already defined there? This may be fine as is, but just wanted to ask in case just using scipy to get the rotation matrices would be cleaner.

pycbc/transforms.py
```diff
@@ -2501,3 +2501,61 @@ def read_transforms_from_config(cp, section="transforms"):
...
+
+    # Extract new spherical coordinates
+    sph_new = numpy.empty(shape=(len(sph_coords), 2), dtype=float)
+    sph_new[:, 0] = numpy.arccos(cart_new[:, 2])
+    sph_new[:, 1] = numpy.arctan2(cart_new[:, 1], cart_new[:, 0])
+
+    return sph_new
+
+
+def decra2polaz(dec, ra):
```
ahnitz, line 2554: Would this make more sense in the conversions module and with naming to match the style there?

ahnitz (commented):
Hey All, I just wanted to raise a couple minor questions on the style / naming in case they haven't been considered.

pycbc/distributions/sky_location.py
titodalcanton: @ahnitz this is actually the Fisher distribution on the sphere (special case of the von Mises-Fisher distribution in 3 dimensions, https://en.wikipedia.org/wiki/Von_Mises%E2%80%93Fisher_distribution) so I think the name is appropriate, but suggestions are welcome of course.

pycbc/transforms.py
titodalcanton: That seems appropriate indeed.

pycbc/distributions/sky_location.py
titodalcanton: I guess you would prefer FisherSky, to be consistent with UniformSky? Since the distribution expects the angular parameters to be called `ra` and `dec`, making it explicit that it is intended for sky locations seems appropriate.

titodalcanton: @Prasia-Pankunni please rename the class to FisherSky.

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +35,56 @@ class UniformSky(angular.UniformSolidAngle):
...
+    Assume kappa = 1 / (0.66*sigma)**2
+    As in UniformSky, the declination (dec) varies from pi/2 to-pi/2
+    and right ascension (ra) varies from 0 to 2pi. And the angles
+    should be provided in (ra,dec) format in radians (mu_radians=True),
+    rather than factors of pi, or in degrees (mu_radians=False).
+    References:
+      * http://en.wikipedia.org/wiki/Von_Mises-Fisher_distribution
+      * http://arxiv.org/pdf/0902.0737v1 (states the Rayleigh limit)
+    """
+    name = 'fisher'
```
titodalcanton, line 52: @Prasia-Pankunni please change this to `fisher_sky` as well.

pycbc/distributions/sky_location.py
Prasia-Pankunni: Yes. the change is made.

pycbc/distributions/sky_location.py
Prasia-Pankunni: This is also done

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +35,56 @@ class UniformSky(angular.UniformSolidAngle):
...
+                                 high = (2*numpy.pi),
+                                 size = size)]).T
+        euler = rotate_euler(arr, self.alpha, self.beta, 0)
+        rot_euler = FieldArray(size,dtype = [('ra','<f8'), ('dec','<f8')])
+        rot_euler['ra'], rot_euler['dec'] = euler[:,0], euler[:,1]
+        return rot_euler
+
 
-__all__ = ['UniformSky']
+__all__ = ['UniformSky', 'Fisher']
```
titodalcanton, line 90: @Prasia-Pankunni this should also be changed to FisherSky.

pycbc/transforms.py
titodalcanton: @Prasia-Pankunni could you move the `decra2polaz` and `polaz2radec` functions to `pycbc/conversions.py` instead? I would put them at line 859, inside the "Extrinsic parameter functions" section. Remember to update the imports accordingly.

pycbc/conversions.py
```diff
@@ -844,7 +844,15 @@ def optimal_ra_from_detector(detector_name, tc):
...
-    ----------
+    ----------def decra2polaz(dec, ra):
+    """Convert (ra,dec) to (polar, azimuthal) angles. All angles in radians."""
+    return (0.5*numpy.pi) - dec, ra
+
+
+def polaz2radec(theta, phi):
+    """Convert (polar, azimuthal) to (ra,dec) angles. All angles in radians."""
+    return (0.5*numpy.pi) - theta, phi
+
```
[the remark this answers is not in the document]
titodalcanton: It actually looks like you are not using these functions anymore. A previous version of the PR had the `mu_radians` flag, but it has been removed. Either reimplement that mechanism, or just remove these functions.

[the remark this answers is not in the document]
titodalcanton: I do think that having the ability to enter the mean position in degrees is useful, so I would recommend having something like a `angle_unit = deg` vs `angle_unit = rad` parameter in the configuration file.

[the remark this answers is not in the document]
Prasia-Pankunni: Have included angle_unit parameter in the config file and modified sky_location.py accordingly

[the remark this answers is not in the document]
Prasia-Pankunni: > Woops something went wrong here. Please fix.

Removed those as it is not used anymore.

Prasia-Pankunni (commented):
The correction is made

pycbc/distributions/__init__.py
```diff
@@ -61,6 +61,7 @@
     FixedSamples.name: FixedSamples,
     MchirpfromUniformMass1Mass2.name: MchirpfromUniformMass1Mass2,
     QfromUniformMass1Mass2.name: QfromUniformMass1Mass2
+    FisherSky.name: FisherSky
```
titodalcanton, line 64: Line 63 needs a comma.

pycbc/conversions.py
```diff
@@ -842,21 +842,20 @@ def optimal_ra_from_detector(detector_name, tc):
...
         The GPS time of the coalescence of the signal in the `ref_frame`.
-
     Returns
     -------
     float :
         The declination of the signal, in radians.
     """
     return optimal_orientation_from_detector(detector_name, tc)[0]
 
+
```
titodalcanton, line 858: Please undo all whitespace changes to `conversions.py`.

pycbc/distributions/__init__.py
Prasia-Pankunni: This is done

pycbc/conversions.py
titodalcanton: I go with 1 empty line before "Parameters" and "Returns" in docstrings, and no particular choice for what comes before `return` statements (depends on context). I guess my main point was to keep this PR short, but if you think this whitespace change makes things more consistent, go ahead.

titodalcanton: I suppose it is fine to remove the empty line on line 845, but the one on line 852 should stay.

Two spaces *after* `return` statements is also fine.

pycbc/distributions/sky_location.py
```diff
@@ -31,5 +35,66 @@ class UniformSky(angular.UniformSolidAngle):
...
+    References:
+      * http://en.wikipedia.org/wiki/Von_Mises-Fisher_distribution
+      * http://arxiv.org/pdf/0902.0737v1 (states the Rayleigh limit)
+    """
+    name = 'fisher_sky'
+    _params = ['ra', 'dec']
+
+    def __init__(self, angle_unit='rad', **params):
+        self.sigma = params['sigma']
+        self.kappa = 1./((0.66*params['sigma']))**2
```
titodalcanton, line 58: Note that sigma must also be converted from deg to rad when `angle_unit == 'rad'`.

Prasia-Pankunni: Sigma conversion included

titodalcanton:
I agree this is now in principle ready. I also agree with @ahnitz's point. Then I would like to see a quick test like what @Prasia-Pankunni did a while ago, i.e. make a few cases with a narrow distribution centered at the poles and at one point on the equator, and show that the points are distributed around the correct position.

pannarale:
To move on with this PR, let's postpone exploring scipy: I opened issue https://github.com/gwastro/pycbc/issues/4305 so we do not lose track of this.

What is missing are the tests requested by @titodalcanton with the code as it is now: "make a few cases with a narrow distribution centered at the poles and at one point on the equator, and show that the points are distributed around the correct position."

titodalcanton:
Today I spent some time doing the tests I suggested previously. There were issues related to incorrect angle conversions and/or RA and dec being swapped. I tried understanding the problem with the idea of producing a minimal change that would fix it, but I got a bit carried away and ended up refactoring the PR to use a Scipy Rotation, as suggested in #4305. This simplified the PR quite a bit. I also added some sanity checks for the distribution parameters.

My tests (currently showing the result after my changes are implemented) can be found here: https://ldas-jobs.ligo.caltech.edu/~tito.canton/pycbc/test_pr4209/

pycbc/distributions/sky_location.py
```diff
@@ -32,4 +38,136 @@ class UniformSky(angular.UniformSolidAngle):
...
+        np_cart[:, 0] = numpy.cos(np_ra) * numpy.sin(np_dec)
+        np_cart[:, 1] = numpy.sin(np_ra) * numpy.sin(np_dec)
+        np_cart[:, 2] = numpy.cos(np_dec)
+
+        # Rotate the samples according to our pre-built rotation
+        rot_cart = self.rotation.apply(np_cart)
+
+        # Convert the samples back to spherical coordinates.
+        # Some unpleasant conditional operations are needed
+        # to get the right angle convention.
```
pannarale, line 158:
```suggestion
        # to get the correct angle convention.
```

pannarale (changes requested):
Just 2 very small changes to documentation/comments.

pannarale (approved):

PR #4213, opened 2022-12-20
Update requirements.txt

ahnitz:


tdent (approved):

titodalcanton:
Does anyone remember why this was done? Can it be reverted now?

ahnitz:
@titodalcanton I think it's a compatibility issue with emcee. We may need to update some examples. Can you open a revert PR anyway? We should get that working, but we'll need to add a few docs changes at least I think to get it working.

tdent:
To get round `numpy.float` suddenly disappearing as an available type, I think. 

The plan was to 'fix this properly after the holidays' (meaning in January 2023).

ahnitz:
@tdent Yeah, which clearly I didn't do. No so surprising since that's around my job change time. I completely forgot about this.

titodalcanton:
It is technically "after the holidays" now, so… Here we are.

PR #4216, opened 2022-12-21
Fix numpy.float deprecation

titodalcanton:


tdent:
As @ahnitz pointed out, there is still usage of np.float inside emcee which is used by inference.

ahnitz:
@titodalcanton I think we need the other fixes you've found in any case. We can't revert the numpy pin though until we decide what to do about some optional dependencies in the test / ci env. Can you take that part out for now so the rest can merge?

tdent (approved):
I think this should be merged ...

Issue #4212, opened 2022-12-20
Make p astro also for snr max followup events


tdent:
It seems LL wants every (at least full bw) event to have a p astro json file, so we should do some plumbing to copy over at least p astro / p terr to the followup input file and possibly combine with p_source from the revised masses/spins.

titodalcanton:
Fixed by #4219.

PR #4219, opened 2023-01-03
Communicate p_terr to snr opt event

tdent:
To address #4212 pass the p_terr from CandidateForGraceDB to the snr max setup by treating it as an extra entry in coinc_results. This allows the snr max event to 'remember' p_terr from the initial event


titodalcanton:
@tdent I see you are not doing the `p_terr` propagation for single-detector candidates. Is that intentional?

titodalcanton (approved):

tdent:
I can probably do some 'minimal' testing on this (1 BNS signal with reduced bank)

tdent:
One bug in getting info out of the hdf5 file ...

PR #4223, opened 2023-01-16
Adding script for plotting histograms of triggers near gates

veronica-villa:
New things:
- I've added a --bin-duration argument so we can control the ammount of seconds represented by each bin (previously it was hardcoded as nbins=80 for every case).
- Now --snr-cut-type and --snr-cut-vals are two different arguments because I think it's more intuitive (same thing could apply to the --template-duration cut, but I haven't changed it yet)

The major problem I see is that the split_in_ifos function relies on the single trigger files to be named as "IFO-...", but it works at the moment with the existing naming system.

### DIFF
--- a/bin/plotting/pycbc_plot_gate_triggers
+++ b/bin/plotting/pycbc_plot_gate_triggers
@@ -126,7 +126,7 @@ for single_file in args.single_trigger_files:
         if args.template_duration:
             if args.duration_ltgt == 'LT':
                 snr_val = snr_val[template_duration <= args.duration_cut]
-            else args.duration_ltgt == 'GT':
+            else:
                 snr_val = snr_val[template_duration >= args.duration_cut]
 
     logging.info("Starting loop over gates...")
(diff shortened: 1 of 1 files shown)


bin/plotting/pycbc_plot_gate_triggers
```diff
@@ -0,0 +1,201 @@
...
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 3 of the License, or (at your
+# option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
```
tdent, line 15: I think these 3 lines are not really worth including any more ...

tdent (approved):

PR #4225, opened 2023-01-16
Change invspec default truncation to Hann

andrew-lundgren:
`inverse_spectrum_truncation` uses boxcar truncation by default. This works OK in the tutorials because the data isn't very long, but it gives a bad filter when used to whiten longer blocks of data. For instance, when whitening 30 minutes of data the PSD is corrupted and gives a bad SNR timeseries. Hann window would be a more sensible default.

In most places in the code, it seems to use Hann window already. The main place where it's used without is lines 1714 and 1720 of [strain.py](https://github.com/gwastro/pycbc/blob/master/pycbc/strain/strain.py#L1714). I'm not sure if this is intentional. It's also used as default in some tests, but I think they're testing length and not default

### DIFF
--- a/pycbc/psd/estimate.py
+++ b/pycbc/psd/estimate.py
@@ -200,7 +200,7 @@ def welch(timeseries, seg_len=4096, seg_stride=2048, window='hann',
     return FrequencySeries(psd, delta_f=delta_f, dtype=timeseries.dtype,
                            epoch=timeseries.start_time)
 
-def inverse_spectrum_truncation(psd, max_filter_len, low_frequency_cutoff=None, trunc_method=None):
+def inverse_spectrum_truncation(psd, max_filter_len, low_frequency_cutoff=None, trunc_method='hann'):
     """Modify a PSD such that the impulse response associated with its inverse
     square root is no longer than `max_filter_len` time samples. In practice
     this corresponds to a coarse graining or smoothing of the PSD.


ahnitz (approved):
I think this is OK, but maybe @titodalcanton has another opinion?

titodalcanton:
As far as I remember, PyCBC-based searches I have worked with have always used the default method, i.e. the hard truncation. A quick glance at the code and O3 offline configuration seems to confirm this. For examples, the lines pointed out by @andrew-lundgren are what PyCBC Live uses.

I never liked the idea of a hard truncation, so I am ok with the change in principle, but I think approving this requires more than a simple opinion. We should do at least some minimal testing to evaluate the effect on searches.

@andrew-lundgren could you link some plots or scripts that show the problem with the hard truncation, and what happens with the Hann one? Would you (or a student) have a chance to rerun one of the O3 offline boxes with the Hann truncation, and compare? I am curious.

andrew-lundgren:
I got some H1 O3 data, made a 16-sec PSD, and invspectrunced it up to both 32 and 2048 sec segments. The PSDs are below, zoomed into the low-frequency region. There's ringing with a period of 16 peaks per Hz. There's similar ringing up at high frequency. The final image shows the SNR for a template that has primarily low-frequency content (up to about 40 Hz), and the background is a lot higher than it should be. However, this doesn't seem to be noticeable if the signal has more normal bandwidth.

We have final-year projects starting next week here, I can ask around and see if anyone wants to run a chunk of O3 for testing. 

image

image

image

titodalcanton:
Thanks @andrew-lundgren, Hann is clearly doing much better.

If you want this change to go in quickly, I propose the following. In the two calls you pointed out (1714 and 1720 of `strain.py`), could you add a kwarg forcing the boxcar truncation, so that PyCBC Live continues to use that for now? I will (find someone to) look into the effect of the Hann change on PyCBC Live, and change that to Hann once we are convinced everything is fine. Maybe you also need to do a few changes to `inverse_spectrum_truncation()` itself.

andrew-lundgren:
Just a quick note to say that this issue didn't occur when I tried to reproduce it with 4kHz data. It only happens with 16 kHz data. So I think it's not an issue in the search because it's downsampled first.

tdent:
In terms of how the function is called I don't get such a good feeling from the default being different from passing None.  As noted in chat, a less pitfally way would be to only allow two explicit values, eg 'boxcar' and 'hann', such that None is no longer a valid option.  This may break things ...

tdent:
> This may break things ...

and I think if a default value is changed it's a _good_ thing that calls which assume the previous value start to break.

tdent:
@PRAVEEN-mnl and I tried running a BBH analysis with "hann invspec truncation" and this is the comparison of the results. 
https://ldas-jobs.ligo.caltech.edu/~praveen.kumar/o3/runs/hlv/c01_highmass_kdestat/a1_kdev2/
https://ldas-jobs.ligo.caltech.edu/~praveen.kumar/o3/runs/hlv/c01_highmass_kdestat/a1_hann2/
The second one has the extra option. 

There are tiny changes in the single detector triggers and non-zero but small effects on the background and sensitivity.

In the loudest background triggers, there are small (few% or less) changes in the SNR & chi_square

tdent:
Overall, it doesn't seem like there is any significant improvement by changing this option. 

> @PRAVEEN-mnl and I tried running a BBH analysis with "hann invspec truncation" and this is the comparison of the results. https://ldas-jobs.ligo.caltech.edu/~praveen.kumar/o3/runs/hlv/c01_highmass_kdestat/a1_kdev2/ https://ldas-jobs.ligo.caltech.edu/~praveen.kumar/o3/runs/hlv/c01_highmass_kdestat/a1_hann2/ The second one has the extra option.
> 
> There are tiny changes in the single detector triggers and non-zero but small effects on the background and sensitivity.
> 
> In the loudest background triggers, there are small (few% or less) changes in the SNR & chi_square

spxiwh:
This one looks like it is ready to merge now? It has an approval and @PRAVEEN-mnl 's test suggests that the search pipeline is not affected by this. Objections to me pushing the merge button??

tdent:
1. my comment about the use of 'None' has not been replied to - basically, when there is a non-None default it is weird and opaque for a user to have to give a `=None` value explicitly if they want boxcar truncation. Better to have a named 'boxcar' option and throw an error on None.
2. the rationale to make the change is not 100% clear yet: is it that bad things don't happen in the search either way, but we want them also not to happen in other contexts like Andy's?

tdent:
oh, and 3. it's still 'Draft' ...

spxiwh:
@tdent I guess the draft status is easily removed. I agree that the "boxcar" transition should be a named kwarg if this gets merged.

Probably need to agree on point 2 though, as then fixing the other 2 points is trivial. I think the rationale is that in any context in which we use a PSD where significant power is present at times larger than the spectrum length the boxcar truncation is going to be bad. I think Andy's plots above motivate this well. It doesn't mean that the Hann window is perfect ... we will lose reliability by truncating such a PSD. But the Hann window does avoid the "bad" features that Andy reports.

tdent:
OK, if you and Andy agree that in general this should be an improvement that should be enough of a reason. 

(NB, if you aren't matched filtering then invspectrunc shouldn't be used in the first place ...)

titodalcanton:
I do not think my comment has been addressed either. I do not want to change this behavior for PyCBC Live without seeing some tests.

spxiwh:
@titodalcanton What tests would you like to see?

If I think more generally, there are already a number of PyCBC Live tests in the CI, which have run. If you mean something more substantive like "run on the MDC", that is not something that we can (or should) require for PRs coming in from people not working on PyCBC Live itself. For e.g. we cannot expect Alex to have to run a potential change that might affect Live on the LVK MDC. So what's a fair process here for being convinced that this doesn't affect Live?

tdent:
The simple way to have this not affect pycbc live in production is to add the 'boxcar' option explicitly to places where trunc is called in pycbc live ..

spxiwh:
@tdent That's an action for this PR then. PyCBC Live does not use the usual strain/PSD CLI code, so would need to have this option added to it and hooked up. Then it could choose whatever it wants in it's config scripts.

titodalcanton:
@spxiwh yes, what I am requesting for this PR is not to do extensive tests of PyCBC Live, I am simply requesting that PyCBC Live's current behavior is not changed by this.

PR #4226, opened 2023-01-17
Fixed phase space coverage

MarcoCusinato:
When calculating tau0 from the two compact object masses, the code used a fixed frequency of 15.0 Hz that prevented a uniform phase space coverage even with an arbitrarily large tau0-end (please see figures). It was also added tau0-end derivation from the lowest chirp mass, assuring total phase space coverage.
before
after

### DIFF
--- a/bin/bank/pycbc_brute_bank
+++ b/bin/bank/pycbc_brute_bank
@@ -89,6 +89,17 @@ if args.input_config is not None:
 
     dists_joint = prior_from_config(cp=config_parser)
 
+if not args.tau0_end:
+    """
+    If tau0_end is not provided, it is derived and rounded up to the
+    upper tens.
+    This is necessary to cover up all the phase space.  
+    """
+    import math
+    from pycbc.conversions import tau0_from_mass1_mass2
+    args.tau0_end = math.ceil(tau0_from_mass1_mass2(args.min[0], args.min[1],
+                                        args.low_frequency_cutoff) / 10) * 10
+
 fdict = {}
 if args.fixed_params:
     for p, v in zip(args.fixed_params, args.fixed_values):
@@ -187,11 +198,12 @@ class TriangleBank(object):
 
         msig = len(r)
 
-        #Apply tua0 threshold
+        #Apply tau0 threshold
         if args.tau0_threshold:
             hp.tau0 = pycbc.conversions.tau0_from_mass1_mass2(
                                             hp.params['mass1'],
-                                            hp.params['mass2'], 15.0)
+                                            hp.params['mass2'],
+                                            args.low_frequency_cutoff)
             hp.tbin = int(hp.tau0 / args.tau0_threshold)
 
             if hp.tbin in self.tbins:
@@ -400,7 +412,8 @@ def cdraw(rtype, ts, te):
 
     p = draw(rtype)
     if  len(p[list(p.keys())[0]]) > 0:
-        t = tau0_from_mass1_mass2(p['mass1'], p['mass2'], 15.0)
+        t = tau0_from_mass1_mass2(p['mass1'], p['mass2'],
+                                  args.low_frequency_cutoff)
         l = (t < te) & (t > ts)
         for k in p:
             p[k] = p[k][l]
@@ -413,7 +426,8 @@ def cdraw(rtype, ts, te):
             p[k] = numpy.concatenate([p[k], tp[k]])
 
         if  len(p[list(p.keys())[0]]) > 0:
-            t = tau0_from_mass1_mass2(p['mass1'], p['mass2'], 15.0)
+            t = tau0_from_mass1_mass2(p['mass1'], p['mass2'],
+                                      args.low_frequency_cutoff)
             l = (t < te) & (t > ts)
             for k in p:
                 p[k] = p[k][l]
--- a/bin/bank/pycbc_brute_bank
+++ b/bin/bank/pycbc_brute_bank
@@ -63,9 +63,6 @@ parser.add_argument('--max-q', type=float)
 parser.add_argument('--tau0-crawl', type=float)
 parser.add_argument('--tau0-start', type=float)
 parser.add_argument('--tau0-end', type=float)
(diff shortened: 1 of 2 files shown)


titodalcanton:
@ahnitz would you be able to have a look at this one? The change looks reasonable to me, but I am not very familiar with the brute bank algorithm.

ahnitz:
@MarcoCusinato I'm not sure I follow the reason for this change. What was the command line in both examples you give? Did you not account for the fixed reference frequency when setting the tau0 limits or are you saying there was some other mistake? I don't see how the code change fixes a  problem atm, unless you are worried about cases where f_final < 15.0 Hz. But that doesn't seems to be your example, hence I think we need some clarification to make sure whatever problem identified was fixed.

MarcoCusinato:
> @MarcoCusinato I'm not sure I follow the reason for this change. What was the command line in both examples you give? Did you not account for the fixed reference frequency when setting the tau0 limits or are you saying there was some other mistake? I don't see how the code change fixes a problem atm, unless you are worried about cases where f_final < 15.0 Hz. But that doesn't seems to be your example, hence I think we need some clarification to make sure whatever problem identified was fixed.

I used the same string to generate both banks. It is based on the bank generation example (bank.sh in /pycbc/examples/search), changing the two masses limits (m1 in [1, 25] and m2 in [1, 3]) and tau0-end (280). 
Now I understand I did not account for the fixed frequency; however, I do not get why we apply a low frequency different from the one in input (also for a matter of computational time). In any case, I think we should keep the function to derive tau0-end changing the last argument from args.low_frequency_cutoff to 15 Hz.

String used:
--minimal-match 0.95 --tolerance .005 --buffer-length 2 --sample-rate 2048  --tau0-threshold 0.5 --approximant IMRPhenomD --tau0-crawl 5 --tau0-start 0 --tau0-end 280 --psd-model aLIGOZeroDetLowPower --min 1 1 0 0 --max 25 3 1. .2 --params mass1 mass2 spin1z spin2z --seed 1 --low-frequency-cutoff 20.0

ahnitz:
@MarcoCusinato The point is to have a fixed coordinate system, so that it's easier to set the values. Otherwise, if you change flow, you have to also change the boundaries. 

Your patch unfortunately, will not generically work as there is no guarantee that m1/m2 are the first two arguments as you've assumed nor that they are provided at all through the command line that way, e.g. one can also provide a prior defining config file to set the space.

bin/bank/pycbc_brute_bank
```diff
@@ -89,6 +89,17 @@ if args.input_config is not None:
...
 
+if not args.tau0_end:
+    """
+    If tau0_end is not provided, it is derived and rounded up to the
+    upper tens.
+    This is necessary to cover up all the phase space.  
+    """
+    import math
+    from pycbc.conversions import tau0_from_mass1_mass2
+    args.tau0_end = math.ceil(tau0_from_mass1_mass2(args.min[0], args.min[1],
```
ahnitz, line 100: The problem here is that this assumes that m1/m2 are actually in these arguments, but that is not necessarily the case. In fact, it may no be provided on the command line at all.

bin/bank/pycbc_brute_bank
```diff
@@ -187,11 +198,12 @@ class TriangleBank(object):
...
         msig = len(r)
 
-        #Apply tua0 threshold
+        #Apply tau0 threshold
         if args.tau0_threshold:
             hp.tau0 = pycbc.conversions.tau0_from_mass1_mass2(
                                             hp.params['mass1'],
-                                            hp.params['mass2'], 15.0)
+                                            hp.params['mass2'],
+                                            args.low_frequency_cutoff)
```
ahnitz, line 206: I'm not opposed to allowing other options than 15 Hz, but I'm not sure it should always be the low frequency cutoff. Maybe add an option?

ahnitz (changes requested):

ahnitz (approved):

bin/bank/pycbc_brute_bank
MarcoCusinato: Ok, I removed this part of the code. However, does it make sense to keep a similar function that could read all the possible options? This way, we wouldn't have to manually derive the tau0-end (which I believe is mandatory in the code, am I correct?), and it would cover the entire phase space.

ahnitz: @MarcoCusinato This would be a nice feature, but I think it is somewhat more complicated that you might imagine, given that one can use somewhat arbitrary coordinates. Coordinates have to be convertible to m1/m2 at a certain point, but there doesn't have to be a simple boundary. The best I can think to suggest if you want it to be generic is to find the min/ max by sampling from the distribution. 

If you think this is reasonable, maybe you could add this in a separate PR after this is merged?

ahnitz: You can pass a prior config file (the same format we use for parameter estimation). That also means you can use things like arbitrary coordinates (as long as you provide the transform back to the ones used for waveform generation) in addition to constraints, etc. It also allows you some control over how proposal points are selected as it will periodically draw from the provided prior in addition to using a kde of previously accepted values. (if you've seen the output, this is the difference between a 'round' marked with 'U' (prior) and 'K' (kde).

ahnitz: A trivial example would be 
```
[variable_params]
mass1 =
mass2 =

[static_params]

[prior-mass1]
name = uniform
min-mass1 = 10.
max-mass1 = 80.

[prior-mass2]
name = uniform
min-mass2 = 10.
max-mass2 = 80.
```

But there are of course many other priors available, and you can provide a [waveform_transform-*] section and constraint sections. See some PE examples for uses of this, to do coordinate transforms from whatever coordinates you want. 

https://pycbc.org/pycbc/latest/html/inference.html#waveform-transforms

https://pycbc.org/pycbc/latest/html/inference.html#constraints

MarcoCusinato: Thank you very much!

PR #4234, opened 2023-01-18
Fom plot kde calc

PRAVEEN-mnl:


bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -8,6 +8,7 @@ from pycbc import libutils
 from pycbc.events import triggers
 akde = libutils.import_optional('awkde')
 kf = libutils.import_optional('sklearn.model_selection')
+import matplotlib.pyplot as plt
```
tdent, line 11: it would be better to import this only if making the plot I think

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -62,17 +71,34 @@ def kfcv_awkde(sample, bwchoice, alphachoice, k=2):
 
 
 def optimizedparam(sampleval, nfold=2):
-    # Test wide range of bandwidths in addition to standard values
-    bwgrid = ['scott', 'silverman'] + numpy.logspace(-2,0,10).tolist()
-    # Fix adaptive sensitivity parameter to 1
-    alphagrid = [1]
+    # bw-grid and alpha-grid are values given as arguments to the function
```
tdent, line 74: this comment is not correct. The arguments of the function are what variables are passed to it inside the brackets, currently these are 'sampleval' and 'nfold'. To make the function self-contained 'bwgrid' and 'alphagrid' should also be arguments of the function. This is not connected with argparse (command line arguments of the whole script)

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -62,17 +71,34 @@ def kfcv_awkde(sample, bwchoice, alphachoice, k=2):
...
         for alphavals in alphagrid:
-            FOM[(gbw, alphavals)] = kfcv_awkde(sampleval, gbw, alphavals, k=nfold)
+            FOM[(gbw, alphavals)] = kfcv_awkde(sampleval, gbw, alphavals,\
+                k=nfold)
     optval = max(FOM.items(), key=operator.itemgetter(1))[0]
     optbw, optalpha = optval[0], optval[1]
     maxFOM = FOM[(optbw, optalpha)]
+
+    # Plotting FOM paramters
+    fig = plt.figure(figsize=(12,8))
```
tdent, line 87: I think you only want to do this if the fom-plot option is given ..

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -62,17 +71,34 @@ def kfcv_awkde(sample, bwchoice, alphachoice, k=2):
...
+    ax = fig.add_subplot(111)
+    for bw in bwgrid:
+        FOMlist = [FOM[(bw, al)] for al in alphagrid]
+        ax.plot(alphagrid, FOMlist, label='{0:.3f}'.format(bw))
+    ax.plot(optalpha, maxFOM, 'ko', linewidth=10, label= \
+            r'$\alpha={0:.3f},bw= {1:.3f}$'.format(optalpha, optbw))
+    ax.set_xlabel(r'$\alpha$', fontsize=15)
+    ax.set_ylabel(r'$FOM$', fontsize=15)
+    ax.set_ylim(maxFOM-20, maxFOM+5)
+    handles, labels = ax.get_legend_handles_labels()
```
tdent, line 97: is this needed? ie do you need to get these values? what happens if you do plt.legend() without using handles/labels?

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -138,16 +167,18 @@ if args.make_signal_kde:
...
-    signal_kde = kde_awkde(signal_samples, temp_samples, alp=optalpha, gl_bandwidth=optbw)
-
-    f_dest.create_dataset("signal_kde", data=signal_kde)
-
+    signal_kde = kde_awkde(signal_samples, temp_samples, \
+                           alp=optalpha, gl_bandwidth=optbw)
+    f_dest.create_dataset("data_kde", data=signal_kde)
+    f_dest.attrs['stat'] = "signal-kde_file"
+  
 f_dest.close()
```
tdent, line 183: so just before this I think you can add the optimal alpha and bw as attrs (this should work equally for template and signal kdes)

tdent (changes requested):

tdent:
One of the tests is failing, can you look at that?

tdent (approved):

PR #4242, opened 2023-01-19
Add second PyCBC Live paper to credit docs page

titodalcanton:
This is untested as I could not get the docs to build on my systems due to the BBHx issues. I guess I can look at the artifacts though.


spxiwh (approved):
Only affects docs, so all happy to merge when ready.

PR #4247, opened 2023-01-25
add PASTRO_READY label

tdent:
Addressing requests from Low Latency chairs

### DIFF
--- a/companion.txt
+++ b/companion.txt
@@ -5,7 +5,7 @@ gwpy>=0.8.1
 healpy
 
 # Needed for GraceDB uploads and skymap generation
-ligo-gracedb>=2.7.8
+ligo-gracedb>=2.10.0
 ligo.skymap
 
 # Needed for ringdown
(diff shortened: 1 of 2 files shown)


titodalcanton:
@tdent do you know if this has a specific requirement in terms of ligo-gracedb version?

tdent:
> @tdent do you know if this has a specific requirement in terms of ligo-gracedb version?

Yes, this is a recent addition to the gracedb interface (https://git.ligo.org/computing/sccb/-/issues/1085) which is now running in playground, do we need to note this somewhere?

titodalcanton:
Yes, please edit `companion.txt` to require ligo-gracedb 2.10.0.

I should make sure that the various environments are up to date with it.

titodalcanton:
I am also thinking that applying the label should come *before* making and uploading the plots, which is just for humans and can take its time. That involves rearranging some code though, so maybe we want to keep that a separate PR.

tdent:
I don't think we make a specific multi-pastro plot, so I guess this is mainly concerned with SNR and PSD plots .. so yes, in general sending tiny .json files should take precedence over doing things with matplotlib. Material for another issue.

titodalcanton (approved):

PR #4251, opened 2023-01-31
PyCBC Live: fail gracefully if single fit file is missing

titodalcanton:


GarethCabournDavies (approved):
Exactly what I thought was needed- thanks!

pycbc/events/single.py
```diff
@@ -123,30 +123,40 @@ def check(self, trigs, data_reader):
...
+        try:
+            with h5py.File(self.fit_file, 'r') as fit_file:
+                bin_edges = fit_file['bins_edges'][:]
+                live_time = fit_file[self.ifo].attrs['live_time']
+                thresh = fit_file.attrs['fit_threshold']
+                dist_grp = fit_file[self.ifo][self.sngl_ifar_est_dist]
+                rates = dist_grp['counts'][:] / live_time
+                coeffs = dist_grp['fit_coeff'][:]
+        except FileNotFoundError:
+            logging.error(
```
GarethCabournDavies, line 155: logging hasnt been imported (caught by codeclimate) - so this will still fail

GarethCabournDavies (changes requested):

titodalcanton:
@GarethCabournDavies to further please CC, one could remove the first check of the function (checking for zero length) as the same check is effectively included in the next one!

GarethCabournDavies:
> @GarethCabournDavies to further please CC, one could remove the first check of the function (checking for zero length) as the same check is effectively included in the next one!

I had thought that would cause an error (comparing an empty array to a float), but apparently it doesn't!

GarethCabournDavies (approved):

PR #4252, opened 2023-02-02
Fix x-axis error in combined fits plotting

GarethCabournDavies:
We noticed an error where the x-axis values were actually counting days since the zero point of GPS time, this resets zero to the first day (as reported in the x label)


tdent (approved):
If this does what is wanted, then I approve ..

PR #4254, opened 2023-02-03
Update GitHub actions to fix Node.js 16 warning

titodalcanton:
I see this warning on one of the Actions pages:

> Node.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions/setup-python@v2. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

I think this PR does what is needed to fix it.

### DIFF
--- a/.github/workflows/distribution.yml
+++ b/.github/workflows/distribution.yml
@@ -11,10 +11,10 @@ jobs:
         os: [ubuntu-20.04, macos-latest]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v4
       - name: Install cibuildwheel
         run: python -m pip install cibuildwheel==2.3.1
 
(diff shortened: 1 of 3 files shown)


ahnitz (approved):
It seems reasonable to me to move to  more recent versions of the actions. Assuming no incompatibilities are found, I am happy with this.

ahnitz:
@titodalcanton The error looks to me like maybe the default python version is being handled differently with the new action?

.github/workflows/distribution.yml
```diff
@@ -11,10 +11,10 @@ jobs:
         os: [ubuntu-20.04, macos-latest]
 
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - uses: actions/setup-python@v2
+      - uses: actions/setup-python@v4
```
ahnitz, line 17: I think the issue is that we need to set a python version here  now

PR #4257, opened 2023-02-04
add LISA time-domain TDI noise generation and signal injection PE example

WuShichao:
@ahnitz This PR adds an example page to demonstrate how to create an SMBHB injection in the time domain, and then use this mock data to run parameter estimation.

One more thing needs to be done:

- [x] merger `imr_duration` PR for waveform plugin, this is needed by signal injection.


ahnitz:
@WuShichao Can you post a link to what this looks like?

WuShichao:
@ahnitz I will add the link tomorrow. For this to be merged, we also need the imr_duration PR for the plugin to be merged.

WuShichao:
@ahnitz This is the link https://www.atlas.aei.uni-hannover.de/work/shichao.wu/doc/inference/examples/lisa_smbhb_inj_pe.html for SMBHB PE injection example, and https://www.atlas.aei.uni-hannover.de/work/shichao.wu/doc/pycbc.psd.html#module-pycbc.psd.analytical_space is the doc for related code.

ahnitz (approved):
@WuShichao Is the corresponding changes needed for the plugin also merged? If not, what's the holdup?

WuShichao:
@ahnitz Oh, I forgot to update, I will update these this week.

WuShichao:
@ahnitz I think this PR is ready to be merged.

pycbc/waveform/waveform.py
```diff
@@ -760,27 +760,20 @@ def _base_get_td_waveform_from_fd(template=None, rwrap=0.2, **params):
     if nparams['approximant'] not in fd_det:
```
ahnitz, line 760: @WuShichao Actuallym, why is there this check here and two paths? The first should work for both cases as all it does is some checks of the arguments and the calls the function in _filter_time_lengths already. Can this duplication be removed?

PR #4258, opened 2023-02-05
Fixed indexing and autochisq argument in pycbc_multi_inspiral

pannarale:
* Wrapping indexes around for coincs detector frame.
* Calling auto-chisquare on the SNR time series (with desired indices) and not on the array of SNRs of triggers.


spxiwh (approved):
@pannarale I'll largely trust you with bug-fix changes in pycbc_multi_inspiral, but I would feel better with some unit-check in place to avoid potential regression.

PR #4259, opened 2023-02-08
Add pruning to the live single trigger fits

GarethCabournDavies:
I noticed that the trigger fits were being affected by loud signals in the data more than I thought they would be - e.g. the 3-6s and 6-11.5s bins here: 
image

This adds the ability to remove all triggers within a window of the loudest trigger(s) in each bin. This largely suppresses the excess triggers
image

### DIFF
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -65,13 +65,12 @@ parser.add_argument('--prune-loudest', type=int,
                     help="Number of loudest triggers to remove from each "
                          "bin (and all triggers within --prune-window "
                          "seconds around it)")
-parser.add_argument("--prune-window", type=float, default=0.1,
+parser.add_argument("--prune-window", type=float,
                     help="Window (seconds) either side of the --prune-loudest "
-                         "loudest triggers in each duration bin to remove. "
-                         "Default=0.1s")
-parser.add_argument("--minimum-prune-stat", type=float, default=8,
+                         "loudest triggers in each duration bin to remove.")
+parser.add_argument("--prune-stat-threshold", type=float,
                     help="Minimum value of --sngl-ranking that a trigger must "
-                         "have before it is used for pruning. Default=8")
+                         "have before it is used for pruning.")
 parser.add_argument("--fit-threshold", type=float, default=5,
                     help="Lower threshold used in fitting the triggers."
                          "Default 5.")
@@ -90,6 +89,14 @@ cuts.insert_cuts_option_group(parser)
 
 #Add some input sanitisation
 args = parser.parse_args()
+
+# These options are mutually required or not needed
+prune_options = [args.prune_loudest, args.prune_window,
+                 args.prune_stat_threshold]
+
+if any(prune_options) and not all(prune_options):
+        parser.error("Require all or none of --prune-loudest, "
+                     "--prune-window and --prune-stat-threshold")
 
 pycbc.init_logging(args.verbose)
 
@@ -292,7 +299,7 @@ times_to_prune = {ifo: [] for ifo in args.ifos}
 logger = logging.getLogger()
 # Remember current logging level
 level = logger.level
-logger.setLevel(logging.WARN)
+logger.setLevel(logging.debug)
 
 for ifo in events.keys():
     # Sort the events into their bins
@@ -314,7 +321,7 @@ for ifo in events.keys():
 
             # Of these events, are any of them above the specified
             # minimum statistic value?
-            above_stat_min = binned_events[cidx] >= args.minimum_prune_stat
+            above_stat_min = binned_events[cidx] >= args.prune_stat_threshold
             cidx = cidx[above_stat_min]
 
             if args.prune_loudest > cidx.size:
@@ -336,7 +343,7 @@ if args.prune_loudest:
     logging.info("Pruning triggers %.2fs either side of the loudest %d "
                  "triggers in each bin if %s > %.2f", args.prune_window,
                  args.prune_loudest, args.sngl_ranking,
-                 args.minimum_prune_stat)
+                 args.prune_stat_threshold)
     for ifo in args.ifos:
         times = events[ifo].data['end_time'][:]
         outwith_window = np.ones_like(times, dtype=bool)
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -318,7 +318,7 @@ for ifo in events.keys():
             cidx = cidx[above_stat_min]
 
             if args.prune_loudest > cidx.size:
-                # There are fewer triggers than the number specified,
+                # There are fewer clusters than the number specified,
                 # so prune them all
                 times_to_prune[ifo] += list(binned_event_times[cidx])
                 continue
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -341,7 +341,7 @@ if args.prune_loudest:
         for t in times_to_prune[ifo]:
             outwith_window &= abs(times - t) > args.prune_window
             # Need to make an (ever-so-small) correction to the live time
-            live_time[ifo] -= args.prune_window
+            live_time[ifo] -= 2 * args.prune_window
 
         # Save the pruned events for reporting
         within_window = np.logical_not(outwith_window)
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -62,8 +62,8 @@ parser.add_argument("--duration-bin-spacing",
                          "if using --duration-bin-start, --duration-bin-end "
                          "and --num-duration-bins.")
 parser.add_argument('--prune-loudest', type=int,
-                    help="[Maximum] number of loudest trigger clusters to "
-                         "remove from each bin."
+                    help="Maximum number of loudest trigger clusters to "
+                         "remove from each bin.")
 parser.add_argument("--prune-window", type=float,
                     help="Window (seconds) either side of the --prune-loudest "
                          "loudest triggers in each duration bin to remove.")
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -285,10 +285,9 @@ counts = {i: np.zeros(n_bins, dtype=np.float32) for i in args.ifos}
 event_bins = {}
 times_to_prune = {ifo: [] for ifo in args.ifos}
 
-# Note: cluster_over_time has quite a lot of logigng, so turn
-# logging off temporarily
+# Note: turn off logging for cluster_over_time which is quite verbose
 logger = logging.getLogger()
-# Get current logging level
+# Remember current logging level
 level = logger.level
 logger.setLevel(logging.WARN)
 
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -63,7 +63,7 @@ parser.add_argument("--duration-bin-spacing",
                          "and --num-duration-bins.")
 parser.add_argument('--prune-loudest', type=int,
                     help="[Maximum] number of loudest trigger clusters to "
-                         "remove from each bin."
+                         "remove from each bin.")
 parser.add_argument("--prune-window", type=float,
                     help="Window (seconds) either side of the --prune-loudest "
                          "loudest triggers in each duration bin to remove.")
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -325,7 +325,7 @@ for ifo in events.keys():
 logger.setLevel(level)
 
 n_pruned = {ifo: [] for ifo in args.ifos}
-pruned_event_times = {}
+pruned_trigger_times = {}
 if args.prune_loudest:
     logging.info("Pruning triggers in a %.2fs window around the loudest %d "
                  "triggers in each bin", args.prune_window,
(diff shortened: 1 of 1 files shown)


tdent:
Can you confirm that these triggers are only due to GW190814?

GarethCabournDavies:
> Can you confirm that these triggers are only due to GW190814?

Not definitively, but trigger around the event were definitely pruned (GW190814 was at 21:10:39) - pruned times were:
```
In [12]: [lal.GPSToUTC(int(t)) for t in f['L1'].attrs['pruned_times']]
Out[12]: 
[(2019, 8, 14, 17, 26, 2, 2, 226, 0),
 (2019, 8, 14, 19, 9, 35, 2, 226, 0),
 (2019, 8, 14, 19, 0, 22, 2, 226, 0),
 (2019, 8, 14, 21, 10, 39, 2, 226, 0),
 (2019, 8, 14, 21, 10, 39, 2, 226, 0),
 (2019, 8, 14, 17, 14, 7, 2, 226, 0),
 (2019, 8, 14, 7, 5, 50, 2, 226, 0),
 (2019, 8, 14, 5, 40, 41, 2, 226, 0),
 (2019, 8, 14, 21, 34, 1, 2, 226, 0)]
```

So it appeared as the loudest event in two of the bins, but any triggers close to the loudest trigger in any bin are removed

GarethCabournDavies:
Having included an output for times of removed triggers, we see that most of the triggers are from this time, but some come from other times - though the number removed from around GW190814 is an order of magnitude more

image

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -273,30 +282,96 @@ logging.info('Sorting events into template duration bins')
...
 alphas = {i: np.zeros(n_bins, dtype=np.float32) for i in args.ifos}
 counts = {i: np.zeros(n_bins, dtype=np.float32) for i in args.ifos}
-binned_sngl_stats = {ifo: {} for ifo in args.ifos}
+event_bins = {}
+times_to_prune = {ifo: [] for ifo in args.ifos}
+
+# Note: cluster_over_time has quite a lot of logigng, so turn
+# logging off temporarily
+logger = logging.getLogger()
+# Get current logging level
```
tdent, line 291: I think 'Remember current level' would be clearer

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -273,30 +282,96 @@ logging.info('Sorting events into template duration bins')
...
+        for bin_num in range(n_bins):
+            inbin = event_bins[ifo] == bin_num
+
+            binned_events = events[ifo].data[args.sngl_ranking][inbin]
+            binned_event_times = events[ifo].data['end_time'][inbin]
+
+            # Cluster the triggers in time (use 2x the pruning window)
+            # to make sure we aren't just getting the same event N times
+            # in different templates within the same bin
+            cidx = cluster_over_time(binned_events, binned_event_times,
```
tdent, line 309: why are we using 2x the prune window? from what I can see in the cluster code, it looks over a given trigger time +- the given window

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -273,30 +282,96 @@ logging.info('Sorting events into template duration bins')
...
+                                     args.prune_window * 2)
+
+            # Deal with the case where there are fewer clustered triggers
+            # than we would prune
+            if args.prune_loudest > cidx.size:
+                # Prune all triggers
```
tdent, line 315: suggest 'There are fewer triggers than the number specified, so prune them all'

but also .. won't the result of this be a nonsensical fit?

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -273,30 +282,96 @@ logging.info('Sorting events into template duration bins')
...
+pruned_trigger_times = {}
+if args.prune_loudest:
+    logging.info("Pruning triggers in a %.2fs window around the loudest %d "
+                 "triggers in each bin", args.prune_window,
+                 args.prune_loudest)
+    for ifo in args.ifos:
+        times = events[ifo].data['end_time'][:]
+        outwith_window = np.ones_like(times, dtype=bool)
+        for t in times_to_prune[ifo]:
+            outwith_window &= abs(times - t) > (args.prune_window / 2.)
```
tdent, line 336: ah, here you are defining prune_window to be the total window duration over both sides, I think usually we take time windows as specified by +-(duration) around a central time
e.g. for autogating '--autogating-width' is half the actual duration/width of the window

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -273,30 +282,96 @@ logging.info('Sorting events into template duration bins')
...
+            # Need to make an (ever-so-small) correction to the live time
+            live_time[ifo] -= args.prune_window
+
+        # want to get the indices of events within the window to
+        # help us with reporting
```
tdent, line 341: think we could remove this comment and put this line of code into the next block under the comment 'Save the pruned events ..'

bin/live/pycbc_live_single_trigger_fits
GarethCabournDavies: Yes it would, the issue came up mostly in the testing using fewer triggers so this didn't work, so it needed a workaround

This is then taken into consideration later on the `if not np.count_nonzero(inbin)` line

bin/live/pycbc_live_single_trigger_fits
GarethCabournDavies: Ah, I have updated, including updating the help messages to reflect this

bin/live/pycbc_live_single_trigger_fits
GarethCabournDavies: I had thought it best to widen the window a little for clustering over pruning, but it makes sense that any trigger which we would consider a part of the same cluster should also be pruned as such - updated

tdent:
Ah I see, the loudest time per bin is pruned regardless how 'loud' it is.  I am assuming the plots are from running with --prune-loudest=1. 
Here I might be a bit concerned that over short periods of time if we have several bins we might be biasing most of them towards lower backgrounds... would it make sense to only prune triggers louder than a given stat value (eg >8, looking at the fits)?

GarethCabournDavies:
> Ah I see, the loudest time per bin is pruned regardless how 'loud' it is. I am assuming the plots are from running with --prune-loudest=1. Here I might be a bit concerned that over short periods of time if we have several bins we might be biasing most of them towards lower backgrounds... would it make sense to only prune triggers louder than a given stat value (eg >8, looking at the fits)?

I will add in a "--minimum-prune-stat" option, and then remake the plots above

GarethCabournDavies:
>  image

Now is image

Out of the 8 bins, 4 are now unaffected (to 2dp), two change by <0.04, and the other two are the ones with big effects


In fact lots are removed from specifically those two bins
```
2023-02-08T06:49:16.585-08:00 Pruned 13 triggers from L1 bin 0
2023-02-08T06:49:16.585-08:00 Pruned 1 triggers from L1 bin 1
2023-02-08T06:49:16.585-08:00 Pruned 0 triggers from L1 bin 2
2023-02-08T06:49:16.585-08:00 Pruned 1309 triggers from L1 bin 3
2023-02-08T06:49:16.585-08:00 Pruned 949 triggers from L1 bin 4
2023-02-08T06:49:16.585-08:00 Pruned 7 triggers from L1 bin 5
2023-02-08T06:49:16.585-08:00 Pruned 0 triggers from L1 bin 6
2023-02-08T06:49:16.585-08:00 Pruned 0 triggers from L1 bin 7
2023-02-08T06:49:16.585-08:00 Pruned 0 triggers from L1 bin 8
```

Histogram of times of removed events:
image

List of pruning times:
```
2019-08-14 17:26:02
2019-08-14 19:09:35
2019-08-14 21:10:39
2019-08-14 21:10:39
```

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -273,30 +285,98 @@ logging.info('Sorting events into template duration bins')
...
+    logging.info("Pruning triggers in a %.2fs window around the loudest %d "
+                 "triggers in each bin", args.prune_window,
+                 args.prune_loudest)
+    for ifo in args.ifos:
+        times = events[ifo].data['end_time'][:]
+        outwith_window = np.ones_like(times, dtype=bool)
+        for t in times_to_prune[ifo]:
+            outwith_window &= abs(times - t) > args.prune_window
+            # Need to make an (ever-so-small) correction to the live time
+            live_time[ifo] -= args.prune_window
```
tdent, line 345: should this now be 2 * args.prune_window ?

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -60,6 +61,17 @@ parser.add_argument("--duration-bin-spacing",
...
                          "and --num-duration-bins.")
+parser.add_argument('--prune-loudest', type=int,
+                    help="Number of loudest triggers to remove from each "
+                         "bin (and all triggers within --prune-window "
+                         "seconds around it)")
+parser.add_argument("--prune-window", type=float, default=0.1,
+                    help="Window (seconds) either side of the --prune-loudest "
+                         "loudest triggers in each duration bin to remove. "
+                         "Default=0.1s")
+parser.add_argument("--minimum-prune-stat", type=float, default=0,
```
tdent, line 72: I would actually put a default of 8. here, on the philosophy that for low latency the default should be at least close to what we want. @titodalcanton what do you think?

tdent (changes requested):
Minor question on the default min prune stat.

bin/live/pycbc_live_single_trigger_fits
GarethCabournDavies: I was 50-50 on whether to go for 8 or 0, so 8 is fine by me

titodalcanton: Do we need to have defaults here, or does it make sense to require these options to be explicit?

Also, I would rename this option `--prune-stat-threshold` to keep the same prefix as the related options.

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -273,30 +285,98 @@ logging.info('Sorting events into template duration bins')
...
 counts = {i: np.zeros(n_bins, dtype=np.float32) for i in args.ifos}
-binned_sngl_stats = {ifo: {} for ifo in args.ifos}
+event_bins = {}
+times_to_prune = {ifo: [] for ifo in args.ifos}
+
+# Note: turn off logging for cluster_over_time which is quite verbose
+logger = logging.getLogger()
+# Remember current logging level
+level = logger.level
+logger.setLevel(logging.WARN)
```
titodalcanton, line 295: Do not like this here. If `cluster_over_time()` is detrimentally verbose, it means it should use `logging.debug()` rather than `logging.info()`.

bin/live/pycbc_live_single_trigger_fits
tdent: We don't _need_ any defaults, but it is up to you if we want to more follow a philosophy of 'have every setting visible in options' or 'stuff should just work with the defaults, save explicit options for things we will actually want to change'

GarethCabournDavies: I think for live stuff, having setting visible in options is probably the best way to go, just so things are kept in config (in case the defaults ever change). I will remove

bin/live/pycbc_live_single_trigger_fits
GarethCabournDavies: I've updated comments and implementation so args.prune_window is correct now

bin/live/pycbc_live_single_trigger_fits
GarethCabournDavies: I've removed it

I think this is a similar situation to where I removed some logging out of the cuts module before.

The implementation is different in different places - here it is used repeatedly over smaller chunks, but in other places it is used (maybe) once, so the logging is appropriate there. Finding/fixing every usage is going to be difficult, so will separate that into a different PR

bin/live/pycbc_live_single_trigger_fits
tdent: yes, but how much live time is actually being lost?

GarethCabournDavies: Ah, good point - sorry I missed the timing of the comment. Updated now

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -60,6 +61,16 @@ parser.add_argument("--duration-bin-spacing",
                     help="How to set spacing for bank split "
                          "if using --duration-bin-start, --duration-bin-end "
                          "and --num-duration-bins.")
+parser.add_argument('--prune-loudest', type=int,
+                    help="Number of loudest triggers to remove from each "
```
tdent, line 65: seems slightly convoluted - how about '[Maximum] number of loudest trigger clusters to remove from each bin' ?
Maximum because if the loudest are not all above the threshold the number removed will be less

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -60,6 +61,16 @@ parser.add_argument("--duration-bin-spacing",
...
+                         "bin (and all triggers within --prune-window "
+                         "seconds around it)")
+parser.add_argument("--prune-window", type=float,
+                    help="Window (seconds) either side of the --prune-loudest "
+                         "loudest triggers in each duration bin to remove.")
+parser.add_argument("--prune-stat-threshold", type=float,
+                    help="Minimum value of --sngl-ranking that a trigger must "
```
tdent, line 72: suggest 'Minimum --sngl-ranking value to consider a trigger for pruning'

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -79,6 +90,14 @@ cuts.insert_cuts_option_group(parser)
 #Add some input sanitisation
 args = parser.parse_args()
 
+# These options are mutually required or not needed
+prune_options = [args.prune_loudest, args.prune_window,
+                 args.prune_stat_threshold]
```
tdent, line 94: I don't think the threshold should be required. So throw an error if you have loudest and not window, also if you have window or threshold but not loudest.  Should be doable in 3 lines or so ..

GarethCabournDavies: It is required after discussion above

tdent (approved):
Seems OK, test failures appear unrelated .. I will just take another pass through for comment verbosity

tdent (approved):

PR #4260, opened 2023-02-08
fix bug where --y-lim-upper was not actually used

GarethCabournDavies:
Fix bug in pycbc_live_plot_single_trigger_fits


titodalcanton (approved):
Yup, that looks better.

PR #4265, opened 2023-02-15
try pinning pegasus version

ahnitz:


ahnitz:
Ok, I think I may have been going in circles. I just checked the pegasus repository. I had looked in that before and seen the old version there, but I just looked closer and found that they aren't listed in the package index. That might be why I was having trouble get the package installer to install the old package. This test just confirms I even had the syntax right. (I don't use apt on my linux machines).

ahnitz:
Ok, that was in fact the case see here : https://github.com/gwastro/pycbc/actions/runs/4185478119/jobs/7252589632. I think I'll report this to the Pegasus team as it's probably reasonable to be able to install old packages if one wants to.

ahnitz:
@spxiwh I've made a post with what we know atm in the Pegasus slack. For now, I think we merge pinning the Pegasus version so that the tests are run for near-term PRs.

spxiwh (approved):

PR #4266, opened 2023-02-15
only run one test suite per PR

ahnitz:
Taking the advice here

https://stackoverflow.com/questions/66335225/how-to-cancel-previous-runs-in-the-pr-when-you-push-new-commitsupdate-the-curre/67939898#67939898

At the moment, the test suite runs *for each commit*. That means if you identify a problem and push a new commit to a branch, you'll have to wait for the previous commit to finish its tests and free up resources to run. I don't think in general we want that behavior and it would tend to bottleneck running the test suite with actively developed PRs. 

This change to our runs of the test suite would mean that only a single set is run per PR (or tag / branch) at a time. The previous workflows from the prior commit are cancelled when a new commit appears.

### DIFF
--- a/.github/workflows/basic-tests.yml
+++ b/.github/workflows/basic-tests.yml
@@ -2,6 +2,10 @@ name: basic tests
 
 on: [push, pull_request]
 
+concurrency:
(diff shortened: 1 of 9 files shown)


spxiwh (approved):
A nice feature to add!

PR #4272, opened 2023-02-22
changed order of gracedb upload

AnaLorenzoMedina:


pycbc/io/live.py
```diff
@@ -477,70 +548,19 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
-                             gid)
-            except Exception as exc:
-                logging.error(
-                    'Failed to upload source probability results for %s', gid)
-                logging.error(str(exc))
+            extra_strings = [] if extra_strings is None else extra_strings
+            for text in extra_strings:
+                self.gracedb.write_log(gid, text, tag_name=['analyst_comments'])
+        except Exception as exc:
+            logging.error('Something failed during the upload/annotation of '
```
tdent, line 559: Here, since the piece of code that is under the 'try' block is only for annotation, remove 'upload/' from the error message and remove the warning 'The event may have not been uploaded!'

pycbc/io/live.py
```diff
@@ -388,18 +388,89 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
                 logging.info("Tagging event %s as an injection", gid)
 
-            # add info for tracking code version
-            gracedb_tag_with_version(self.gracedb, gid)
-
-            extra_strings = [] if extra_strings is None else extra_strings
-            for text in extra_strings:
-                self.gracedb.write_log(gid, text, tag_name=['analyst_comments'])
         except Exception as exc:
             logging.error('Something failed during the upload/annotation of '
```
tdent, line 392: Here, since the preceding piece of code only deals with the upload, remove '/annotation' from the error message.

titodalcanton:
For reference, the goal of this is making sure the PASTRO_READY label is assigned as soon as possible. Since the plots and comments are for human consumption, they are not critical to the latency of the alerts, and can be uploaded later.

pycbc/io/live.py
tdent: It looks like you have still left the message 'The event may not have been uploaded!' in here, although this would not be an accurate warning.

titodalcanton:
The pie chart code should also be either removed, or used consistently whenever any kind of probability is present. And in the latter case, it should be made after the other plots are made. I have a slight preference for removing it in order to minimize the interaction with GraceDB, but plots are useful… So leaving it up to you.

pycbc/io/live.py
```diff
@@ -388,18 +388,89 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
+                    gid, '2-component p_astro JSON file upload',
+                    filename=self.pastro_file,
+                    tag_name=['sig_info']
+                )
+                logging.info('Uploaded p_astro for %s', gid)
+            except Exception as exc:
+                logging.error('Failed to upload p_astro file for %s', gid)
+                logging.error(str(exc))
+
+        # If there is no multi p_astro, upload source probabilities in JSON
```
tdent, line 438: So following Tito's latest comment this block of code should be moved to after the block dealing with SNR timeseries and PSD plots but just before the 'extra strings' block.  Also remove the 'if self.astro_probs is None' switch, ie if self.prob_file exists then the plot should always be made and uploaded.

pycbc/io/live.py
```diff
@@ -318,15 +318,15 @@ def save(self, fname):
...
-            return
 
         # Save source probabilities in a json file
         if self.probabilities is not None:
             self.prob_file = os.path.join(save_dir, 'src_probs.json')
             with open(self.prob_file, 'w') as probf:
                 json.dump(self.probabilities, probf)
             logging.info('Source probabilities file saved as %s', self.prob_file)
+            # Don't save any other files!
+            return
```
titodalcanton, line 329: I am not clear why you are moving the return from line 322 to here. Could it not be simply removed?

tdent: Because the subsequent code that creates a pastro/pterr only file should only be carried out in the case that there is no source probability calculation and no multi-component pastro.  This is the same behaviour as before except that we are now creating the pie plot and associated json file as you requested.

tdent: I.e. I think we still prefer not to create an extra .json file that is fully redundant with other ones and does not correspond to a plot

titodalcanton: Ok.

pycbc/io/live.py
```diff
@@ -529,18 +531,18 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
-                logging.info('Uploaded p_astro for %s', gid)
-            except Exception as exc:
-                logging.error('Failed to upload p_astro file for %s', gid)
-                logging.error(str(exc))
+        try:
+            # add info for tracking code version
+            gracedb_tag_with_version(self.gracedb, gid)
+
+            extra_strings = [] if extra_strings is None else extra_strings
+            for text in extra_strings:
```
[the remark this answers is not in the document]
tdent: This is not something we have changed in the code, but if the shorter version works for you we can do it

pycbc/io/live.py
```diff
@@ -388,18 +388,53 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
-                self.gracedb.write_log(gid, text, tag_name=['analyst_comments'])
         except Exception as exc:
-            logging.error('Something failed during the upload/annotation of '
+            logging.error('Something failed during the upload of '
                           'event %s on GraceDB. The event may not have been '
                           'uploaded!', fname)
             logging.error(str(exc))
 
+        # Upload em_bright properties JSON
+        if self.hasmassgap is not None:
```
titodalcanton, line 398: You should also check that `gid is not None` here, and in the similar blocks below, as the upload could have failed, but we still want to plot things in that case.

titodalcanton: And you may want to correspondingly remove the "Don't try to do anything else!" block further down.

pycbc/io/live.py
```diff
@@ -477,39 +497,21 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
+            prob_plot = {k: v for (k, v) in self.probabilities.items()
+                         if v != 0.0}
+            labels, sizes = zip(*prob_plot.items())
+            colors = [source_color(label) for label in labels]
+            fig, ax = pl.subplots()
+            ax.pie(sizes, labels=labels, colors=colors, autopct='%1.1f%%',
+                   textprops={'fontsize': 15})
+            ax.axis('equal')
+            fig.savefig(self.prob_plotf)
+            pl.close()
```
titodalcanton, line 514: So it seems the pie chart would still only show the astrophysical classes right? I.e. if a trigger has a total p_astro of 0.5 divided equally between BNS and NSBH, would the pie chart show 0.5 BNS and 0.5 NSBH? This may be ok provided that it is extremely clear that the plot is only showing the astrophysical classes.

titodalcanton: Personally I would plot *every* class including terrestrial, if available and if nonzero, but maybe this can be a later PR if agreed.

tdent: This code to make an 'astrophysical class' pie chart has always done only that, ie it ignores p_terr, the choice of a pie format was made specifically to differentiate it from the standard horizontal bar multi component p astro plot.

tdent: (I don't think we want a pie that just duplicates the content of the gwcelery style bar plot)

pycbc/io/live.py
tdent: Hmm, OK, this aspect (what to do in the absence of gid) should probably be handled by a general upload method, but that is a bit more of a rewrite.  ATM if gid is `None` then the uploads should fail and the error messages will be logged, correct?

titodalcanton: True, assuming `write_log()` does not do anything catastrophically unpleasant when given None… Which is probably the case.

pycbc/io/live.py
titodalcanton: Fair enough, though I might request adding a comment to explicitly state the intention (for when I will forget this).

titodalcanton: Also, I would not move this block from where it was - it should still happen before the "Don't try to do anything else!" check.

pycbc/io/live.py
tdent: I think anything short of a segmentation fault should be covered

tdent:
@titodalcanton can you make the suggestion for removing the redundant `return` statement at the end of the `save()` method?

titodalcanton:
Could you remove this useless `return`: https://github.com/gwastro/pycbc/pull/4272/files#diff-37b3d08eeec68316651629fee5e47e343a415cd2866c702a622dc799951fe27fL338

pycbc/io/live.py
tdent: Well, you asked for this plot to come after the SNR time series and ASD uploads. So then the solution would be to move or remove the `if gid is None` block

tdent: I guess the cleanest way to handle this would be remove that block, but put a test for `gid is not None` in the final annotation.

AnaLorenzoMedina (commented):
Actually is 'if gid is None', sorry I named it wrong

tdent:
I think all the requested changes have been made ..

AnaLorenzoMedina (commented):
'codeclimate' automatic review says there are 2 lines too long, I don't know if that's an issue?

titodalcanton (approved):
Thanks @AnaLorenzoMedina and @tdent, I will merge when tests pass and do a short test at IJCLab before deploying.

PR #4278, opened 2023-03-07
Truncate SNR/FAR for live pastro calcs

tdent:
This will require changes / additional info in the .json config file for the live p astro calculation, eg 
```
"method": "template_param_bins_types_farlim",
...
"limit_snr": 12.0,
"limit_far": 3.17e-10
```
for a limiting FAR of (just over) 1/100yr and SNR of 12.

### DIFF
--- a/pycbc/population/live_pastro_utils.py
+++ b/pycbc/population/live_pastro_utils.py
@@ -27,18 +27,23 @@ def insert_live_pastro_option_group(parser):
 
 # Choices of p astro calc method
 _check_spec = {
-      'template_param_bins': livepa.check_template_param_bin_data,
-      'template_param_bins_types': livepa.check_template_param_bin_data
+    'template_param_bins': livepa.check_template_param_bin_data,
+    'template_param_bins_types': livepa.check_template_param_bin_data,
+    'template_param_bins_types_farlim': \
+        livepa.check_template_param_bin_farlim_data
 }
 
 _read_bank = {
-      'template_param_bins': livepa.read_template_bank_param,
-      'template_param_bins_types': livepa.read_template_bank_param
+    'template_param_bins': livepa.read_template_bank_param,
+    'template_param_bins_types': livepa.read_template_bank_param,
+    'template_param_bins_types_farlim': livepa.read_template_bank_param
 }
 
 _do_calc = {
-      'template_param_bins': livepa.template_param_bin_pa,
-      'template_param_bins_types': livepa.template_param_bin_types_pa
+    'template_param_bins': livepa.template_param_bin_pa,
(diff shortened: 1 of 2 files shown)


pycbc/population/live_pastro_utils.py
```diff
@@ -71,6 +76,24 @@ def __init__(self, specfile, bank):
...
+        if len(trigger_data['triggered']) > 1:
+            farlim = self.spec['limit_far']
+            snrlim = self.spec['limit_snr']
+            if trigger_data['far'] < farlim and \
+                    trigger_data['network_snr'] > snrlim:
+                logging.debug('Truncating FAR and SNR from %f, %f to %f, %f',
+                              trigger_data['far'], trigger_data['network_snr'],
+                              farlim, snrlim)
+                trigger_data['network_snr'] = snrlim
+                trigger_data['far'] = farlim
```
titodalcanton, line 94: My usual suggestion of negating these if statements, so as to return as early as possible and reduce indentation and nesting.

titodalcanton:
One small comment but otherwise this looks good to me. Is there more coming? Should I test this at IJCLab?

tdent:
Hopefully the return logic is still correct after 'negation', please check ..

pycbc/population/live_pastro_utils.py
```diff
@@ -71,6 +76,32 @@ def __init__(self, specfile, bank):
...
+                return trigger_data
+
+            logging.debug('Truncating FAR and SNR from %f, %f to %f, %f',
+                          trigger_data['far'], trigger_data['network_snr'],
+                          farlim, snrlim)
+            trigger_data['network_snr'] = snrlim
+            trigger_data['far'] = farlim
+            return trigger_data
+
+        raise RuntimeError('Number of triggered ifos must be >0 !')
```
titodalcanton, line 103: I am pretty sure `len()` will not return a number less than zero, so this should look like this:
```suggestion
        farlim = self.spec['limit_far']
        snrlim = self.spec['limit_snr']
        # Only do anything if FAR and SNR are beyond given limits
        if trigger_data['far'] > farlim or \
                trigger_data['network_snr'] < snrlim:
            return trigger_data

        logging.debug('Truncating FAR and SNR from %f, %f to %f, %f',
                      trigger_data['far'], trigger_data['network_snr'],
                      farlim, snrlim)
        trigger_data['network_snr'] = snrlim
        trigger_data['far'] = farlim
        return trigger_data
```

tdent: No, the error is in the (unexpected) case where it is an empty list with length 0, which is not negative but also not >0.

titodalcanton:
Testing at IJCLab.

titodalcanton (approved):

PR #4283, opened 2023-03-08
Allow arbitrary labels to events uploaded by PyCBC Live

titodalcanton:
For some applications we need to immediately apply a label to a GraceDB event. This adds a `--gracedb-label` CLI option which enables this. I took the chance and cleaned up the GraceDB interface code and log messages a bit.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -95,6 +95,9 @@ class LiveEventManager(object):
         self.use_date_prefix = args.day_hour_output_prefix
         self.ifar_upload_threshold = args.ifar_upload_threshold
         self.pvalue_livetime = args.pvalue_combination_livetime
+        self.gracedb_server = args.gracedb_server
+        self.gracedb_search = args.gracedb_search
+        self.gracedb_labels = args.gracedb_labels
         self.gracedb_testing = not args.enable_production_gracedb_upload
         self.enable_gracedb_upload = args.enable_gracedb_upload
         self.run_snr_optimization = args.run_snr_optimization
@@ -328,8 +331,9 @@ class LiveEventManager(object):
 
         cmd += '--params-file {} '.format(curr_fname)
         cmd += '--approximant {} '.format(apr)
-        cmd += '--gracedb-server {} '.format(args.gracedb_server)
-        cmd += '--gracedb-search {} '.format(args.gracedb_search)
+        cmd += '--gracedb-server {} '.format(self.gracedb_server)
+        cmd += '--gracedb-search {} '.format(self.gracedb_search)
+        # FIXME should we also pass the GraceDB labels here?
         if not self.gracedb_testing:
             cmd += '--production '
         cmd += '--verbose '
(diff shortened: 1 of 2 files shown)


pycbc/io/live.py
```diff
@@ -373,27 +375,40 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
-            if not hasattr(self, 'gracedb'):
-                from ligo.gracedb.rest import GraceDb
-                gdbargs = {'reload_certificate': True, 'reload_buffer': 300}
-                self.gracedb = GraceDb(gracedb_server, **gdbargs) \
-                    if gracedb_server else GraceDb(**gdbargs)
-            # create GraceDB event
-            group = 'Test' if testing else 'CBC'
-            r = self.gracedb.create_event(group, "pycbc", fname, search).json()
-            gid = r["graceid"]
+            response = self.gracedb.create_event(
```
tdent, line 400: why are you getting rid of the case where self.gracedb was not defined? I guess it is harder to diagnose/debug if the creation of the gdb session can happen in two different places .. anyway, if we can reach this point and self.gracedb does not exist, we might want to produce an exception earlier on in this function

pycbc/io/live.py
```diff
@@ -373,27 +375,40 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
+                group,
+                "pycbc",
+                fname,
+                search=search,
+                labels=labels
+            )
+            json_response = response.json()
+            gid = json_response["graceid"]
```
tdent, line 408: If the variable json_response is only used for this one line, best not to create it. I.e do just `gid = response.json()["graceid"]`

maxtrevor:
I agree with Tom about the json_response , otherwise I think this is good to go

titodalcanton:
This is now tested both with and without labels.

titodalcanton:
@maxtrevor @tdent this just needs one of your approvals.

tdent:
I know, I have been preparing and giving a GW lecture

bin/pycbc_live
```diff
@@ -328,8 +331,9 @@ class LiveEventManager(object):
 
         cmd += '--params-file {} '.format(curr_fname)
         cmd += '--approximant {} '.format(apr)
-        cmd += '--gracedb-server {} '.format(args.gracedb_server)
-        cmd += '--gracedb-search {} '.format(args.gracedb_search)
+        cmd += '--gracedb-server {} '.format(self.gracedb_server)
+        cmd += '--gracedb-search {} '.format(self.gracedb_search)
+        # FIXME should we also pass the GraceDB labels here?
```
tdent, line 336: This is far from self-explanatory - ie why we are not passing default labels to the snr opt instance. Suggest a much more explicit reasoning, eg 'Don't pass labels to the SNR optimization job: currently labels are only required for Early Warning analysis, which does not use SNR opt'

tdent: (That at least will enable someone to come along and work out what the answer to this question should be in future)

tdent, line 336:
```suggestion
        # Don't pass labels to the SNR optimization job:
        # currently labels are only required for Early Warning
        # analysis, which does not use SNR opt
```

tdent (approved):

PR #4287, opened 2023-03-11
Bug fix in pycbc_grb_inj_finder

pannarale:
The trigger dictionary needs the trigger with index eid to be appended, not with index i (which is enumerating the injections)


spxiwh (approved):
I'll trust you on this! Would be good to have some GRB unit tests at some point.

PR #4288, opened 2023-03-13
Pygrb offline workflow

pannarale:
This PR enables the webpage generation workflow pycbc_pygrb_pp_workflow as a SubWorkflow in pycbc_make_offline_grb_workflow.

It is now possible to run a triggered search, with timeslides, on a single sky point, and produce the results webpage.  The idea is to bring this to master so that people can carry out further development, push fixes, etc. on the basis of plots they can automatically generate and see. 

-  An example run script is available here: https://git.ligo.org/ligo-cbc/pycbc-config/-/blob/master/O4/pygrb/dev/make_workflow.sh
- Splitting this down in fewer commits/PR would not allow for a full run/example to work, and it would probably break several tests based on running an executable with `--help`.
- Lots of cleaning up took place (e.g., legacy_grb.py is gone) and some blocks of code are commented out as they do not seem necessary: more cleaning up will definitely follow
- I am aware there are many TODOs and FIXMEs, but I would rather not wait longer to put those in place.  Some of them actually depend upon other development happenning.  Another major thing to do is to convert a lot of the material now in `pycbc/workflow/*pygrb*` to classes (+their methods).
- All pygrb executable are, as always, in a dedicated directory in bin/ with the exception of pycbc_multi_inspiral.  All pygrb utilities are in dedicated .py files containing _pygrb_ in their names.  This ensures PyCBC is not broken and that pygrb developers know where to look to work further on the code.
- Looking forward to code climate response...


pannarale:
[This](https://ldas-jobs.ligo.caltech.edu/~francesco.pannarale/LVC/pygrb_dev_webpage/) is an example of what comes out at the moment when running from scratch on public data around GRB 170817A.  The number of injections is intentionally small, as is the template bank, to expedite the test.
For comparison [here](https://ldas-jobs.ligo.caltech.edu/~francesco.pannarale/LVC/py3_pygrb_webpage_3) is a "full" webpage generated prior to this PR, by providing xml files from an old analysis as input to the webpage workflow generator.  Missing plots/tables in the new webpage will be populated after more work is carried out on pycbc_multi_inspiral and its output files.

pannarale:
Looks like I picked up the changes for https://github.com/gwastro/pycbc/pull/4287 in here too...  I actually spotted that bug by trying out new and different workflows with the infrastructure proposed here.

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -101,21 +101,14 @@ parser.add_argument("-g", "--glitch-check-factor", action="store",
...
 
 # Check options
-if (opts.found_file is None) and (opts.missed_file is None):
-    do_injections = False
-elif (opts.found_file) and opts.missed_file:
-    do_injections = True
-else:
-    err_msg = "Must provide both found and missed file if running injections."
-    parser.error(err_msg)
+do_injections = True if opts.found_missed_file is not None else False
```
titodalcanton, line 111:
```suggestion
do_injections = opts.found_missed_file is not None
```

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -124,8 +117,7 @@ if not opts.newsnr_threshold:
 outdir = os.path.split(os.path.abspath(opts.background_output_file))[0]
 trig_file = opts.offsource_file
 onsource_file = opts.onsource_file
-found_file = opts.found_file
-missed_file = opts.missed_file
```
titodalcanton, line 128: This will not work, will it? `missed file` is still being used in the next line.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -76,8 +76,10 @@ def load_data(input_file, vetoes, ifos, opts, injections=False):
         if snr_type in ['coherent', 'null', 'reweighted']:
             data[snr_type] = trigs_or_injs['network/%s_snr' % snr_type][:]
         elif snr_type == 'single':
-            att = opts.ifo[0].lower()
-            data[snr_type] = trigs_or_injs['%s/snr_%s' % (opts.ifo, att)][:]
+            key = opts.ifo + '/snr_' + opts.ifo.lower()
+            if opts.ifo.lower() != 'h1':
```
titodalcanton, line 80: Why is H1 being hardcoded here? Is this temporary? Add a comment please.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -88,20 +90,37 @@ def load_data(input_file, vetoes, ifos, opts, injections=False):
...
+        data[veto_type] = trigs_or_injs[chisq_key][:]
 
         # Floor single IFO chi-square at 0.005
         numpy.putmask(data[veto_type], data[veto_type] == 0, 0.005)
 
-        data['dof'] = numpy.unique(
-            trigs_or_injs['%s/%s_dof' % (opts.ifo, veto_tags[veto_type])][:])
+        if not injections:
+            data['dof'] = numpy.unique(
+                trigs_or_injs['%s/%s_dof' % (opts.ifo, veto_tags[veto_type])][:])
```
titodalcanton, line 104: Hard to read this. I suggest defining the key in a separate variable and then using that.

bin/pygrb/pycbc_pygrb_plot_null_stats
```diff
@@ -75,8 +75,10 @@ def load_data(input_file, vetoes, ifos, opts, injections=False):
         if null_stat_type == 'null':
             data[null_stat_type] = trigs_or_injs['network/%s_snr' % null_stat_type][:]
         elif null_stat_type == 'single':
-            att = opts.ifo[0].lower()
-            data[null_stat_type] = trigs_or_injs['%s/snr_%s' % (opts.ifo, att)][:]
+            key = opts.ifo + '/snr_' + opts.ifo.lower()
+            if opts.ifo.lower() != 'h1':
```
titodalcanton, line 79: Same comment as above.

bin/pygrb/pycbc_pygrb_plot_snr_timeseries
```diff
@@ -133,34 +132,35 @@ ifos, vetoes = ppu.extract_ifos_and_vetoes(trig_file, opts.veto_files,
...
-inj_data = load_data(found_file, vetoes, injections=True)
+inj_data = load_data(inj_file, vetoes, injections=True)
 
 # Specify HDF file keys for x quantity (time) and y quantity (SNR)
 if snr_type == 'single':
-    x_key = opts.ifo + '/' + 'end_time'
-    y_key = opts.ifo + '/' + 'snr_' + opts.ifo.lower()
+    x_key = opts.ifo + '/end_time'
+    y_key = opts.ifo + '/snr_' + opts.ifo.lower()
+    if opts.ifo.lower() != 'h1':
```
titodalcanton, line 141: See before. Also, this code shows up a few times, maybe a function would be useful?

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -248,11 +248,31 @@ else:
     IPN = False
 
 # Get bank_veto_bank.xml if running bank veto
-# TODO: Finalize config file location of bank veto bank
 if wflow.cp.has_option('workflow-inspiral', 'bank-veto-bank-file'):
     bank_veto_file_path = wflow.cp.get('workflow-inspiral', 'bank-veto-bank-file')
```
spxiwh, line 252: Do this (on `bank_veto_file_path`):

https://github.com/gwastro/pycbc/blob/master/bin/workflows/pycbc_make_uberbank_workflow#L162

it will convert it to a File object (assuming there is only one file). It will work "properly" if the file is in CVMFS, and can pull files from the github repo.

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -261,7 +281,7 @@ if IPN:
     datafind_veto_files.extend(_workflow.FileList([search_pts_file]))
 
 # Make ExtTrig xml file (needed for lalapps_inspinj and summary pages)
-# TODO: how will this work in O4?
+# TODO: how will this work with the new code
```
spxiwh, line 284: If you've moved away from lalapps_inspinj, is it needed at all?

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -88,20 +90,37 @@ def load_data(input_file, vetoes, ifos, opts, injections=False):
...
 
+# Function to calculate chi-square weight for the reweighted SNR
+def new_snr_chisq(snr, new_snr, chisq_dof, chisq_index=4.0, chisq_nhigh=3.0):
+    """Returns the chi-square value needed to weight SNR into new SNR"""
+
+    chisqnorm = (snr/new_snr)**chisq_index
+    if chisqnorm <= 1:
+        return 1E-20
+
+    return chisq_dof * (2*chisqnorm - 1)**(chisq_nhigh/chisq_index)
```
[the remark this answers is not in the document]
spxiwh: The GRB "new_SNR" is the same as all_sky, except the DOF is sometimes different (for the same number of bins), and the "index" and "nhigh" parameters were different.

pycbc/results/legacy_grb.py
```diff
@@ -1,110 +0,0 @@
...
-
-    fig.axes[0].set_title('Science Segments for GRB%s' % trigger_name)
-    plt.tight_layout()
-    fig.subplots_adjust(hspace=0)
-
-    plot_name = 'GRB%s_segments.png' % trigger_name
-    plot_url = 'file://localhost%s/%s' % (out_dir, plot_name)
-    fig.savefig('%s/%s' % (out_dir, plot_name))
-
-    return [ifos, plot_name, extent, plot_url]
```
spxiwh, line 110: Yay for deleting old legacy stuff!

pycbc/workflow/grb_utils.py
```diff
@@ -575,3 +510,404 @@ def create_node(self, input_file, out_dir, ifo_tag, segment, tags=None):
                         os.path.join(out_dir, out_name), tags=tags)
         node.add_output_opt('--output-file', out_file)
         return node, out_file
+
+
+# [Never invoked outside this file]
+def opt_to_file(workflow, section, option):
```
spxiwh, line 516: Consider adding this to core. You could use it elsewhere (see comment on bank_veto_bank, which can basically be replaced with "use this"). If adding to core, please do use a clearer name (e.g. configparser_value_to_file) to distinguish from command line options. Also add in the Inputs and Returns fields in the docstring as for other functions there.

bin/pygrb/pycbc_pygrb_efficiency
pannarale: I cannot try the efficiency code with the new hdf5 output from `pycbc_multi_inspiral` yet (it doesn't contain all the necessary information).  But you are right, this way it is really broken :-). I changed the `missed_file` left to `found_missed_file` so this will fail in a more educated way when we try to use it on more complete hdf5 files.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
pannarale: It is undoing this: https://github.com/gwastro/pycbc/blob/64ecd78aeea6cf2605a5198684c30782c93448c0/pycbc/events/eventmgr.py#L710.  I will add a comment.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
pannarale: Done.

bin/pygrb/pycbc_pygrb_plot_snr_timeseries
pannarale: Commented this too.

bin/pygrb/pycbc_pygrb_plot_null_stats
pannarale: Added a TODO here as well.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
spxiwh: Change eventmgr.py if you need to! The Class you point to was written for the GRB analysis, and is only used by GRB. It looks like it's following an old XML standard (as changing XML column names was always a disaster), and we don't need to follow those rules in O4.

.... I think the idea was that the SNR would be stored in columns `snr_h1`, `snr_h2`, `snr_l`, `snr_v`, and hence H1 was treated specially (and we hope no-one ever tries to run this on old H2 data, I guess!!) I would prefer using the full ifo name (`snr_l1`, `snr_v1`, to avoid having to do conversions all the time).

pycbc/workflow/grb_utils.py
pannarale: @spxiwh, I followed up on this and I am not marking it as resolved as it touches core. I.e., let's look at it carefully :-)

bin/pygrb/pycbc_make_offline_grb_workflow
pannarale: @spxiwh, this led to a major cleanup.  I removed lalapps_inspinj support from PyGRB.  This knocked off the need to support dark_vs_bright jobs, jitter jobs, align spin jobs from PyCBC in general (the related executables are no longer on master or they are in pylal, which we dropped!).  Finally, all this + `configparser_value_to_file` rendered a couple of functions from `grb_utils.py` obsolete.

spxiwh:
@pannarale Any update with how this is going?

pannarale:
@spxiwh , I had missed the notification of your message, sorry!  Going back to my todo list:
* The automated tests were all fine yesterday, with the exception of code climate.
* Codeclimate complains about a bunch of fixme's and todo's (but this PR is intended to also allow others to start working on those) and about some code repetition (but if I were to loop there, it would probably turn into a complaint on too many indentations).
* The most recent pygrb test run with results page is [here](https://ldas-jobs.ligo.caltech.edu/~francesco.pannarale/LVC/pygrb_dev_webpage2/) and it looks in order (empty subpages are exactly where work is needed and are related to the fixme's above)
* The ranking statistic issue flagged above.  You and @titodalcanton noted how newSNR for PyGRB is very close to the all-sky one: do you want me to move the function that returns the chi-square value needed to weight SNR into new SNR (`new_snr_chisq`) from `bin/pygrb/pycbc_pygrb_plot_chisq_veto` to ranking.py?  It basically inverts https://github.com/gwastro/pycbc/blob/master/pycbc/events/ranking.py#L21 with the caveat that the chi-square is set to a small non-zero number because `bin/pygrb/pycbc_pygrb_plot_chisq_veto` produces log-log plots so we are staying away from zeros.

pycbc/results/pygrb_plotting_utils.py
```diff
@@ -70,9 +54,109 @@ def contour_plotter(axis, snr_vals, contours, colors, vert_spike=False):
 
 
 #
-# Used (also) in executables
+# Functions used in executables
 #
 
+# =============================================================================
+# Plot trigger time and offsource extent over segments
+# Copyright (C) 2015 Andrew R. Williamson
```
spxiwh, line 62: Don't think we need this copyright/comment (copyrights can go at the top, if needed) .... I also think the original version of this plot is courtesy of Alex Dietz (it exists even in the pre-coh_PTF GRB search results).)

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -32,18 +31,16 @@
...
 # All/most of these final imports will become obsolete with hdf5 switch
 try:
     from ligo import segments
     from ligo.lw import utils, lsctables
     from ligo.lw.table import Table
     from ligo.segments.utils import fromsegwizard
-    # Handle MultiInspiral xml-talbes with glue,
+    # Handle MultiInspiral xml-tables with glue,
     # as ligo.lw no longer supports them
     from glue.ligolw import lsctables as glsctables
```
spxiwh, line 42: Is `glsctables` used anywhere still?

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -165,43 +159,17 @@ def pygrb_add_bestnr_opts(parser):
...
 
 
-def pygrb_add_missed_injs_input_opt(parser):
-    """Add to parser object the arguments for missed injection file."""
-    if parser is None:
-        parser = argparse.ArgumentParser()
-    parser.add_argument("-m", "--missed-file", action="store",
-                        default=None,
-                        help="Location of the missed injections file.")
+# def pygrb_add_missed_injs_input_opt(parser):
```
spxiwh, line 162: Is it still needed?

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -165,43 +159,17 @@ def pygrb_add_bestnr_opts(parser):
...
+#     """Add to parser object the arguments for missed injection file."""
+#     if parser is None:
+#         parser = argparse.ArgumentParser()
+#     parser.add_argument("-m", "--missed-file", action="store",
+#                         default=None,
+#                         help="Location of the missed injections file.")
 
 
 # =============================================================================
-# Functions to create appropriate FileLists of veto/segment files
```
spxiwh, line 178: There's numerous comments of "never invoked outside this file". I would remove this comment every time it appears. Instead either add the `__all__` variable at the end of the module to ensure these functions are not imported when doing `import pygrb_postprocessing_utils`, or use an underscore in the filename `def _build_veto_filelist` suggests an internal function.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -390,9 +360,10 @@ def load_triggers(input_file, vetoes):
 # * Function to calculate the antenna distance factor
 # =============================================================================
 # The call, e.g., Detector("H1", reference_time=None), will not always work in
-# Python 2.7 as is needs an old version of astropy which cannot download
+# Python 2.7 as it needs an old version of astropy which cannot download
```
spxiwh, line 363: We don't support python2 any more anyway, so remove comment.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -390,9 +360,10 @@ def load_triggers(input_file, vetoes):
...
 # recent enough IERS tables. TEMPORARILY use the default time (GW150914) as
 # reference, thus approximating the sidereal time.
+# [Never invoked outside this file]
 def get_antenna_factors(antenna, ra, dec, geocent_time):
```
spxiwh, line 367: Is there a real need to have this function at all??

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -601,13 +573,13 @@ def extract_basic_trig_properties(trial_dict, trigs, slide_dict, seg_dict,
 # =============================================================================
```
spxiwh, line 573: Remove `get_grb_time` if no longer needed?

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -849,60 +821,60 @@ def mc_cal_wf_errs(num_mc_injs, inj_dists, cal_err, wf_err, max_dc_cal_err):
...
-                multis.extend(multi_inspiral_table)
-            else:
-                multis = multi_inspiral_table
-        except Exception as exc:
-            err_msg = "Unable to read a time-slid multiInspiral table "
-            err_msg += f"from {this_file}."
-            raise RuntimeError(err_msg) from exc
-
-    return multis, time_slides
+# def read_multiinspiral_timeslides_from_files(file_list):
```
spxiwh, line 824: Remove all of this if no loner needed.

pycbc/workflow/grb_utils.py
```diff
@@ -575,3 +408,397 @@ def create_node(self, input_file, out_dir, ifo_tag, segment, tags=None):
...
+    seg_dir = workflow.cp.get('workflow', 'segment-dir')
+    file_names = ["bufferSeg.txt", "offSourceSeg.txt", "onSourceSeg.txt"]
+    seg_files = [os.path.join(seg_dir, fn) for fn in file_names]
+    seg_files = [resolve_url_to_file(sf) for sf in seg_files]
+    seg_files = FileList(seg_files)
+
+    return seg_files
+
+
+# exclude=None, require=None,
```
spxiwh, line 436: Remove commented line here?

spxiwh (commented):
This basically all looks good to me. I'm happy with the CodeClimate status. I would use the `get_new_snr` functions defined in the event module *if* it makes sense. I think most of my comments are "please don't leave old commented code in". @titodalcanton Do you want to give final approval, or are you happy for me to do it?

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -418,43 +448,52 @@ ppDir = os.path.join(currDir, "post_processing")
...
-    # Retrieve style files for webpage
-    summary_files = _workflow.get_coh_PTF_files(wflow.cp, ifos, ppDir,
-                                                summary_files=True)
-
-    pp_files.extend(_workflow.FileList([cp_file]))
-    pp_files.extend(summary_files)
+    # pp_files[1] ALLTIMES_CLUSTERED File
+    # pp_files[3] ONSOURCE_CLUSTERED File
+    # pp_files[-2] FOUNDMISSED FileList
+    # pp_files[-1] FOUNDMISSED-FILTERED FileList
```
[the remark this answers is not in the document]
titodalcanton: This comment makes me wonder if different data structure may be beneficial here in terms of code clarity, maybe a dict?

titodalcanton:
@spxiwh I agree with your comments above and I trust your approval.

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -124,9 +117,8 @@ if not opts.newsnr_threshold:
...
-inj_set_name = os.path.split(os.path.abspath(missed_file))[1].split('INSPINJ')[1].split('_')[1]
+found_missed_file = opts.found_missed_file
+inj_set_name = os.path.split(os.path.abspath(found_missed_file))[1].split('INSPINJ')[1].split('_')[1]
```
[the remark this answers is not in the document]
pannarale: We are not using `lalapps_inspinj` but the naming of these files comes from the line `split_inspinj=${which:pycbc_hdf_splitinj}` in the configuration file.  Shall I go for `splitinj=${which:pycbc_hdf_splitinj}`?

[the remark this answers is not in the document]
spxiwh: The preference would be that we don't assume information is being sent through the filename. So I would recommend a new option for `--injection-set` or similar.

bin/pygrb/pycbc_pygrb_pp_workflow
```diff
@@ -551,29 +315,38 @@ for inj_set in inj_sets:
...
+    inj_layout = list(layout.grouper(files, 2)) # + [(mf_table,), (qf_table,)]
     layout.two_column_layout(out_dir, inj_layout)
 
     # Follow up of loudest N quiet-found injections:
     # qf_h5 includes quiet and missed found injections
-    out_dir = rdir['injections/'+inj_set+'loudest_quiet_found_followups']
-    setup_pygrb_minifollowups(wflow, qf_h5,
-                              'daxes', out_dir,
-                              tags=[inj_set, 'loudest_quiet_found_injs'])
+    #out_dir = rdir['injections/'+inj_set+'loudest_quiet_found_followups']
```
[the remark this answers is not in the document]
pannarale: Turned this in a TODO

bin/pygrb/pycbc_pygrb_pp_workflow
```diff
@@ -597,18 +371,24 @@ for stat in stats:
...
-lofft_layout = list(layout.grouper(files, 2)) + [(lofft_table,)]
-layout.two_column_layout(out_dir, lofft_layout)
+#FIXME
+#html_node, [lont_table, lont_h5, lofft_table, lofft_h5] =\
+#     _workflow.make_pygrb_injs_tables(wflow, out_dir, inj_set=None)
+#html_nodes.append(html_node)
+#lofft_layout = list(layout.grouper(files, 2)) + [(lofft_table,)]
+#layout.two_column_layout(out_dir, lofft_layout)
 
 # Follow up N loudest offsource triggers (parent-child)
```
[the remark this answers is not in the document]
pannarale: I turned this and other blocks of commented out codes into TODOs with a 1 line description.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: This will have to be handled with hdf so it can definitely go.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: Yes, it's gone: this was a very dangerous function with a hard-coded +1s!

pycbc/results/pygrb_plotting_utils.py
pannarale: I had copied that function from a file with Andrew in the Copyright string on the top, so I thought I would keep track of that in this new file.  I now added `courtesy of Alex Dietz` instead.

bin/pygrb/pycbc_make_offline_grb_workflow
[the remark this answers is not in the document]
pannarale: I will always need a `_workflow.FileList([])` but I guess I can build a dictionary in parallel, use it as a "living documentation" structure and go through it when I need individual `pp_files` elements.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: There are some post-processing codes we still need to revamp that in principle do use `glsctables` via `pycbc/results/pygrb_postprocessing_utils.py`.  See: https://github.com/gwastro/pycbc/issues/4419.

This is one of those cases where these scripts have been pulled out of the workflow generator, but you can run them from command line on old xml results file.  However, if we were to remove `glsctables` here, for example, automated tests on those scripts would fail; if we were to remove those scripts entirely, then 80% of their lines would be readded later.  We are in the middle of a transition.  I'd rather keep them, update them according to the issue I linked, and progressively clean up `pycbc/results/pygrb_postprocessing_utils.py`.  This is in line with what was done with `pycbc/results/grb_utils.py` for example.

I hope that makes sense.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: Good point.  I was tracking these because I think some of these functions may end up disappearing, and I wanted to be sure to keep an eye on them.  I opted for single underscore convention.

pannarale:
Thanks, @spxiwh and @titodalcanton.  While I work on the last remaining review items raised above, I wanted to go back to 

> [...] I would use the `get_new_snr` functions defined in the event module _if_ it makes sense. [...]

The thing is that `new_snr_chisq` returns the chi-square value needed to weight SNR into new SNR, it does not reweight the SNR to return new SNR.  So it's consistent with `get_new_snr` but serves a different purpose.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: One line function... so no :-)

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -389,9 +421,7 @@ all_files.extend(_workflow.FileList([inspiral_cache]))
 inspiral_cache_entries = inspiral_files.convert_to_lal_cache()
 inspiral_cache_entries.tofile(open(inspiral_cache.storage_path, "w"))
 
-# LONG TIME SLIDES
-# TODO: Eventually remove this once new time slides are finished
-# Long slides will likely not be done in the future
+# LONG TIME SLIDES: unlikely to be done in the (near) future
```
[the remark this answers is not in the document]
pannarale: I think it's best to remove this block and place a TODO.  It is consistent with what was done for the code lines that were commented out and are now removed.

bin/pygrb/pycbc_make_offline_grb_workflow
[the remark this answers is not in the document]
pannarale: Okay, thinking about this more, it is probably better to document this in `pycbc/workflow/grb_utils.py` where `setup_pygrb_pp_workflow` is defined, and here place a reminder to look over there.

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -248,25 +249,46 @@ else:
     IPN = False
 
 # Get bank_veto_bank.xml if running bank veto
-# TODO: Finalize config file location of bank veto bank
 if wflow.cp.has_option('workflow-inspiral', 'bank-veto-bank-file'):
```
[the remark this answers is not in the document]
pannarale: And would you agree with sticking this (ideally its updated version...) in `pycbc-config`?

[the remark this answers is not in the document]
spxiwh: If the file is small, yes. If it's larger we could use the [soon to exist] large file server that we will use for other large input files.

pannarale:
@spxiwh, I'm handing back to you.  The points I have not marked as resolved are the ones where there is some conversation going on, but there generally are code changes in reaction to those points too.

bin/pygrb/pycbc_make_offline_grb_workflow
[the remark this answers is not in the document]
spxiwh: On second thought (not required for this PR, as it's time to merge this, but something to fix going forward):

I think it's a bad idea to assume anything about the order here, especially things like `[-1]` or `[-2]`. The FileLists were never intended to have a special order. That said, the case of a job that produces multiple output files was never supported well, and dealing with cases like this is ugly ..... Suggestions for improvement here are welcome!

However, I would suggest using one of the methods of the FileList class (https://pycbc.org/pycbc/latest/html/_modules/pycbc/workflow/core.html#FileList) to get these files. It's possible that a new method will be needed, as I think you'll want to search by "description", rather than "tag" (but I'm not sure). I think that `file_list. categorize_by_attr('description')` may produce what @titodalcanton suggests ... and would then be easier and clearer to work with.

spxiwh (approved):
Okay. I think there are still some [identified] things to fix in the GRB workflow, and perhaps a couple of discussions here still open, but I think it is time to merge this to allow work on pygrb to continue.

pannarale:
https://github.com/gwastro/pycbc/issues/4430 tracks the open discussion from this PR.

PR #4289, opened 2023-03-13
add coordinates_space.py

WuShichao:
@ahnitz This PR is for the coordinate transformation between space-borne GW detectors and ground-based ones.

Need to finish more things:
- [x] documentation
- [x] test with .ini in PE
- [x] use some more accurate functions from Astropy
- [x] polarization test
- [x] implement LDC convention correction
- [x] SMBHB PE/likelihood test

In order to make the code easier to review, I separated parts of the code into the following PRs:

1. apply coordinate transform for `fid_params` in the `relbin` model https://github.com/gwastro/pycbc/pull/4396
2. add standard names into `parameters.py` https://github.com/gwastro/pycbc/pull/4397
3. multiband likelihood  https://github.com/gwastro/pycbc/pull/4505


ahnitz:
@WuShichao Don't create a new module for this. I think it would be better to put into the existing ones.

WuShichao:
@ahnitz I have considered this before. I think creating a standalone module for the space detector related is not a bad idea, just like what we did for PSD.

pycbc/coordinates_space.py
```diff
@@ -0,0 +1,196 @@
...
+# =============================================================================
+#
+"""
+This module provides coordinate transform between space-borne detectors and
+ground-based detectors.
+"""
+
+import numpy as np
+from scipy.spatial.transform import Rotation
+from sympy import symbols, nsolve, sin, cos
```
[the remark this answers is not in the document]
WuShichao: Actually, this function is fast, but I will check if this can be done with Astropy.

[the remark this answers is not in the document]
WuShichao: The latest doc for this module https://www.atlas.aei.uni-hannover.de/work/shichao.wu/doc/pycbc.html#module-pycbc.coordinates_space

ahnitz:
And the orbital assumptions part should still be made clear. I think the earth orbit is fine, but the LISA part will need to be changeable.

WuShichao:
@ahnitz OK, I will change the names of functions or variables. For LISA orbit, it depends on two functions: (1) the rotation matrix one specifies the self-rotation and LISA plane orientation; (2) the mass center's motion one specifies LISA barycenter's motion. If we want to make this more flexible, then I think we should read in this information from a file (e.g. a .hdf), this will depend on how we store these information in this file, I prefer leave this to a future PR once we implement our detector response functions.

pycbc/coordinates/__init__.py
```diff
@@ -0,0 +1,36 @@
...
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+"""
+This modules provides functions for basic coordinate transformations, and
+more advanced transformations between ground-based detectors and space-borne
+detectors.
+"""
+
+from pycbc.coordinates.basic import *
```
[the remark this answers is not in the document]
ahnitz: Or give it a useful name

ahnitz:
@WuShichao Please separate the change to coordinates from all the others. I think those should be reviewed separately.

WuShichao:
@ahnitz I can separate, then other parts depending on it will be merged after than coordinates PR (otherwise their build tests will fail).

pycbc/coordinates/base.py
```diff
@@ -140,8 +141,3 @@ def spherical_to_cartesian(rho, phi, theta):
     y = rho * numpy.sin(phi) * numpy.sin(theta)
     z = rho * numpy.cos(theta)
     return x, y, z
-
```
ahnitz, line 143: Don't remove this. It should explicitly say what should be imported by * which is what __all__ does. Your module should also use this feature.

pycbc/transforms.py
```diff
@@ -1468,6 +1468,509 @@ def from_config(cls, cp, section, outputs):
...
+        return self.format_output(maps, out)
+
+    @classmethod
+    def from_config(cls, cp, section, outputs):
+        tag = outputs
+        skip_opts = []
+        additional_opts = {}
+
+        # get custom variable names
+        if cp.has_option("-".join([section, outputs]), 'tc-geo'):
```
ahnitz, line 1576: A lot of this code looks very repetitive and is doing the same type of thing at each line. Can't you compactify this by iterative over a list of the names rather than duplicating for each?

pycbc/transforms.py
```diff
@@ -1468,6 +1468,509 @@ def from_config(cls, cp, section, outputs):
...
+            cp, section, outputs, skip_opts=skip_opts,
+            additional_opts=additional_opts
+        )
+
+
+class LISAToGEO(BaseTransform):
+    """Converts arrival time, sky localization, and polarization angle in the
+    LISA frame to the corresponding values in the geocentric frame."""
+
+    name = "lisa_to_geo"
```
ahnitz, line 1810: Can't you simplify this by using a dictionary of the defaults? And then in your init you don't need separate if statements, you can just iterate over a list? It's the same comment for each transform so I won't repeat it for each.

pycbc/coordinates/space.py
```diff
@@ -0,0 +1,893 @@
...
+#
+"""
+This module provides coordinate transformations related to space-borne
+detectors, such as coordinate transformations between space-borne detectors
+and ground-based detectors.
+"""
+
+import numpy as np
+from scipy.spatial.transform import Rotation
+from scipy.optimize import fsolve
```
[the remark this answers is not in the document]
ahnitz: As mentioned, we should have  unittests for these functions that verify where possibly that they give answers we've validated to be correct and that they are revsesible. Make a new file in the test directory modeled after the test_conversions.py one. It will be automatically run by the test suite. You should first verify it works on your local machine though.

ahnitz (commented):
@WuShichao I think this is getting in good shape. See my few comments on documentation and unit tests as you are adding a large chunk of code here. Once those are addressed, I think we can merge.

pycbc/coordinates/base.py
WuShichao: @ahnitz Will this be duplicated? Because `__init__.py` has `__all__` already.

pycbc/coordinates/space.py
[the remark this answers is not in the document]
WuShichao: @ahnitz I add the comment on https://github.com/gwastro/pycbc/pull/4289/files#diff-7a1da599956f99d94f2c70707533b585de4d61399182d8b58954fe73ed1e15dcR27

pycbc/transforms.py
```diff
@@ -1468,6 +1468,389 @@ def from_config(cls, cp, section, outputs):
...
+            additional_opts=additional_opts
+        )
+
+
+class LISAToSSB(BaseTransform):
+    """Converts arrival time, sky localization, and polarization angle in the
+    LISA frame to the corresponding values in the SSB frame."""
+
+    name = "lisa_to_ssb"
+    default_params_name = {
```
WuShichao, line 1603: @ahnitz I have a question here, you can see the default names of `lisa` and `ssb` are the same; will this cause some potential issues if the user doesn't set custom names for them? Last time in PE slack channel, you both mentioned not hard code det-related names.

pycbc/coordinates/base.py
```diff
@@ -14,7 +14,7 @@
 # You should have received a copy of the GNU General Public License along
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-""" Coordinate transformations.
+""" Base coordinate transformations.
```
ahnitz, line 17: I'm not sure this really clarifies anything? What is meant here? What kind of coordinate transformations are in this module?

pycbc/transforms.py
ahnitz: Yes, it might. Perhaps we should require that the user set these, rather than make an assumption? If we do find a convention that allows an obvious default we can always put one in.

ahnitz:
@WuShichao I couple of the CC comments can be fixed, but some of the ones about snakecase / duplication in this case I think cannot easily. I'd address the ones you can, and then double check your unittests. Once the unittest are in place that at least excersize the key functions, I think this can be merged.

ahnitz (approved):
@WuShichao Provisionally approved, see the other comments, but once those are addressed, I think we can merge this.

pycbc/coordinates/space.py
```diff
@@ -0,0 +1,906 @@
...
+        latitude_ssb_array = np.array([latitude_ssb])
+    if isinstance(polarization_ssb, np.ndarray):
+        polarization_ssb_array = polarization_ssb.copy()
+    else:
+        polarization_ssb_array = np.array([polarization_ssb])
+    num = len(t_ssb_array)
+    t_geo_array, longitude_geo_array = np.zeros(num), np.zeros(num)
+    latitude_geo_array, polarization_geo_array = np.zeros(num), np.zeros(num)
+
+    for i in range(num):
```
ahnitz, line 647: Why are you looping over num here rather than just doing vector / ndarray operations as appropriate? This would be a much slower implementation if you want to calculate for many points.

ahnitz (dismissed):

pycbc/coordinates/space.py
```diff
@@ -0,0 +1,906 @@
...
+        In the unit of 'radian'.
+    latitude_geo : float or numpy.array
+        The latitude of a GW signal in geocentric frame.
+        In the unit of 'radian'.
+    polarization_geo : float or numpy.array
+        The polarization angle of a GW signal in geocentric frame.
+        In the unit of 'radian'.
+    """
+    if isinstance(t_ssb, np.ndarray):
+        t_ssb_array = t_ssb.copy()
```
ahnitz, line 628: Why do you have these copies? That is only  needed if some function you call modifies the vector in place. Otherwise, you don't need this. Most numpy operations  return a new array.

test/test_coordinates_space.py
```diff
@@ -0,0 +1,325 @@
...
+
+        dist_amp = curve_similarity(amp_E3_psi/numpy.mean(amp_E3_psi),
+                                    amp_LISA3_psi/numpy.mean(amp_LISA3_psi))
+        dist_phase = curve_similarity(numpy.array(phase_E3_psi),
+                                      numpy.array(phase_LISA3_psi))
+        if (numpy.abs(dist_amp - 0.2838670151034317) < 1e-2) and \
+            (numpy.abs(dist_phase - 151.77197349820668) < 1e-2):
+            passed = True
+        else:
+            passed = False
```
ahnitz, line 315: Can you add some inline documentation here to explain this test a bit more? It's a bit opaque to someone  not already familiar with what you are trying to do. 

Separately, where do these random numbers come from, for example, how would one update these if needed? It would be good if this is also clear in the inline docs

pycbc/coordinates/space.py
```diff
@@ -0,0 +1,938 @@
...
+        # Earth is moving, when GW arrives at Earth center,
+        # time is t_geo, not t_ssb.
+        p = earth_position_ssb(t_geo)[0]
+        return t_geo - t_ssb - np.vdot(k, p) / c.value
+
+    return fsolve(equation, t_ssb)[0]
+
+
+def t_ssb_from_t_geo(t_geo, longitude_ssb, latitude_ssb,
+                     use_astropy=True, frame=None):
```
ahnitz, line 573: Is the 'use_astropy' option now working either way? If so, maybe this should be part of the unittest to make sure given it gave so many troubles before? Or should we settle on always using astropy?

ahnitz (commented):
@WuShichao I think this PR is ready to go barring the last couple comments and getting the test suite to pass (the priority). Once the test suite passes, give me a final ping and we can merge.

pycbc/coordinates/space.py
WuShichao: @ahnitz In the unittest, using astropy is the default.

test/test_coordinates_space.py
WuShichao: @ahnitz I have added a paragraph, please check.

pycbc/coordinates/space.py
ahnitz: @WuShichao This issue actually still needs to be addressed.

WuShichao:
@ahnitz In order to make `unittest` work, we also need to update the waveform plugin https://github.com/gwastro/BBHX-waveform-model/pull/1

ahnitz:
@WuShichao What is the status of this PR? It seems you need to rebase and address the CC issues still in addition to fixing any remaining tests (assuming they don't pass on a rebase).

WuShichao:
@ahnitz There is still an issue related to the BBHx setup. Other than that, I'm waiting for Connor's tests for SMBHB and debugging the `MultibandRelativeTimeDom`, currently I'm testing it within this PR. Maybe we can separate it as a new PR.

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +605,227 @@ def _loglikelihood(self):
...
+        Returns
+        -------
+        float
+            The value of the log likelihood ratio.
+        """
+        # calculate <d-h|d-h> = <h|h> - 2<h|d> + <d|d> up to a constant
+
+        # note that for SOBHB signals, ground-based detectors dominant SNR
+        # and accuracy of (tc, ra, dec)
+        sh_ground, hh_ground = self.primary_model.loglr(just_sh_hh=True)
```
ahnitz, line 697: Don't explicitly call things ground / space-based

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +605,227 @@ def _loglikelihood(self):
...
+        sh_total = sh_ground + sh_space
+        hh_total = hh_ground + hh_space
+        loglr = self.primary_model.marginalize_loglr(sh_total, hh_total)
+        return loglr
+
+    def _loglikelihood(self):
+        other_models_lognl = 0
+        for lbl, model in self.submodels.items():
+            # Update the parameters of each
+            model.update(**{p.subname: self.current_params[p.fullname]
```
ahnitz, line 722: I think this may not work correctly. You should only update the variable parameters of the primary model.

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +605,227 @@ def _loglikelihood(self):
...
+                cp, 'waveform_transforms')
+            wfoutputs = set.union(*[t.outputs
+                                    for t in waveform_transforms])
+            wfparam_map = map_params(hpiter(wfoutputs, submodel_lbls))
+        else:
+            wfparam_map = {lbl: [] for lbl in submodel_lbls}
+        # initialize the models
+        submodels = {}
+        logging.info("Loading submodels")
+        for lbl in submodel_lbls:
```
ahnitz, line 773: Set primarly model explicitly in the config file. 

[mdoel]
....
primary_model = ce 
other_models = model2 model3

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +605,227 @@ def _loglikelihood(self):
...
+    name = 'multiband_relative_time_dom'
+
+    def __init__(self, variable_params, submodels, **kwargs):
+        super().__init__(variable_params, submodels, **kwargs)
+
+        # We assume the ground-based submodel as the primary model.
+        lbl_list = list(self.submodels.keys())
+        lbl_primary = lbl_list[0]
+        self.primary_model = self.submodels[lbl_primary]
+        if self.primary_model.still_needs_det_response:
```
ahnitz, line 636: You shouldn't need that check

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +605,227 @@ def _loglikelihood(self):
...
+            lbl_primary = lbl_list[1]
+            self.primary_model = self.submodels[lbl_primary]
+        self.other_models = self.submodels.copy()
+        self.other_models.pop(lbl_primary)
+        self.other_models = list(self.other_models.values())
+
+        # marginalize on multiband likelihood level
+        marginalize_phase = self.primary_model.marginalize_phase
+        variable_params, kwargs = self.setup_marginalization(
```
ahnitz, line 645: You don't need to call this.

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +605,227 @@ def _loglikelihood(self):
...
+                               variable_params,
+                               marginalize_phase=marginalize_phase,
```
ahnitz, line 647: You do need to take out marginalized parameters from your variable params and from the prior. The eaisest would be to just copy from the primary model, don't recalculate.

pycbc/inference/models/hierarchical.py
WuShichao: There are no errors from here in my test runs, but I will ensure that.

WuShichao:
> @WuShichao Can you confirm that if the plugin PR, the tests that are currently failing here will work? (e.g. check that there are not unrelated failures). If that is the case, we can merge the plugin PR first and then rerun the tests here and then merge this when they pass.

@ahnitz Yes, we need to review and merge the waveform plugin PR first, because you can see one of the failed tests is caused by not setting the LISA position correctly https://github.com/gwastro/pycbc/actions/runs/6511157215/job/17688598659?pr=4289#step:6:192 so in our polarization test, LISA and ET can not align correctly, I have run the same test (with plugin PR) in a fresh env on ATLAS, it works:
image

test/test_coordinates_space.py
```diff
@@ -0,0 +1,358 @@
...
+                                polarization_lisa, params['t_offset'])
+
+                nparams = {'mass1':params['mass1'], 'mass2':params['mass2'],
+                            'spin1z':params['spin1z'],
+                            'spin2z':params['spin2z'],
+                            'f_lower':params['f_lower']}
+
+                bbhx_fd = get_fd_det_waveform(
+                            ifos=['LISA_A','LISA_E','LISA_T'], **params)
+                lisa_X_fd = -numpy.sqrt(2)/2 * bbhx_fd['LISA_A'] + \
```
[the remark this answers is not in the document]
WuShichao: Thanks for the link; this `test_coordinates_space.py` is just testing for my transform, which overlaps LISA's X channel with ET's E1, when long wavelength approximation holds for the same signal, two V-shape detectors should have very similar detector response, so this can check if my frame transform is correct or not. It's not very matter here, just a test.

test/test_coordinates_space.py
```diff
@@ -0,0 +1,358 @@
...
+import unittest
+from utils import simple_exit
+
+
+seed = 8202
+numpy.random.seed(seed)
+
+# ranges to draw random numbers for each parameter
+RANGES = {
+    "tc_geo" : (1126259462.43, 1426259462.43),
```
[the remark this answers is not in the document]
WuShichao: This part randomly generates parameters set from those ranges, to see if my transform functions are reversible or not because LISA's orbit is periodic, so it's OK to choose a random value. Another reason is that this range covers the GPS time when 3G detectors will be online if I remember correctly.

test/test_transforms.py
```diff
@@ -44,6 +44,12 @@
     "xi1" : (0.0, 1.0),
     "xi2" : (0.0, 1.0),
     "chirp_distance" : (2.0, 10.0),
+    "tc" : (1126259462.43, 1526259462.43),
```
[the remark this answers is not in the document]
WuShichao: Those two are just wide enough time ranges to make sure it works for "any choosen time", also this `tc` is in the SSB frame, not the geocentric one `tc_geo`.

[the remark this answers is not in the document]
ConWea: Thanks for the explanation!

ConWea (approved):

PR #4291, opened 2023-03-16
Flag cases when PyCBC Live's IFAR and/or followup p-values are saturated

titodalcanton:
Loud signals in PyCBC Live can saturate the 2-coincident IFAR and potentially saturate the p-value calculated for a followup detector, in the sense that the on-source SNR is louder than every off-source SNR. It would be useful to report this situation downstream, an immediate example being the P(astro) calculation. So this PR adds a flag that detects this occurrence and propagates the information downstream along with the IFAR and p-value themselves.

Note that most of the change is actually just code refactoring/cleanup and docstrings.

@ahnitz let me know if you think there is a better way to do this.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -245,7 +245,7 @@ class LiveEventManager(object):
             triggers[f'foreground/{ifo}/{name}'] = triggers[key]
 
         # Set the detector-specific fields for which we have data
-        snr_series_peak = snr_series.at_time(
+        snr_series_peak = pvalue_info['snr_series'].at_time(
             pvalue_info['peak_time'],
             nearest_sample=True
         )
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -188,20 +188,26 @@ class LiveEventManager(object):
             if snr_series is not None:
                 out[ifo] = {'snr_series': snr_series}
 
-        # Determine if the other ifos can contribute to the coincident event
+        # Determine if the other ifos can contribute to the coincident event,
+        # if so then update the info in `triggers` appropriately
         for ifo in followup_ifos:
-            snr_series, ptime, pvalue, sigma2 = followup_event_significance(
+            pvalue_info = followup_event_significance(
                 ifo,
                 self.data_readers[ifo],
                 self.bank,
                 template_id,
                 coinc_times
             )
-            if snr_series is not None:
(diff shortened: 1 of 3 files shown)


bin/pycbc_live
```diff
@@ -216,31 +222,47 @@ class LiveEventManager(object):
...
 
-    def get_followup_info(self, coinc_ifo, ifo, triggers, snr_series, ptime,
-                          pvalue, sigma2, recalculate_ifar=False):
-        # Copy the common fields from the other detector
+    def get_followup_info(
+        self,
+        coinc_ifo,
+        ifo,
+        triggers,
+        pvalue_info,
```
tdent, line 230: this is reasonable, but to make the code more readable (eg  repetition of `pvalue_info['peak_time']`) I would unpack the dict entries immediately .. which presumably also makes the diff smaller

tdent:
Does it make sense to deal at the same time with the case that the 2-ifo (without followup) FAR obtained directly from time slides is saturated?

titodalcanton:
@tdent yes, it should probably be included here (I thought it was already being handled, but it does not seem to be).

titodalcanton:
Ok the PR has been rescoped to generally flag saturations either in the original 2-coincidence, or in any of the followup p-values.

tdent:
OK .. so we have 2 separate flags 'ifar_saturated' and 'pvalue_saturated'.  Then I think in 2-ifo time we do a fix for anything with ifar_saturated, in 3-ifo time we fix anything with ifar_saturated _and_ pvalue_saturated. 

Logically it's possible to have pvalue_saturated and _not_ ifar_saturated, but this seems fairly unlikely for the chosen event (?)

titodalcanton:
Right, I think it could *in principle* happen if a source is located near a blind spot of LHO/LLO but not in a blind spot of Virgo: then we may have a relatively quiet LHO-LLO coincidence (with a non-saturated IFAR) and a pretty loud SNR peak in Virgo (resulting in a saturated p-value). It surely seems pretty unlikely though.

tdent (approved):
Looks OK, though I have not thought in detail about how it would hook up to (eg) the p astro calculation.

tdent:
Is this still 'work in progress' ?

titodalcanton:
@tdent it needs more thorough testing than the unit test (hopefully this week).

PR #4296, opened 2023-03-21
add page to help people understand the different likelihood models

ahnitz:
This adds a page to the documentation. 

See here https://www.atlas.aei.uni-hannover.de/work/ahnitz/doctest/_build/html/inference/models.html

This is also linked from the page PyCBC inference page.

Depends on #4203


WuShichao:
@ahnitz I see you added a link to our LISA LDC signal example page, it would be also helpful to add our new LISA signal injection and PE page (once my PR  https://github.com/gwastro/pycbc/pull/4257 has been merged).

cdcapano (approved):
Looks good. It'd be good to add cards for the gated models too, but that can be done another time.

PR #4263, opened 2023-02-10
Added remnant threshold mass to pycbc_brute_bank

MarcoCusinato:
Features added:
- Maximum mass sustained by a Neutron star
- EOS selection
- Possibility to discriminate NS-BH events with a remnant disk mass below a certain threshold (using foucart18 in pycbc/pycbc/neutron_stars/eos_util.py)
- replaced `for` cycles with Python list and dictionary comprehensions (a bit faster)

Two images showing what the threshold does are attached.
The first one is a "plain" bank generation with the following parameters
```
--minimal-match 0.97 \
--tolerance .005 \
--buffer-length 2 \
--sample-rate 2048 \
--tau0-threshold 0.5 \
--approximant IMRPhenomD \
--tau0-crawl 5 \
--tau0-start 0 \
--tau0-end 600 \
--psd-file /home/marco.cusinato/banks/brute_banks/O3a/H1L1-AVERAGE_PSD-1239184818-128800.txt \
--min 1 1 0. 0. \
--max 25 2.8346480922 .998 .05 \
--params mass1 mass2 spin1z spin2z \
--seed 1915 \
--low-frequency-cutoff 27.0
```
In the second one, we consider only BNS events and NS-BH with a remnant mass strictly greater than 0
```
--minimal-match 0.97 \
--tolerance .005 \
--buffer-length 2 \
--sample-rate 2048 \
--tau0-threshold 0.5 \
--approximant IMRPhenomD \
--tau0-crawl 5 \
--tau0-start 0 \
--tau0-end 600 \
--psd-file /home/marco.cusinato/banks/brute_banks/O3a/H1L1-AVERAGE_PSD-1239184818-128800.txt \
--min 1 1 0. 0. \
--max 25 2.8346480922 .998 .05 \
--params mass1 mass2 spin1z spin2z \
--seed 1915 \
--low-frequency-cutoff 27.0\
--remnant-mass-threshold 0.0
```

no_threshold
threshold

PR #4299, opened 2023-03-26
EM brute

pannarale:
Making `remnant_mass_from_mass1_mass2_spherical_spin_eos` and `remnant_mass_from_mass1_mass2_cartesian_spin_eos` functions more flexible so they can be used as constraints with `pycbc_brute_bank`.  This includes adding the new function `ensure_obj1_is_primary` to conversions.py.


pannarale:
This PR was triggered by the suggestion in PR #4263 to leave `pycbc_brute_bank` as it is and get the EM-Bright cut to work via configuration file.
This required making functions dedicated to remnant mass calculations more flexible.  The user can now enforce a threshold mass beyond which the object is considered a BH (so the NSBH is treated as a BBH and "labelled" as dark).  The user can also require that (rather than hitting and error) NSBHs with mass1 < mass2 are treated as NSBH with all properties swapped around.  As far as I can tell, this does not break the old behaviour as the default choices fall back to the original behaviour.  @titodalcanton, @MarcoCusinato: comments are welcome on this!
@ahnitz, @WuShichao: when distributions are sampled, it may happen that mass1 < mass2, hence I created the new function in conversions.py.  If you think it should be relocated, let me know.  As an aside: I was wondering if it would make sense to change [this line](https://github.com/gwastro/pycbc/blob/master/pycbc/distributions/joint.py#L240) to `result[result] &= constraint(params[result])`, the idea being that if a constraint fails, there is no need to check all the following ones.
Last but not least, I have test jobs running...  I'll update when they finish.

pannarale:
Example configuration file syntax
```
[variable_params]
mass1 =
mass2 =
spin1_a =
spin2_a =

[static_params]
eos = 2H

[prior-mass1]
name = uniform
min-mass1 = 1
max-mass1 = 25

[prior-mass2]
name = uniform
min-mass2 = 1
max-mass2 = 2.8

[prior-spin1_a]
name = uniform
min-spin1_a = 0.0
max-spin1_a = 0.998

[prior-spin2_a]
name = uniform
min-spin2_a = 0.0
max-spin2_a = 0.05

[constraint-1]
name = custom
constraint_arg = remnant_mass_from_mass1_mass2_spherical_spin_eos(mass1, mass2, spin1a=spin1_a, spin1pol=0, eos=eos, spin2a=spin2_a, spin2pol=0, swap_companions=True, max_ns_mass=2.8, extrapolate=True) > 0
```

pycbc/conversions.py
```diff
@@ -421,8 +423,60 @@ def lambda_from_mass_tov_file(mass, tov_file, distance=0.):
...
+        output_properties.append(formatreturn(s, input_is_array))
+    # Release output
+    return output_properties
+
+
 def remnant_mass_from_mass1_mass2_spherical_spin_eos(
-        mass1, mass2, spin1a=0.0, spin1pol=0.0, eos='2H'):
+        mass1, mass2, spin1a=0.0, spin1pol=0.0, eos='2H',
+        spin2a=0.0, spin2pol=0.0, swap_companions=False,
+        max_ns_mass=None, extrapolate=False):
```
GarethCabournDavies, line 479: This is super useful! I had a quick look through (as I was considering using an em-bright cut in a new early warning bank).

There is one thing I would suggest: The `max_ns_mass` should be determined by the function you give it as well, i.e. `max_ns_mass` should be the minimum of the supplied value and the maximum supported value in the equation

pannarale: Thanks, @GarethCabournDavies!
This would require reading the EOS data twice: once to create the `mask` with mass2 <= min(max_ns_mass_given_by_the_user, max_ns_mass_given_by_the_eos), and then a second time when the EOS data is needed to calculate the baryonic mass and the compactness.  This is not ideal, but the rise in computational cost of this can be avoided by, say, maintaining a dictionary of maximum masses.
The thing that concerns me more is the case of a user that does not read the warnings I would send out instead of raising a `ValueError` (as is the case right now) when min(max_ns_mass_given_by_the_user, max_ns_mass_given_by_the_eos) = max_ns_mass_given_by_the_eos.  That user could, for example, extract N values of NS mass between 1 and 10 solar masses, calculate the remnant mass with all other parameters being fix, and conclude that a high (wrong!) fraction of NSBHs does not produce a remnant mass.  This is pretty extreme, of course, but if you get around ~3 solar masses things get dangerous.  The first time one extracts NS masses between 1 and 3 Msun, with the 2H EOS, and performs remnant mass calculations, they should learn from a code error that this EOS does not go above >~ 2.8 Msun, and set `max_ns_mass`.  Maybe this needs a better name, e.g., `ns_bh_mass_boundar`.  I just don't want the code to handle NSBHs as BBHs without the user knowing.

ahnitz:
> @ahnitz, @WuShichao:
> 
> * when distributions are sampled, it may happen that mass1 < mass2, hence I created the new function in conversions.py.  If you think it should be relocated, let me know.

I think this is something @cdcapano may have an opinion on, but the impotus seems fine. Keep in mind that we explicilty allow m1 / m2 to be defined without restriction because there are cases where you can have overlapping priors but the objects need to be correctly tracked separately e.g. certain NSBH models or formation-history influenced models were each object may come from a different distribution. 

> * As an aside: I was wondering if it would make sense to change [this line](https://github.com/gwastro/pycbc/blob/master/pycbc/distributions/joint.py#L240) to `result[result] &= constraint(params[result])`, the idea being that if a constraint fails, there is no need to check all the following ones.
> 

I think we want to move to where these can all be vectors, in which case that won't work. @cdcapano may have other thoughts. 

> Last but not least, I have test jobs running... I'll update when they finish.

pannarale:
Figure_1
Figure_2
Figure_3

Looks like it works (i.e., the run completes and the plots look reasonable).  The arguments I passed were `--verbose --buffer-length 2 --sample-rate 2048 --tau0-threshold 0.5 --tau0-crawl 5 --tau0-start 20 --tau0-end 600 --seed 1915 --tolerance .005 --psd-file /home/francesco.pannarale/mc_debug/brute_bank/H1L1-AVERAGE_PSD-1239184818-128800.txt \
--output-file O3abank_IMRPhenomD_H1L1_th_massconstraint.hdf --approximant IMRPhenomD --minimal-match 0.95 --low-frequency-cutoff 40.0 --input-config /home/francesco.pannarale/mc_debug/brute_bank/embright_massconstraint.ini

pycbc/conversions.py
```diff
@@ -446,30 +500,62 @@ def remnant_mass_from_mass1_mass2_spherical_spin_eos(
...
+    spin2a : float, optional
+        The dimensionless magnitude of the spin of mass2. Default = 0.
+    spin2pol : float, optional
+        The tilt angle of the spin of mass2. Default = 0 (aligned w L).
+    swap_companions : boolean, optional
+        If mass2 > mass1, swap mass and spin of object 1 and 2 prior
+        to applying the fitting formula (otherwise fail). Default is False.
+    ns_bh_mass_boundary : float, optional
+        If mass2 is greater than this value, the NS is effectively
+        treated as a BH and the returned value is 0. Default is None.
```
pannarale, line 512: Suggest to the user to open the eos file for the EOS they chose and take the maximum NS mass from there.

pycbc/conversions.py
```diff
@@ -495,15 +581,40 @@ def remnant_mass_from_mass1_mass2_cartesian_spin_eos(
...
+    spin2x : float, optional
+        The dimensionless x-component of the spin of mass2. Default = 0.
+    spin2y : float, optional
+        The dimensionless y-component of the spin of mass2. Default = 0.
+    spin2z : float, optional
+        The dimensionless z-component of the spin of mass2. Default = 0.
+    swap_companions : boolean, optional
+        If mass2 > mass1, swap mass and spin of object 1 and 2 prior
+        to applying the fitting formula (otherwise fail). Default is False.
+    ns_bh_mass_boundary : float, optional
```
pannarale, line 593: Same applies here: suggest where to read off a possible value.

pannarale:
@ahnitz, should we follow @MarcoCusinato's suggestion to replace for cycles with comprehensions? This was one of the things he was doing in the original PR https://github.com/gwastro/pycbc/pull/4263.

ahnitz:
@pannarale Seems reasonable, yeah. Those should be logically equivalent.

MarcoCusinato:
Here is a comparison between this PR and the original one ( #4263). With this method, ~20K have been generated, and therefore there is a more uniform space coverage. The only downside is the generation time for this one, ~120h compared to the ~12h of the original one. However, I have only tested the file without replacing the 'for' cycles, so this should hopefully shave a consistent amount of generation time.
compare_banks_m1m2
compare_banks_m2spin
compare_banks_m1spin
The arguments I passed for the two generations are:
- Original PR: `pycbc_brute_bank.py --verbose --output-file O3a.hdf --minimal-match 0.97 --tolerance .005 --buffer-length 2 
--sample-rate 2048 --tau0-threshold 0.5 --psd-file H1L1-AVERAGE_PSD-1239184818-128800.txt --approximant MRPhenomD 
--tau0-crawl 5 --tau0-start 0 --min 1 1 0 0 --max 25 2.8346480922 .998 .05 --params mass1 mass2 spin1z spin2z --seed 1915
--low-frequency-cutoff 27.0 --remnant-mass-threshold 0.`
- This PR: `pycbc_brute_bank --verbose --output-file O3a_init.hdf --minimal-match 0.97 --tolerance .005 --buffer-length 2 --sample-rate 2048 --tau0-threshold 0.5 --psd-file ../H1L1-AVERAGE_PSD-1239184818-128800.txt --approximant IMRPhenomD --tau0-crawl 5 --tau0-start 0 --tau0-end 600 --seed 1915 --low-frequency-cutoff 27.0 --input-config embright.ini`

pannarale:
Hi @ahnitz , @MarcoCusinato reminded me that you had requested to update the names for spin components in this PR.  Let us know if this is what you had in mind.

pannarale:
@ahnitz, is there a way to complete the 2 required tests that are missing (Build wheels on macos-latest and Build wheels on ubuntu-20.04)?

ahnitz (approved):

PR #4304, opened 2023-04-01
Set a minimal verson of BBHx

WuShichao:
@ahnitz This PR uses a fixed version number for BBHx, so any changes in its main branch will not affect PyCBC build tests. I contacted Katz, and he released a new release (v1.0.5) for PyCBC https://github.com/mikekatz04/BBHx/releases/tag/v1.0.5

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -53,7 +53,7 @@ commands = bash tools/pycbc_test_suite.sh
 deps =
     {[base]deps}
     ; Needed for `BBHx` package to work with PyCBC
-    git+https://github.com/mikekatz04/BBHx.git; sys_platform == 'linux'
+    git+https://github.com/mikekatz04/BBHx.git#1.0.5; sys_platform == 'linux'
     git+https://github.com/ConWea/BBHX-waveform-model.git; sys_platform == 'linux'
 conda_deps=
     mysqlclient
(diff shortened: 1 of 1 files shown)


ahnitz (changes requested):
Don't fix an exact version. If we need to specify a minimum that is Ok, but we absolutely do want to catch if upstream needs to be fixed or vice-versa.

ahnitz:
@WuShichao @spxiwh It hink it might be worth now to think about contributing some things back upstream. Maybe the plugin itself in addition to some unittests and maybe a cibuildwheel config so we don't have to rebuild BBHx most of the time. Thoughts?

WuShichao:
@ahnitz Good suggestions! Katz is on vacation until May, so we have enough time to think about this.

ahnitz:
@WuShichao Does this syntax actually get the required version, or does it just checkout master in all cases?

WuShichao:
@ahnitz I have searched how to specify a minimum release number for pip install from GitHub (not just from master branch, but from a formal release) https://stackoverflow.com/questions/68315468/pip-install-latest-release-from-github. It seems not very elegant.

ahnitz:
@WuShichao Think we should maybe abandon this then and instead just help BBHx to start doing formal releases with wheels. That way we can use regular pip commands.

WuShichao:
@ahnitz That's correct, `wheels` can do that. Let me close this PR, and think about how to contribute to BBHx itself.

Issue #4300, opened 2023-03-28
Do not hardcode mass-gap definition in PyCBC Live


titodalcanton:
The definition of the "mass gap" in PyCBC Live is currently hardcoded in the io/live module, specifically here: https://github.com/gwastro/pycbc/blob/master/pycbc/io/live.py#L271-L273. This is a scientific parameter that should ideally not be defined in that module, and it would be nice to have it controllable by the user in a clear way. Design a clearer way to organize the source classification parameters, e.g. similarly to what is done for P(astro).

tdent:
(Noting that this is only done for an extra 'EmBright' calculation beyond the mass-based classification.)

titodalcanton:
Fixed by #4306.

PR #4306, opened 2023-04-04
Modify HasMassGap to be optional and customizable

veronica-villa:
Proposal of code changes to make the HasMassGap estimation optional (`--enable-embright-has-massgap`) and controllable (`--embright-massgap-max`). I think these options should be in the `pycbc_live` module since this calculation is different to the source classification. 

In case we do not want the mass gap max limit argument to be in `pycbc_live` it can be in `mchirp_area` as `--src-class-mass-gap-max`, but we would need a `--src-class-enable-mass-gap` set to False so the source classification is done without the MG category. 

I can also work in a `src_class_arguments.ini` file as the `p_astro` one.

PR related to the #4300 issue.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -122,7 +122,6 @@ class LiveEventManager(object):
                 self.fu_cores = 1
 
         if args.enable_embright_has_massgap:
-            # check if HasMassGap is enabled and set mass gap max limit
             if args.embright_massgap_max < self.mc_area_args['mass_bdary']['ns_max']:
                 parser.error('MAX_GAP value cannot be lower than MAX_NS limit')
             self.mc_area_args['embright_mg_max'] = args.embright_massgap_max
@@ -894,7 +893,9 @@ parser.add_argument('--snr-opt-label', type=str, default='SNR_OPTIMIZED',
 
 
 parser.add_argument('--enable-embright-has-massgap', action='store_true', default=False,
-                    help='Estimate HasMassGap probability for EMBright info.')
+                    help='Estimate HasMassGap probability for EMBright info. Lower limit '
+                         'of the mass gap is equal to the maximum NS mass used for '
+                         'the source classification.')
 parser.add_argument('--embright-massgap-max', type=float, default=5.0, metavar='SOLAR MASSES',
                     help='Upper limit of the mass gap, used for estimating '
                          'HasMassGap probability.')
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -121,6 +121,12 @@ class LiveEventManager(object):
                                 available_cores)
                 self.fu_cores = 1
 
+        if args.enable_embright_has_massgap:
+            # check if HasMassGap is enabled and set mass gap max limit
+            if args.embright_massgap_max < self.mc_area_args['mass_bdary']['ns_max']:
+                parser.error('MAX_GAP value cannot be lower than MAX_NS limit')
+            self.mc_area_args['embright_mg_max'] = args.embright_massgap_max
+
     def commit_results(self, results):
         logging.info('Committing triggers')
         self.comm.gather(results, root=0)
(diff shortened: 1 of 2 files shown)


bin/pycbc_live
```diff
@@ -855,6 +861,12 @@ parser.add_argument('--snr-opt-timeout', type=int, default=400, metavar='SECONDS
                     help='Maximum allowed duration of followup process to maximize SNR')
 
 
+parser.add_argument('--enable-embright-has-massgap', action='store_true', default=False,
```
tdent, line 864: Is this argument needed? Eg could the code do the HasMassGap calculation only if the embright-massgap-max arg is given? 
I guess the issue is that if massgap-max has a default of 5 then effectively the value is always given, so we can't switch on that.
So in practice we don't have to give the massgap-max option only the 'enable' one?

tdent (commented):
Seems OK as long as the logic of the command line args is clear (ie why we specifically need these options)

bin/pycbc_live
```diff
@@ -855,6 +861,12 @@ parser.add_argument('--snr-opt-timeout', type=int, default=400, metavar='SECONDS
...
+                    help='Estimate HasMassGap probability for EMBright info.')
```
tdent, line 865: Maybe also say that the gap lower limit is equal to the max NS mass?

bin/pycbc_live
```diff
@@ -120,6 +120,12 @@ class LiveEventManager(object):
                                 available_cores)
                 self.fu_cores = 1
 
+        if args.enable_embright_has_massgap:
+            # check if HasMassGap is enabled and set mass gap max limit
```
tdent, line 124: I think the line of comment is not needed, the code is basically self explanatory

tdent:
Can you deal with the 2 codeclimate complaints about lines being over 100 characters long?  (81 characters does not matter but we should shorten anything over 85 or so)

tdent (approved):

pycbc/io/live.py
```diff
@@ -267,15 +267,18 @@ def __init__(self, coinc_ifos, ifos, coinc_results, **kwargs):
...
+            if 'embright_mg_max' in kwargs['mc_area_args']:
+                kwargs['hasmg_args'] = copy.deepcopy(kwargs['mc_area_args'])
+                kwargs['hasmg_args']['mass_gap'] = True
+                kwargs['hasmg_args']['mass_bdary']['gap_max'] = \
+                    kwargs['mc_area_args']['embright_mg_max']
+                self.hasmassgap = calc_probabilities(
+                                      coinc_inspiral_row.mchirp,
+                                      coinc_inspiral_row.snr,
+                                      min(eff_distances),
+                                      kwargs['hasmg_args'])['Mass Gap']
```
titodalcanton, line 279: Is it necessary to store the args as a new key in the `kwargs` dict? It seems to me you could just use a local variable `hasmg_args`, I think that would make the code more readable.

pycbc/io/live.py
```diff
@@ -267,15 +267,18 @@ def __init__(self, coinc_ifos, ifos, coinc_results, **kwargs):
...
+            else:
+                self.hasmassgap = None
         else:
             self.probabilities = None
             self.hasmassgap = None
```
titodalcanton, line 284: Another code-complexity comment: if you initialize `self.probabilities` and `self.hasmassgap` to None before the `if 'mc_area_args' in kwargs:` block, then you can remove both `else` blocks.

titodalcanton:
Hi @veronica-villa I left two minor clean-code suggestions, otherwise this looks ok to me.

titodalcanton:
@veronica-villa could you please rebase on master? There is a change there needed to pass the CI tests.

veronica-villa:
Apologies for the mess @titodalcanton, I did some commits directly in github and it was a bit tricky to do the rebasing, but I hope it's ok now.

PR #4310, opened 2023-04-07
add option to set number of sky points in precalculation

ahnitz:
This is a small change to add an option to set the number of sky points used for the precalculation done when doing sky marginalization. It adds an optional argument 

```
marginalize_sky_initial_samples = 1e7
```


cdcapano (approved):
Looks fine

PR #4311, opened 2023-04-09
Live single trigger fits clustering Bugfix

GarethCabournDavies:
Bug where the `--cluster` option has no effect as the change is not propagated any further

### DIFF
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -194,7 +194,6 @@ for filename in files:
                              events[ifo].data['snr'].size, live_time[ifo])
 
     f = os.path.join(date_directory, filename)
-    skipping_file = False
     # If there is an IOerror with the file, don't fail, just carry on
     try:
         h5py.File(f, 'r')
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -266,7 +266,7 @@ for filename in files:
         # If we are clustering, take the max sngl_ranking value
         if args.cluster:
             max_idx = sngls_value.argmax()
-            triggers_cut = triggers_da.select(max_idx)
+            triggers_da = triggers_da.select(max_idx)
 
         if ifo in events:
             events[ifo] += triggers_da
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -266,7 +266,7 @@ for filename in files:
         # If we are clustering, take the max sngl_ranking value
         if args.cluster:
             max_idx = sngls_value.argmax()
-            triggers_da = triggers_da.select(max_idx)
+            triggers_da = triggers_da.select([max_idx])
 
         if ifo in events:
             events[ifo] += triggers_da
--- a/bin/live/pycbc_live_single_trigger_fits
+++ b/bin/live/pycbc_live_single_trigger_fits
@@ -176,7 +176,7 @@ if not os.path.exists(date_directory):
     raise FileNotFoundError(f"The directory {date_directory} does not exist.")
 
 files = [f for f in os.listdir(date_directory)
-         if args.file_identifier in f]
+         if args.file_identifier in f and f.endswith('hdf')]
 
 events = {}
 counter = 0


bin/live/pycbc_live_single_trigger_fits
```diff
@@ -176,7 +176,7 @@ if not os.path.exists(date_directory):
     raise FileNotFoundError(f"The directory {date_directory} does not exist.")
 
 files = [f for f in os.listdir(date_directory)
-         if args.file_identifier in f]
+         if args.file_identifier in f and f.endswith('hdf')]
```
GarethCabournDavies, line 179: Some safety against files called e.g.: `.H1L1V1-Live-1252782892.07-8.hdf.Oo5mdy`

tdent:
This seems straightforward, I'm somewhat surprised that the didn't have a visibly large effect on outputs ...

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -266,7 +266,7 @@ for filename in files:
         # If we are clustering, take the max sngl_ranking value
         if args.cluster:
             max_idx = sngls_value.argmax()
-            triggers_cut = triggers_da.select(max_idx)
+            triggers_da = triggers_da.select([max_idx])
```
tdent, line 269: why is it better to index on a list containing max_idx rather than max_idx itself?

GarethCabournDavies: If we use `max_idx` on its own, the returned DictArray's `data` is a dictionary of floats, rather than arrays, which breaks the addition of the triggers [here](https://github.com/gwastro/pycbc/blob/master/bin/live/pycbc_live_single_trigger_fits#L272)

bin/live/pycbc_live_single_trigger_fits
titodalcanton: This is good. Those are probably temporary files being written and being renamed with the final file name after the files have been properly written and closed.

titodalcanton: Which makes me wonder, do you do anything important with the `counter` variable below? It will count files even if they are not the files you are looking for.

GarethCabournDavies: This is basically to make it so that progress is reported each 1000 files [check here](https://github.com/gwastro/pycbc/blob/master/bin/live/pycbc_live_single_trigger_fits#L186), so that we still get progress reports with verbose, but without spamming stderr

GarethCabournDavies:
> This seems straightforward, I'm somewhat surprised that the didn't have a visibly large effect on outputs ...

It does, but I think I forgot to test the clustering in #4157 - The effects are here:

No clustering (i.e. same as before bugfix):
image

With clustering after bugfix (note the reduction away from Gaussian at low newsnr, so for practical uses, we need to raise fit threshold):
image


With clustering and raising fit threshold:
image

Add in pruning (shows that it is actually still useful, which I wasn't sure that clustering + pruning together would be useful):
image



Note for later: It may be best to move to a cluster_over_time implementation so that we can use the low-ish fit threshold and have decent statistics, but that doesn't match what is done in the `check_singles` function [here](https://github.com/gwastro/pycbc/blob/1ad800852d850df7199c354814faa03e3beab12b/pycbc/events/single.py#L123)

bin/live/pycbc_live_single_trigger_fits
tdent: OK, I was going to request a comment to explain this. Can add that as a separate commit I guess

tdent: it's a list isn't it? if so would it not be simpler to always initialize the list (`events[ifo] = []`) rather than have a dictionary that may have an entry or may not?

GarethCabournDavies: > it's a list isn't it? if so would it not be simpler to always initialize the list (`events[ifo] = []`) rather than have a dictionary that may have an entry or may not?

triggers_da is a DictArray - this needs the keys from the triggers to be known before going into the loop over files. I suppose we could open file 0 and read them initially (something like events = {ifo: DictArray(data={k: [] for k in keys})}`, but this way is a bit cleaner to code IMO

tdent: Oh, I see (I think). Is this not a shortcoming of DictArray that it allows `data` to be a type that breaks its own `__add__` method?

PR #4313, opened 2023-04-10
Allow precomputation of antenna factors for sky marginalization

ahnitz:
This somewhat simplifies the sky marginalization code and adds the option to precompute antenna factors. Likelihood functions have to access them if they want to use that, so  no changes are needed if they can't (or won't) do so. This has a 20-30% performance improvement as the antenna factor calculations were somewhat dominant. 

Depends on #4310


cdcapano (approved):
Looks fine.

PR #4316, opened 2023-04-11
DictArray fixes: adding to empty, ensuring array

tdent:


pycbc/io/hdf.py
```diff
@@ -158,6 +156,9 @@ def __len__(self):
         return len(self.data[tuple(self.data.keys())[0]])
 
     def __add__(self, other):
+        if self.data is None:  # special case: add to empty DictArray
```
GarethCabournDavies, line 159: This maybe shouldnt be a silent fix, and if we are doing this, I think we should mirror this check for `other`

GarethCabournDavies: I think this should be flagged at the logging.debug level

tdent: I mean it's already a try-except converted into logging.info .. unless the exception for `other=None` is not a KeyError but something else?

GarethCabournDavies: A DictArray can't be initialised if neither `data` or `files` are given (fails the setattr on [this line](https://github.com/gwastro/pycbc/blob/2b4233272e47b17d8b4be3914ad01b289235f3d7/pycbc/io/hdf.py#L150))

GarethCabournDavies: Or are you thinking of a `DictArray(data={})`?

tdent: Ah, so either we should allow that by not doing the setattr in that case, or we should make it fail immediately with a clear error, ie add another check for either one or the other to exist.

tdent: But indeed, initializing with an empty dict would be a reasonable way of setting up a placeholder object ..

GarethCabournDavies: yup that does make sense

GarethCabournDavies (approved):

PR #4317, opened 2023-04-11
Some simplifications to the relative model 

ahnitz:
These are some simplifications to the relative model to help make way for future changes. This combines some of the complication introduced to handles LISA.


ahnitz:
Deferring to #4319

PR #4318, opened 2023-04-12
fix small issues with LISA sensitivity curve

WuShichao:
@ahnitz This PR fixes issues reported by Xisco @frcojimenez:

- [x] Fix Zenodo's `CERTIFICATE_VERIFY_FAILED` issue.
- [x] Fix a `UnboundLocalError: local variable 'base_curve' referenced before assignment` issue.

### DIFF
--- a/pycbc/psd/analytical_space.py
+++ b/pycbc/psd/analytical_space.py
@@ -379,8 +379,8 @@ def averaged_lisa_fplus_sq_approx(f, len_arm=2.5e9):
     cwd = getcwd()
     if path.exists(cwd+"/AvFXp2_Raw.npy") is False:
         url = "https://zenodo.org/record/7497853/files/AvFXp2_Raw.npy"
-        context = ssl._create_unverified_context()
-        request.urlretrieve(url, cwd+"/AvFXp2_Raw.npy", context=context)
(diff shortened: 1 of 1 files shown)


ahnitz (dismissed):

frcojimenez:
@ahnitz @WuShichao The code now works fine. You can close the review. Thanks!

pycbc/psd/analytical_space.py
```diff
@@ -378,6 +379,7 @@ def averaged_lisa_fplus_sq_approx(f, len_arm=2.5e9):
     cwd = getcwd()
     if path.exists(cwd+"/AvFXp2_Raw.npy") is False:
         url = "https://zenodo.org/record/7497853/files/AvFXp2_Raw.npy"
+        ssl._create_default_https_context = ssl._create_unverified_context
```
ahnitz, line 382: This seems a bit dangerous since you are changing this in the library. Every use would be affected. I think you need to use a way that only changes this for this case.  Also, why does this require no verification? Is that a bug on zenodo's end with a certificate? If so, would it resolve itself shortly?

ahnitz (approved):
Assuming this now works, this seems like an ok approach.

frcojimenez:
Hi, fix seems to be working. Thanks 👍

Issue #4193, opened 2022-11-28
Adjust the signal density from kde with extra term representing 'unexpected' events


tdent:
At present the KDE signal density calculation only considers the masses/spins of previous detections, however we want to allow for 'unexpected' events which lie outside the previously seen distribution. 

The simplest way of doing this is adding another distribution, which should cover the whole search space, to the KDE, ie producing a mixture of the previous detection KDE and the added distribution.

tdent:
The idea is that we take a modified signal density which is a mixture of the KDE (with fraction _f_) and another component (with fraction 1-_f_).  This will be a properly normalised density if the KDE is normalised (which we assume to be true) and the other component is also. 

The simplest choice for the extra component is a constant over the parameter space covered by the KDE.  For instance a constant over ln(mchirp), chi_eff, eta.  To normalize this we need to integrate a constant over the whole parameter space, ie find the volume of the space.

tdent:
For specifying how to add the constant extra component we need two extra command line options, one giving the fraction of the mixture that the component should be, the other giving the volume covered by the template bank in the coordinates used by the KDE.

tdent:
Closing as has been addressed by a code change

PR #4324, opened 2023-04-18
unexpected events

PRAVEEN-mnl:


PRAVEEN-mnl:
The template volume calculated here is not appropriate yet.

PRAVEEN-mnl:
#4193

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -26,6 +27,8 @@ parser.add_argument('--signal-file', help='File with parameters of GW signals '
                     'for KDE calculation')
 parser.add_argument('--template-file', required=True, help='Hdf5 file with '
                     'template masses and spins')
+parser.add_argument('--attr-filename', required=True, help='Saved as an attr '
```
tdent, line 30: I am not seeing what this should be used for. Can you give an example of what you might give for this option?  Why is it called 'filename' ?

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -159,12 +193,23 @@ if args.make_template_kde:
                              gl_bandwidth=optbw)
     f_dest.create_dataset("data_kde", data=template_kde)
     f_dest.attrs['stat'] = "template-kde_file"
+    f_dest.attrs['template-file'] = args.attr_filename
+
+
+def new_signal_kde(signal_samples, temp_samples, signal_kde, frac, volume, alp, gl_bandwidth):
```
tdent, line 199: The calculation here looks OK but this function has many arguments which are not used, why are they there?

bin/all_sky_search/pycbc_template_kde_calc
PRAVEEN-mnl: This attr stores the file name used in the signal-kde or template-kde output file. 
for example, in the latest template-kdevals.hdf, this attr gives "sbbh2_Op985.hdf"
& in case of signal-kdevals.hdf, this attr gives "4OGC"

tdent: ok, but I am not seeing why this needs to be a command line option at all if the signal file or template file names are already options. Can you just set the attr to the actual name of the input file?

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -13,6 +13,7 @@
 # Public License for more details.
 
 import numpy, h5py, operator, argparse, logging
+from scipy.integrate import quad
```
tdent, line 16: This is not required any more

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -159,12 +165,23 @@ if args.make_template_kde:
                              gl_bandwidth=optbw)
     f_dest.create_dataset("data_kde", data=template_kde)
     f_dest.attrs['stat'] = "template-kde_file"
+    f_dest.attrs['template-file'] = args.template_file
+
+
+def new_signal_kde(signal_kde, frac, volume):
```
tdent, line 171: the function name is not very descriptive - suggest 'signal_kde_extra_cpt'

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -197,7 +214,9 @@ if args.make_signal_kde:
     logging.info('Evaluating signal KDE')
     signal_kde = kde_awkde(signal_samples, temp_samples, \
                            alp=optalpha, gl_bandwidth=optbw)
-    f_dest.create_dataset("data_kde", data=signal_kde)
+    modified_kde = new_signal_kde(signal_kde, frac=args.extra_cpt_fraction, \
```
tdent, line 217: what if the user doesn't give the extra cpt fraction and volume arguments?  should the code still be able to run?

tdent (approved):

PR #4327, opened 2023-04-20
Allow cuts on a multiple of the median sigma value

GarethCabournDavies:
When running analyses for https://arxiv.org/abs/2203.08545, we saw some times with furious autogating, which destroyed most of the PSD and so the instantaneous sensitive distance was measured to be huge, meaning the statistic blew up.

This cut allows us to discard any triggers at the findtrigs stage where the measured sensitive distance at the time is unphysical, e.g. 50 times the median sigma from triggers in the trigger_merge file.

The check for ReadByTemplate is because we also use the cuts module in the pycbc_live_single_trigger_fits code, but use a dictionary there

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_findtrigs
+++ b/bin/all_sky_search/pycbc_coinc_findtrigs
@@ -181,7 +181,9 @@ if args.randomize_template_order:
     shuffle(template_ids)
     template_ids = template_ids[tmin:tmax]
 else:
-    template_ids = range(tmin, tmax)
+    template_ids = np.array([range(tmin, tmax)])
+
+original_bank_len = len(template_ids)
 
 # Only analyze templates which might have coincs
 if tids_with_trigs is not None:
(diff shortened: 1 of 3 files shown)


bin/all_sky_search/pycbc_coinc_findtrigs
```diff
@@ -196,7 +198,7 @@ template_ids = cuts.apply_template_cuts(
     template_ids=template_ids)
 
 logging.info("%d out of %d templates kept after applying template cuts",
-             len(template_ids), len(trigs.singles[0].bank))
+             len(template_ids), original_bank_len)
```
GarethCabournDavies, line 201: This fixes a small bug in the logging, where the original bank length was not correct

bin/all_sky_search/pycbc_sngls_findtrigs
```diff
@@ -114,14 +116,29 @@ template_ids = cuts.apply_template_cuts(
     template_ids=template_ids)
 
 logging.info("%d out of %d templates kept after applying template cuts",
-             len(template_ids), len(trigs.bank))
+             len(template_ids), original_bank_len)
 
 logging.info('Clustering events over %s s window within each template',
              args.cluster_window)
+
+# Stat class instance to calculate the ranking statistic
```
GarethCabournDavies, line 124: Move the `extra_kwargs` definition out of the loop - not much saving computationally but best practice

PR #4329, opened 2023-04-20
Add corner plotting script for template banks

GarethCabournDavies:
Add a corner plot generator for template banks, allows parameters to be read from the file, converted using the tmpltbank bank_conversions code, and scatter plotted.

Allows a parameter to be used as the color of the scatter points, for parameters to be plotted in log scale, and for parameters to be manually overridden (useful for e.g. working out a template duration if it is not in the bank by adding f_lower)


GarethCabournDavies:
I realised this really should use the `pycbc/results/scatter_histograms.py` module, so closing this until I have implemented thats

PR #4330, opened 2023-04-23
Fix logging bug in PyCBC Live

titodalcanton:
I got this error today:
```
[full traceback messed up by MPI]
UnboundLocalError: local variable 'valid_mask' referenced before assignment
```
This is a case where a variable was renamed (or rather, moved to a dict) but we forgot to update one line using it. We did not notice the problem before because we normally use both the low-latency state vector and DQ flags, so a variable with the same name ends up in the namespace from previous code. This results in the wrong bitmask being logged, but no practical consequences, and the exception is not raised. However, the analysis I was trying to start today uses the DQ flags and *not* the state vector, triggering the exception.


maxtrevor (approved):
Looks good

PR #4332, opened 2023-04-24
Do not try uploading stuff if event creation failed

titodalcanton:
When PyCBC Live fails to create an event on GraceDB due to an occasional error, it logs the error and then happily carries on and tries to upload the various associated data products, as if no error happened. This is harmless, but it creates confusing log messages, so this PR fixes it.


maxtrevor (approved):
Looks good

PR #4334, opened 2023-04-24
more physical rwrap for LISA SMBHB/SOBHB injection (follow-up fix)

WuShichao:
@ahnitz This PR is a follow-up fix for https://github.com/gwastro/pycbc/pull/4322


WuShichao:
@ahnitz Poke :)

PR #4339, opened 2023-04-26
Add bank corner plot

GarethCabournDavies:
Add a corner plot for template banks, using a combination of `bank_conversions.py` from `tmpltbank` and the `create_multidim_plot` function used in inference results plotting

Some examples in https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/add_bank_corner_plot/testing/, with the bash script used to create them (and some specific failure cases) and the output of the script in the .err file


GarethCabournDavies:
Poke @titodalcanton, though if youre busy / dont feel able to comment on this code, then someone else may be able to review

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,282 @@
...
+from pycbc.tmpltbank import bank_conversions as bconv
+from pycbc.pnutils import get_imr_duration
+from pycbc.results.scatter_histograms import create_multidim_plot
+from pycbc.results import metadata
+
+conversion_options = bconv.conversion_options
+
+parser = argparse.ArgumentParser(usage='pycbc_plot_bank_corner [--options]',
+                    description='Plots various parameters against one '
+                          'another for pycbc banks in an hdf file.')
```
titodalcanton, line 38: There is a "feed two birds with one scone" case here: add a docstring to the entire file with this description, then just use `description=__doc__` when you instantiate the ArgumentParser.

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,282 @@
...
+         "the smallest parameter value in the posterior "
+         "will be used.")
+parser.add_argument('--maxs',
+    nargs='+',
+    metavar='PARAM:VAL',
+    default=[],
+    help="Same as mins, but for the maximum values to "
+         "plot.")
+parser.add_argument("--color-parameter",
+    nargs=1,
```
titodalcanton, line 97: Why is this needed if you only take 1 parameter?

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,282 @@
...
+bank = {}
+with h5py.File(args.bank_file, 'r') as bankf:
+    for p in bankf.keys():
+        if not isinstance(bankf[p], h5py.Dataset): continue
+        # Ignore things which aren't numbers, which cannot
+        # be histogrammed - e.g. approximant
+        if not np.issubdtype(bankf[p].dtype, np.number): continue
+        bank[p] = bankf[p][:]
+
+banklen = list(bank.values())[0].size
```
titodalcanton, line 129: This may require more RAM than is reasonable for such an operation. Can you instead set `banklen` in the `for` loop above like `banklen = len(bankf[p])`?

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,282 @@
...
+logging.info("Got %d templates from the bank", banklen)
+
+# If no parameters are given - just plot whatever is in the bank
+if not args.parameters:
+    # Plot anything which is not singular
+    args.parameters = [k for k in bank.keys()
```
titodalcanton, line 135:
```suggestion
    args.parameters = [k for k in bank
```

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,282 @@
...
+                       if len(np.unique(bank[k])) > 1]
+    args.parameters_labels = {k: k for k in list(bank.keys())}
```
titodalcanton, line 137: Can this be simplified like so:
```suggestion
    args.parameters_labels = {k: k for k in bank}
```

titodalcanton: Also, one may perhaps want to not give args.parameters but still give args.parameters_labels?

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,282 @@
...
+    if len(sharex_axes) > 1:
+        for s0, s1 in zip(sharex_axes[:-1], sharex_axes[1:]):
+            s0.sharex(s1)
+
+logging.info("Plot generated")
+fig.set_dpi(args.dpi)
+
+# save
+caption = "Template bank as a corner plot with "
+caption += "scatter points for each waveform."
```
titodalcanton, line 276:
```suggestion
caption = ("Template bank as a corner plot with "
           "scatter points for each waveform.")
```

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,286 @@
...
+
+    logging.info("Converting %s", p)
+    bank[p] = bconv.get_bank_property(p, bank, np.arange(banklen))
+
+# Durations can be negative where get_imr_duration is not perfect - 
+# set a minimum value of 0.15s
+if 'duration' in bank:
+    bank['duration'] = np.maximum(bank['duration'], 0.15)
+if 'template_duration' in bank:
+    bank['template_duration'] = np.maximum(bank['template_duration'], 0.15)
```
titodalcanton, line 191: So is there no way of plotting durations shorter than 150 ms, even if computed correctly? I would say that a plotting code should just plot whatever comes out of the duration calculation…

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,286 @@
...
+if args.mpl_style == 'dark_background':
+    hist_color = 'white'
+else:
+    hist_color = 'black'
+
+# All parameters should be in the bank now, check they are the right size:
+assert all([len(bank[p]) == banklen for p in bank])
+
+# Change the bank to the format required for the plot:
+bank_fa = FieldArray.from_arrays(bank.values(), names=list(bank.keys()))
```
titodalcanton, line 202: It is guarantees that `bank.values()` and `bank.keys()` will always return keys and values in a consistent order?

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,286 @@
...
+logging.info("Generating corner plot")
+
+fig, axis_dict = create_multidim_plot(
+    args.parameters,
+    bank_fa,
+    labels=args.parameters_labels,
+    plot_marginal=args.plot_histogram,
+    plot_scatter=True,
+    plot_contours=False,
+    scatter_cmap="rainbow",
```
titodalcanton, line 228: I would avoid the rainbow colormap because of its [weird lightness behavior](https://matplotlib.org/stable/_images/sphx_glr_colormaps_014.png). Use one of the [perceptually uniform](https://matplotlib.org/stable/_images/sphx_glr_colormaps_008.png) colormaps instead.

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -0,0 +1,286 @@
...
+    show_colorbar=cpar is not False,
+    cbar_label=cpar_label,
+    vmin=mins[cpar] if cpar else 0,
+    vmax=maxs[cpar] if cpar else 0,
+    hist_color=hist_color,
+    mins=mins,
+    maxs=maxs
+)
+
+fig.suptitle(f"Bank length: {banklen}")
```
titodalcanton, line 242: Suggestion for an alternative title:
```suggestion
fig.suptitle(f"{os.path.basename(args.bank_file)} - {banklen} templates")
```

pycbc/tmpltbank/bank_conversions.py
```diff
@@ -88,20 +83,56 @@ def get_bank_property(parameter, bank, template_ids,
...
         values = bank[parameter][:][template_ids]
-
-    # These things may be in the bank, but if not, we need to calculate
-    elif parameter in ['template_duration', 'duration']:
-        if 'template_duration' in bank:
+    # Duration may be in the bank, but if not, we need to calculate
+    elif parameter.endswith('duration'):
+        duration_approximant = bank['template_duration'][:][template_ids] \
+            if 'approximant' in bank \
+            else np.repeat("SPAtmplt", len(template_ids))
```
titodalcanton, line 90: This is effectively changing the default approximant for duration calculation from SEOBNRv4 to SPAtmplt. Not sure this is necessarily a good idea, and if it is, a plotting PR is probably not the place to do it. Can you keep the `duration_approximant` as a kwarg with the current default?

titodalcanton: Also, I guess line 88 meant to use `approximant` as index, not `template_duration`?

bin/plotting/pycbc_plot_bank_corner
GarethCabournDavies: This was just a way of getting around the negative duration issue, and was valid for the purpose I used it for, then forgot to fix it. I'm not sure what is best to say to do for this though

titodalcanton: Treat the duration as any other value. If the user tries to use a log duration scale, and some durations are negative, the code will produce an error. This seems a perfectly reasonable behavior to me.

bin/plotting/pycbc_plot_bank_corner
GarethCabournDavies: > Changed in version 3.7: Dictionary order is guaranteed to be insertion order

From the [documentation](https://docs.python.org/3/library/stdtypes.html#dictionary-view-objects)

bin/plotting/pycbc_plot_bank_corner
GarethCabournDavies: Updated to use plasma - many others have points with brightness that goes too close to white (in my opinion)

GarethCabournDavies:
I think this is ready for re-review / merge now

titodalcanton:
@GarethCabournDavies could you please rerun your examples with the latest version?

GarethCabournDavies:
> @GarethCabournDavies could you please rerun your examples with the latest version?

See https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/add_bank_corner_plot/testing/2023-06-13/

PR #4342, opened 2023-04-27
Threading race condition bugfix for optimize_snr in pycbc_live

xangma:
**Problem**: 
There was a potential race condition in `bin/pycbc_live` where:

- if a coinc or single occurs,
- `optimize_snr_checks == True`, 
- and the thread starts the optimize snr subprocess,

the main thread could have moved the analysis chunk on before the `setup_optimize_snr` function re-reads some data from self - the `event` class instance,. This causes lines like this: `curr_data = self.data_readers[ifo].overwhitened_data(delta_f)` to potentially read the wrong data.

**Solution**:
Split the `setup_optimize_snr` function, which normally accesses/modifies data in `event`, into two functions: `setup_optimize_snr`  and `run_optimize_snr`. The `setup` function reads and saves all the data in the main thread before any GraceDB uploads or optimizing ensuring the correct data is saved before the main thread moves the analysis on. It returns the command to be run as a subprocess and the output directory. The `run ` function takes these as inputs and runs after the event is uploaded.  
One extra modification I had to include is because the `setup_optimize_snr` function was dependent on the `gid` from GraceDB, the `run_optimize_snr` function has to re-open and close the attributes hdf to add the `gid` in.

Happy for any suggestions/modifications. This is the best idea I've had since I found this was happening.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -425,8 +425,8 @@ class LiveEventManager(object):
         gid: str
             GraceDB ID of the candidate
         """
-        with h5py.File(attribute_fname, 'a') as hdfp:
-            if gid is not None:
+        if gid is not None:
+            with h5py.File(attribute_fname, 'a') as hdfp:
                 hdfp['gid'] = gid
         log_fname = os.path.join(out_dir_path, 'optimize_snr.log')
 
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -633,10 +633,12 @@ class LiveEventManager(object):
             comment = comment.format(ifo, ppdets(followup_ifos))
 
             # Has a coinc event at this time been uploaded recently?
-            # If so, skip upload
+            # If so, skip upload - Note that this means that we _always_
+            # prefer an uploaded coincident event over a single
             coinc_tdiffs = abs(event.merger_time -
                                numpy.array(self.last_few_coincs_uploaded))
             nearby_coincs = coinc_tdiffs < 0.1
+
             if any(nearby_coincs):
                 logging.info(
                     "Single-detector candidate at time %.3f was already "
@@ -656,14 +658,14 @@ class LiveEventManager(object):
             if not upload_checks:
                 event.save(fname)
 
-            if nearby_coincs:
-                # Don't save the snr optimization stuff for singles where
-                # there is already a coinc
-                continue
-
             # Save the event info/data and what _would_ be run for SNR optimizer,
             # even if not running it - do this before the thread so no
             # data buffers move on in a possible interim period
+
+            if nearby_coincs:
+                # Don't even save the snr optimization stuff for singles
+                # where there is already a coinc
+                continue
 
             # Which IFOs were active?
             live_ifos = [ifo for ifo in sld if 'snr_series' in sld[ifo]]
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -662,7 +662,7 @@ class LiveEventManager(object):
             # even if not running it - do this before the thread so no
             # data buffers move on in a possible interim period
 
-            if nearby_coincs:
+            if any(nearby_coincs):
                 # Don't even save the snr optimization stuff for singles
                 # where there is already a coinc
                 continue
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -284,7 +284,7 @@ class LiveEventManager(object):
             triggers[f'foreground/pvalue_{ifo}_saturated'] = pv_sat
 
     def setup_optimize_snr(
-        self, results, live_ifos, triggering_ifos, fname, gid
+        self, results, live_ifos, triggering_ifos, fname
     ):
         """Setup and start the network SNR optimization process for a
         candidate event. See arXiv:2008.07494 for details.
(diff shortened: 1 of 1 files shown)


bin/pycbc_live
```diff
@@ -514,14 +515,21 @@ class LiveEventManager(object):
...
+                # Run the setup to gather/save all the data before threading
+                cmd, out_dir_path = self.setup_optimize_snr(
+                    coinc_results,
+                    live_ifos,
+                    coinc_ifos,
+                    fname,
+                )
             thread_args = (
-                event, fname, comment, coinc_results, live_ifos, coinc_ifos,
+                event, fname, comment, cmd, out_dir_path
```
GarethCabournDavies, line 532:
```suggestion
                event, fname, comment, cmd, out_dir_path,
```

xangma: Ah, yes, sorry, I fixed that in my test but forgot to commit again.

xangma: Ok fixed, and rebased etc. Always a wild ride for me.

xangma:
Is there any appetite for putting a comment in `upload_in_thread` stating something like this? :

> #Warning: there be dragons ahead. Within threads created by the main thread, race conditions must be considered when accessing any data from unlocked objects that the main thread could be changing - e.g `self.data_readers`.

bin/pycbc_live
```diff
@@ -276,7 +276,7 @@ class LiveEventManager(object):
             triggers[f'foreground/pvalue_{ifo}_saturated'] = pv_sat
 
     def setup_optimize_snr(
-        self, results, live_ifos, triggering_ifos, fname, gid
+        self, results, live_ifos, triggering_ifos, fname
     ):
         """Setup and start the network SNR optimization process for a
```
GarethCabournDavies, line 281:
```suggestion
        """Setup the network SNR optimization process for a
```

tdent:
Seems like we could now use the MDC to test this, at least for a short time, given that it served its pre-O4 review purpose.

titodalcanton:
I plan to finally look into this next week.

GarethCabournDavies:
Poke on this

GarethCabournDavies:
Poke on this - this would be good to get in soon(ish)

titodalcanton:
No, only stuff on master is being tested on the MDC. This is now (again) near the top of my priority list.

bin/pycbc_live
```diff
@@ -403,6 +401,32 @@ class LiveEventManager(object):
...
+        out_dir_path: os.path or str
+            Place for storing the SNR optimization results and log
+        attribute_fname: str
+            The filename of the attributes of the candidate
+        gid: str
+            GraceDB ID of the candidate
+        """
+        with h5py.File(attribute_fname, 'a') as hdfp:
+            if gid is not None:
+                hdfp['gid'] = gid
```
titodalcanton, line 429: The `with` and `if` blocks should be swapped. We do not want to open the file unless we are sure we are going to use it.

bin/pycbc_live
```diff
@@ -108,6 +108,8 @@ class LiveEventManager(object):
         # Keep track of which events have been uploaded
         self.last_few_coincs_uploaded = []
 
+        # Give this a nominal value, it is requried but not used
+        self.fu_cores = None
```
titodalcanton, line 112: Typo: requried -> required. Also, there should probably be a `if self.fu_cores is not None` condition down below in `setup_optimize_snr` where `fu_cores` is used, otherwise I do not think the resulting command (`--cores None`) would be valid.

bin/pycbc_live
```diff
@@ -605,17 +646,33 @@ class LiveEventManager(object):
...
+            if hasattr(event, 'p_terr') and event.p_terr is not None:
+                single['p_terr'] = event.p_terr
+
+            # Do the optimizer setup
+            cmd, out_dir_path = self.setup_optimize_snr(
+                single,
+                live_ifos,
+                [ifo],
+                fname,
+            )
```
titodalcanton, line 671: This will run even if there was a nearby coinc, and will generate a bunch of files as a consequence, which does not sound like a desirable behavior. Maybe this call should go into the `if upload_checks or optimize_snr_checks` condition below? And I guess the same should happen for the similar code above (watch out for code duplication…)

titodalcanton: I.e. what is the use case for doing anything if we are neither uploading nor running the SNR optimizer?

titodalcanton: Hold on, I am conflating two different decisions in my two comments above. One decision is what to do in terms of SNR optimization of singles that have a nearby coinc. The other decision is what to do in terms of generating SNR optimization input if we are neither uploading nor SNR-optimizing.

titodalcanton: Ok, I think we generally do want to call `setup_optimize_snr` even if `upload_checks` and `optimize_snr_checks` are both false: this would allow us to generate some files for testing the SNR optimizer later on.

However, if there is a nearby coinc, then `upload_checks` and `optimize_snr_checks` will both be false, and I argue that we should *not* call `setup_optimize_snr`.

GarethCabournDavies: I have added in a `if nearby_coincs: continue` statement so that this bit isn't reached in that case.

GarethCabournDavies: I don't see the "similar code above" though, I think that setup_optimize_snr should be called for _all_ coincs, right?

titodalcanton: Yes nevermind, that was a thinko.

titodalcanton (approved):
I have no more comments, let's merge this and get it running on the MDC.

PR #4347, opened 2023-05-03
page_foreground bugfix

GarethCabournDavies:
Bug found by @PRAVEEN-mnl, where the time was not correct in the page_foreground hdf outputs, this was as the dtype was being set to be a float32 (I don't know why I thought this was a good idea before), but that isn't suitable for gps times

### DIFF
--- a/pycbc/io/hdf.py
+++ b/pycbc/io/hdf.py
@@ -1031,35 +1031,24 @@ class ForegroundTriggers(object):
         ofd['time'] = time
 
         if self._inclusive:
-            ifar = self.get_coincfile_array('ifar')
-            ofd['ifar'] = ifar
+            ofd['ifar'] = self.get_coincfile_array('ifar')
+            ofd['p_value'] = self.get_coincfile_array('fap')
 
-        ifar_exc = self.get_coincfile_array('ifar_exc')
-        ofd['ifar_exclusive'] = ifar_exc
-
-        if self._inclusive:
-            fap = self.get_coincfile_array('fap')
-            ofd['p_value'] = fap
-
-        fap_exc = self.get_coincfile_array('fap_exc')
-        ofd['p_value_exclusive'] = fap_exc
+        ofd['ifar_exclusive'] = self.get_coincfile_array('ifar_exc')
+        ofd['p_value_exclusive'] = self.get_coincfile_array('fap_exc')
 
         # Coinc fields
         for field in ['stat']:
-            vals = self.get_coincfile_array(field)
-            ofd[field] = vals
+            ofd[field] = self.get_coincfile_array(field)
 
         logging.info("Outputting template information")
         # Bank fields
         for field in ['mass1','mass2','spin1z','spin2z']:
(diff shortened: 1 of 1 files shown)


pycbc/io/hdf.py
```diff
@@ -1028,41 +1028,38 @@ def to_coinc_hdf_object(self, file_name):
         # Some fields are special cases
         logging.info("Outputting search results")
         time = self.get_end_time()
-        ofd.create_dataset('time', data=time, dtype=np.float32)
+        ofd['time'] = time
 
         if self._inclusive:
```
tdent, line 1033: would it be neater to roll both the `if self._inclusive` switches into one?

pycbc/io/hdf.py
```diff
@@ -1028,41 +1028,38 @@ def to_coinc_hdf_object(self, file_name):
...
         fap_exc = self.get_coincfile_array('fap_exc')
-        ofd.create_dataset('p_value_exclusive', data=fap_exc,
-                           dtype=np.float32)
+        ofd['p_value_exclusive'] = fap_exc
 
         # Coinc fields
         for field in ['stat']:
             vals = self.get_coincfile_array(field)
-            ofd.create_dataset(field, data=vals, dtype=np.float32)
+            ofd[field] = vals
```
tdent, line 1050: I don't think we need to define 'vals' any more, just assign the dataset directly

pycbc/io/hdf.py
GarethCabournDavies: fair, will do

pycbc/io/hdf.py
```diff
@@ -1090,20 +1086,18 @@ def to_coinc_hdf_object(self, file_name):
                 vals = vals_valid[ifo][0]
                 valid = vals_valid[ifo][1]
                 vals[np.logical_not(valid)] = -1.
-                ofd.create_dataset(ifo + '_'+field, data=vals,
-                                   dtype=np.float32)
+                ofd[f'{ifo}_{field}'] = vals
```
tdent, line 1089: maybe good to put a line of comment before the 'vals_valid' lines above to explain what this is intended to do?

pycbc/io/hdf.py
```diff
@@ -1028,41 +1028,38 @@ def to_coinc_hdf_object(self, file_name):
         # Some fields are special cases
         logging.info("Outputting search results")
         time = self.get_end_time()
-        ofd.create_dataset('time', data=time, dtype=np.float32)
+        ofd['time'] = time
```
tdent, line 1031: as for other fields, no need to define a variable which is only used once

tdent: ok .. maybe put a brief inline comment on that, eg `time will be used to determine active ifos`

tdent (approved):

pycbc/io/hdf.py
GarethCabournDavies:
```suggestion
        # time will be used later to determine active ifos
        ofd['time'] = time
```

PR #4350, opened 2023-05-03
more physical rwrap for LISA SMBHB/SOBHB injection (follow-up fix 2)

WuShichao:
@ahnitz This PR is a follow-up fix for https://github.com/gwastro/pycbc/pull/4322 and https://github.com/gwastro/pycbc/pull/4334 
I found that when `rwrap` is shorter than 0.2s, the TD waveform for ground-based detectors is still a little bit wrapped around, which is not an issue in https://github.com/gwastro/pycbc/pull/4322 because that PR has a check https://github.com/gwastro/pycbc/pull/4322/files#diff-9e155d2730c717f169ec7ea98eae6bd053eeae580a7e8643b09084c4af7fdbb4R767, while this check is missing in https://github.com/gwastro/pycbc/pull/4334


ahnitz (commented):
@WuShichao Can you please scale the 0.2 by total mass then, so this works properly for higher mass  sources? Maybe remove the pykerr parts if this is now triggering alwasy?

WuShichao:
> @WuShichao Can you please scale the 0.2 by total mass then, so this works properly for higher mass sources? Maybe remove the pykerr parts if this is now triggering alwasy?

This PR works fine for both LISA and aLIGO sources, so I guess we don't need to modify it again? If we rescale 0.2, then we don't know the total mass of this 0.2, and also can't include the spin effect.

pycbc/waveform/waveform.py
```diff
@@ -766,6 +766,8 @@ def _base_get_td_waveform_from_fd(template=None, rwrap=None, **params):
                 mass1=nparams['mass1'], mass2=nparams['mass2'],
                 spin1z=nparams['spin1z'], spin2z=nparams['spin2z'])
             rwrap = tau_from_final_mass_spin(m_final, spin_final) * 10
+            if rwrap < 0.2:
```
ahnitz, line 769: Why not do something like 

rwrap += max(0.2, 0.2 * (mtotal - 60)

That would make sure this is a minimum while providing some safety scaling.

WuShichao: Here I use 5 seconds as the minimal instead, so don't assume any rescaling relation for BBHs in ground-based detectors, 5s is long enough for XG, and still not a computational burden for the current 2G study:
image

WuShichao: @ahnitz Poke :)

ahnitz (approved):

PR #4351, opened 2023-05-03
Update at_time method to allow float / vector types and extrapolation

ahnitz:
This mergers the at_time and at_times which had mixed functionality previously. Some vector functionality was only in at_time for example when at_times was added. This addresses that incongruity and so at_times now points to at_time.  And extrapolation fill value is also now possible to give. 

This also adds a unittest for this function.

### DIFF
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -240,7 +240,7 @@ class TimeSeries(Array):
     sample_times = property(get_sample_times,
                             doc="Array containing the sample times.")
 
-    def at_time(self, time, nearest_sample=False, interpolate=None):
+    def at_time(self, time, nearest_sample=False, interpolate=None, extrapolate=None):
         """ Return the value at the specified gps time
 
         Parameters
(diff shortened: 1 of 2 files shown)


pycbc/types/timeseries.py
```diff
@@ -251,51 +252,64 @@ def at_time(self, time, nearest_sample=False, interpolate=None):
         interpolate: str, None
             Return the interpolated value of the time series. Choices
             are simple linear or quadratic interpolation.
+        extrapolate: str or float, None
```
maxtrevor, line 255: Why is extrapolate allowed to be a str? Wouldn't giving a str trigger the ValueError on line 280?

test/test_timeseries.py
```diff
@@ -477,6 +477,33 @@ def test_duration(self):
             self.assertAlmostEqual(self.b.duration, 0.3)
             self.assertAlmostEqual(self.bad3.duration, 0.6)
 
+    def test_at_time(self):
```
maxtrevor, line 480: I don't really know how our github tests work, will this be run automatically everytime a commit is pushed?

pycbc/types/timeseries.py
```diff
@@ -251,51 +252,64 @@ def at_time(self, time, nearest_sample=False, interpolate=None):
...
             xl = self.data[i - 1] - c
             a = 0.5 * (xr + xl)
             b = 0.5 * (xr - xl)
             ans = a * di**2.0 + b * di + c
-            return ans
-
-        if nearest_sample:
-            time += self.delta_t / 2.0
-        return self[int((time-self.start_time)*self.sample_rate)]
+        else:
```
maxtrevor, line 297: Jus making sure I know what this is doing, if no interpolate option is given then it will return the value of the closest previous point in the time series?

test/test_timeseries.py
ahnitz: Yes, that's right. These are all run automatically.

pycbc/types/timeseries.py
ahnitz: Yes, at the moment it would, but the idea was that probably some methods would make sense in the future, but won't be implemented yet.

pycbc/types/timeseries.py
ahnitz: Yes, that's the same behavior as before, so there should be no change change in output.

ahnitz:
@maxtrevor Any reason you don't think this can be merged?

maxtrevor (approved):
I think everything is good to go, sorry for the delay

PR #4354, opened 2023-05-05
Rework trigger file finding in pycbc_live_single_trigger_fits

GarethCabournDavies:
Use gps start and end times to find files for triggers, rather than a single directory where we find all files

This should be something fairly easily changed in future to allow us to use file attributes rather than filenames

~Long term thing, not needed for O4 analysis~
Needed for the trigger collation script


GarethCabournDavies:
All output datasets are the same to within a numpy.isclose, some are re-ordered but the same once sorted. Keys change, but that is largely in the addition of fit-start and fit-end, as well as the `input` as arguments have changed

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -136,25 +148,36 @@ elif args.duration_bin_spacing == 'linear':
...
 
+logging.info("Setting up the cut dictionaries")
 # Add template duration cuts according to the bin inputs
 args.template_cuts = args.template_cuts or []
 args.template_cuts.append(f"template_duration:{min(duration_bin_edges)}:lower")
 args.template_cuts.append(f"template_duration:{max(duration_bin_edges)}:upper_inc")
 
+# Ensure that triggers are within the time limits
+args.trigger_cuts = args.trigger_cuts or []
+args.trigger_cuts.append(f"end_time:{args.gps_start_time}:lower_inc")
```
GarethCabournDavies, line 171: This cut will help in the case the the gps_start_time (and equivalently gps_end_time) are in the middle of a file

bin/live/pycbc_live_single_trigger_fits
```diff
@@ -170,20 +193,12 @@ logging.info("Getting events which meet criteria")
...
-
-if not os.path.exists(date_directory):
-    raise FileNotFoundError(f"The directory {date_directory} does not exist.")
-
-files = [f for f in os.listdir(date_directory)
-         if args.file_identifier in f and f.endswith('hdf')]
-
 events = {}
 counter = 0
+no_triggers_yet = {ifo: True for ifo in args.ifos}
```
GarethCabournDavies, line 198:
```suggestion
```

GarethCabournDavies:
I've rebased and added some changes to this.

Mainly the changes are to move the argument parsing into functions in a module so they can be reused in https://github.com/gwastro/pycbc/pull/4697

GarethCabournDavies:
> All output datasets are the same to within a numpy.isclose, some are re-ordered but the same once sorted. Keys change, but that is largely in the addition of fit-start and fit-end, as well as the `input` as arguments have changed

This is still the case with the updates since the rebase, for more concrete discussion of changes
 - pruned_trigger_times has changed order
 - All datasets under the triggers group change order
 - fit-start, fit-end have been added, and analysis_date is removed
 - fit_coeff datasets change by small amounts (<1e-4)

ArthurTolley (approved):
Just a few comment changes but other than that I am happy to approve this

GarethCabournDavies:
quite a few changes since last review need looking at

ArthurTolley (approved):
Changes since last review look good to me

ArthurTolley:
@titodalcanton Could you have a quick look over the changes Gareth has made to the supervise scripts? Then hopefully this can be merged 😁

bin/live/pycbc_live_combine_single_significance_fits
```diff
@@ -93,19 +97,20 @@ for f in args.trfits_files:
...
-            if ifo not in fits_f:
-                continue
-            trig_times = fits_f[ifo]['triggers']['end_time'][:]
-            gps_last = max(gps_last, trig_times.max())
-            gps_first = min(gps_first, trig_times.min())
-        trigger_file_starts.append(gps_first)
-        trigger_file_ends.append(gps_last)
+        # Get the start/end times of the trigger_fits file
+        fit_start = fits_f.attrs['fit_start']
+        fit_end = fits_f.attrs['fit_end']
```
titodalcanton, line 102: I suggest renaming `fit_start` / `fit_end` to `fit_start_time` / `fit_end_time`, at least in the HDF file, so as to have HDF keys be as self-documenting as possible.

bin/live/pycbc_live_plot_single_significance_fits
```diff
@@ -172,7 +177,7 @@ for ifo in all_ifos:
     )
     ax.set_xlabel(sngl_ranking)
     ax.set_ylabel("Number of louder triggers per live time")
-    title = f"{ifo} {analysis_date} trigger fits"
+    title = f"{ifo} {start_str}-{end_str} singles significance fits"
```
titodalcanton, line 180: This string is going to be difficult to read:
```
H1 2024-02-03 02:12:34-2024-02-04 02:12:34 singles significance fits
```
How about this:
```
H1 singles significance fits from 2024-02-03 02:12:34 to 2024-02-04 02:12:34
```

titodalcanton: Maybe add a newline before "from" if too long.

bin/live/pycbc_live_single_significance_fits
```diff
@@ -424,7 +326,8 @@ with h5py.File(args.output, 'w') as fout:
     fout['bins_lower'] = tbins.lower()
 
     fout.attrs['ifos'] = ','.join(args.ifos)
-    fout.attrs['analysis_date'] = args.analysis_date
+    fout.attrs['fit_start'] = args.gps_start_time
+    fout.attrs['fit_end'] = args.gps_end_time
```
titodalcanton, line 330: See my comment above.

titodalcanton: Might want to be even more explicit in fact, and use `fit_start_gps_time` / `fit_end_gps_time`.

bin/live/pycbc_live_supervise_single_significance_fits
```diff
@@ -59,7 +59,14 @@ def mail_volunteers_error(controls, mail_body_lines, subject):
...
     """
-    with open(controls['mail-volunteers-file'], 'r') as mail_volunteers_file:
+    if 'mail_volunteers_file' not in controls:
+        logging.info(
+            "No file containing people to email, logging the error instead"
+        )
+        for line in mail_body_lines:
+            logging.warning(line)
+        return
+    with open(volunteers_file, 'r') as mail_volunteers_file:
```
titodalcanton, line 69: `volunteers_file` is not defined.

bin/live/pycbc_live_supervise_single_significance_fits
```diff
@@ -389,28 +388,27 @@ def do_fitting(args, day_dt, day_str):
...
+        try:
+            shutil.copyfile(
+                combined_options['output'],
+                controls['variable-trigger-fits']
+            )
+            logging.info(
+                "%s updated to link to %s",
+                controls['variable-trigger-fits'],
+                combined_options['output']
+            )
```
titodalcanton, line 404: This can be left where it was; in case of an exception, it will not happen anyway.

pycbc/io/live.py
```diff
@@ -558,4 +564,321 @@ def gracedb_tag_with_version(gracedb, event_id):
...
+
+    # Find the string at the start of the gps time which will match all
+    # files in this range - this helps to cut which ones we need to
+    # compare later
+    num_match = maximum_string([gps_start_time, gps_end_time])
+
+    # ** means recursive, so for large directories, this is expensive.
+    # It is not too bad if date_directories is set, as we don't waste time
+    # in directories where there cant be any files.
+    glob_string = f'**/*{id_string}*{num_match}*.hdf'
```
titodalcanton, line 803: I might be missing the point here, but I do not think it is worth the complication of trying to be smart with large directories. Storing more than a few thousand files in a single directory is terrible in terms of user experience, so it should never be done.

pycbc/io/live.py
```diff
@@ -558,4 +564,321 @@ def gracedb_tag_with_version(gracedb, event_id):
...
+            numpy.log10(max_dur),
+            args.num_duration_bins + 1
+        )
+    if args.duration_bin_spacing == 'linear':
+        return numpy.linspace(
+            min_dur,
+            max_dur,
+            args.num_duration_bins + 1
+        )
+    raise RuntimeError("Invalid duration bin specification")
```
titodalcanton, line 872: pycbc.io.live feels like the wrong place for this function. Does it make sense to start a pycbc.live.single_fits module instead? Or maybe we should move the old stuff to pycbc.io.gracedb instead, since it is not actually that specific to Live anymore?

titodalcanton (changes requested):
Somehow I managed to go through the entire PR fairly quickly :) See comments above.

pycbc/io/live.py
GarethCabournDavies: I was thinking specifically of the early-warning trigger files, which have 86400 per directory - 10800 in the fullband ones is still a lot!

GarethCabournDavies:
testing for latest changes is ongoing

GarethCabournDavies:
I think the remaining codeclimate complaints are previously-existing problems

GarethCabournDavies:
@titodalcanton does this look good-to-go?

pycbc/io/live.py
titodalcanton: Still not entirely convinced about the utility-complexity tradeoff, but fair enough.

pycbc/io/live.py
titodalcanton: This suggestion backfired as this is now a huge API-breaking PR :) But fine.

titodalcanton (approved):
More than a careful re-review, I think we need to get this running and tested on the MDC, so approving. Note that there is a conflict to resolve now.

GarethCabournDavies:
Latest force-push is to resolve clashes with the singles-statistic live fits PR

PR #4358, opened 2023-05-11
add lognl for hierarchical models to output files

ahnitz:
This enables people to make plots with z-arg snr again with hierarchical based likelihoods. For the regular hierarchical it adds the lognl from each submodel (assumption of independent noise). For multisignal, it picks one (assumption of identical noise). These assumptions match the models, so I think this is correct. It also is run in a try/except in cases where a submodel for its own purposes can't produce a lognl. 

@WuShichao This should enable you to plot with SNR for your tests. Please give it a try! (requires running the inference with this patch included)

### DIFF
--- a/pycbc/inference/models/hierarchical.py
+++ b/pycbc/inference/models/hierarchical.py
@@ -217,6 +217,14 @@ class HierarchicalModel(BaseModel):
         for lbl, model in self.submodels.items():
             model.write_metadata(fp, group=prefix+lbl)
 
+        # if all submodels support it, write a combined lognl parameter
+        try:
+            sampattrs = fp.getattrs(group=fp.samples_group)
+            lognl = [self.submodels[k].lognl for k in self.submodels]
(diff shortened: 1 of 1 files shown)


pycbc/inference/models/hierarchical.py
WuShichao, line 1: This solution is better than mine. I'm running my test by using this patch. Can you explain more about "For multisignal, it picks one (assumption of identical noise)"? If there are two overlapping signals in the same data, although the detector noise n(t) is the same, the total noise n(t) + hj(t) is different for each, right?

ahnitz:
@WuShichao Can this now be merged?

WuShichao:
@ahnitz Yes, let me merge it with **squash**.

ahnitz:
@WuShichao You don't need to merger. You need to approve a review.

WuShichao (approved):

PR #4359, opened 2023-05-11
make contains method of joint slightly less slow

ahnitz:
@WuShichao I noticed the 'contains' method was quite slow. The way you were retrieving the index was *very* brute force and I think unnecessarily slow. This fixes that. While it's not perfect, since we should directly support vectors, it is a lot better than searching for the index.


pycbc/distributions/joint.py
WuShichao, line 1: Thanks for fixing this issue! Some tests failed, which seems caused by a network issue, so I re-launch the failed jobs.

WuShichao (approved):
Thanks for fixing this issue! Some tests failed, which seems caused by a network issue, so I re-launch the failed jobs.

PR #4362, opened 2023-05-16
Fix edge case in PyCBC Live's check for horizon distance

titodalcanton:
I just noticed the following behavior in PyCBC Live:
```
2023-05-16T13:25:43.870-07:00 node1466 77 Analyzing from 1368303948.069336
2023-05-16T13:25:47.320-07:00 node1466 77 Recalculating H1 PSD, nan
2023-05-16T13:25:47.320-07:00 node1466 77 Filtering H1
```
Here the horizon distance turned out to be nan for whatever reason (probably h(t) being identically zero, and we are not being very strict with the state vector at the moment) but PyCBC Live happily went ahead and tried to do matched filtering on the data anyway. It turns out that the check for the horizon distance does not work correctly for the case where the distance is nan. This PR changes the logic so as to produce a correct result in this case, and moves the check away from `pycbc_live` and into the StrainBuffer class.

### DIFF
--- a/pycbc/strain/strain.py
+++ b/pycbc/strain/strain.py
@@ -1713,14 +1713,14 @@ class StrainBuffer(pycbc.frame.DataBuffer):
 
     def check_psd_dist(self, min_dist, max_dist):
         """Check that the horizon distance of a detector is within a required
-        range.
+        range. If so, return True, otherwise log a warning and return False.
         """
         if self.psd is None:
             # ignore check
             return True
         # Note that the distance can in principle be inf or nan, e.g. if h(t)
         # is identically zero. The check must fail in those cases.  Be careful
-        # with how the logic works out!
+        # with how the logic works out when comparing inf's or nan's!
         good = self.psd.dist >= min_dist and self.psd.dist <= max_dist
         if not good:
             logging.info(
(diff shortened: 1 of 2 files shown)


pycbc/strain/strain.py
```diff
@@ -1711,6 +1711,28 @@ def recalculate_psd(self):
...
+        good = self.psd.dist >= min_dist and self.psd.dist <= max_dist
+        if not good:
+            logging.info(
+                "%s PSD dist %s outside acceptable range [%s, %s]",
+                self.detector,
+                self.psd.dist,
+                min_dist,
+                max_dist
+            )
+            return False
```
GarethCabournDavies, line 1733: Just return good rather than the two statements?

pycbc/strain/strain.py
```diff
@@ -1711,6 +1711,28 @@ def recalculate_psd(self):
...
+        range.
+        """
+        if self.psd is None:
+            # ignore check
+            return True
+        # Note that the distance can in principle be inf or nan, e.g. if h(t)
+        # is identically zero. The check must fail in those cases.  Be careful
+        # with how the logic works out!
+        good = self.psd.dist >= min_dist and self.psd.dist <= max_dist
+        if not good:
```
GarethCabournDavies, line 1725: if not using good as the return parameter, then we can save a variable here by juts using the logic in the if

GarethCabournDavies (approved):
Approved - minor comments but they aren't important for functionality

PR #4363, opened 2023-05-17
Allow a force override character in config files

GarethCabournDavies:
Allow a character to be placed (currently !), in order to tell the config to overwrite any previously-given options

Some of the changes are so that the current checks for whether there are duplicates in the config aren't applied

Tests performed/planned:
 - [x] ValueError with same option twice and no `!`
 - [x] ValueError on attempt to provide a `!` for a section with a subsection (i.e. [fit_by_template!-h1])
 - [x] ValueError if given an overriding subsection to an already-forced section
 - Interpolation:
    - [x] Can the section be used (with or without the !) for later interpolation?
    - [x] Will the interpolation be performed for a subsection with a `!`?
 - [x] Are other sections unaffected by this change when they shouldn't be?
 - [x] Is everything the same given an identical ini file with no `!` set?

(please add any that you would like to see)

### DIFF
--- a/pycbc/types/config.py
+++ b/pycbc/types/config.py
@@ -244,16 +244,20 @@ class InterpolatingConfigParser(DeepCopyableConfigParser):
 
     def get_subsections(self, section_name):
         """Return a list of subsections for the given section name"""
-        # Keep only subsection names
+        # Keep only subsection names, ignore if the _force_override
+        # character is set. Override sections are duplicated without
+        # the character so ignore these as well
         subsections = [
-            sec[len(section_name) + 1:].rstrip(_force_override)
+            sec[len(section_name) + 1:]
             for sec in self.sections()
             if sec.startswith(section_name + "-")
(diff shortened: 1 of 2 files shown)


spxiwh (changes requested):
I am a little concerned about this being quite convoluted, and needing tests and changes in multiple places. I wonder if this could be inverted and have a special tag, something like:

```
[section1-defaultvalues]
```

whose values would only be added to a job (or used by the workflow parser) *if* that value isn't already given by the tags used.

Then if one gives a `[section-subsection!]` type command, the workflow would respond by moving options into `[section-subsection]` and putting all options in `[section]` into `[section-defaultvalues]` ... ALthough I think using the `defaultvalues` syntax might be more commonly used in this model.

GarethCabournDavies:
I'm going to close and open a new PR which is based on default values, rather than overwriting specific options

PR #4365, opened 2023-05-19
Add argument verification for singles

GarethCabournDavies:
Make sure that the argument to enable single-detector events is set when setting options designed to control them. As well as the upload flag

### DIFF
--- a/pycbc/events/single.py
+++ b/pycbc/events/single.py
@@ -133,7 +133,6 @@ class LiveSingle(object):
         # The checks already performed mean that all(sngl_opts) is okay
         return all(sngl_opts)
 
-
     @classmethod
     def from_cli(cls, args, ifo):
         return cls(
--- a/pycbc/events/single.py
+++ b/pycbc/events/single.py
@@ -8,7 +8,6 @@ from pycbc.types import MultiDetOptionAction
 from pycbc import conversions as conv
 from pycbc import bin_utils
 
-
 class LiveSingle(object):
     def __init__(self, ifo,
                  newsnr_threshold=10.0,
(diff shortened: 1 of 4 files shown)


tdent:
Do we actually need the `--enable-single-detector-background` option? I.e. if all the relevant quantitative options are given, why not turn the singles estimation on anyway ..

GarethCabournDavies:
> Do we actually need the `--enable-single-detector-background` option? The alternative is if all the relevant quantitative options are given, just turn the singles estimation on ..
> 
> I.e. if the value that doesn't cause an error is always determined by other options, it's not a meaningful option.

It was set up and named to mirror the `--enable-background-estimation` coinc background estimator. I don't see a problem with removing it in favour of allowing the other options to set whether singles are used, but will check with @titodalcanton first

pycbc/events/single.py
```diff
@@ -66,20 +66,35 @@ def insert_args(parser):
...
+                         "given, but --enable-single-detector-background "
+                         "is not set.")
+        if args.enable_single_detector_upload \
+                and not all([args.enable_single_detector_background,
+                             args.enable_gracedb_upload]):
+            parser.error("--enable-single-detector-upload does not make "
+                         "sense without both --enable-gracedb-upload "
+                         "and --enable-single-detector-background set.")
+
+
```
GarethCabournDavies, line 95:
```suggestion
```

GarethCabournDavies:
> > Do we actually need the `--enable-single-detector-background` option? The alternative is if all the relevant quantitative options are given, just turn the singles estimation on ..
> > I.e. if the value that doesn't cause an error is always determined by other options, it's not a meaningful option.
> 
> It was set up and named to mirror the `--enable-background-estimation` coinc background estimator. I don't see a problem with removing it in favour of allowing the other options to set whether singles are used, but will check with @titodalcanton first

I have this ready to go - if its what we want to do

pycbc/events/single.py
```diff
@@ -66,20 +66,33 @@ def insert_args(parser):
...
+                         "--single-duration-threshold, "
+                         "--single-newsnr-threshold")
+        if args.enable_single_detector_background and not all(sngl_opts):
+            parser.error(f"Single detector trigger options ({sngl_opts_str}) "
+                         "must ALL be given if "
+                         "--enable-single-detector-background is set.")
+        if any(sngl_opts) and not args.enable_single_detector_background:
+            parser.error(f"Single detector trigger options ({sngl_opts_str}) "
+                         "given, but --enable-single-detector-background "
+                         "is not set.")
```
titodalcanton, line 86: This particular check does not look correct, what if the user wants to use `--sngl-ifar-est-dist fixed`? Then `--single-fit-file` is not needed, right?

GarethCabournDavies: That... is actually a bug right now. So if the fits_file is not given, it will fail even if the `--sngl-ifar-est-dist` is fixed

I'll see if I can fix that

GarethCabournDavies: This has been fixed, and also changes have been made to allow us to use `--sngl-ifar-est-dist` fixed (and so `--single-fixed-ifar`) for some IFOs, but `mean` or `conservative` for others, which we couldn't do before.

This is then a natural way to switch off single-detector uploads for a specific detector if we wished

tdent:
This looks actually somewhat complicated, I don't know if I will check all of the logic just now, so maybe this is a change that we would look for deploying in O4b, particularly as it implies a config file (bash command line) change. 

It is good to have proper input checking in any case, and I would say that equivalent checks should be put in for the coinc IFAR calculation options, but particularly in that case it seems like an O4b horizon for such changes.

titodalcanton (approved):
This looks ok to me now. @GarethCabournDavies can you confirm that the only CLI change is the removal of `--enable-single-detector-background`?

GarethCabournDavies:
> This looks ok to me now. @GarethCabournDavies can you confirm that the only CLI change is the removal of `--enable-single-detector-background`?

That is correct. There are a couple of fixes which mean we can now use more varied options, but for the standard use-case, it is just this removal

GarethCabournDavies:
Did we want to wait to merge this until closer to the O4b changes or are we good to go?

PR #4369, opened 2023-05-22
Live single-trigger combined fits date order bug

GarethCabournDavies:
Also use dates as xtick labels rather than "Days since X" in the plots, and align grid to date boundaries

I had a look at setting the plot to be horizontal between the start/end of each file, but it was impossible to see anything, so didn't make that change in the end

### DIFF
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -110,11 +110,11 @@ for f in args.trfits_files:
 
 trigger_file_starts = np.array(trigger_file_starts)
 trigger_file_ends = np.array(trigger_file_ends)
-ad_order = np.argsort(np.array(trigger_file_starts))
-start_date_n = trigger_file_starts[ad_order[0]]
-ad = trigger_file_ends[ad_order] - start_date_n
+ad_order = np.argsort(trigger_file_starts)
+start_time_n = trigger_file_starts[ad_order[0]]
+ad = trigger_file_ends[ad_order] - start_time_n
 
-# Get the counts and alphas sorted by date
+# Get the counts and alphas
 counts_bin = {ifo: [c for c in zip(*counts_all[ifo])] for ifo in args.ifos}
 alphas_bin = {ifo: [a for a in zip(*alphas_all[ifo])] for ifo in args.ifos}
 
@@ -129,7 +129,7 @@ fout.attrs['fit_threshold'] = fit_thresh
 fout.attrs['conservative_percentile'] = args.conservative_percentile
 fout.attrs['ifos'] = args.ifos
 fout['bins_edges'] = list(bl) + [bu[-1]]
-fout['fits_dates'] = ad + start_date_n
+fout['fits_dates'] = ad + start_time_n
 
 for ifo in args.ifos:
     fout.create_group(ifo)
@@ -146,7 +146,7 @@ for ifo in args.ifos:
     alpha_all = np.mean(counts_bin[ifo], axis=0) / invalphan_all
     meant = l_times.mean()
 
-    fout_ifo[f'separate_fits/live_times'] = l_times
+    fout_ifo[f'separate_fits/live_times'] = l_times[ad_order]
     fout_ifo[f'separate_fits/start_time'] = trigger_file_starts[ad_order]
     fout_ifo[f'separate_fits/end_time'] = trigger_file_ends[ad_order]
 
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -60,7 +60,8 @@ with h5py.File(args.trfits_files[0], 'r') as fit_f0:
 
 live_times = {ifo : [] for ifo in args.ifos}
 
-trigger_file_times = []
+trigger_file_starts = []
+trigger_file_ends = []
 
 n_files = len(args.trfits_files)
 logging.info("Checking through %d files", n_files)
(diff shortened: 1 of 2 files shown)


GarethCabournDavies:
Examples in https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/pycbclive/plot_combined_fits_date_bug/testing/

titodalcanton:
The curves in `L1-2023_05_09-2023_05_21-TRIGGER_FITS_COMBINED-counts_NEW.png` are quite different from the previous version of the plot (even accounting for the x-flip). Is it clear why?

Also, a small request to increase the number of columns in the legend, so the legend becomes wider and fills all the available horizontal space. Then you can make the actual plot a bit taller, giving more space for the curves.

bin/live/pycbc_live_combine_single_fits
```diff
@@ -102,16 +105,16 @@ for f in args.trfits_files:
...
+# Set up the date array, this is stored as an offset from the first trigger time of
+# the first file to the last trigger of the file
 
-trigger_file_times = np.array(trigger_file_times)
-ad_order = np.argsort(np.array(trigger_file_times))
-start_date_n = trigger_file_times[ad_order[0]]
-ad = trigger_file_times[ad_order] - start_date_n
+trigger_file_starts = np.array(trigger_file_starts)
+trigger_file_ends = np.array(trigger_file_ends)
+ad_order = np.argsort(np.array(trigger_file_starts))
```
titodalcanton, line 113: Looks like `np.array()` is not needed here.

bin/live/pycbc_live_combine_single_fits
```diff
@@ -102,16 +105,16 @@ for f in args.trfits_files:
...
+start_date_n = trigger_file_starts[ad_order[0]]
```
titodalcanton, line 114: Optional nitpick: these are timestamps, so name the variable start_time_n.

bin/live/pycbc_live_combine_single_fits
```diff
@@ -102,16 +105,16 @@ for f in args.trfits_files:
...
+ad = trigger_file_ends[ad_order] - start_date_n
 
-# Get the counts and alphas sorted by bin rather than by date
+# Get the counts and alphas sorted by date
```
titodalcanton, line 117: If I understand lines 157-158 right, these are *not* sorted by date, right? Is that not the wrong assumption at the origin of the bug?

titodalcanton: Ah, I guess the comment is meant to also describe what happens in lines 157-158?

bin/live/pycbc_live_plot_combined_single_fits
```diff
@@ -119,24 +133,25 @@ for ifo in ifos:
...
         counts = separate_counts[ifo][i]
 
+        # replace zeros with infinity, so that it is
+        # not plotted rather than plotted as zero
         valid = np.logical_and(alphas > 0, np.isfinite(alphas))
-
+        alphas[np.logical_not(valid)] = np.inf
 
         if not any(valid):
+            logging.info("No valid fit coefficients for %s", ifo)
```
titodalcanton, line 148: Use `logging.warning()` here?

bin/live/pycbc_live_combine_single_fits
```diff
@@ -144,12 +147,15 @@ for ifo in args.ifos:
     meant = l_times.mean()
 
     fout_ifo[f'separate_fits/live_times'] = l_times
-    fout_ifo[f'separate_fits/date'] = ad + start_date_n
+    fout_ifo[f'separate_fits/start_time'] = trigger_file_starts[ad_order]
+    fout_ifo[f'separate_fits/end_time'] = trigger_file_ends[ad_order]
```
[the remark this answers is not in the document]
GarethCabournDavies: No implication, as all of this is not used downstream, only for plotting.

GarethCabournDavies:
> The curves in `L1-2023_05_09-2023_05_21-TRIGGER_FITS_COMBINED-counts_NEW.png` are quite different from the previous version of the plot (even accounting for the x-flip). Is it clear why?

I forgot to additionally reorder the live times, so the wrong ones were being taken into account. This has been updated now

bin/live/pycbc_live_plot_combined_single_fits
```diff
@@ -168,21 +182,23 @@ for ifo in ifos:
...
-    ax_count.set_xlabel('Days since ' + start_date)
     ax_count.set_ylabel('Counts per live time')
-    ax_count.set_xlim([ad[0], ad[-1]])
     count_labels = [l.get_label() for l in count_lines]
     ax_count.legend(count_lines, count_labels, loc='lower center',
                     ncol=3, bbox_to_anchor=(0.5, 1.01))
     ax_count.grid(zorder=-30)
+    ax_count.set_xticks((xtix - min_start) / 86400)
+    ax_count.set_xticklabels(xtix_labels, rotation=90)
+    ax_count.set_xlim(-0.5, len(xtix) - 0.5)
```
titodalcanton, line 202: Loop over `[ax_alpha, ax_count]` and set the x tick properties once.

bin/live/pycbc_live_plot_combined_single_fits
```diff
@@ -168,21 +182,23 @@ for ifo in ifos:
                                             c=bin_colour, linestyle=':',
                                             label=count_lab))
 
-    ax_alpha.set_xlabel('Days since ' + start_date)
     ax_alpha.set_ylabel('Fit coefficient')
-    ax_alpha.set_xlim([ad[0], ad[-1]])
+    ax_alpha.set_xticks((xtix - min_start) / 86400)
```
titodalcanton, line 186: I am a bit confused by this, what guarantees that a time offset divided by 86400 will fall exactly at midnight? Or do we not care that ticks are at midnight?

bin/live/pycbc_live_combine_single_fits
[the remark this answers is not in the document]
titodalcanton: Fair enough.

titodalcanton:
Looks good to me now, I would like to see plots remade after the changes.

titodalcanton (approved):

Issue #4366, opened 2023-05-19
spa_tmplt understands f upper but not f final


tdent:
See https://github.com/gwastro/pycbc/blob/master/pycbc/waveform/spa_tmplt.py#L217 vs. eg https://github.com/gwastro/pycbc/blob/master/pycbc/waveform/waveform.py#L633 

While the waveform code will work if someone calling it knows to use f_upper when using spa_tmplt and f_final when using any lalsimulation waveform, and puts in the appropriate (and probably messy) switches, it will not work properly otherwise.

titodalcanton:
As far as I can tell, spa_tmplt is the only place where `f_upper` is used. PyCBC Live's bank uses a `f_final` dataset, which is consistent with what the PreTaylorF2 approximant understands.

I guess this could be "fixed" by adding support for `f_final` in spa_tmplt, keeping the support for `f_upper`, but printing a warning when that is used.

PR #4371, opened 2023-05-23
Allow f_final as well as f_upper, a bit more PEP8

tdent:
Addresses #4366

### DIFF
--- a/pycbc/waveform/spa_tmplt.py
+++ b/pycbc/waveform/spa_tmplt.py
@@ -22,7 +22,7 @@
    vectors.
 """
 from math import sqrt, log
-import logging
+import warnings
 import numpy, lal, pycbc.pnutils
 from pycbc.scheme import schemed
 from pycbc.types import FrequencySeries, Array, complex64, float32, zeros
@@ -226,8 +226,8 @@ def spa_tmplt(**kwds):
             fISCO = kwds['f_final']
         elif 'f_upper' in kwds:
             fISCO = kwds['f_upper']
-            logging.warn('You used f_upper for spa_tmplt, this is deprecated '
-                         'in favour of f_final !')
+            warnings.warn('f_upper is deprecated in favour of f_final !',
+                          DeprecationWarning)
         else:
             vISCO = 1. / sqrt(6.)
             fISCO = vISCO * vISCO * vISCO / piM
--- a/pycbc/waveform/spa_tmplt.py
+++ b/pycbc/waveform/spa_tmplt.py
@@ -27,6 +27,7 @@ import numpy, lal, pycbc.pnutils
 from pycbc.scheme import schemed
 from pycbc.types import FrequencySeries, Array, complex64, float32, zeros
 from pycbc.waveform.utils import ceilpow2
+from .waveform import NoWaveformError
 
 lalsimulation = pycbc.libutils.import_optional('lalsimulation')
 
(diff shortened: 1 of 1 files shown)


pycbc/waveform/spa_tmplt.py
```diff
@@ -211,11 +220,17 @@ def spa_tmplt(**kwds):
...
-        fISCO = vISCO * vISCO * vISCO / piM
 
-        if 'f_upper' in kwds:
+        # Get ISCO frequency one way or another
+        if 'f_final' in kwds:
+            fISCO = kwds['f_final']
+        elif 'f_upper' in kwds:
             fISCO = kwds['f_upper']
+            logging.warn('You used f_upper for spa_tmplt, this is deprecated '
+                         'in favour of f_final !')
```
titodalcanton, line 230: I meant an actual deprecation warning,
```python
import warnings
warnings.warn('f_upper is deprecated in favour of f_final', DeprecationWarning)
```

pycbc/waveform/spa_tmplt.py
```diff
@@ -95,60 +98,66 @@ def spa_length_in_time(**kwds):
...
+
 def spa_amplitude_factor(**kwds):
     m1 = kwds['mass1']
     m2 = kwds['mass2']
 
     _, eta = pycbc.pnutils.mass1_mass2_to_mchirp_eta(m1, m2)
 
-    FTaN = 32.0 * eta*eta / 5.0
-    dETaN = 2 * -eta/2.0
+    FTaN = 32. * eta*eta / 5.
```
titodalcanton, line 108:
```suggestion
    FTaN = 32. * eta * eta / 5.
```

tdent:
Hmm, seems the unit test is actually failing - not sure why atm

titodalcanton:
Got it, the problem is that by default `f_final` is set to zero. So I propose to include a check that `f_final` > 0 before using it. I also propose to include a check that whatever termination frequency is worked out, it has to be larger than the starting frequency:
```python
        if fISCO <= f_lower:                                         
            raise ValueError(                                         
                f'termination frequency ({fISCO} Hz) must be larger '
                f'than starting frequency ({f_lower} Hz)'
            )
```

tdent:
Hmm, OK.  The code that actually assigns the default is slightly convoluted, but I see in parameters.py `The default (0) indicates that the choice is made by the respective approximant` ... which is not yet the case here. 

The 'standard' fd waveform interface already does a 'final greater than lower' frequency check, see https://github.com/gwastro/pycbc/blob/master/pycbc/waveform/waveform.py#LL633C19-L633C19 
so we should probably copy that aspect.

tdent:
Hopefully the checks now pass.

pycbc/waveform/spa_tmplt.py
```diff
@@ -211,14 +221,25 @@ def spa_tmplt(**kwds):
...
+        elif 'f_upper' in kwds:
+            fmax = kwds['f_upper']
+            warnings.warn('f_upper is deprecated in favour of f_final !',
+                          DeprecationWarning)
+        else:
+            # Schwarzschild ISCO frequency
+            vISCO = 1. / sqrt(6.)
+            fmax = vISCO * vISCO * vISCO / piM
+        if fmax <= f_lower:
+            raise NoWaveformError("cannot generate waveform: f_lower >= "
```
titodalcanton, line 238: Looks like you cannot import the waveform module here because that creates a circular import. Just use a ValueError or RuntimeError, I think they are good enough here.

Also, I find having two variables named `f_max` and `fmax` confusing. I would just use `f_max` all the way.

tdent: Ah, ugh, that is a bit unpleasant.  I guess the whole package would have to be restructured to allow these specialized errors to be used by every module, or they would have to be put in __init__.py (?)

Issue #3585, opened 2021-01-18
Speed up or parallelize `pycbc_fit_sngls_over_multiparam`


titodalcanton:
`pycbc_fit_sngls_over_multiparam` seems to involve N^2 operations, with N the number of templates in the bank. With large template banks, `fit_sngls_over_multiparam` jobs can become bottlenecks for the all-sky coincident search workflow. Can the operation be parallelized across 10 or 100 jobs and the results merged, like we do for optimal SNRs?

ahnitz:
A little more info is needed on how much speed-up (wall-time) is desired and what kind of parameters are used. In general, I think it's better to optimize the code than to try to parallelize it. 

It could be sped up by simply creating and pre-assigning templates to 'buckets' based on a sorted grid. This would make the operation count a lot closer to optimal for at least the 2-d case. I'm not sure about 3-d, but it would be certainly faster if one dimension is much longer than the others.

GarethDaviesGW:
I'm not sure whether parallelization or buckets could even work here, as the information for each template is used for all nearby templates without any hard borders

Unless we had 'fuzzy' buckets/parallelization? Where some templates are in more than one job/buckets, but that may just be moving the problem

tdent:
well, if there are some dimensions that are many smoothing-lengths, you can make bins that are 1 smoothing length and then you need to calculate distances over at most 2 bins at once

titodalcanton:
@GarethDaviesGW it seems to me that this for loop is embarassingly parallel,
```python
else:
    for i in range(len(nabove)):
        dsq = dist(i, rang)
```
but I agree with Alex on optimization.

titodalcanton:
> A little more info is needed on how much speed-up (wall-time) is desired and what kind of parameters are used.

The jobs that completed from the workflow I am referring to took about a day and used 3 fitting parameters. Another one, still running but hopefully close to completion, has accumulated a comparable amount of CPU time so far. So having them complete in a couple hours instead would help the workflow a lot.

tdent:
Is this still desired?

GarethCabournDavies:
I think it is, as the distance calculation is still done , if you have time/someone to add this that would be great!

GarethCabournDavies:
That makes sense to me, Though I think it would have to be modified for the different types of smoothing. (the "closest N templates" is particularly confusing me right now

Does it have to be in 1D actually? It should be the case in all dimensions

GarethCabournDavies:
[...]

PR #4374, opened 2023-05-24
Implement pre-cuts in fit_over_multiparam for efficiency

GarethCabournDavies:
Implementing @tdent's suggestion in #3585, I put in a way to pre-cut templates we are not interested in given the restrictions of each method of smoothing.

### Timing

As a quick and dirty test, I ran before/after codes on the same bank (O3 hyperbank, 428725 templates) on CIT pcdev5, with the following timing results:

|   | `tophat_smoothing`  | `distance_weighted` |
| --- | --- | --- |
| Before this change | 57 mins | 178 mins |
| After this change | 12 mins | 150 mins |
| Fraction of original | 21% | 84% |

Note that:
- `n_closest` is unaffected by this change (by design)
- It is right that there are more savings for `tophat_smoothing` than `dist_weighted`, as this cuts at 1 in weighted distance, whereas `distance_weighted` cuts at 3 (i.e. 3 times the size)

### Accuracy

- The `tophat_smoothing` outputs were compared and found to be identical to sha256 accuracy.
- The `distance_weighted` outputs were not, however all datasets matched to numpy.isclose accuracy (this is not obvious why this was not exact)

### DIFF
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -138,6 +138,19 @@ def smooth(nabove, invalphan, ntotal, dists, smoothing_method, **kwargs):
     return _smooth_dist_func[smoothing_method](nabove, invalphan,
                                                ntotal, dists, **kwargs)
 
+# This is used to pre-calculate the templates which could be cut
+# before calculating distances
+# The value is used to determine the 1D-extent of the used templates;
+# number of smoothing lengths to cut anything outwith this range.
+# n_closest has no cut as it needs to contain enough
+# templates to contain n triggers, which we cannot know beforehand
+
+_smooth_cut = {
+    'smooth_tophat': 1,
+    'n_closest': numpy.inf,
+    'distance_weighted': 3,
+}
+
 
 parser = argparse.ArgumentParser(usage="",
     description="Smooth (regress) the dependence of coefficients describing "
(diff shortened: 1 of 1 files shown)


bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -284,13 +314,66 @@ if len(parvals) == 1 and args.smoothing_method == 'smooth_tophat':
...
+    # For each template, find the range of nearby templates which fall within
+    # the chosen window.
+    lefts = numpy.searchsorted(parvals[sort_dim],
+            parvals[sort_dim] - cut_lengths[sort_dim])
+    rights = numpy.searchsorted(parvals[sort_dim],
+            parvals[sort_dim] + cut_lengths[sort_dim])
+    n_removed = len(parvals[0]) - rights + lefts
+    logging.info("Cutting between %d and %d templates for each smoothing",
+                 n_removed.min(), n_removed.max())
+    # Sort the values to be smoothed into paramaeter's order
```
tdent, line 346: small typo - I would write 'Sort the values to be smoothed by parameter value'

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -284,13 +314,66 @@ if len(parvals) == 1 and args.smoothing_method == 'smooth_tophat':
...
+    nabove = nabove[par_sort]
+    invalphan = invalphan[par_sort]
+    ntotal = ntotal[par_sort]
```
tdent, line 349: code is easier to read if the quantities are generally in the same order - I would do ntotal, nabove, invalpha

tdent: or change the code from 'Precompute the sums ..' onward to use the order they are given to the functions, ie nabove, invalpha, ntotal

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -138,6 +138,35 @@ def smooth(nabove, invalphan, ntotal, dists, smoothing_method, **kwargs):
     return _smooth_dist_func[smoothing_method](nabove, invalphan,
                                                ntotal, dists, **kwargs)
 
+# This is used to pre-calculate the templates which could be cut
+# before calculating distances
+# The value is used to determine the 1D-extent of the used templates;
+# number of smoothing lengths to cut anything outwith this range.
```
tdent, line 144: this is a bit repetitive .. suggest 'Number of smoothing lengths around the current template where distances will be calculated'

tdent (changes requested):
Minor changes requested, otherwise it looks like it does what we want

tdent (approved):

PR #4376, opened 2023-05-26
add in several missing marginalization combinations

ahnitz:
This adds in several missing marginalization combinations for the relative model. These include pol + earth rotation and pol + time + earth rotation.


spxiwh (approved):
I'm not sure I can meaningfully review the correctness of the new code here, but I am happy that it won't interfere with anything else and so am happy to merge and fix later if needed. I think codeclimate makes some valid points here to be resolved.

PR #4379, opened 2023-05-29
Rejection samp

PRAVEEN-mnl:


bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -158,7 +161,33 @@ f_dest.attrs['fit_param'] = args.fit_param
 f_dest.attrs['log_param'] = args.log_param
 with h5py.File(args.template_file, "r") as f_src:
     f_src.copy(f_src["./"], f_dest["./"], "input_template_params")
-temp_samples = numpy.vstack((template_pars)).T
+
+
+# Rejection Sampling
```
tdent, line 166: This can only be done if 'mchirp' is one of the fit params. So I would add a check here to produce an error if the fit params list does not include mchirp

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -158,7 +161,33 @@ f_dest.attrs['fit_param'] = args.fit_param
...
+beta = args.mchirp_downsample_power
+power_law = lambda x: x**(beta)
+mchirp_index = args.fit_param.index('mchirp') # Index of mchirp in fit parameters
```
tdent, line 169: I would remove the comment as it doesn't add anything compared to reading the code

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -158,7 +161,33 @@ f_dest.attrs['fit_param'] = args.fit_param
...
+# Get the maximum mchirp value dynamically based on its position in template_pars
```
tdent, line 170: again, the comment doesn't really add anything beyond reading the code

bin/all_sky_search/pycbc_template_kde_calc
tdent: I also suggest adding an attribute to the output file to record the value of the power, ie `attrs['mchirp_downsample_power']`

tdent:
You are making the template KDE with the downsampled set of templates, then the result will not be an accurate representation of the density of the original templates.  Then the idea is to compensate for the downsampling by multiplying the KDE by the inverse factor of mchirp.  So you will need the mchirp^beta values of the original templates to apply this factor.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -51,6 +51,8 @@ parser.add_argument('--extra-cpt-fraction', type=float,
                     help='Fraction of the extra component in the signal density')
 parser.add_argument('--temp-volume', type=float, 
                     help='Volume covered by the template bank')
+parser.add_argument('--mchirp-downsample-power', type=float, default=1.0,
```
tdent, line 54: I think it would be better to have a default of None, then the code does not do the downsampling unless a value is given

tdent:
So I think the current code will output a file giving KDE values _at the downsampled template points_. However, we want to be able to know the KDE values at _all_ template points in order to calculate the statistic. The downsampling is only done to reduce the computational cost of making the KDE.  So, you will need to have two sets of arrays which are the template values which are for _training_ the KDE, which might be downsampled, and the values for _evaluating_ the KDE which should be the original.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -161,17 +163,39 @@ with h5py.File(args.template_file, "r") as f_src:
 temp_samples = numpy.vstack((template_pars)).T
 
 
+# Rejection Sampling
+if 'mchirp' not in args.fit_param:
+    raise ValueError("mchirp does not exist in args.fit_param")
```
tdent, line 168: But it's OK to not have mchirp as a parameter _if we are not doing the rejection sampling_. So these lines and the assignment of mchirp_index should be inside the block conditional on `mchirp_downsample_power is not None`.

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -161,17 +163,39 @@ with h5py.File(args.template_file, "r") as f_src:
...
+mchirp_index = args.fit_param.index('mchirp')
+if args.mchirp_downsample_power is not None:
+    power_vals = template_pars[mchirp_index] ** args.mchirp_downsample_power
+    f_dest.attrs['mchirp_downsample_power'] = args.mchirp_downsample_power
```
tdent, line 172: This should be done before any calculation just to keep the lines that do the calculation together

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -161,17 +163,39 @@ with h5py.File(args.template_file, "r") as f_src:
...
+    ind = rand_nums < probabilities
+    new_mchirp = template_pars[mchirp_index][ind]
+    additional_params = numpy.delete(temp_samples, mchirp_index, axis=1)[ind]
+    new_temp_samples = numpy.column_stack((new_mchirp, additional_params))
+
+
 if args.make_template_kde:
     logging.info('Starting optimization of template KDE parameters')
-    optbw, optalpha = optimizedparam(temp_samples, alphagrid=args.alpha_grid,
+    optbw, optalpha = optimizedparam(new_temp_samples, alphagrid=args.alpha_grid,
```
tdent, line 183: This will fail if we are not doing the downsampling. I would call the result of the downsampling something like kde_train_samples, then if we do _not_ do the downsampling you can set kde_train_samples = temp_samples.

tdent: (since 'new_temp_samples' is rather vague - not clear what is 'new' about them)

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -161,17 +163,39 @@ with h5py.File(args.template_file, "r") as f_src:
...
     logging.info('Bandwidth %.4f, alpha %.2f' % (optbw, optalpha))
     logging.info('Evaluating template KDE')
-    template_kde = kde_awkde(temp_samples, temp_samples, alp=optalpha,
+    template_kde = kde_awkde(new_temp_samples, temp_samples, alp=optalpha,
                              gl_bandwidth=optbw)
+    
+    compensation_factor = numpy.ones_like(template_kde)
+    if args.mchirp_downsample_power is not None:
+        mchirp_downsample_factor = power_vals[ind]
+        compensation_factor[ind] = 1 / mchirp_downsample_factor
```
tdent, line 193: you don't need to define another variable 'mchirp_downsample_factor' I think you can just deal with this in a single line

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -161,17 +163,39 @@ with h5py.File(args.template_file, "r") as f_src:
...
+    raise ValueError("mchirp does not exist in args.fit_param")
+mchirp_index = args.fit_param.index('mchirp')
+if args.mchirp_downsample_power is not None:
+    power_vals = template_pars[mchirp_index] ** args.mchirp_downsample_power
+    f_dest.attrs['mchirp_downsample_power'] = args.mchirp_downsample_power
+    probabilities = power_vals / numpy.max(power_vals)
+    rand_nums = numpy.random.uniform(0, 1, len(mass1))
+    ind = rand_nums < probabilities
+    new_mchirp = template_pars[mchirp_index][ind]
+    additional_params = numpy.delete(temp_samples, mchirp_index, axis=1)[ind]
```
tdent, line 177: here, is it necessary to treat mchirp separately from the other template parameters?

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -161,17 +163,39 @@ with h5py.File(args.template_file, "r") as f_src:
...
                                      bwgrid=args.bw_grid, nfold=args.nfold_template)
     logging.info('Bandwidth %.4f, alpha %.2f' % (optbw, optalpha))
     logging.info('Evaluating template KDE')
-    template_kde = kde_awkde(temp_samples, temp_samples, alp=optalpha,
+    template_kde = kde_awkde(new_temp_samples, temp_samples, alp=optalpha,
                              gl_bandwidth=optbw)
+    
+    compensation_factor = numpy.ones_like(template_kde)
+    if args.mchirp_downsample_power is not None:
+        mchirp_downsample_factor = power_vals[ind]
```
tdent, line 192: why are you indexing on 'ind' here? we evaluate the KDE on all the templates, the KDE itself is a smooth function that needs to be compensated for the downsampling factor everywhere, I don't think it makes sense to do the compensation for only a subset of templates

bin/all_sky_search/pycbc_template_kde_calc
```diff
@@ -223,8 +247,7 @@ if args.make_signal_kde:
     if args.extra_cpt_fraction is not None and args.temp_volume is not None:
         modified_kde = signal_kde_extra_cpt(signal_kde, frac=args.extra_cpt_fraction,
                                             volume=args.temp_volume)
-    else:
-        modified_kde = signal_kde
+    else: modified_kde = signal_kde
```
tdent, line 250: not sure why this has been changed - better to have a new line after 'else'

tdent (approved):

PR #4380, opened 2023-05-30
Allow pycbc_multi_inspiral to search over a sky grid

hoangstephanie:
At the moment, `pycbc_multi_inspiral` can only search over a single sky position. However, it needs to search over a sky grid covering the GRB error box.

This pull request implements this change. `pycbc_multi_inspiral` now takes an hdf file containing two datasets, `latitude` and `longitude`. The command-line arguments `--ra` and `--dec` are thus removed. After matched-filtering is done, it loops over the sky positions and time slides to perform the coincident and coherent search.

It also fixes the output values of `latitude` and `longitude` which were not correctly written before (set to 0 for all triggers).

This plot below shows, when clustering is disabled, that we indeed get the same triggers when running `pycbc_multi_inspiral` only once over 2 sky points and running `pycbc_multi_inspiral` twice over a single sky position each time : 

image

The CPU time for the search over a sky grid has also been checked : 
image

### DIFF
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -26,6 +26,7 @@ import time
 from collections import defaultdict
 import argparse
 import numpy as np
+import h5py
 from pycbc import (
     detector, fft, init_logging, inject, opt, psd, scheme, strain, vetoes,
     waveform, DYN_RANGE_FAC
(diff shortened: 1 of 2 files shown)


titodalcanton:
@hoangstephanie as we discussed, the tests are failing because the example still requires the `--ra` and `--dec` options. I propose you keep the ability to use either `--ra` and `--dec` *or* `--sky-grid`. If they are given at the same time, an error should be printed using `parser.error()`.

hoangstephanie:
The tests are still failing because I have renamed the command lines `--ra` and `--dec` to `--latitude` and `--longitude` to match the datasets contained in the sky grid and the names of the ouputs. I am thinking of doing a pull request for the example to change the name of the parameters given.

titodalcanton:
@hoangstephanie rather than opening a new PR, you should change the example in *this* PR, so that the example continues to match the code.

bin/pycbc_multi_inspiral
```diff
@@ -324,24 +324,17 @@ with ctx:
...
+   
     # Antenna patterns
-    fp = {
-            position_index: {
-                ifo: detector.Detector(ifo).antenna_pattern(sky_positions[0][position_index], sky_positions[1][position_index], polarization=0, t_gps=t_gps)[0]
-                for ifo in args.instruments
-                }
-            for position_index in positions_array
-            }
+    antenna_patterns = [[[0 for i in range(2)] for position_index in positions_array] for i in range(3)]
```
pannarale, line 329: The 3 should not be hardcoded.

pannarale (changes requested):

pannarale (approved):

PR #4385, opened 2023-05-31
Add a possibility for defaultvalues sections in config files

GarethCabournDavies:
After the discussion on #4363, a defaultvalues section seemed more sensible than specifically overriding certain options

### DIFF
--- a/examples/search/analysis.ini
+++ b/examples/search/analysis.ini
@@ -182,11 +182,12 @@ loudest-keep-values = [-1:5,1:5]
 
 [coinc-full_data-3det]
 loudest-keep-values = [-3:5,-1:5]
+timeslide-interval = 0.11
 
 [coinc-fullinj&coinc-injfull]
 cluster-window = ${statmap|cluster-window}
 loudest-keep-values = 15.0:9999999999999
-timeslide-interval = 0.11
+timeslide-interval = ${coinc-defaultvalues|timeslide-interval}
 
 [coinc-injinj]
 
--- a/examples/search/analysis.ini
+++ b/examples/search/analysis.ini
@@ -184,9 +184,9 @@ loudest-keep-values = [-1:5,1:5]
 loudest-keep-values = [-3:5,-1:5]
 
 [coinc-fullinj&coinc-injfull]
-timeslide-interval = ${coinc-full_data|timeslide-interval}
 cluster-window = ${statmap|cluster-window}
 loudest-keep-values = 15.0:9999999999999
+timeslide-interval = 0.11
 
 [coinc-injinj]
 
--- a/examples/search/analysis.ini
+++ b/examples/search/analysis.ini
@@ -172,8 +172,10 @@ sngl-ranking = newsnr_sgveto_psdvar
 randomize-template-order =
 statistic-files = ${resolve:./statHL.hdf} ${resolve:./statLV.hdf} ${resolve:./statHV.hdf} ${resolve:./statHLV.hdf}
 
-[coinc-full_data]
+[coinc-defaultvalues]
 timeslide-interval = 0.1
+
+[coinc-full_data]
 
 [coinc-full_data-2det]
 loudest-keep-values = [-1:5,1:5]
--- a/examples/search/analysis.ini
+++ b/examples/search/analysis.ini
@@ -150,9 +150,11 @@ autogating-pad = ${inspiral|autogating-pad}
 
 [bank2hdf]
 
+[fit_by_template-defaultvalues]
+sngl-ranking = newsnr_sgveto_psdvar_scaled_threshold
+
 [fit_by_template]
 fit-function = exponential
-sngl-ranking = newsnr_sgveto_psdvar_scaled_threshold
 stat-threshold = 4.0
 prune-param = mtotal
 log-prune-param =
@@ -187,7 +189,6 @@ timeslide-interval = 0.11
 [coinc-fullinj&coinc-injfull]
 cluster-window = ${statmap|cluster-window}
 loudest-keep-values = 15.0:9999999999999
-timeslide-interval = ${coinc-defaultvalues|timeslide-interval}
 
 [coinc-injinj]
 
--- a/pycbc/types/config.py
+++ b/pycbc/types/config.py
@@ -476,7 +476,7 @@ class InterpolatingConfigParser(DeepCopyableConfigParser):
                 continue
 
             if section.endswith('-defaultvalues') and \
-                    not len(section.split()) == 2:
+                    not len(section.split('-')) == 2:
                 # Only allow defaultvalues for top-level sections
                 raise NotImplementedError(
                     "-defaultvalues subsections are only allowed for "
--- a/pycbc/types/config.py
+++ b/pycbc/types/config.py
@@ -602,7 +602,7 @@ class InterpolatingConfigParser(DeepCopyableConfigParser):
             overwrite = False
             if self.has_section(f"{section}-defaultvalues"):
                 return_vals.append(
-                    self.get(f"{section}-{sub}", option)
+                    self.get(f"{section}-defaultvalues", option)
                 )
                 overwrite = True
 
--- a/pycbc/types/config.py
+++ b/pycbc/types/config.py
@@ -475,11 +475,19 @@ class InterpolatingConfigParser(DeepCopyableConfigParser):
             if section == "pegasus_profile":
                 continue
 
+            if section.endswith('-defaultvalues') and \
+                    not len(section.split()) == 2:
+                # Only allow defaultvalues for top-level sections
+                raise NotImplementedError(
+                    "-defaultvalues subsections are only allowed for "
+                    "top-level sections; given %s" % section
+                )
+
             # Loop over the sections again
             for section2 in self.sections():
                 # Check if any are subsections of section
                 if section2.startswith(section + "-"):
-                    if section2 == "defaultvalues":
+                    if section2.endswith("defaultvalues"):
                         # defaultvalues is storage for defaults, and will
                         # be over-written by anything in the sections-proper
                         continue
@@ -573,7 +581,7 @@ class InterpolatingConfigParser(DeepCopyableConfigParser):
         Supplement to ConfigParser.ConfigParser.get(). This will search for an
         option in [section] and if it doesn't find it will also try in
         [section-defaultvalues], and [section-tag] for every value of tag
-        in tags. [section-tag] will overwrite any [section-defaultvalues]
+        in tags. [section-tag] will be preferred to [section-defaultvalues]
         values. Will raise a ConfigParser.Error if it cannot find a value.
 
         Parameters
--- a/pycbc/types/config.py
+++ b/pycbc/types/config.py
@@ -517,11 +517,16 @@ class InterpolatingConfigParser(DeepCopyableConfigParser):
                 "Section %s not present in ConfigParser." % (section2,)
             )
 
+        section_and_default = (section1 == f"{section2}-defaultvalues" or
+                    section2 == f"{section1}-defaultvalues")
+
         if section1.endswith("-defaultvalues") or \
                 section2.endswith("-defaultvalues"):
-            # Override the raise_error variable not to error when
-            # defaultvalues are given
-            raise_error = False
+            if not section_and_default:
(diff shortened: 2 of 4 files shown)


examples/search/analysis.ini
```diff
@@ -172,19 +172,22 @@ sngl-ranking = newsnr_sgveto_psdvar
...
 timeslide-interval = 0.1
 
+[coinc-full_data]
+
 [coinc-full_data-2det]
 loudest-keep-values = [-1:5,1:5]
 
 [coinc-full_data-3det]
 loudest-keep-values = [-3:5,-1:5]
+timeslide-interval = 0.11
```
spxiwh, line 185: Is this just here as a demonstration, not as a scientific change? ... I have noticed some "demonstration settings" from here getting into places they shouldn't, so just need to be careful with this :-)

examples/search/analysis.ini
```diff
@@ -172,19 +172,22 @@ sngl-ranking = newsnr_sgveto_psdvar
...
 
 [coinc-fullinj&coinc-injfull]
-timeslide-interval = ${coinc-full_data|timeslide-interval}
 cluster-window = ${statmap|cluster-window}
 loudest-keep-values = 15.0:9999999999999
+timeslide-interval = ${coinc-defaultvalues|timeslide-interval}
```
spxiwh, line 190: Wait, as a defaultvalue, why would I be specifying it here? I thought we should get the defaultvalue by default??

spxiwh: Or just another illustration?? Perhaps a confusing one without comments if so.

pycbc/types/config.py
```diff
@@ -513,6 +517,12 @@ def check_duplicate_options(self, section1, section2, raise_error=False):
                 "Section %s not present in ConfigParser." % (section2,)
             )
 
+        if section1.endswith("-defaultvalues") or \
+                section2.endswith("-defaultvalues"):
+            # Override the raise_error variable not to error when
+            # defaultvalues are given
+            raise_error = False
+
```
spxiwh, line 525: I think there are some sanity checks to catch here.

The case:

```
[coinc]
option1 = 10

[coinc-defaultvalues]
option1=15
```

should not be allowed. The first means the second can never be used.

pycbc/types/config.py
```diff
@@ -587,6 +598,14 @@ def get_opt_tags(self, section, option, tags):
             if not tags:
                 raise ConfigParser.Error(err_string + ".")
             return_vals = []
+            # First, check if there are any default values set:
+            overwrite = False
```
spxiwh, line 602: Shouldn't this be `has_defaultvalue` not `overwrite`

pycbc/workflow/core.py
```diff
@@ -315,16 +315,45 @@ def add_ini_profile(self, cp, sec):
             key = opt.split('|')[1]
             self.add_profile(namespace, key, value)
 
+    def add_or_replace_common_options(self, opt, value):
+        """
+        Check whether a value opt already exists in the list opts,
+        if it does, then update the next enrty in the list to the
+        new value, if not, append it to the list
```
spxiwh, line 322: New line needed before `Parameters`

pycbc/workflow/core.py
```diff
@@ -380,7 +409,8 @@ def add_ini_opts(self, cp, sec):
                 # stuff later.
                 self.unresolved_td_options[opt] = value
             else:
-                self.common_options += [opt, value]
+                # This option comes from the config file(s)
+                self.add_or_replace_common_options(opt, value)
```
spxiwh, line 413: I don't understand the change here. I don't think I should be allowing to *replace* options here, and I don't know why a similar change for the other two cases above wouldn't also be needed? I think if we enforce that:
```
[coinc]
option1 = 10

[coinc-defaultvalues]
option1=15
```
is not allowed, then is this needed?

spxiwh (changes requested):
Mostly happy with this, thanks! Some minor tweaks requested, and I would like to see it work with those tweaks in place.

examples/search/analysis.ini
GarethCabournDavies: This is just a demo

examples/search/analysis.ini
GarethCabournDavies: So this is something I'm not sure should be fixed, which is that the interpolation is done before the default values are set, so in order to use this feature, the `-defaultvalues` subsection must be specified

pycbc/workflow/core.py
GarethCabournDavies: This would be needed in the case of

```
[coinc-defaultvalues]
option1 = 10

[coinc-subsec]
option1 = 15
```

As the defaultvalues would be set first, and then the subsec would be overwriting it

GarethCabournDavies: You're right there should be the same for the above examples though - that was missed

spxiwh: This should be fine:
```
[coinc-defaultvalues]
option1 = 10

[coinc-subsec]
option1 = 15
```

This should not be:

```
[coinc]
option1 = 10

[coinc-defaultvalues]
option1=15
```

examples/search/analysis.ini
spxiwh: But why *would* you do this. This section should get the defaultvalue anyway. Wouldn't the code work fine with this line just removed?

pycbc/workflow/core.py
spxiwh: But for the bigger question of does this need changing:

If this line needs a change, the other two cases (time-dependent options and File options) might need changes as well. I think the time-dependent option is fine, whereas the File option probably will not be.

Let's look closer at where this section is used, in particular it is *only* called from here:

https://github.com/gwastro/pycbc/blob/158e038a8e1186c686523c3da686aeaede5ad90b/pycbc/workflow/core.py#L571

(perhaps this should be _add_ini_opts to avoid anyone trying to call it from anywhere else, which I think would break things).

Now this will be called on all sections and subsections (not defaultvalues explicitly), and you then call `_defaultvalues` again for all of these .... This does open up the possibility of `[section-tag1-defaultvalues]` and `[section-tag2-defaultvalues]` which would conflict and we probably want to avoid.

You are also assuming that `[section]` is always sent to `add_ini_opts` before any subsections. It is, but if anyone changes that this will break in 'interesting' ways.

I would try and implement a model where you use a new dict defined here:

https://github.com/gwastro/pycbc/blob/158e038a8e1186c686523c3da686aeaede5ad90b/pycbc/workflow/core.py#L568

which would keep track of all options added (not values). Then I would explicitly call `add_ini_opts` after this for loop:

https://github.com/gwastro/pycbc/blob/158e038a8e1186c686523c3da686aeaede5ad90b/pycbc/workflow/core.py#L575

with `self.name + "-defaultvalues"`. This would also have a new keyword argument for "don't override existing values". The default behaviour would be that we fail if an option is already in the new dict.

examples/search/analysis.ini
GarethCabournDavies: However in this specific case, it is an illustration, and so is not needed (and I am adding a comment)

pycbc/workflow/core.py
spxiwh: I'll look more closely later, but banksim and faithsim do not use *this* add_ini_opts function, they use a different, but same named, function in glue (and need to move over to pegasus).

examples/search/analysis.ini
spxiwh: (I also don't see the indicated comment)

pycbc/types/config.py
```diff
@@ -463,22 +463,34 @@ def add_options_to_section(self, section, items, overwrite_options=False):
...
+                raise NotImplementedError(
+                    "-defaultvalues subsections are only allowed for "
+                    "top-level sections; given %s" % section
+                )
+
             # Loop over the sections again
             for section2 in self.sections():
                 # Check if any are subsections of section
                 if section2.startswith(section + "-"):
+                    if section2.endswith("defaultvalues"):
```
spxiwh, line 490: How can I get here now? Isn't this case going to trapped at line 481 and raise a failure?

pycbc/types/config.py
```diff
@@ -513,17 +525,29 @@ def check_duplicate_options(self, section1, section2, raise_error=False):
                 "Section %s not present in ConfigParser." % (section2,)
             )
 
+        section_and_default = (section1 == f"{section2}-defaultvalues" or
+                               section2 == f"{section1}-defaultvalues")
```
spxiwh, line 529: I don't know that we'd ever run this function on two separate sections (ie. `coinc` and `plotcoinc`) as opposed to subsections. Still, I don't object to this fallback, but what to do if `defaultvalues` collide is undetermined because one should not take options from two separate sections.

spxiwh: Oh wait, that's an `or` not an `and`

pycbc/types/config.py
```diff
@@ -513,17 +525,29 @@ def check_duplicate_options(self, section1, section2, raise_error=False):
...
+
+        if section1.endswith("-defaultvalues") or \
+                section2.endswith("-defaultvalues"):
+            if not section_and_default:
```
spxiwh, line 533: Why two layers of logic here? Surely if one of the sections has `defaultvalues` it needs to be a subsection of the other, or I don't know what this function is doing.

spxiwh: Is it the case of `section1-subsection1` and `section1-defaultvalues` vs `section1` and `section1-defaultvalues`?? (Add comments to explain if so).

spxiwh: Maybe the comment here does explain, but I did get confused by this!

pycbc/workflow/core.py
```diff
@@ -315,17 +315,22 @@ def add_ini_profile(self, cp, sec):
...
         -----------
         cp : ConfigParser object
-            The ConfigParser object holding the workflow configuration settings
+            The ConfigParser object holding the workflow configuration
+            settings
         sec : string
             The section containing options for this job.
         """
         for opt in cp.options(sec):
+            if not ignore_existing and opt in self.all_added_options:
```
spxiwh, line 330: This doesn't look right! Shouldn't we have:
```
if opt in self.all_added_options:
    if ignore_existing:
        continue
    else:
        raise ValueError("Option %s has already been added" % opt)
self.all_added_options.append(opt)
```

spxiwh: The `continue` feels important.

pycbc/workflow/core.py
```diff
@@ -562,18 +567,24 @@ def update_current_tags(self, tags):
 
         # collect the options and profile information
         # from the ini file section(s)
+        self.all_added_options = []
```
spxiwh, line 570: Make this a set not a list. All you're doing is `is XX in this_list` and that is *waaaaayyyyy* faster with sets (a dictionary's keys works the same as a set, but we don't have values here, so don't want a dict).

pycbc/types/config.py
GarethCabournDavies: You would get here e.g. in the case that `section` is `coinc` and `section2` is `coinc-defaultvalues`.

The logic above catches when you try to provide defaults to a subsection, e.g. `coinc-fullinj-defaultvalues` through the `len()` check

examples/search/analysis.ini
GarethCabournDavies: I will change this, I couldnt see a better example before, but found one for `fit_by_template` and `plot_binnedhist`

pycbc/types/config.py
GarethCabournDavies: I have added the comment that it is the section-defaultvalues and section-subsection case that is caught here

GarethCabournDavies:
This is erroring on main CI now, the problem looks to be something which I checked in earlier iterations, which is that the `plot_binnedhist` is now run for the section and subsection (`mtotal`) whereas before it only spawned jobs from the subsection. This doesn't really make sense to me why it would have changed, or why this behaviour was in place in the first place

GarethCabournDavies:
I found what I think is the source and am testing a fix

GarethCabournDavies:
I missed that this wasn't a 'required' test - lesson learned!

spxiwh:
It's only required if you're touching search functionality, but it's hard to automate that sentence!

PR #4391, opened 2023-06-06
Tweak GitHub actions

josh-willis:
See if moving some installations from the system to conda can fix recent breakage.  Do not merge!!!


josh-willis:
Superseded by #4394

PR #4392, opened 2023-06-08
use uniform colormap for simple kde plots

tdent:


bin/all_sky_search/pycbc_plot_kde_vals
PRAVEEN-mnl, line 1: while updating stat.py code in the past, we made some changes in kde calculation code also accordingly and updated as well but we didn't implement on plotting code then. for example,
signal_kde = signal_data['signal_kde'][:]
template_kde = template_data['template_kde'][:]

changes to:
signal_kde = signal_data['data_kde'][:]
template_kde = template_data['data_kde'][:]

tdent: Yes, so what does that have to do with this request to change the colormap?  If you want to you can also edit the code via the web browser to make these changes and combine them with the colormap change in this pull request, but it is not something I am concerned with.

tdent:
Closing as it was too much effort to update this feature branch & it should be removed anyway

PR #4396, opened 2023-06-11
add frame transform for fid_params in relbin

WuShichao:
@ahnitz According to your good suggestion, I separated parts of the code from https://github.com/gwastro/pycbc/pull/4289 as separated PRs, this PR applies the frame transform for `fid_params` in the relbin model.


ahnitz:
@WuShichao Just to put this discussion here, we had discussed in the slack if we want to have these functions here. A more general solution would be better and so for the moment, the best may be to simply leave it to the user to set the reference parameters, until one can provide  a transformation like already done for the sampling itself. 

@WuShichao If you agree, perhaps we can close this PR.

WuShichao:
> @WuShichao Just to put this discussion here, we had discussed in the slack if we want to have these functions here. A more general solution would be better and so for the moment, the best may be to simply leave it to the user to set the reference parameters, until one can provide a transformation like already done for the sampling itself.
> 
> @WuShichao If you agree, perhaps we can close this PR.

@ahnitz Yes, we can close this PR for now.

PR #4401, opened 2023-06-14
Replace LOSC with GWOSC

titodalcanton:
Cleanup PR to replace the remaining LOSC cases with GWOSC. Note that this introduces a backwards-incompatible API change. I could make it more backward-compatible (at the price of adding code) if there is strong desire.

Side note: `examples/gw150914/PyCBCInspiral.ipynb` does not really work anymore. I added a small fix for a trivial error, but there is less trivial stuff as well which I did not try fixing.


pycbc/frame/frame.py
```diff
@@ -326,16 +326,16 @@ def query_and_read_frame(frame_type, channels, start_time, end_time,
...
         if not isinstance(channels, list):
             channels = [channels]
-        data = [read_strain_losc(c[:2], start_time, end_time)
+        data = [read_strain_gwosc(c[:2], start_time, end_time)
                 for c in channels]
         return data if len(data) > 1 else data[0]
-    if frame_type == 'LOSC':
-        from pycbc.frame.losc import read_frame_losc
-        return read_frame_losc(channels, start_time, end_time)
+    if frame_type in ['LOSC', 'GWOSC']:
```
ahnitz, line 336: Thumbs up from me for maintaining backwards compatibility!

pycbc/frame/gwosc.py
```diff
@@ -0,0 +1,185 @@
...
+    -------
+    frame_files: list
+        A dictionary containing information about the files that span the
+        requested times.
+    """
+    data = gwosc_frame_json(ifo, start_time, end_time)['strain']
+    return [d['url'] for d in data if d['format'] == 'gwf']
+
+
+def read_frame_gwosc(channels, start_time, end_time):
```
ahnitz, line 126: Would it be worth setting an alias to losc  with a deprecation warning? I'll let  you decide on that.

ahnitz (approved):
@titodalcanton This looks good to me. Thanks for maintaining backwards compatibility for config / command lines. The only question I have if it is worth adding an alias / deprecation warning for the old named functions. Otherwise, please merge when you are happy.

pycbc/frame/gwosc.py
titodalcanton: I might open a new PR for that.

PR #4403, opened 2023-06-14
Make gate-and-inpainting starting time precise [WIP]

yi-fan-wang:
In this PR, I timeshift GW strains such that the gating starting time lands exactly on a specific data sample.

### DIFF
--- a/pycbc/inference/models/gated_gaussian_noise.py
+++ b/pycbc/inference/models/gated_gaussian_noise.py
@@ -561,8 +561,8 @@ class GatedGaussianNoise(BaseGatedGaussian):
             rtilde = res.to_frequencyseries()
             #get the subsampling offset time
             gateenddelay = gatestartdelay + dgatedelay
-            gateidx = numpy.floor(float(gateenddelay - res.start_time) * res.sample_rate)
-            shifttime = gateenddelay - res.get_sample_times[gateidx]
+            gateidx = numpy.floor(float(gateenddelay - res.start_time) * res.sample_rate).astype(int)
+            shifttime = gateenddelay - res.sample_times[gateidx]
             #apply time shift such that the gating start time lands on a data sample
             rtilde = apply_fd_time_shift(rtilde, -shifttime, copy=False)
             res = rtilde.to_timeseries()
--- a/pycbc/inference/models/gated_gaussian_noise.py
+++ b/pycbc/inference/models/gated_gaussian_noise.py
@@ -27,7 +27,7 @@ from pycbc.waveform import (NoWaveformError, FailedWaveformError)
 from pycbc.types import FrequencySeries
 from pycbc.detector import Detector
 from pycbc.pnutils import hybrid_meco_frequency
-from pycbc.waveform.utils import time_from_frequencyseries
+from pycbc.waveform.utils import (time_from_frequencyseries, apply_fd_time_shift)
 from pycbc.waveform import generator
 from pycbc.filter import highpass
 from .gaussian_noise import (BaseGaussianNoise, create_waveform_generator)
@@ -559,6 +559,14 @@ class GatedGaussianNoise(BaseGatedGaussian):
             ht = h.to_timeseries()
             res = data - ht
             rtilde = res.to_frequencyseries()
+            #get the subsampling offset time
+            gateenddelay = gatestartdelay + dgatedelay
+            gateidx = numpy.floor(float(gateenddelay - res.start_time) * res.sample_rate)
+            shifttime = gateenddelay - res.get_sample_times[gateidx]
+            #apply time shift such that the gating start time lands on a data sample
+            rtilde = apply_fd_time_shift(rtilde, -shifttime, copy=False)
+            res = rtilde.to_timeseries()
+            #gating
             gated_res = res.gate(gatestartdelay + dgatedelay/2,
                                  window=dgatedelay/2, copy=True,
                                  invpsd=invpsd, method='paint')
--- a/pycbc/inference/models/marginalized_gaussian_noise.py
+++ b/pycbc/inference/models/marginalized_gaussian_noise.py
@@ -229,7 +229,7 @@ class MarginalizedTime(DistMarg, BaseGaussianNoise):
                 generator_class=generator.FDomainDetFrameTwoPolNoRespGenerator,
                 gates=self.gates, **kwargs['static_params'])
         else:
-            # create a waveform generator for each ifo respestively
+            # create a waveform generator for each ifo respectively
             self.waveform_generator = {}
             for det in self.data:
                 self.waveform_generator[det] = create_waveform_generator(
@@ -386,7 +386,7 @@ class MarginalizedPolarization(DistMarg, BaseGaussianNoise):
                 generator_class=generator.FDomainDetFrameTwoPolGenerator,
                 gates=self.gates, **kwargs['static_params'])
         else:
-            # create a waveform generator for each ifo respestively
+            # create a waveform generator for each ifo respectively
             self.waveform_generator = {}
             for det in self.data:
                 self.waveform_generator[det] = create_waveform_generator(


cdcapano (changes requested):
This looks good. But can you put lines 563-575 into a separate function that can be called? Maybe make it a class method of `BaseGatedGaussian`. That will make it easier to implement in the MarginalizedPolarization model. So, for example, in the GatedGaussian you would pass `rtilde` to do the shift, while in the MarginalizedPolarization model you would pass hp, hc, and the data.

yi-fan-wang:
@cdcapano After a second though, now I implement the "making the gating time precise" differently. All the inference's models are left untouched, the time shifting is only implemented in `pycbc.types.timeseries` when calling `TimeSeries.gate()`. 

What the codes do is basicaly firstly shifting data to an earlier time by an amount equal to `gate_end_time - the closest (floor) discrete data sample's time`, then gating and inpainting, then shifting the data back to the original time (by shifting the same amount of time but to the later direction). I also change the inpainting array's index from [lindex:rindex] to [lindex:rindex+1], this makes the right index is included.  

I test this PR by running GW150914's ringdown part with only 220 mode. This [plot](https://www.atlas.aei.uni-hannover.de/work/yifan.wang/pycbcdoc/ringdown/GW150914-220/0711-subgating/comparison.png) is a comparison of results with and without this PR applied. I see small but visible differences.

yi-fan-wang:
Got a figure to visulize it. The second figure is a zooming-in
subgating-data
subgating-data-zoomin

pycbc/strain/gate.py
```diff
@@ -168,16 +168,16 @@ def gate_and_paint(data, lindex, rindex, invpsd, copy=True):
     # Copy the data and zero inside the hole
     if copy:
         data = data.copy()
-    data[lindex:rindex] = 0
+    data[lindex:rindex+1] = 0
```
cdcapano, line 171: As we discussed on the call, remove the +1 to stay consistent with PyCBC's convention of using right-open intervals. Add a comment about the ambiguity of the gate start though.

pycbc/strain/gate.py
```diff
@@ -168,16 +168,16 @@ def gate_and_paint(data, lindex, rindex, invpsd, copy=True):
...
     tdfilter = invpsd.astype('complex').to_timeseries() * invpsd.delta_t
     owhgated_data = (data.to_frequencyseries() * invpsd).to_timeseries()
 
     # remove the projection into the null space
-    proj = linalg.solve_toeplitz(tdfilter[:(rindex - lindex)],
-                                 owhgated_data[lindex:rindex])
-    data[lindex:rindex] -= proj
+    proj = linalg.solve_toeplitz(tdfilter[:(rindex + 1 - lindex)],
+                                 owhgated_data[lindex:rindex+1])
+    data[lindex:rindex+1] -= proj
```
cdcapano, line 180: Remove the +1s here too.

pycbc/types/timeseries.py
```diff
@@ -595,15 +595,31 @@ def gate(self, time, window=0.25, method='taper', copy=True,
...
             lindex = int((time - window - self.start_time) / self.delta_t)
-            rindex = lindex + int(2 * window / self.delta_t)
+            rindex = int((time + window - self.start_time) / self.delta_t)
             lindex = lindex if lindex >= 0 else 0
             rindex = rindex if rindex <= len(self) else len(self)
-            return gate_and_paint(data, lindex, rindex, invpsd, copy=False)
+            # shift data such that gate_end_time lands on a specific data sample
+            fdata = data.to_frequencyseries()
+            rindex_time = data.sample_times[rindex]
+            offset = rindex_time - (time + window)
```
cdcapano, line 610: As we discussed on the call, calculate offset at line 607. Then only do the time shifting if offset != 0. Also make `copy=False` at all points here, since the copy was already done at the beginning of the function.

cdcapano (changes requested):
Just a few more changes, as we discussed on the call. Looks good otherwise.

yi-fan-wang:
@cdcapano all comments addressed. I tested with GW150914 with ringdown 220 ([config file](https://www.atlas.aei.uni-hannover.de/work/yifan.wang/pycbcdoc/ringdown/GW150914-220/0713-removerindex+1/inference-2048.ini)). This [plot](https://www.atlas.aei.uni-hannover.de/work/yifan.wang/pycbcdoc/ringdown/GW150914-220/0713-removerindex+1/comparison.png) compares results with and without this PR. Without this PR, A_220 = 2.8E-21; while results with this PR reduce A_220 a little bit to 2.77E-21. I think this is expected, because a little bit more pre-ringdown data are gated by this PR, hence smaller A220.

yi-fan-wang:
@ahnitz Collin and I would like to let you review this PR before merging, because it affects gate and inpainting everywhere. 

The idea is that the previous gating start and end time are not precise when it is located at a subsample time. Say the LIGO data are sampled at ..., t_-2, t_-1, t_0, t_1, ... and the gating end time is somewhere between t_0 and t_1, previous gate-and-inpainting would inpaint only until t_-1, sometime even t_-2 if the window length is not a integer times of 1/sampling rate. This PR effectively reconstruct the time domain data at epoch t_gate_end and t_gate_end plus/minus n times of 1/sampling rate, then inpaint until one element before the t_gate_end. The reconstruction is done by time shifting in frequency domain and then shifting back after inpainting. We think this makes the gating epoch more precise, this will affect the results of ringdown overtone analysis where the characteristic decay time is even less than 1 ms.

pycbc/types/timeseries.py
```diff
@@ -595,15 +595,34 @@ def gate(self, time, window=0.25, method='taper', copy=True,
...
                 # These are some bare minimum settings, normally you
                 # should probably provide a psd
                 invpsd = 1. / self.filter_psd(self.duration/32, self.delta_f, 0)
             lindex = int((time - window - self.start_time) / self.delta_t)
-            rindex = lindex + int(2 * window / self.delta_t)
+            rindex = int((time + window - self.start_time) / self.delta_t)
             lindex = lindex if lindex >= 0 else 0
             rindex = rindex if rindex <= len(self) else len(self)
-            return gate_and_paint(data, lindex, rindex, invpsd, copy=False)
+            rindex_time = data.sample_times[rindex]
```
ahnitz, line 607: This is a slow way to calculate this, since sample_times call arange. You should just do the math directly.

pycbc/types/timeseries.py
```diff
@@ -595,15 +595,34 @@ def gate(self, time, window=0.25, method='taper', copy=True,
...
+            offset = rindex_time - (time + window)
+            if offset == 0:
+                return gate_and_paint(data, lindex, rindex, invpsd, copy=False)
+            else:
+                # time shift such that gate end time lands on a specific data sample
+                fdata = data.to_frequencyseries()
+                fdata = apply_fd_time_shift(fdata, offset + fdata.epoch, copy=False)
```
ahnitz, line 614: Have you double checked the sign on these things? I think the logic makes sense, but I always screw up signs.

pycbc/types/timeseries.py
```diff
@@ -595,15 +595,34 @@ def gate(self, time, window=0.25, method='taper', copy=True,
...
+                # gate and paint in time domain
+                data = fdata.to_timeseries()
+                data = gate_and_paint(data, lindex, rindex, invpsd, copy=False)
+                # shift back to the original time
+                fdata = data.to_frequencyseries()
+                fdata = apply_fd_time_shift(fdata, -offset + fdata.epoch, copy=False)
+                tdata = fdata.to_timeseries()
+                # fill in projslc information
+                tdata.projslc = data.projslc
```
ahnitz, line 623: This seems like a hack, would it not be better to have an argument that tells the function to return some additional things instead of just the data?

yi-fan-wang: @ahnitz This indeed looks like a hack to me as well! The two attributes, `tdata.projslc` and `tdata.proj` are so hidden. However, I'm imitating the return of the `gate_and_inpaint`, which was there originally. See here: https://github.com/gwastro/pycbc/blob/master/pycbc/strain/gate.py#L181

This returned format,  `tdata.projslc` and `tdata.proj`, is also later called by the gated_gaussian model, for example see this line: https://github.com/gwastro/pycbc/blob/master/pycbc/inference/models/gated_gaussian_noise.py#L565

I'd rather not touch any other files, so implement it in the current way to be consistent everywhere. Does this make sense? I hope I understand what you suggest correctly.

cdcapano: @yi-fan-wang This can be removed. If you search for `current_proj` in `gated_gaussian_noise.py`, you'll see it isn't actually used anywhere; it was just for debugging purposes (e.g., see the comment [here](https://github.com/gwastro/pycbc/blob/master/pycbc/inference/models/gated_gaussian_noise.py#L49).) I would just comment out all the places `current_proj` is used in `gated_gaussian_noise.py`, then comment out those two lines of code in `timeseries.py` that adds the attribute. If @ahnitz prefers the lines be removed entirely you can do that too.

pycbc/types/timeseries.py
cdcapano: @yi-fan-wang Please address this comment. As Alex said, `rindex_time` should just be `rindex * data.delta_t + data.start_time`.

cdcapano:
@yi-fan-wang It looks like you've addressed all the comments that @ahnitz and I had, is that correct? Do you think this is ready to merge?

pycbc/types/timeseries.py
yi-fan-wang: I'm careful to take `offset` less than 0 and equal to the difference between the genuine gate_end_time and the nearest (floor) discrete sample's time. The reasoning can be explained as that the previous codes always inpaint data until earlier than we want. Putting another way, although we want to inpaint until gate_end_time, it only stops until a floor time. Therefore we shift the data back to earlier, such that more later data can be inpainted. I hope this understanding makes sense.

pycbc/types/timeseries.py
yi-fan-wang: done

pycbc/types/timeseries.py
yi-fan-wang: done. I removed all the `projslc` and `proj`

yi-fan-wang:
@ahnitz @cdcapano All comments addressed. I have launched a few runs and confirm that the commits after Alex's comments has the same results as before. There is a "line too long" complaint from codeclimate (85 characters > 79). I don't know. I kind of think a single line has better readability. If you also think this is minor, then I'll leave as what it is. 

(Plots to compare runs with commits after Alex's comments and before https://hypatia.aei.mpg.de/~yiwang/ringdown/atlas-result/srate2018-4th-result.png and https://hypatia.aei.mpg.de/~yiwang/ringdown/atlas-result/srate2018-5th-result.png. They look close enough)

Also note that (1) this PR changes gate and inpainting everywhere, and it doesn't have backward compatibility, so we should add a notice in GW190521 data release repository and anywhere relevant. (2) after the fork of this branch, there were many new updates for PyCBC inference workflow, for example, take care of https://github.com/gwastro/pycbc/issues/4453. Also, the most recent pegasus, 5.0.6 only works with the most updated (at the time of writing) PyCBC

yi-fan-wang:
@ahnitz @cdcapano if you don't have objections, this is ready to merge

yi-fan-wang:
In case the above hypatia link doesn't work I drop the two figures here. These are runs with ringdown mode 220. The two figures have different starting times. The label "old runs in atlas" is the result before Alex's comments, while "after PR" is after. They look close enough.
srate2018-4th-result
srate2018-5th-result

cdcapano (approved):
Thanks for making the changes, look fine to me.

PR #4407, opened 2023-06-16
significance bug for trigger_fit

GarethCabournDavies:
The significance was not being handled correctly for events below the fit threshold in trigger_fit cases for the significance calculations

We see this in e.g. 
image Where the basic issue is that the events above the fit threshold were not being added to the count_n_louder fallback below the threshold

Workflow ongoing to get an updated verison of the plot


tdent:
I might need a bit more context to see what this ought to be doing ... Is there an equation or section of a paper this is implementing?

GarethCabournDavies:
There is no specific discussion of what is done below the fit threshold, but section 2.3 of https://arxiv.org/pdf/2203.08545.pdf shows what is done above the threshold.

In the plot above, the fit threshold is -12.5; below this we count the number of higher-ranked events and above this we use the exponential fit.

To clarify the above description, the bug was that the count of events above the fit threshold were being added to the FAN for _all_ events rather than just those below threshold, leading to this jump in FAR.

The effect is in place for the foreground events, but is not so obvious on the cumulative plot, you can see it in the non-cumulative one

With bug: 

image

Without bug:

image

GarethCabournDavies:
Not that this also gets the number of events back into the expected range: 

With bug:

image

After fix:

image

tdent:
Seems OK to me, but I am not the expert in single detector significance methods ;) 

Just noting that the initial description said 'significance was not being handled correctly for events below the fit threshold ' but presumably the intention was to say 'above the fit threshold'

tdent (approved):

PR #4413, opened 2023-06-26
Simple bugfix for optimal snr

bhooshan-gadre:
It is tested and work as expected.


GarethCabournDavies (approved):
Slightly surprised this if statement wasn't True for eight years (I think it was [this commit](https://github.com/gwastro/pycbc/commit/1be96a484e85dc7e8b08542c701948f163e0d2be) which missed the required change), but it looks to be doing the right thing

PR #4418, opened 2023-06-28
fftw_fix

neeresh:
Made a few changes to the scripts: **bin/pycbc_inspiral**, **pycbc/fft/npfft.py**, and **examples/inspiral/run.sh**. Earlier I couldn't run the batch file **run.sh** and I was able to run the after making these changes.

### DIFF
--- a/bin/pycbc_inspiral
+++ b/bin/pycbc_inspiral
@@ -301,19 +301,21 @@ def template_triggers(t_num):
     return out_vals_all, tparam
 
 with ctx:
-    # The following FFTW specific options needed to wait until
-    # we were inside the scheme context.
+    if opt.fft_backends == 'fftw':
 
-    # Import system wisdom.
-    if opt.fftw_import_system_wisdom:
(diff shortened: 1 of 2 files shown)


cdcapano:
A bunch of extra files were included here that wasn't intended. I'll work with @neeresh offline to get the PR fixed.

cdcapano:
@spxiwh This is from an MS student that's working with me. He's been trying to run pycbc_inspiral on his M2 Macbook Pro. When he installed pycbc fftw didn't install (going to try to figure out why, but that's a separate issue), leaving only numpy as backend. That should work, but we ran into errors because `pycbc_inspiral` was assuming options were in the argument parser that are not actually there if fftw isn't installed. This should fix that.

spxiwh (approved):

cdcapano:
Thanks @spxiwh. @neeresh I'll walk you through what you need to do when we meet on Wednesday.

PR #4421, opened 2023-06-29
Prevent duplicates of the HLV page_ifar plot on the summary page

GarethCabournDavies:
On the summary page, the snrifar / page_ifar plots are given for each detector combination, but the page_ifar plot for singles was never made.

So the workflow was grabbing the page_ifar plot from the last coincidence (HLV)

### DIFF
--- a/bin/workflows/pycbc_make_offline_search_workflow
+++ b/bin/workflows/pycbc_make_offline_search_workflow
@@ -521,6 +521,9 @@ for key in final_bg_files:
         # as it is just blank
         ifar_cb = wf.make_ifar_plot(workflow, bg_file, closed_dir,
                                     tags=bg_file.tags + ['closed_box'])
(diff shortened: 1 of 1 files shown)


GarethCabournDavies:
An example results page can be seen here (when the workflow completes):
https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/lightning_plot_duplicated/

Whereas a results page without this fix is at:
https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/lightning_plot_duplicated/

GarethCabournDavies:
Not working at the moment, so I will convert to draft until I figure out what has gone wrong here

GarethCabournDavies:
I've also added a couple of fixes to rename things which were "coincident" to show that they may not be

ArthurTolley (approved):
Looks good to me

PR #4424, opened 2023-07-05
Range_temp kde

PRAVEEN-mnl:


tdent:
Should it be necessary to specify both a minimum and a maximum?  What I was thinking is that we specify a maximum *range* of values between the maximum KDE and the minimum.  E.g. the minimum should be at least a factor of 1e-6 *times* the maximum.

PRAVEEN-mnl:
well yes, we don't need to fix the maximum. The maximum value is of the order of 1e+4 so if we fix the minimum 1e-2, the maximum is itself almost 1e+6 times higher. So I can take out the --max-val.

tdent:
Ok, but what I was actually thinking was that we set a minimum kde value _relative to the maximum_
Ie if you get the max kde value you can divide all the other values by it and then set a floor on the _ratio_ which would be eg 1e-6

tdent (approved):

PR #4425, opened 2023-07-06
mistake in example search config file

ahnitz:
@spxiwh @GarethCabournDavies 

There was a mistake in the example search configuration introduced in https://github.com/gwastro/pycbc/pull/4385

This caused the search workflow to suddenly take > 3x longer. This happened because timeslide-interval was in 'defaultvalues' for 'coinc'. The 'injinj' jobs were running timeslides! This means they'll produce a huge number of triggers for later stages due to the very large number of coincidences between different injections. 

I think the comments in #4420 related to performance are mostly due to this configuration error, although if it's not used, it's not a terrible thing to clean  up anyway.

### DIFF
--- a/.github/workflows/search-workflow.yml
+++ b/.github/workflows/search-workflow.yml
@@ -9,6 +9,7 @@ concurrency:
 jobs:
   build:
     runs-on: ubuntu-20.04
+    timeout-minutes: 90
     steps:
     - uses: actions/checkout@v1
     - name: Set up Python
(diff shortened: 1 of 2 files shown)


ahnitz:
This also makes the search fail if it runs too long. In the future that should make it easier to spot if a PR causes a similar issue.

GarethCabournDavies:
Thanks! I saw this and was fixing it as a part of #4420, but it definitely is worth adding in straight away

GarethCabournDavies (approved):

PR #4426, opened 2023-07-06
Change decisive to network optimal snr in foundmissed

GarethCabournDavies:
For a search including single-detector events, the decisive SNR for whether a signal is detected is actually the maximum SNR

See e.g. [comment here](https://github.com/gwastro/pycbc/blob/4488619e969fd36142a9a4f9fe6bc93d15fbcb6b/bin/plotting/pycbc_page_foundmissed#L123)

### DIFF
--- a/examples/search/plotting.ini
+++ b/examples/search/plotting.ini
@@ -139,14 +139,14 @@ log-dist =
 far-type = exclusive
 
 [plot_foundmissed-sub_mchirp_grad&plot_foundmissed-all_mchirp_grad&plot_foundmissed-summary]
-distance-type = max_optimal_snr
+distance-type = comb_optimal_snr
 axis-type = mchirp
 log-x =
 log-distance =
 gradient-far =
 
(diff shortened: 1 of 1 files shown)


ahnitz:
@GarethCabournDavies I'm not sure I follow. This makes sense if you just look at singles, but coincs won't have the same detection threshold in terms of max SNR which blurs out the detection boundary. 

If we want comparable numbers, wouldn't network optimal SNR make more sense as a default?

GarethCabournDavies:
> @GarethCabournDavies I'm not sure I follow. This makes sense if you just look at singles, but coincs won't have the same detection threshold in terms of max SNR which blurs out the detection boundary.
> 
> If we want comparable numbers, wouldn't network optimal SNR make more sense as a default?

It does - I had thought that max_optimal_snr would be better, but trying to think of arguments _why_ that was true convinced me that combined is better

tdent:
Noting that this requires a review to be merged.

ahnitz (approved):

PR #4432, opened 2023-07-11
SNR optimisation options for pycbc_live

ArthurTolley:
This is a recreation of #4398 which is now closed due to my mistake of creating the PR using Gareth's branch and then not being able to make changes to that branch.

This PR includes a new directory containing two new files:  
`pycbc/live` -> `__init__.py` + `snr_optimizer.py`

`snr_optimizer.py` contains the functions required to do the snr optimization (the snr optimizers themselves, but _not_ the boundary calculations)

The following two files have also been modified:  
`pycbc_live` - includes the new snr_optimizer options group. 
`pycbc_optimize_snr` - removing the functionality that has been moved to `snr_optimizer.py`

I am yet to fully test these changes and will do so pronto.


pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,340 @@
...
+import pycbc.conversions as cv
+
+try:
+    import pyswarms as ps
+except:
+    ps = None
+
+# Set a minimum mass for points tried in optimization allowing for
+# minimal slop relative to the lightest template
+MIN_CPT_MASS = 0.99
```
tdent, line 21: I would not be super happy about this special variable being defined here _and_ in the executable - SSOT. If it's defined here it should be used via import (`snr_optimizer.MIN_CPT_MASS` I think) in the script

pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,340 @@
...
+# Set a large maximum total mass
+MAX_MTOTAL = 500.
+
+Nfeval = 0
+start_time = time.time()
+
+def callback_func(Xi, convergence=0):
+    global Nfeval
+    logging.info("Currently at %d %s", Nfeval, convergence)
+    if (time.time() - start_time) > 360:
```
tdent, line 32: Actually I'm not sure why this is here and why the specific value 360. Would be good to add a comment at least

pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,340 @@
...
+        return True
+    Nfeval += 1
+
+
+def compute_network_snr_core(v, *argv, raise_err=False):
+    """
+    Compute network SNR as a function over mchirp, eta and two aligned
```
tdent, line 39: IMO this would be a good time to properly document the input values, as 'v' and what has to be in argv are currently mysterious

pycbc/live/snr_optimizer.py
ArthurTolley: The comment in `pycbc_optimize_snr` is helpful for a definition of `MIN_CPT_MASS` so the way I've done it isn't the prettiest, what do you think?

pycbc/live/snr_optimizer.py
ArthurTolley: The optimization has an effective time-out of 6 minutes with this line, maybe that's a ridiculously long time but it helps to kill the optimizer if there is an issue causing hanging.

ArthurTolley:
Currently the example in `examples/live/run.sh` works with `--optimizer pso` but when using `differential_evolution` there is an issue with the command line options. Giving the two differential evolution options, maxiter and popsize, there is a bug when trying to run the optimizer which says:
`pycbc_optimize_snr: error: unrecognized arguments: --differential_evolution-maxiter 50.0 --differential_evolution-popsize 100.0`

I've looked at how the option group is generated and the option group looks fine, I'm thinking maybe there is an issue with how the option is being passed from the command line in to the program. Not sure and i can't figure it out.

pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,383 @@
...
+        default='differential_evolution',
+        choices=optimizer_choices,
+        help='The optimizer to use, choose from ' + ', '.join(optimizer_choices))
+
+    # Add the generic options
+    opt_opt_group.add_argument('--include-candidate-in-optimizer',
+        action='store_true',
+        help='Include parameters of the candidate event in the initialized array for the optimizer. '
+             'Only relevant for --optimizer pso or differential_evolution')
+    opt_opt_group.add_argument('--seed',
```
GarethCabournDavies, line 325: As this is also going to go through to the pycbc_live options, it would be good to update this to be `--optimizer-seed` (or similar), just to be obvious that it is going to be passed through

GarethCabournDavies (commented):
As this is moving a lot of code into a new module, we will lose the blame information. So it is easy(ish) to find, here's the [LINK](https://github.com/gwastro/pycbc/blame/b2c67fd7240243a8b08338feea2f0d56f75e89ad/bin/pycbc_optimize_snr)

(It may be good to add this to the description)

ArthurTolley:
@GarethCabournDavies They do but I don't know why they do. Surely unrecognized arguments means it's being given the wrong arguments but I'm giving it arguments which have hyphens and not underscores.

examples/live/run.sh
```diff
@@ -124,7 +124,6 @@ mpirun \
 -host localhost,localhost \
 -n 2 \
 --bind-to none \
- -x PYTHONPATH -x LD_LIBRARY_PATH -x OMP_NUM_THREADS -x VIRTUAL_ENV -x PATH -x HDF5_USE_FILE_LOCKING \
```
GarethCabournDavies, line 127: Undo this change once you are ready to merge

pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,383 @@
...
+            setattr(args, key_name, value[1])
+
+def args_to_string(args):
+    """
+    Convert the supplied arguments for SNR optimization config into
+    a string - this is to be used when running subprocesses
+    """
+    argstr = f' --optimizer {args.optimizer} '
+    for opt in option_dict[args.optimizer]:
+        option_fullname = f"--{args.optimizer}-{opt}"
```
GarethCabournDavies, line 379: args.optimizer here will use the value of args.optimizer from pycbc_live, which will be `pso`, `differential_evolution` or `shgo`. This is where the underscore comes in.

pycbc/live/snr_optimizer.py
ArthurTolley: Needed to actually add the seed in too, do you know if setting the seed in `pycbc_optimize_snr` will pass that numpy arg seed through to `snr_optimizer`?

GarethCabournDavies: I dont think so, the seed was added after I did the initial options change

GarethCabournDavies: (though you can test this by running twice and checking outputs are the same)

ArthurTolley: They have the same outputs when running twice so I'm going to say it is passed through.

pycbc/live/snr_optimizer.py
tdent: well, comments are cheap .. it seems OK atm

pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,384 @@
...
+
+def args_to_string(args):
+    """
+    Convert the supplied arguments for SNR optimization config into
+    a string - this is to be used when running subprocesses
+    """
+    argstr = f' --optimizer {args.optimizer} '
+    optimizer_name = args.optimizer.replace('_','-')
+    for opt in option_dict[args.optimizer]:
+        option_fullname = f"--{optimizer_name}-{opt}"
```
tdent, line 380: this is slightly messy in that some of these strings have spaces at the end and/or the beginning and some don't .. can it be redone so that they all have exactly 1 space at the beginning? 
That way it should be easier to check / parse. 

Also this line has double quotes whereas all the others have single .. hopefully this is not for any reason, then it's easier to read with all single ;)

tdent (changes requested):
Only a few formatting / comment requests

ArthurTolley:
@tdent All done I believe 😄

tdent (approved):

pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,383 @@
...
+        return True
+    Nfeval += 1
+
+
+def compute_network_snr_core(v, *argv, raise_err=False):
+    """
+    Compute network SNR as a function over mchirp, eta, and two aligned
+    spin components, stored in that order in the sequence v.
+
+    Parameters:
```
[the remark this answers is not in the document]
GarethCabournDavies: See e.g. [here](https://github.com/gwastro/pycbc/blob/f037f98b914483693b906b4b951332b408b7e893/pycbc/io/live.py#L34)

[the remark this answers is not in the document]
ArthurTolley: Ah, thank you very much. I did a quick search for a standard PyCBC docstring but couldn't find anything concrete. I will change it now

pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,383 @@
...
+    opt_opt_group.add_argument('--optimizer',
+        type=str,
+        default='differential_evolution',
+        choices=optimizer_choices,
+        help='The optimizer to use, choose from ' + ', '.join(optimizer_choices))
+
+    # Add the generic options
+    opt_opt_group.add_argument('--include-candidate-in-optimizer',
+        action='store_true',
+        help='Include parameters of the candidate event in the initialized array for the optimizer. '
```
GarethCabournDavies, line 323: Some of the codeclimate issues are worth fixing - this line being too long is one of them.

Please have a look at the codeclimate issues. Some of the are due to standard things for/from the optimizers, but others should be fixed (e.g. adding module docstring)

pycbc/live/snr_optimizer.py
```diff
@@ -40,19 +42,23 @@ def compute_network_snr_core(v, *argv, raise_err=False):
...
-    - network_snr (float): The computed network SNR (Signal-to-Noise
-                           Ratio) value.
-    - snr_series_dict (dict): A dictionary containing the snr timeseries
-                              from each ifo.
+    Parameters
+    ----------
+    v : list
+        A list containing the input values for mchirp, eta, and spin
+        components.
+    *argv
```
GarethCabournDavies, line 50: argv has to be a dictionary, right?

GarethCabournDavies: That looks to be a specifically ordered list of arguments for the function, which I am not a fan of, but changing it is outside the scope of this PR

pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,427 @@
+# Copyright (C) 2023
```
GarethCabournDavies, line 1:
```suggestion
# Copyright (C) 2023 Arthur Tolley
```

pycbc/live/snr_optimizer.py
ArthurTolley: Agreed, I feel like it somewhat defeats the purpose of the *argv being flexible in the number of arguments that can be passed and then forcing a specific number of ordered arguments within it!

pycbc/live/snr_optimizer.py
ArthurTolley: I did start doing this but didn't know whether to add the names of everyone who has worked on the snr optimization or just my own? Felt weird to just put myself

GarethCabournDavies: Too late, oh well

ArthurTolley:
To fix the docstring problem I just implemented the change suggested by @GarethCabournDavies above, removing the `*argv` input to `compute_network_snr_core` and instead include the actual components. This still works without changing any of the other scripts that call the snr optimizer as long as they remain as they are.

examples/live/run.sh
```diff
@@ -195,13 +195,27 @@ python -m mpi4py `which pycbc_live` \
 --src-class-eff-to-lum-distance 0.74899 \
 --src-class-lum-distance-to-delta -0.51557 -0.32195 \
 --run-snr-optimization \
+--optimizer differential_evolution \
+--differential-evolution-maxiter 50 \
+--differential-evolution-popsize 100 \
+--include-candidate-in-optimizer \
+--optimizer-seed 42 \
```
titodalcanton, line 202: Can we make it a bit clearer that those options are related to the SNR optimization? Maybe prefixing everything with `--snr-opt`?

bin/pycbc_optimize_snr
```diff
@@ -286,67 +60,36 @@ parser.add_argument('--output-path', required=True,
...
-if args.snr_threshold != 4:
-    parser.error("Sorry, the SNR threshold option doesn't work yet")
-if args.optimizer == 'pso' and ps == None:
-    parser.error('You need to install pyswarms to use the pso optimizer.')
+
 pycbc.init_logging(args.verbose)
 
-if args.seed:
-    numpy.random.seed(args.seed)
+if args.optimizer_seed:
```
titodalcanton, line 71: What if I want to use a seed of zero? This should be `if args.optimizer_seed is not None:`

ArthurTolley: ah very true, I forget that python `0 == False`!

ArthurTolley:
I also realised now that I am probably using an old version of the pycbc live conda environment so I will reclone and add install these changes to do some more testing

bin/pycbc_optimize_snr
GarethCabournDavies: In fact the current default is 42, so we may not even need the if statement

examples/live/run.sh
tdent: I would worry that argparse doesn't cope well with very long option names so we might have to prefer eg `--snr-opt-di-maxiter`

pycbc/live/snr_optimizer.py
```diff
@@ -0,0 +1,402 @@
...
+        help='SNR Optimizer choices: ' + ', '.join(optimizer_choices))
+
+    # Add the generic options
+    opt_opt_group.add_argument('--include-candidate-in-optimizer',
+        action='store_true',
+        help='Include parameters of the candidate event in the initialized '
+             'array for the optimizer. Only relevant for --optimizer pso or '
+             'differential_evolution')
+    opt_opt_group.add_argument('--optimizer-seed',
+        type=int,
```
tdent, line 344: Suggest to leave this as a string (the default type is str, so no need to specify that) and then if the string is 'None' the seed is random, otherwise it is to be converted to int

bin/pycbc_optimize_snr
tdent: See my suggestion. (Basically allow 'None' or a string which can be converted to integer, and the default answer should be '42')

GarethCabournDavies: Okay, so you would specifically state `--optimizer-seed None` in the arguments if you want it turned off?

examples/live/run.sh
titodalcanton: Some examples which do not seem that much longer to me:
```
--snr-opt-method differential_evolution
--snr-opt-di-max-iter
--snr-opt-di-pop-size
--snr-opt-include-candidate
--snr-opt-seed
```

pycbc/live/snr_optimizer.py
titodalcanton: See above, I prefer the string "random" rather than "None".

bin/pycbc_optimize_snr
tdent: Yes, that's what I said. 'None' would access the default behaviour of numpy.random.seed when given None ... maybe 'random' would be a good alias for that

pycbc/live/snr_optimizer.py
tdent: or why not both / either ..

ArthurTolley: I have included the option of both but the help for the option explicitly says use 'random' if you wish to use a random seed. I think this logic is correct: `if args.snr_opt_seed is not None and args.snr_opt_seed != 'random':`

examples/live/run.sh
ArthurTolley: I have updated the argument names to all include the prefix 'snr-opt'. This requires a bit of code changes because the argument given to 'snr-opt-method' is 'differential-evolution' but the options pertaining to that method are all labelled 'snr-opt-di-*' to keep them shortened.

tdent:
@ArthurTolley @GarethCabournDavies where are we with this PR?

GarethCabournDavies:
So far as I can tell, we just need someone to merge it - @ArthurTolley, is that right?

tdent:
Codeclimate has some weird looking error  .. 
```
warning: Could not find remote branch live_options to clone.
fatal: Remote branch live_options not found in upstream origin
```

ArthurTolley:
That is a really weird error, as far as I can tell the branch should be available 🤷🏻‍♂️

pycbc/live/snr_optimizer.py
```diff
@@ -392,11 +396,13 @@ def args_to_string(args):
...
-    for opt in option_dict[args.optimizer]:
-        option_fullname = f'--{optimizer_name}-{opt}'
-        key_name = f'{args.optimizer}_{opt}'
+    argstr = f'--snr-opt-method {args.snr_opt_method} '
+    optimizer_name = args.snr_opt_method.replace('_', '-')
+    if optimizer_name == 'differential-evolution':
+        optimizer_name = 'di'
+    for opt in option_dict[args.snr_opt_method]:
+        option_fullname = f'--snr-opt-{optimizer_name}-{opt}'
+        key_name = f'snr_opt_{optimizer_name}_{opt}'
```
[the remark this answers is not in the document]
GarethCabournDavies: that is okay actually, at the moment

examples/live/run.sh
```diff
@@ -195,13 +195,26 @@ python -m mpi4py `which pycbc_live` \
 --src-class-eff-to-lum-distance 0.74899 \
 --src-class-lum-distance-to-delta -0.51557 -0.32195 \
 --run-snr-optimization \
+--snr-opt-di-maxiter 50 \
+--snr-opt-di-popsize 100 \
+--snr-opt-include-candidate \
+--snr-opt-seed 42 \
```
tdent, line 201: I would remove this option, as it is equal to the default and IMO we prefer to use default values for checking

tdent: (i.e. if at some point something bad happens to the default, we want the test to be sensitive to it)

bin/pycbc_optimize_snr
```diff
@@ -68,8 +68,9 @@ args = parser.parse_args()
 
 pycbc.init_logging(args.verbose)
 
-if args.optimizer_seed:
-    numpy.random.seed(args.optimizer_seed)
+if args.snr_opt_seed is not None and args.snr_opt_seed != 'random':
```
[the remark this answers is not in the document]
tdent: Yes, but just in case ..

examples/live/run.sh
ArthurTolley: I can remove it, should I also remove the commented out options which indicate how to use particle swarm optimization?

pycbc/live/snr_optimizer.py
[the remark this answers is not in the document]
ArthurTolley: I think the design for potential future options would be: If they're long enough to require a hypen then they will probably take a shorted version similar to how 'differential-evolution' becomes 'di'

spxiwh:
... Although it's probably the case that most of these issues are from me, not from this patch. So maybe not fair to enforce that.

examples/live/run.sh
ArthurTolley: I included them to make the example a bit more helpful for those that would like to test and use the pso optimizer (as it is the superior optimizer).

ArthurTolley:
I did clean the codeclimate of any issues relating to this PR.

tdent:
Out of the remaining codeclimate whines, I think removing the unused argparse import is worth noting

GarethCabournDavies:
> @spxiwh @ArthurTolley why are we using `numpy.random.mtrand._rand` ? Isn't there a more 'standard' way to get a random number?

Thats from a previous PR - https://github.com/gwastro/pycbc/pull/4393, but its basically one of a few ways of getting the global random state (see https://stackoverflow.com/a/41985898)

tdent:
(should be fixed now)

ArthurTolley:
@tdent Thank you for that, I'm getting some strange problems with trying to use VSCode today and I'm really rusty using the terminal to do changes.

PR #4438, opened 2023-07-14
Automatically make files for comprehensive offline gracedb uploads

GarethCabournDavies:
In order to avoid manually making skymaps for the data release (as in O3), we want to include files in the offline uploads as standard.

This includes:
 - the XML being uploaded _including the PSD and SNR timeseries_
 - ~an HDF file being uploaded, alongside the PSD (as is done for live)~
 - a plot of the SNR timeseries _and PSDs for each detector (active or not)_
 - _The fits file and a skymap plot, made using information which would be lost if it relied on GraceDB information_

NB - edits to this description are in _italic text_


pycbc/results/snr.py
```diff
@@ -0,0 +1,73 @@
...
+#
+#                                   Preamble
+#
+# =============================================================================
+#
+"""
+Module to generate SNR figures
+"""
+import numpy as np
+import matplotlib; matplotlib.use('Agg')
```
titodalcanton, line 30: Please do not set the matplotlib backend at import time.

pycbc/results/snr.py
```diff
@@ -0,0 +1,73 @@
...
+    ref_time : number, GPS seconds
+        Reference time which will be used as the zero point of the plot
+        This should be an integer value, but doesn't need to be an integer
+
+    Returns
+    -------
+        None
+    """
+    pl.figure()
+    for ifo in sorted(snrdict.keys()):
```
titodalcanton, line 57: `.keys()` is not necessary.

pycbc/results/snr.py
```diff
@@ -0,0 +1,73 @@
...
+        curr_snrs = snrdict[ifo]
+
+        pl.plot(curr_snrs.sample_times - ref_time, abs(curr_snrs),
+                c=ifo_color(ifo), label=ifo)
+        if ifo in triggers:
+            pl.plot(triggers[ifo][0] - ref_time,
+                    triggers[ifo][1], marker='x', c=ifo_color(ifo))
+
+    pl.legend()
+    pl.xlabel(f'GPS time from {int(ref_time)} (s)')
```
titodalcanton, line 67: This risks introducing a bug if `ref_time` violates the requirement that it is an integer value (which I can easily see happening - docstrings are easy to ignore): the actual zero will be `ref_time`, but the axis label will claim it is `int(ref_time)`. I recommend sticking to what the old code did, i.e. *force* `ref_time` to be integer-valued once and then assume it is throughout the function (and print it with the `:d` formatter).

pycbc/workflow/minifollowups.py
```diff
@@ -340,10 +344,125 @@ class PlotQScanExecutable(PlotExecutable):
...
+            node.add_opt('--spin2x',"%.6f" % params['spin2x'])
+            node.add_opt('--spin2y',"%.6f" % params['spin2y'])
+            node.add_opt('--inclination',"%.6f" % params['inclination'])
+            try:
+                node.add_opt('--u-val',"%.6f" % params['u_vals'])
+            except:
+                node.add_opt('--u-val',
+                             "%.6f" % params['u_vals_%s' % ifo])
+
+    # str(numpy.float64) restricts to 2d.p. BE CAREFUL WITH THIS!!!
```
titodalcanton, line 444: What does this comment refer to? I do not see that conversion used in the next few lines.

pycbc/workflow/minifollowups.py
```diff
@@ -340,10 +344,125 @@ class PlotQScanExecutable(PlotExecutable):
...
+    if params[ifo + '_end_time'] > 0 and not use_mean_time:
+        str_trig_time = f"{params[ifo + '_end_time']:.6f}"
+    else:
+        str_trig_time = f"{params['mean_time']:.6f}"
+    node.add_opt('--trigger-time', str_trig_time)
```
titodalcanton, line 449: I would do this like so:
```suggestion
    if params[ifo + '_end_time'] > 0 and not use_mean_time:
        trig_time = params[ifo + '_end_time']
    else:
        trig_time = params['mean_time']
    node.add_opt('--trigger-time', f'{trig_time:.6f}')
```
so there is only one string conversion.

pycbc/workflow/minifollowups.py
GarethCabournDavies: Copied from previous usage - removed

GarethCabournDavies:
Now trying a wrapper script workaround as bayestar doesnt look like it will be able to output individual filenames any time soon

Search will fail as we don't include ligo-make-skymap or bayestar-localize-coincs in pycbc. I will remove the changes from the config files before merging

GarethCabournDavies:
There is an error with `run small search using pegasus + condor`, which will be fixed once I revert the changes to the config files.

There are also a couple minor changes I plan to make to the single-event-upload a script, but I think this is ready to review otherwise.

pycbc/workflow/minifollowups.py
```diff
@@ -842,3 +948,192 @@ def make_skipped_html(workflow, skipped_data, out_dir, tags):
     workflow += node
     files = node.output_files
     return files
+
+
+def make_upload_files(workflow, psd_files, snr_timeseries, xml_all,
+                      event_id, approximant, out_dir, tags=None):
+    """
+    Make files including coinc and skymap fits for uploading to gracedb
+    """
```
GarethCabournDavies, line 957:
```suggestion
    """
    Make files including xml, skymap fits and plots for uploading to gracedb
    for a given event
    
    Parameters
    ----------
    psd_files: FileList([])
        PSD Files from MERGE_PSDs for the search as appropriate for the
        event
    snr_timeseries: FileList([]) 
        SNR timeseries files, one from each IFO, to add to the XML and plot
        output from pysbs_single_template
    xml_all: File
        XML file containing all events from the search
    event_id: string
        an integer to describe the event's position in the xml_all file 
    approximant: byte string
        The approximant used for the template of the event, to be passed
        to bayestar for sky location
    out_dir:
        The directory where all the output files should go
    tags: {None, optional}
        Tags to add to the minifollowups executables

    Returns
    -------
    all_output_files: FileList
        List of all output files from this process
    """
```

pycbc/workflow/minifollowups.py
```diff
@@ -842,3 +948,192 @@ def make_skipped_html(workflow, skipped_data, out_dir, tags):
...
+    out_dir: path
+        The directory to store minifollowups result plots and files
+    tags: {None, optional}
+        Tags to add to the minifollowups executables
+
+    Returns
+    -------
+    layout: list
+        A list of tuples which specify the displayed file layout for the
+        minifollops plots.
```
GarethCabournDavies, line 1079:
```suggestion
        minifollowups plots.
```

pycbc/workflow/psd.py
```diff
@@ -73,10 +73,7 @@ def setup_psd_calculate(workflow, frame_files, ifo, segments,
                                     segment_name, out_dir,
                                     tags=tags + ['PART%s' % i])]
 
-    if num_parts > 1:
-        return merge_psds(workflow, psd_files, ifo, out_dir, tags=tags)
-    else:
-        return psd_files[0]
+    return merge_psds(workflow, psd_files, ifo, out_dir, tags=tags)
```
GarethCabournDavies, line 76: Always use merged psds, even if there is only one calculate_psd job

GarethCabournDavies (commented):
Added a few comments from myself for description / fixing a couple of things

GarethCabournDavies:
Example upload:

https://gracedb-playground.ligo.org/events/G1190222/view/

bin/all_sky_search/pycbc_make_bayestar_skymap
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
titodalcanton: Please indicate in the docstring that this wrapper has to exist because the XML format does not indicate the f-low and waveform information.

bin/all_sky_search/pycbc_make_bayestar_skymap
```diff
@@ -0,0 +1,102 @@
...
+logging_level = args.verbose + 1 if args.verbose else None
+init_logging(logging_level)
+logging.info("Starting")
+
+bayestar_exe = args.bayestar_executable if args.bayestar_executable \
+    else 'bayestar-localize-coincs'
+
+# Use a seed to always make the temp directory the same
+# This is the same as the seed passed to bayestar
+if args.seed:
```
titodalcanton, line 54:
```suggestion
if args.seed is not None:
```
Also, can `random.seed()` really take a string argument?

bin/all_sky_search/pycbc_make_bayestar_skymap
```diff
@@ -0,0 +1,102 @@
...
+    random.seed(args.seed)
+
+# Set up a temporary directory so that the fits file shouldnt clash
+# with any existing ones
+tmpdir_name = 'bayestar_tmp_output_%030x' % random.randrange(16**30)
+tmpdir = os.path.join(os.getcwd(), tmpdir_name)
```
titodalcanton, line 60: This seems overly complicated. I recommend using `tempfile.mkdtemp()` and not worrying about the name.

bin/all_sky_search/pycbc_make_bayestar_skymap
```diff
@@ -0,0 +1,102 @@
...
+
+# Make the directory if it doesnt exist:
+# Don't delete the temporary directory if it already existed
+rmtmpdir = False
+try:
+    os.mkdir(tmpdir)
+    rmtmpdir = True
+except OSError:
+    logging.info("Directory %s already exists", tmpdir)
```
titodalcanton, line 69: Remove this if you switch to `tempfile.mkdtemp()` above.

bin/all_sky_search/pycbc_make_bayestar_skymap
```diff
@@ -0,0 +1,102 @@
...
+except subprocess.CalledProcessError as e:
+    logging.warn("Error in running bayestar")
+    logging.warn(e.output)
+    sys.exit(1)
+
+# Find the fits file in the temporary directory:
+# It would be nice to do this better! - maybe use the input xml
+# file to find the number which is going to be used?
+fits_filename = [os.path.join(tmpdir, f) for f in os.listdir(tmpdir)
+                 if f.endswith('fits')][0]
```
titodalcanton, line 94: This is easier with the glob module:
```
fits_filenames = glob.glob(os.path.join(tmpdir, '/*.fits*'))
if len(fits_filenames) != 1:
    raise ValueError(f'argh, got {len(fits_filenames)} FITS files after running BAYESTAR, I want one!')
```

bin/all_sky_search/pycbc_make_bayestar_skymap
titodalcanton: …apparently it can! But seem my next comment, I think you can remove this?

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,230 @@
...
+from pycbc.types import MultiDetOptionAction
+from pycbc.results import generate_asd_plot, generate_snr_plot
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument("--verbose", action='count',
+                    help="Increase logging level, default=info")
+parser.add_argument("--psd-files", nargs='+', required=True,
+                    help='HDF file(s) containing the PSDs to upload')
+parser.add_argument("--snr-timeseries", nargs='+', required=True,
+                    help='HDF file(s) containing the SNR timeseries to upload')
```
titodalcanton, line 53: Is the idea here to pass one path per event? With hundreds of events that is going to be a long command line, no? Does it actually work?

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,230 @@
...
+parser.add_argument('--input-file', required=True, type=str,
+                    help='Input LIGOLW XML file of coincidences.')
+parser.add_argument('--event-id', type=int,
+                    help='Only make XML file for this event')
+parser.add_argument('--output-file', default=os.getcwd(),
+                    help="Output directory for locally stored XML and PSD "
+                         "files. Default: current directory")
```
titodalcanton, line 60: Rename the option to `--output-dir` or `--output-base`? `--output-file` implies it takes a *file* path.

bin/all_sky_search/pycbc_make_bayestar_skymap
```diff
@@ -0,0 +1,102 @@
...
+parser.add_argument('--waveform', default='TaylorF2',
+                    help="Waveform used in the matched-filtering "
+                         "to generate the SNR timeseries.")
+parser.add_argument('--low-frequency-cutoff', type=float, default=20,
+                    help="Low-frequency cutoff used in the matched-filtering "
+                         "to generate the SNR timeseries")
+parser.add_argument('--seed',
+                    help="Seed to pass to bayestar-localize-coincs")
+parser.add_argument('--output-file', required=True,
+                    help="Filename to output the fits file to.")
```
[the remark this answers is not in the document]
GarethCabournDavies: Unrecognised options may be passable, which may allow us to set things in the config files as normal

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,230 @@
...
+parser.add_argument('--snr-timeseries-plot', required=True,
+                    help="Output filename for the plot of SNR timeseries. "
+                         "Must be a plot filetype as used by matplotlib.")
+parser.add_argument('--psd-plot', required=True,
+                    help="Output filename for the plot of the PSD. "
+                         "Must be a plot filetype as used by matplotlib.")
```
titodalcanton, line 66: Given the lack of precise requirements, I suggest making these options not required, and actually making the plots only if the options are given. In this way we can control what gets uploaded from the config files without having to change the code.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,230 @@
...
+    def __init__(self, psd, *args):
+        self.psd = psd
+
+psds = {}
+for psd_file in args.psd_files:
+    (ifo, group), = h5py.File(psd_file, "r").items()
+    psd = [group["psds"][str(i)] for i in range(len(group["psds"].keys()))]
+    psds[ifo] = segmentlist(psd_segment(*segargs) for segargs in zip(
+        psd, group["start_time"], group["end_time"]))
+
```
titodalcanton, line 95: Add a log line "Reading SNR timeseries" here.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
titodalcanton: I now see that this script is probably intended to handle *a single* event. That was not clear at all from the docstring, I assumed "prepare for upload to gracedb" referred to multiple events. Please clarify if necessary.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,230 @@
...
+    if coinc_map.coinc_event_id == event.coinc_event_id:
+        coinc_event_map_table_curr.append(coinc_map)
+        sngl_ids.append(coinc_map.event_id)
+
+# Get the SNR timeseries and PSDs at the time of this event
+# these may not have contributed a trigger
+snr_timeseries_time_match = [snr_key for snr_key in snr_timeseries.keys()
+                             if abs(snr_key[1] - time) < 0.1]
+ifos = list(zip(*snr_timeseries_time_match))[0]
+time_key = list(zip(*snr_timeseries_time_match))[1][0]
```
titodalcanton, line 148: List comprehensions are easier to read here IMO. I cannot tell if the second line has a typo.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,230 @@
...
+# These are the SNR timeseries for this event only:
+
+snr_curr_event = {}
+for ts_key in snr_timeseries_time_match:
+    snr_curr_event[ts_key[0]] = TimeSeries(
+        snr_timeseries[ts_key],
+        delta_t=snr_delta_t[ts_key[1]],
+        epoch=snr_epoch[ts_key[1]],
+        dtype=np.csingle,
+    )
```
titodalcanton, line 158: See my comment above about directly loading a TimeSeries object, then I think you can remove this loop.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,230 @@
...
+    except KeyError:
+        parser.error(
+            "--psd-files {0}: no PSDs found for detector {1}".format(
+            " ".join(args.psd_files), ifo))
+
+    try:
+        psd = psd[psd.find(time)].psd
+    except ValueError:
+        logging.warning("SNR timeseries but no PSD exists for event at %.3f",
+            time)
```
titodalcanton, line 175: What happens after the warning, everything just carries on? Is that what we want? Another case where I would just let the exception crash the code.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,230 @@
...
+sample_freqs = {}
+
+for sngl in sngl_inspiral_table:
+    if sngl.event_id not in sngl_ids:
+        continue
+    sngl_inspiral_table_curr.append(sngl)
+    sngl.channel = args.channel_name[sngl.ifo]
+
+    # Two versions of the PSD dictionary have useful info here
+    psd = psds_curr_event[ifo]
```
titodalcanton, line 195: Typo?
```suggestion
    psd = psds_curr_event[sngl.ifo]
```

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,230 @@
...
+            "--psd-files {0}: no PSDs found for detector {1}".format(
+            " ".join(args.psd_files), ifo))
+
+    try:
+        psd = psd[psd.find(time)].psd
+    except ValueError:
+        logging.warning("SNR timeseries but no PSD exists for event at %.3f",
+            time)
+    # resample the psd to 0.25Hz spacing
+    df = 0.25
```
titodalcanton, line 177: `df` is defined inside a loop but is also used below, outside the loop. Maybe better to use a CLI option with a default instead?

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
GarethCabournDavies: The help was wrong on this occasion, the output is a single file. Help string updated

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
GarethCabournDavies: I will have to work out how to do the 'optionally making the plots' in the workflow, but this seems sensible

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
titodalcanton: Then the default value is also wrong. The option should be required and not have a default if it takes an actual file name.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
titodalcanton: I think for now just make the options optional in this script.

bin/all_sky_search/pycbc_make_bayestar_skymap
```diff
@@ -0,0 +1,102 @@
...
+if args.seed:
+    cmd += ['--seed', args.seed]
+
+logging.info("Running %s", ' '.join(cmd))
+try:
+    subprocess.check_output(cmd)
+except subprocess.CalledProcessError as e:
+    logging.warn("Error in running bayestar")
+    logging.warn(e.output)
+    sys.exit(1)
```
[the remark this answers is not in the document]
GarethCabournDavies: Good point, and this actually loses some information

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
GarethCabournDavies: I have updated to use a dictionary comprehension method which seems much better / more comprehensible

bin/all_sky_search/pycbc_make_bayestar_skymap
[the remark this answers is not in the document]
GarethCabournDavies: This is done on the below line - unknown options are passed to `unknown`, and simply added to the bayestar command later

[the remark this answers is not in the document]
titodalcanton: And I learned a new feature of argparse, cool.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,216 @@
...
+xmldoc.childNodes[-1].removeChild(sngl_inspiral_table)
+xmldoc.childNodes[-1].removeChild(coinc_event_map_table)
+xmldoc.childNodes[-1].removeChild(coinc_inspiral_table)
+xmldoc.childNodes[-1].removeChild(coinc_table)
+
+for event in coinc_table:
+    coinc_event_id = event.coinc_event_id
+    if args.event_id == coinc_event_id:
+        # This is the event we want
+        break
```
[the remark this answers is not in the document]
GarethCabournDavies: I don't see how event could be None, as it would be changed in the loop?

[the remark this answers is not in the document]
titodalcanton: If the user provided `--event-id` is not actually contained in the XML, this will be detected by `event` being None. Without this change, the code would happily pick an effectively random event and proceed.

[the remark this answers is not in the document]
GarethCabournDavies: I see that it would get `event` from the end of the loop in that case, which is bad, but the event would still not be None, it would be the last event in the table. I've added another variable to allow what you're thinking of though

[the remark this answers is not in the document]
titodalcanton: :picard: that was a major thinko, sorry. However I think Python has an even better solution: https://book.pythontips.com/en/latest/for_-_else.html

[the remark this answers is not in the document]
GarethCabournDavies: Surprised I missed this thought - for/else is probably my favourite python feature (its a bit weird that I have one)

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,226 @@
...
+            for sngl in sngl_inspiral_table_curr}
+base_time = int(np.floor(time))
+if args.snr_timeseries_plot:
+    generate_snr_plot(
+        snr_ts,
+        args.snr_timeseries_plot,
+        triggers,
+        base_time
+    )
+    logging.info("Saved SNR plot %s", args.snr_timeseries_plot)
```
titodalcanton, line 226: Turn the "Saved…" messages into "Saving…" messages just before the corresponding operation (in case of an error we want to know what the code was doing, and we would not see the "Saved" message).

Also, add the usual "Done" at the end.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,185 @@
...
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+"""
+Take a coinc xml file containing multiple events and upload to gracedb.
```
titodalcanton, line 20: This docstring contradicts the script name, update as appropriate.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,226 @@
...
+    psds_event[ifo] = psd
+    psddict[ifo] = psd_fs
+
+lal_psddict = {}
+sample_freqs = {}
+
+for sngl in sngl_inspiral_table:
+    if sngl.event_id not in sngl_ids:
+        continue
+    sngl_inspiral_table_curr.append(sngl)
```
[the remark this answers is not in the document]
GarethCabournDavies: Not including subthreshold detectors here, this is inherited from the pycbc_page_foreground output, so does not have this info added. I'm not sure whether we'd want that though or not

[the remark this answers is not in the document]
GarethCabournDavies: I found one from the github example run, and have uploaded to https://gracedb-playground.ligo.org/events/G1268054/view/

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,185 @@
+#!/usr/bin/env python
```
GarethCabournDavies, line 1: I should probably say that this code would be run after the workflow has finished, and the box is opened.

We would use something like:
```
pycbc_upload_single_event_to_gracedb \
  --xml-file-for-upload \
    $upload_data_dir/H1L1V1-GENERATE_XML_FULL_DATA_UPLOAD_3-1186740100-3400.xml \
  --snr-timeseries-plot \
    $upload_data_dir/H1L1V1-GENERATE_XML_FULL_DATA_UPLOAD_3_SNR-1186740100-3400.png \
  --asd-plot \
    $upload_data_dir/H1L1V1-GENERATE_XML_FULL_DATA_UPLOAD_3_PSD-1186740100-3400.png \
  --skymap-plot \
    $upload_data_dir/H1L1V1-SKYMAP_PLOT_FULL_DATA_UPLOAD_3-1186740100-3400.png \
  --skymap-fits-file \
    $upload_data_dir/H1L1V1-BAYESTAR_FULL_DATA_UPLOAD_3-1186740100-3400.fits \
  --labels \
    GRACEDB_LABEL
```

With the upload number (3 here) being looped over and timestamp updated as appropriate

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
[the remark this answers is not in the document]
titodalcanton: Ok, I suspect we do need the single inspiral entries to be present in order for BAYESTAR to use the subthreshold detectors, but let's keep things as they are for now. We can fix that with a later PR if needed.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,185 @@
...
+import logging
+from ligo.gracedb.rest import GraceDb
+import numpy as np
+
+import pycbc
+import lal
+from pycbc.io.live import gracedb_tag_with_version
+
+logging.basicConfig(format='%(asctime)s:%(levelname)s : %(message)s',
+                    level=logging.INFO)
```
titodalcanton, line 35: Use PyCBC's `logging_init()` function here.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,185 @@
...
+
+if args.production_server:
+    gracedb = GraceDb()
+else:
+    gracedb = GraceDb(service_url='https://gracedb-playground.ligo.org/api/')
+
+labels = [l.upper() for l in args.labels]
+allowed_labels = gracedb.allowed_labels
+
+if not all([l in allowed_labels for l in labels]):
```
titodalcanton, line 89:
```suggestion
if set(labels) - set(allowed_labels):
```

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,185 @@
...
+    + ' '.join(sys.argv)
+gracedb.write_log(r['graceid'], command_str)
+
+# add the custom log message, if provided
+if args.log_message is not None:
+    gracedb.write_log(
+        r['graceid'],
+        args.log_message,
+        tag_name=['analyst_comments']
+    )
```
titodalcanton, line 127: I suggest writing all those as a single comment. Concatenate all the comments into a single string separated by a blank line for clarity, and call `write_log()` just once.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,185 @@
...
+    displayname=''
+    upload_file(args.source_probabilities, displayname, comment, tag)
+
+if args.source_probabilities_plot:
+    logging.info("Found source probabilities plot %s", args.source_probabilities_plot)
+    comment = "Source probabilities plot upload"
+    tag = "pe"
+    displayname=''
+    upload_file(args.source_probabilities_plot, displayname, comment, tag)
+
```
titodalcanton, line 185: Log "Done" here.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,185 @@
...
+        displayName=[displayname]
+    )
+
+
+if args.asd_plot:
+    logging.info("Found ASD plot %s", args.asd_plot)
+    comment = "PyCBC ASD estimate from the time of event"
+    tag = "psd"
+    displayname='ASDs'
+    upload_file(args.asd_plot, displayname, comment, tag)
```
titodalcanton, line 149: The use of variables here is a bit verbose, I suggest this:
```suggestion
    upload_file(
        args.asd_plot,
        displayname='ASDs',
        comment='PyCBC ASD estimate from the time of event',
        tag='psd'
    )
```

titodalcanton: Same comment for the case below of course.

bin/minifollowups/pycbc_foreground_minifollowup
```diff
@@ -108,26 +108,32 @@ if args.sort_variable not in f[file_val]:
...
 
+# Save the time / trigger / template ids for the events
 times = {}
 tids = {}
-# Version used for multi-ifo coinc code
 ifo_list = f.attrs['ifos'].split(' ')
 for ifo in ifo_list:
     times[ifo] = f['{}/{}/time'.format(file_val,ifo)][:][event_idx]
     tids[ifo] = f['{}/{}/trigger_id'.format(file_val, ifo)][:][event_idx]
+bank_ids = f['{}/template_id'.format(file_val)][:][event_idx]
```
titodalcanton, line 136:
```suggestion
bank_ids = f[file_val]['template_id'][:][event_idx]
```
though you can remove quite a bit of verbosity with an auxiliary variable `f_cat = f[file_val]` around line 102.

bin/minifollowups/pycbc_foreground_minifollowup
```diff
@@ -178,7 +184,7 @@ while event_count < num_events and curr_idx < (events_to_read - 1):
                         tags=['SKIP_{}'.format(event_count)]),)
         skipped_data = []
 
-    bank_id = f['{}/template_id'.format(file_val)][:][sorting][curr_idx]
+    bank_id = bank_ids[curr_idx]
```
titodalcanton, line 187: Removing the `[sorting]` indexing is intentional right?

GarethCabournDavies: Yeah that's right, bank_ids is already sorted by `event_idx` above

bin/minifollowups/pycbc_foreground_minifollowup
```diff
@@ -188,40 +194,33 @@ while event_count < num_events and curr_idx < (events_to_read - 1):
...
+    params = mini.get_single_template_params(
+        curr_idx,
+        times,
+        bank_data,
+        bank_ids[curr_idx],
+        fsdt,
+        tids
+    )
+
+    sngl_tmplt_files, sngl_tmplt_plots = mini.make_single_template_plots(
```
titodalcanton, line 206: Is the variable `sngl_tmplt_files` actually used below? If not, turn to an underscore.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
GarethCabournDavies: I tried this [this event](https://gracedb-playground.ligo.org/events/G1268234) and it looks quite ugly - seems that the gracedb UI doesn't deal with newlines at all. So I will leave as-is

titodalcanton: :picard: ok

GarethCabournDavies:
Updated the PR name as the scope has widened somewhat

GarethCabournDavies:
We are splitting this PR into two: #4534 and #4535, as the changes were getting too long-winded for a single review

PR #4443, opened 2023-07-20
PyGRB bugfixes/corrections

jakeb245:
A couple bugs have appeared in the new PyGRB workflow.

The first appears in the remnant mass calculation for NSBH injections. There was an issue with some tricky indexing that was a straightforward fix (note that the same sort of call is done in the same function).

The second involved `fp` and `fc` being calculated within one branch of an if statement, but the output is needed elsewhere. Moving it outside of the conditional statement resolves the issue.

### DIFF
--- a/bin/pygrb/pycbc_pygrb_grb_info_table
+++ b/bin/pygrb/pycbc_pygrb_grb_info_table
@@ -74,10 +74,10 @@ data[0].append(utc_time)
 headers.append('Coordinated Universal Time')
 
 data[0].append(str(opts.ra))
-headers.append('R.A. (degrees)')
+headers.append('R.A. (deg)')
 
 data[0].append(str(opts.dec))
-headers.append('Dec (degrees)')
+headers.append('Dec (deg)')
 
 data[0].append(str(opts.sky_error))
 headers.append('Sky Error')
--- a/bin/pygrb/pycbc_pygrb_grb_info_table
+++ b/bin/pygrb/pycbc_pygrb_grb_info_table
@@ -74,10 +74,10 @@ data[0].append(utc_time)
 headers.append('Coordinated Universal Time')
 
 data[0].append(str(opts.ra))
-headers.append('R.A.')
+headers.append('R.A. (degrees)')
 
 data[0].append(str(opts.dec))
-headers.append('Dec')
+headers.append('Dec (degrees)')
 
 data[0].append(str(opts.sky_error))
 headers.append('Sky Error')
--- a/bin/pygrb/pycbc_pygrb_plot_injs_results
+++ b/bin/pygrb/pycbc_pygrb_plot_injs_results
@@ -100,7 +100,7 @@ def load_mass_data(input_file_handle, key, tag):
     return data
 
 
-def load_sky_error_data(input_file_handle, key, tag, trig_data):
+def load_sky_error_data(input_file_handle, key, tag):
     """Extract data related to sky_error from raw injection and trigger data"""
     if tag == 'missed':
         # Missed injections are assigned null values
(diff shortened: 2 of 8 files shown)


jakeb245:
I have added another fix in `pycbc_pygrb_pp_workflow` to adapt to the changes made in #4431. The various names of workflow objects were mixed up, and needed some correction.

The test suite is reporting an interesting error that seems unrelated to my changes here...

jakeb245:
Adding fixes to plotting scripts since inspiral output files were switched to output ra/dec instead of latitude/longitude. These fixes make the code run, but I can't claim the units are 100% correct. We should go through the sky position units for the entire workflow in the near future.

This seems to fix the last of the bugs and the test workflow runs to completion.

pannarale (commented):
The two initial fixes look fine: the bug with `.size` and defining `fc` and `fp`.

Regarding `plot_cho_ifosnr`, `plot_injs_results` and `plot_skygrid` we should consider filling in both ra-dec and longitude-latitude if we indeed want to have access to or plots with both pairs: the output file is determined in `pycb_multi_inspiral`.  Right now you uncovered an inconsistency in the very first page of the results webpage where we have always called the same *pair of numbers* ra-dec and longitude-latitude.  We need to settle this.

bin/pygrb/pycbc_pygrb_plot_injs_results
```diff
@@ -100,7 +100,7 @@ def load_mass_data(input_file_handle, key, tag):
     return data
 
 
-def load_sky_error_data(input_file_handle, key, tag, trig_data):
+def load_sky_error_data(input_file_handle, key, tag):
```
pannarale, line 103: Unless I am missing something, we want to keep injections and trigger separated, hence the need for `input_file_handle` and `trig_data`.

bin/pygrb/pycbc_pygrb_plot_skygrid
pannarale, line 1: If ra and dec are stored in trig_data in radians, this is fine.  But do cross-check with `pycbc_multi_inspiral`

pannarale (changes requested):

bin/pygrb/pycbc_pygrb_plot_injs_results
jakeb245: The `trig_data` in this code contains the triggers from the no-injections side of the workflow. The found-missed injections file (`input_file_handle` here) contains the triggers associated with found injections along with the parameters of those injections.

The error is that it was trying to compare different sets of triggers, which are not necessarily the same size so index errors can happen.

jakeb245:
I think I've satisfied myself that these changes are using the proper units. As @pannarale said, we will input RA and dec into the workflow in degrees and convert into radians when possible, since that is what PyCBC's modules use. For outward facing stuff (i.e. visible items on the results webpage) we should convert back to degrees as necessary. RA/dec values within trigger files will be stored in radians, so `pycbc_pygrb_page_tables` may require a conversion and/or a clarification of units.

Assuming the unit tests pass, I think this is back to being ready to merge.

bin/pygrb/pycbc_pygrb_grb_info_table
```diff
@@ -74,10 +74,10 @@ data[0].append(utc_time)
 headers.append('Coordinated Universal Time')
 
 data[0].append(str(opts.ra))
-headers.append('R.A.')
+headers.append('R.A. (degrees)')
 
 data[0].append(str(opts.dec))
-headers.append('Dec')
+headers.append('Dec (degrees)')
```
titodalcanton, line 80: I suggest abbreviating in "deg" instead of "degrees", as done commonly.

bin/pygrb/pycbc_pygrb_plot_skygrid
```diff
@@ -91,8 +91,8 @@ trig_data = load_data(trig_file, vetoes)
 # Generate sky grid plot
 #
 
-xlabel = "Longitude (Degrees)"
-ylabel = "Latitude (Degrees)"
+xlabel = "RA (Degrees)"
+ylabel = "Dec (Degrees)"
```
titodalcanton, line 95: Same comment as above, "deg" is more common.

bin/pygrb/pycbc_pygrb_plot_injs_results
pannarale: Can you post an example of how you get the error you are referring to?  I'm trying to understand why this error was never encountered before.

This function is meant to calculate the sky errors by considering the injected RA and dec values and the recovered RA and dec values.  So `inj` needs to contain the injected values, while `trig` the recovered ones and we need to ensure we are comparing them correctly (i.e., the recovered values must correspond to the correct injection).

I think we may have to open a separate issue for this.

pannarale (changes requested):
Please see my comment about `bin/pygrb/pycbc_pygrb_plot_injs_results`.

bin/pygrb/pycbc_pygrb_plot_injs_results
jakeb245: Ok, I'll open a new PR for this and put together a case where the error can be seen. It's now been removed from this PR.

pannarale (approved):

PR #4449, opened 2023-08-04
Live single fits combined bug

GarethCabournDavies:
Rates plot was showing percentile values above the largest rate, which makes no sense

https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/pycbclive/combine_average_weighting/testing/H1-2023_07_14-2023_08_02-TRIGGER_FITS_COMBINED-counts-original.png

This fix changes to use the percentile of the _rate_ rather than counts, so that we get the correct value out

https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/pycbclive/combine_average_weighting/testing/H1-2023_07_14-2023_08_02-TRIGGER_FITS_COMBINED-counts.png

Change made to mean as well as conservative

PR #4450, opened 2023-08-04
Live cleanup combine single fits

titodalcanton:
Some cleanup I thought about when reviewing #4449. I broke the change down into a few commits to make checking easier if needed.

### DIFF
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -139,9 +139,9 @@ for ifo in args.ifos:
     l_times = np.array(live_times[ifo])
     fout_ifo.attrs['live_time'] = l_times.sum()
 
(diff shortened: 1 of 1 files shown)


bin/live/pycbc_live_combine_single_fits
```diff
@@ -132,23 +133,15 @@ fout['bins_edges'] = list(bl) + [bu[-1]]
 fout['fits_dates'] = ad + start_time_n
 
 for ifo in args.ifos:
+    logging.info(ifo)
     fout.create_group(ifo)
```
GarethCabournDavies, line 137: `fout_ifo = fout.create_group(ifo)` will avoid a line - I cant see a way to suggest this

GarethCabournDavies (approved):
Thanks for this - I think some of these unused variables were from before the code was separated for combining / plotting, so thanks for that. One (very) minor suggestion, but this looks good to go

PR #4455, opened 2023-08-08
Run conversion on all hdf injection files, allowing LVK injection format hdf files to be used

GarethCabournDavies:
I don't know if the file specification changed, or if someone changed from py2 to py3 without thinking, but this attribute is bytes in some of the examples we have been given

### DIFF
--- a/pycbc/workflow/injection.py
+++ b/pycbc/workflow/injection.py
@@ -147,8 +147,9 @@ def cut_distant_injections(workflow, inj_file, out_dir, tags=None):
     return node.output_files[0]
 
 def inj_to_hdf(workflow, inj_file, out_dir, tags=None):
-    """ Convert injection file to hdf format, if it is already one,
-    this just makes a copy
+    """ Convert injection file to hdf format.
+
+    If the file is already PyCBC HDF format, this will just make a copy.
     """
     if tags is None:
         tags = []
--- a/pycbc/workflow/injection.py
+++ b/pycbc/workflow/injection.py
@@ -147,12 +147,9 @@ def cut_distant_injections(workflow, inj_file, out_dir, tags=None):
     return node.output_files[0]
 
 def inj_to_hdf(workflow, inj_file, out_dir, tags=None):
-    """ Convert injection file to hdf format if not already one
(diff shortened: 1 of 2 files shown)


spxiwh:
Would be good to get a check from Tom or the Rates team in terms of what to expect here .... and to check nothing *else* has changed!

ahnitz:
[:] would be wrong as these should be scalars.

GarethCabournDavies:
Okay it is [this line](https://github.com/gwastro/pycbc/blob/b711ca3fea274d5f1351ca7c9ce3e893971b0d7e/pycbc/workflow/injection.py#L153), which just returns the file without converting if it is a hdf file which we want to change as well

GarethCabournDavies:
This now passes injections as appropriate to the inspiral jobs

GarethCabournDavies:
> Would be good to get a check from Tom or the Rates team in terms of what to expect here .... and to check nothing _else_ has changed!

I asked the rates team about this, they suggested to allow both bytes or string if we can

pycbc/workflow/injection.py
```diff
@@ -147,12 +147,9 @@ def cut_distant_injections(workflow, inj_file, out_dir, tags=None):
     return node.output_files[0]
 
 def inj_to_hdf(workflow, inj_file, out_dir, tags=None):
-    """ Convert injection file to hdf format if not already one
+    """ Convert injection file to hdf format, if it is already one,
+    this just makes a copy
```
spxiwh, line 151: Docstring naming convention says that line 1 should be a single sentence.

So this should really be:

```
""" Convert injection file to hdf format.

If the file is already PyCBC HDF format, this will just make a copy.
"""

spxiwh (approved):
Thanks, merge when ready.

This PR *also* runs conversion on all HDF files, allowing us to actually use LVK injection files, which is more than just a metadata naming correction patch as the title now suggests.

GarethCabournDavies:
> Thanks, merge when ready.
> 
> This PR _also_ runs conversion on all HDF files, allowing us to actually use LVK injection files, which is more than just a metadata naming correction patch as the title now suggests.

Good point - I have updated the title of the PR to reflect this

PR #4460, opened 2023-08-14
Idq workflow times

GarethCabournDavies:
The PyCBCCalculateDQFlagExecutable output file was initialised with the workflow analysis times.

This meant that the outputs from each segment were all named the same as each other, so the run would not start.

This seems to fix things, but will wait for @maxtrevor to confirm that his run is working


spxiwh (approved):
Clear bugfix. Approved, thanks!

GarethCabournDavies:
> Clear bugfix. Approved, thanks!

Cheers - I will merge once the search CI completes

PR #4461, opened 2023-08-18
default bug in cluster-window in pycbc_inspiral

GarethCabournDavies:
pycbc_inspiral's --cluster-window option has a default of -1, but expects 0 to indicate no clustering in [this line](https://github.com/gwastro/pycbc/blob/1e360a077992d851c579d7fee3600ebe05dc07fe/bin/pycbc_inspiral#L397)

Updating to use 0 as default to fix this

Found by accident when I introduced a different bug in testing :wink:


ahnitz:
@GarethCabournDavies For this, we should wait to make sure all the tests pass, even the optional ones.

GarethCabournDavies:
> @GarethCabournDavies For this, we should wait to make sure all the tests pass, even the optional ones. The search is normally listed as optional only because it is slow, but in this case, it should be exercised.

Good point - sorry!

PR #4462, opened 2023-08-21
Bugfix for multiple dqflag files

maxtrevor:
This bugfix is needed for the use of dq flags with multiple segments in the O4 production PyCBC offline search


GarethCabournDavies (approved):
bugfix. Looks to do what it is supposed to

PR #4463, opened 2023-08-23
Add KDE/DQ statistic

GarethCabournDavies:
Add in a statistic which uses both KDE and DQ reweighting

Checking with @tdent and @PRAVEEN-mnl that the KDE bit has been taken in properly, and with @maxtrevor that the DQ stuff has been brought in okay.

The summary is as follows:
 - copy `__init__` from the DQ statistic, but add in the stuff needed for kde - `assign_kdes` basically, but needs to pass this through to the other class.
 - inherit `logsignalrate` from the KDE statistic - Please confirm there is no DQ reweighting here
 - inherit `lognoiserate` from the DQ statistic - Please confirm there is no KDE reweighting here

### DIFF
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -2093,12 +2093,6 @@ class DQExpFitFgBgKDEStatistic(DQExpFitFgBgNormStatistic):
         """
         DQExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
                                            ifos=ifos, **kwargs)
-        self.find_kdes()
-
-    def find_kdes(self):
-        """
-        Inherited, see docstring for ExpFitFgBgKDEStatistic.find_kdes
-        """
         ExpFitFgBgKDEStatistic.find_kdes(self)
 
     def assign_kdes(self, kname):
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -1829,8 +1829,6 @@ class ExpFitFgBgKDEStatistic(ExpFitFgBgNormStatistic):
         """
         ExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
                                          ifos=ifos, **kwargs)
-        # Initialize variable to hold event template id(s)
-        self.curr_tnum = None
         self.find_kdes()
         # Initialize variable to hold event template id(s)
         self.curr_tnum = None
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -1829,6 +1829,8 @@ class ExpFitFgBgKDEStatistic(ExpFitFgBgNormStatistic):
         """
         ExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
                                          ifos=ifos, **kwargs)
+        # Initialize variable to hold event template id(s)
+        self.curr_tnum = None
         self.find_kdes()
         # Initialize variable to hold event template id(s)
         self.curr_tnum = None
@@ -2100,8 +2102,17 @@ class DQExpFitFgBgKDEStatistic(DQExpFitFgBgNormStatistic):
 
     def __init__(self, sngl_ranking, files=None, ifos=None, **kwargs):
         """
-        Inherited from DQExpFitFgBgNormStatistic, see docstring there
-        but with find_kdes fucntion from ExpFitFgBgKDEStatistic added
+        Parameters
+        ----------
+        sngl_ranking: str
+            The name of the ranking to use for the single-detector triggers.
+        files: list of strs, needed here
+            A list containing the filenames of hdf format files used to help
+            construct the coincident statistics. The files must have a 'stat'
+            attribute which is used to associate them with the appropriate
+            statistic class.
+        ifos: list of strs, not used here
+            The list of detector names
         """
         DQExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
                                            ifos=ifos, **kwargs)
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -1406,6 +1406,8 @@ class ExpFitFgBgNormStatistic(PhaseTDStatistic,
                                        for ifo in ref_ifos], axis=0)
         self.benchmark_logvol = 3.0 * numpy.log(hl_net_med_sigma)
         self.single_increasing = False
+        # Initialize variable to hold event template id(s)
+        self.curr_tnum = None
 
     def assign_median_sigma(self, ifo):
         """
@@ -1484,14 +1486,14 @@ class ExpFitFgBgNormStatistic(PhaseTDStatistic,
         singles['sigmasq'] = trigs['sigmasq'][:]
         singles['snr'] = trigs['snr'][:]
         try:
-            tnum = trigs.template_num  # exists if accessed via coinc_findtrigs
+            self.curr_tnum = trigs.template_num  # exists if accessed via coinc_findtrigs
         except AttributeError:
-            tnum = trigs['template_id']  # exists for SingleDetTriggers
+            self.curr_tnum = trigs['template_id']  # exists for SingleDetTriggers
             # Should only be one ifo fit file provided
             assert len(self.ifos) == 1
         # Store benchmark log volume as single-ifo information since the coinc
         # method does not have access to template id
-        singles['benchmark_logvol'] = self.benchmark_logvol[tnum]
+        singles['benchmark_logvol'] = self.benchmark_logvol[self.curr_tnum]
         return numpy.array(singles, ndmin=1)
 
     def rank_stat_single(self, single_info,
@@ -1830,8 +1832,6 @@ class ExpFitFgBgKDEStatistic(ExpFitFgBgNormStatistic):
         ExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
                                          ifos=ifos, **kwargs)
         self.find_kdes()
-        # Initialize variable to hold event template id(s)
-        self.curr_tnum = None
 
     def find_kdes(self):
         """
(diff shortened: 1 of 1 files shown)


GarethCabournDavies:
Quick question on previous implementations: does the maximum KDE value for the signal rate need to be added in to `coinc_lim_for_thresh`? and the same with DQ for the noise rate? I feel like this is going to be messing with some edge cases

pycbc/events/stat.py
```diff
@@ -2083,6 +2083,83 @@ def lognoiserate(self, trigs):
...
+            A list containing the filenames of hdf format files used to help
+            construct the coincident statistics. The files must have a 'stat'
+            attribute which is used to associate them with the appropriate
+            statistic class.
+        ifos: list of strs, not used here
+            The list of detector names
+        """
+        DQExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
+                                           ifos=ifos, **kwargs)
+        # The stat file attributes are hard-coded as 'signal-kde_file'
```
tdent, line 2110: I expect codeclimate complained about this, and (IMO) correctly - these lines should rather be refactored as another method of the ExpFitFgBgKDEStatistic (eg `find_kdes` ?).  Hopefully this also won't require you to define `assign_kdes` redundantly here.

tdent:
> does the maximum KDE value for the signal rate need to be added in to coinc_lim_for_thresh?

I don't follow the question, where are you seeing a maximum KDE value?

pycbc/events/stat.py
```diff
@@ -2083,6 +2083,83 @@ def lognoiserate(self, trigs):
...
+        -----------
+        kname: str
+            Used to label the kde files.
+        """
+        ExpFitFgBgKDEStatistic.assign_kdes(self, kname)
+
+
+    def logsignalrate(self, stats, shift, to_shift):
+        """
+        Calculate the normalized log rate density of signals via lookup.
```
tdent, line 2140: I would replace this docstring by eg 'See docs for ExpFitFgBgKDEStatistic.logsignalrate' to avoid duplication

GarethCabournDavies:
> > does the maximum KDE value for the signal rate need to be added in to coinc_lim_for_thresh?
> 
> I don't follow the question, where are you seeing a maximum KDE value?

I'm not seeing one, but `coinc_lim_for_thresh` assumes a best-case scenario for the phasetd histogram. Is the same needed for KDE?

GarethCabournDavies:
> > I'm not seeing one, but coinc_lim_for_thresh assumes a best-case scenario for the phasetd histogram. Is the same needed for KDE?
> 
> Oh, I see. But you can consult what's already coded for the KDE statistic: since the lim_for_thresh calculation is done separately for each template we take the existing ExpFitFgBg(etc) lim value and add the KDE term for the current template - same as in the BBH case basically.

Yup, that makes sense, and as the DQ statistic only affects how the lognoiserate is calculated into snglstat, I can inherit this from KDE. I think that will work

maxtrevor:
I confirm that the DQ stat only modifies the lognoiserate calculation

pycbc/events/stat.py
```diff
@@ -2084,6 +2090,41 @@ def lognoiserate(self, trigs):
...
+    The ExpFitFgBgKDEStatistic with DQ-based reranking.
+
+    This is the same as the DQExpFitFgBgNormStatistic except the signal
+    rate is adjusted according to the KDE statistic files
+    """
+
+    def __init__(self, sngl_ranking, files=None, ifos=None, **kwargs):
+        """
+        Inherited from DQExpFitFgBgNormStatistic, see docstring there
+        but with find_kdes fucntion from ExpFitFgBgKDEStatistic added
```
tdent, line 2104: Not sure we want to squash this docstring if it is not a straight inheritance without change .. probably better to keep as it was.

GarethCabournDavies:
> I put in one minor edit based on CC. However I'm not sure this will work yet because the KDE statistic requires `self.curr_tnum` to be set to look up the KDE values, this is done at present inside the KDE stat `single()` method which I think is not accessible to the new stat at present.

Ah - okay, that may be difficult then, as that will be the only place there is a difference in the same function for the two statistics

GarethCabournDavies:
> > I put in one minor edit based on CC. However I'm not sure this will work yet because the KDE statistic requires `self.curr_tnum` to be set to look up the KDE values, this is done at present inside the KDE stat `single()` method which I think is not accessible to the new stat at present.
> 
> Ah - okay, that may be difficult then, as that will be the only place there is a difference in the same function for the two statistics

Think I have sorted this

pycbc/events/stat.py
```diff
@@ -1856,29 +1864,6 @@ def assign_kdes(self, kname):
         with h5py.File(self.files[kname+'-kde_file'], 'r') as kde_file:
             self.kde_by_tid[kname+'_kdevals'] = kde_file['data_kde'][:]
 
-    def single(self, trigs):
```
GarethCabournDavies, line 1859: give that Ive moved the curr_tnum to the parent class, this can now be removed

pycbc/events/stat.py
```diff
@@ -2084,6 +2069,61 @@ def lognoiserate(self, trigs):
...
+        """
+        Inherited, see docstring for ExpFitFgBgKDEStatistic.assign_kdes
+        """
+        ExpFitFgBgKDEStatistic.assign_kdes(self, kname)
+
+    def logsignalrate(self, stats, shift, to_shift):
+        """
+        Inherited, see docstring for ExpFitFgBgKDEStatistic.logsignalrate
+        """
+        # Inherit the function from the KDE statistic, with no changes
```
tdent, line 2114: Don't need this comment line IMO

pycbc/events/stat.py
```diff
@@ -2084,6 +2069,61 @@ def lognoiserate(self, trigs):
...
+            A list containing the filenames of hdf format files used to help
+            construct the coincident statistics. The files must have a 'stat'
+            attribute which is used to associate them with the appropriate
+            statistic class.
+        ifos: list of strs, not used here
+            The list of detector names
+        """
+        DQExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
+                                           ifos=ifos, **kwargs)
+        self.find_kdes()
```
tdent, line 2096: I am not quite seeing why we need to call self.find_kdes here and define the method rather than calling `ExpFitFgBgKDEStatistic.find_kdes` directly

pycbc/events/stat.py
```diff
@@ -2084,6 +2069,61 @@ def lognoiserate(self, trigs):
...
+
+    def find_kdes(self):
+        """
+        Inherited, see docstring for ExpFitFgBgKDEStatistic.find_kdes
+        """
+        ExpFitFgBgKDEStatistic.find_kdes(self)
+
+    def assign_kdes(self, kname):
```
tdent, line 2104: is a redefinition needed here, as ExpFitFgBgKDEStatistic.find_kdes calls the method defined within that class?

GarethCabournDavies: I've checked through testing, and assign_kdes needs explicit inheritance, but find_kdes can be run directly

pycbc/events/stat.py
GarethCabournDavies: You're right - I had thought all definitions were needed, but it is just find_kdes

pycbc/events/stat.py
GarethCabournDavies: (`AttributeError: 'DQExpFitFgBgKDEStatistic' object has no attribute 'assign_kdes'`)

tdent: Ah, as find_kdes calls `self.assign_kdes` and 'self' is still this class instance. 
Maybe your rewrite of the KDE init method was over-complex, it could just do find_kdes and then assign_kdes without having one call the other.

pycbc/events/stat.py
```diff
@@ -1828,6 +1833,12 @@ def __init__(self, sngl_ranking, files=None, ifos=None, **kwargs):
         """
         ExpFitFgBgNormStatistic.__init__(self, sngl_ranking, files=files,
                                          ifos=ifos, **kwargs)
+        self.find_kdes()
+
```
tdent, line 1837: I think the `for kname in self.kde_names:` can be moved here rather than being put inside find_kdes, then when reusing from another class one can have the same loop and `ExpFitFgBgKDEStatistic.assign_kdes(kname)` inside the loop

GarethCabournDavies: That's the exact same change I am just testing!

tdent: Hmm, it seems not ... I said move the loop over self.kde_names _here_, ie into the init method, not roll it into find_kdes.  If we wanted to remove the assign_kdes function entirely I would have requested that.

GarethCabournDavies: Ah okay, I have reverted the last change and added in your suggestion.

It is a little more repetition, but seems to be a bit clearer as to what is happening

tdent (approved):
Now basically happy with the method reuse, which I think is the best we could do short of subclassing the KDE statistic (might work but would probably end up as too complicated)

pycbc/events/stat.py
tdent: Right, I am thinking we want to be as simple as possible if reusing methods from another class, so avoiding nested methods and keeping each one short is preferable.

tdent:
So this looks OK to go with .. if it works with basic testing

GarethCabournDavies:
A couple of minor tests, comparing the IDQ-KQE (IDK) statistic with the O3 statistic, the KDE statistic and the IDQ statistic:

image

image

image

Flipping between the images, it looks like the changes between the IDK stat and the O3 stat are excatly the ones in the KDE and iDQ stat, which shows that this is combining them pretty well

PR #4464, opened 2023-08-23
Bugfix: allow two detectors exactly in offline search

GarethCabournDavies:
The offline search was erroring when it was trying to run exclude_zerolag without any other statmap files.

This adds some logic to allow that search configuration

### DIFF
--- a/bin/workflows/pycbc_make_offline_search_workflow
+++ b/bin/workflows/pycbc_make_offline_search_workflow
@@ -534,6 +534,7 @@ for key in final_bg_files:
     else:
         closed_page = [(snrifar_cb,)]
     if len(ifo_sets) > 1:
+        # If there is only one ifo_set, then this table has already been made
         table = wf.make_foreground_table(workflow, bg_file, hdfbank, open_dir,
                                          singles=insps, extension='.html',
                                          tags=bg_file.tags)
--- a/bin/workflows/pycbc_make_offline_search_workflow
+++ b/bin/workflows/pycbc_make_offline_search_workflow
@@ -278,7 +278,11 @@ for ifocomb in ifo_sets:
     else:
         ordered_ifo_list = ifocomb[0]
         coinctag= '1det'
-    if len(ifo_sets) > 1:
+
+    if len(ifo_sets) == 1:
+        # Just one event type - pass it through
+        final_bg_files[ordered_ifo_list] = no_fg_exc_files[ordered_ifo_list]
+    else:
         # Combining the statmap files, multiple possible event types
         other_ifo_keys = list(no_fg_exc_files.keys())
         other_ifo_keys.remove(ordered_ifo_list)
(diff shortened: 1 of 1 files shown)


bin/workflows/pycbc_make_offline_search_workflow
```diff
@@ -278,20 +278,29 @@ for ifocomb in ifo_sets:
...
-    final_bg_files[ordered_ifo_list] = wf.setup_exclude_zerolag(
-        workflow,
-        no_fg_exc_files[ordered_ifo_list],
-        wf.FileList(other_bg_files.values()),
-        output_dir, ordered_ifo_list,
-        tags=ctagcomb
-    )
 
-combined_bg_file = wf.setup_combine_statmap(
+    if len(ifo_sets) == 1:
```
maxtrevor, line 282: I think this block of code all looks good

bin/workflows/pycbc_make_offline_search_workflow
```diff
@@ -523,9 +533,10 @@ for key in final_bg_files:
         closed_page = [(snrifar_cb, ifar_cb)]
     else:
         closed_page = [(snrifar_cb,)]
-    table = wf.make_foreground_table(workflow, bg_file, hdfbank, open_dir,
-                                     singles=insps, extension='.html',
-                                     tags=bg_file.tags)
+    if len(ifo_sets) > 1:
```
maxtrevor, line 536: what exactly is the foreground table? something in the report? are we sure its something we only want when running with 3 (or more) ifos?

bin/workflows/pycbc_make_offline_search_workflow
```diff
@@ -278,20 +278,29 @@ for ifocomb in ifo_sets:
...
+            no_fg_exc_files[ordered_ifo_list],
+            wf.FileList(other_bg_files.values()),
+            output_dir, ordered_ifo_list,
+            tags=ctagcomb
+        )
+
+if len(ifo_sets) == 1:
+    combined_bg_file = no_fg_exc_files[ordered_ifo_list]
+else:
+    combined_bg_file = wf.setup_combine_statmap(
```
maxtrevor, line 303: I think this block of code makes sense

bin/workflows/pycbc_make_offline_search_workflow
```diff
@@ -523,9 +533,10 @@ for key in final_bg_files:
         closed_page = [(snrifar_cb, ifar_cb)]
     else:
         closed_page = [(snrifar_cb,)]
-    table = wf.make_foreground_table(workflow, bg_file, hdfbank, open_dir,
-                                     singles=insps, extension='.html',
-                                     tags=bg_file.tags)
+    if len(ifo_sets) > 1:
+        table = wf.make_foreground_table(workflow, bg_file, hdfbank, open_dir,
+                                         singles=insps, extension='.html',
+                                         tags=bg_file.tags)
```
maxtrevor, line 539: what exactly is the foreground table? something in the report? are we sure its something we only want if we have 3 or more ifos?

GarethCabournDavies: I've added a comment to explain this - If there is only one ifo_set, then this table has already been made, so the workflow crashes given that two jobs are doing the same thing

maxtrevor: ok that makes sense

maxtrevor (approved):

GarethCabournDavies:
@maxtrevor could you merge once the search test completes? Automerge doesnt take that into accoutn and it will take a while so it will be a bit late for me

maxtrevor:
> @maxtrevor could you merge once the search test completes? Automerge doesnt take that into accoutn and it will take a while so it will be a bit late for me

Sure

PR #4466, opened 2023-08-25
stat.py PEP8 cleanup

tdent:


spxiwh (approved):
Looks good. Do be aware that PEP8 does specify that it should be ignored where it impacts readibility (and highlights maths expressions as one example of that).

So `2**3 / 4` may be clearer than `2 ** 3 / 4` in that it more readily indicates that you weren't trying to do `2**(3/4)` and PEP8 says that "readability always trumps the guidelines". .... I'm not explicitly saying anything should change in this regard, but that I'm not entirely sure that spaces in the `**` operator is necessarily helping here.

tdent:
Sure, that is debatable (& I found a couple more places-with-no-spaces).  My personal take is that being able to spot binary operators as easily as possible (ie don't make people hunt for a * or + in the middle of a chunk of symbols) is step 1 of readability, putting in (extra) brackets would be step 2.

PR #4469, opened 2023-08-30
Stop testing on Python 3.7

titodalcanton:
Python 3.7 reached end of life 2 months ago. This stops testing under it and changes the distribution action to use Python 3.8.


ahnitz (approved):
@titodalcanton I think this is a reasonable time to stop testing 3.7. I won't ask for it here, but we'll want to consider after the next release (I doubt we'll introduce any 3.7 bugs in the interim) if we want to formally declare that the packages do not allow use with 3.7.

PR #4470, opened 2023-08-30
Allow background binning by seobnrv5 duration

maxtrevor:
Does what it says in the title


tdent (approved):

yi-fan-wang:
@maxtrevor  I added in https://github.com/gwastro/pycbc/pull/4451 the time estimation of SEOBNRv5, with the approximant name `SEOBNRv5_ROM`, i.e., with a underscore. While this PR's approximant name doesn't have that underscore. Have you checked this PR works as intended?

maxtrevor:
> @maxtrevor I added in #4451 the time estimation of SEOBNRv5, with the approximant name `SEOBNRv5_ROM`, i.e., with a underscore. While this PR's approximant name doesn't have that underscore. Have you checked this PR works as intended?

You're right. I see that some runs I launched earlier today have failed. Whoops

PR #4472, opened 2023-08-31
Make number of dq bins used configurable

maxtrevor:
This small fix makes the number of dq bins used configurable.
Not having this was not causing any errors, but when using binary dq flags, the bin that flagged data would be assigned to was dependent on the flag's active time. By setting the number of bins to 2 with a config setting, a binary dq flag should always result in flagged times getting assigned to bin 1, while unflagged times are assigned to bin 0.


spxiwh (approved):
Looks good to me. Definitely better to use a CLI option with a default value that corresponds to a "good" value than to have things hardcoded!

PR #4474, opened 2023-09-01
Fixed bug created when making number of dq bins configurable

maxtrevor:
When I got rid of the hardcoded `percent_bin = 0.5 ` I did not remove a later reference to the percent_bins variable, which causes an error.

Previously the code had ` n_bins = 100 / percent_bin ` so I have replaced the remaining the reference to `percent_bin/100` to `1/n_bins`


tdent (approved):
Approve if this is tested to work

PR #4477, opened 2023-09-05
[WIP] Use pregenerated injection frame file to perform injections for O4

bhooshan-gadre:
1. We read injection frames and preprocess injection strain the same as strain data 
2. Then add injection strain to the strain data. 
3. One can do injection frames and injection file together as well.
4. Finally we delete injections. 

I am still testing `pycbc_inspiral` runs for both. When satisfied I will remove WIP.


pycbc/strain/strain.py
```diff
@@ -512,6 +566,33 @@ def insert_strain_option_group(parser, gps_times=True):
                             help="(optional), Only use frame files where the "
                                  "URL matches the regular expression given.")
 
+    ## Read frame containing premade O4 injections
```
GarethCabournDavies, line 569: Make sure you state that these are noiseless injection frames.

GarethCabournDavies (commented):
This looks good to me! - if you could add some images of e.g. qscans from frames made by this, that would be great

The next part of this would be to add the ability to search / add these in the workflow, but that can probably come later

GarethCabournDavies:
> > You probably want something like #4360 to avoid requiring an injection file later
> 
> Isn't it require to run full bank over all the injections? This will significantly increase computational cost, especially if we have injection parameter file available. Still, one cannot use `injection_filter_rejector_match_threshold` with this.

that is actually assuming that we _don't_ have an injection file available. I haven't worked out how to to "injections in frames, but we know when / what they are" yet.

bhooshan-gadre:
This PR is now ready to be merged. I have made code-climate as happy as I can.

pycbc/inject/inject.py
```diff
@@ -531,11 +531,24 @@ def apply(self, strain, detector_name, f_lower=None, distance_scale=1,
         TypeError
             For invalid types of `strain`.
         """
-        if strain.dtype not in (float32, float64):
-            raise TypeError("Strain dtype must be float32 or float64, not " \
-                    + str(strain.dtype))
-
-        lalstrain = strain.lal()
+        # if not generate_injections and inj_filter_rejector:
```
ahnitz, line 534: Debugging??

pycbc/strain/strain.py
```diff
@@ -299,27 +338,57 @@ def from_cli(opt, dyn_range_fac=1, precision='single',
...
-
-    if injector is not None:
+        injection_strain = resample_to_delta_t(injection_strain,
+                                     1. / opt.sample_rate, method='ldas') \
+                                             if injection_strain else None
+
+    # Check if we need to use provided injection file
+    # FIXME: Shoud this be done at the begining even before creating injector
+    # instance using parser options?
+    if (injector and not injection_strain):
```
ahnitz, line 356: Code duplication can be avoided here.

pycbc/strain/strain.py
```diff
@@ -299,27 +338,57 @@ def from_cli(opt, dyn_range_fac=1, precision='single',
...
+        injections = \
+            injector.apply(strain, opt.channel_name.split(':')[0],
+                           distance_scale=opt.injection_scale_factor,
+                           injection_sample_rate=opt.injection_sample_rate,
+                           inj_filter_rejector=inj_filter_rejector,
+                           generate_injections=False)
 
     if opt.sgburst_injection_file:
+        if injection_strain is not None:
+            logging.warn("Burst injection file and frames containing injection"
```
ahnitz, line 378: Why is this a warning?

pycbc/strain/strain.py
```diff
@@ -200,7 +200,42 @@ def from_cli(opt, dyn_range_fac=1, precision='single',
     gating_info = {}
 
     injector = InjectionSet.from_cli(opt)
+    # Read noiseless injections from frame file
+    if opt.injection_frame_cache or opt.injection_frame_files or \
```
ahnitz, line 204: This section is nearly a complete duplication with the only change really being a couple argument names. Could this be combined or modularized?

pycbc/strain/strain.py
```diff
@@ -284,12 +319,16 @@ def from_cli(opt, dyn_range_fac=1, precision='single',
     if opt.zpk_z and opt.zpk_p and opt.zpk_k:
```
ahnitz, line 319: Similar to early sections, the conditioning is largely duplciated. Can this not be turned into a loop / modular function, so the lines aren't duplicated?

pycbc/inject/inject.py
```diff
@@ -936,7 +965,8 @@ def slice_and_taper(inj, ts):
         return ts
 
     def apply(self, strain, detector_name, distance_scale=1,
-              injection_sample_rate=None, inj_filter_rejector=None):
+              injection_sample_rate=None, inj_filter_rejector=None,
+              generate_injections=True):
         if inj_filter_rejector is not None:
             raise NotImplementedError("IncoherentFromFile injections do not "
```
ahnitz, line 971: The 'generate_injections' part seems like a bit of hack. Is there a reason to do this? It seems like the logic is that some setup is needed for the injection filter rejection, but it seems to add a lot of fiddly bits. It seems very unintuitive what is happening when the 'generate_injections' option is set to false. I would prefer this option be removed. 

For injection frames, it would seem you should simply not call this method. And for injection filter rejector you essentially use all the parts of the apply method, so the if statements just complicate what is happening. I'm not sure any computation is really saved since the full injection has to be made in this case. 

Overall, the logic seems like it would be more clear if not hidden in this function, and it would take up less space.

ahnitz (changes requested):
@spxiwh @bhooshan-gadre 

I think I see what this is doing. My request is that one consider cutting down on a lot of the code duplication in general here. Secondly, that one rethink the changes the the 'apply' method a bit as it's somewhat intuitive from an API perspective.

ahnitz:
@bhooshan-gadre FYI, there do appear to be CI failures due to the change in command line options.  Here's an example.

```
  File "/home/runner/work/pycbc/pycbc/.tox/py-docs/lib/python3.8/site-packages/pycbc/strain/strain.py", line 204, in from_cli
    if opt.injection_frame_cache or opt.injection_frame_files or \
AttributeError: 'Namespace' object has no attribute 'injection_frame_cache'
```

GarethCabournDavies:
It would be nice to see the command lines for with pregenerated injections and not, that will help us to understand how this would be included in the workflow (and also so that I can test the changes myself more easily)

spxiwh:
@bhooshan-gadre What's the state of this PR, are you going to be able to address comments/suggestions? The functionality here is worth having, but I think the comments/suggestions on implementation are fair, as this is a [necessarily] invasive change. Would prefer not to see it get lost though!

bhooshan-gadre:
> @bhooshan-gadre What's the state of this PR, are you going to be able to address comments/suggestions? The functionality here is worth having, but I think the comments/suggestions on implementation are fair, as this is a [necessarily] invasive change. Would prefer not to see it get lost though!

@spxiwh I am on it. I tried to incorporate suggestions in a couple of ways. I will update PR in a couple of days.

GarethCabournDavies:
> This is near the bottom of my to-do list, but anyone else wanting to take this forward would be great!

It's at the top of mine, but on the second column 😉 . This would align nicely with other LVK searches, so maybe @rahuldhurkunde ?

PR #4480, opened 2023-09-12
Plot trigger rates during dq flags

maxtrevor:
This commit adds a new executable for plotting the likelihood factor found from computing trigger rates during dq flags, and adds it to the offline workflow. This should be the last code change needed to support the use of dq flags in O4.

### DIFF
--- a/bin/plotting/pycbc_plot_dq_flag_likelihood
+++ b/bin/plotting/pycbc_plot_dq_flag_likelihood
@@ -1,0 +1,64 @@
+#!/usr/bin/env python
+""" Plot the log likelihood percentiles for a DQ bin
+"""
+import sys
+import argparse
+import numpy
+import pycbc
+import h5py
+from matplotlib import use; use('Agg')
+from matplotlib import pyplot
+from pycbc.version import git_verbose_msg as version
+import pycbc.results
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument('--version', action='version', version=version)
+parser.add_argument('--verbose', action="store_true")
+parser.add_argument("--dq-file", required=True)
+parser.add_argument("--ifo", type=str, required=True)
+parser.add_argument("--output-file", required=True)
+args = parser.parse_args()
+
+pycbc.init_logging(args.verbose)
+
+ifo = args.ifo
+
+f = h5py.File(args.dq_file, 'r')
+dq_name = f.attrs['stat'].split('-')[1]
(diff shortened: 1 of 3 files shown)


spxiwh (approved):
Looks good to me, assuming it's tested etc.

bin/plotting/pycbc_plot_dq_flag_likelihood
```diff
@@ -0,0 +1,64 @@
...
+pycbc.init_logging(args.verbose)
+
+ifo = args.ifo
+
+f = h5py.File(args.dq_file, 'r')
+dq_name = f.attrs['stat'].split('-')[1]
+logrates_grp = f[ifo + '/dq_percentiles']
+bin_names = logrates_grp.keys()
+
+# when using a flag, times when the flag was on go to the last dq bin
```
tdent, line 31: I can't parse this comment. You take each bin in the group and take the last element of it, but I can't see what this has to do with using a flag (eg, what happens if not using a flag?)
The result is assigned to 'logrates' but I don't know what physically that should be. Logarithm of something? Rate of what?  Is this the segments when a flag was on or just the number of times (number of time samples?)

tdent:
If you can clarify the code comment that would be great, but I don't consider it crucial

bin/plotting/pycbc_plot_dq_flag_likelihood
maxtrevor: So the bins referred to by bin_names are the template bins. Within each of the template bins, we have some number of dq bins. The file being parsed here contains the natural log of the normalized trigger rates (the trigger rates in each dq bin divided by the mean trigger rate of the whole template bin ) for each dq bin within each template bin, where the dq bins are ordered by increasing dq severity.

Effectively this code plots the log of normalized trigger rate in the last dq bin of each template bin. The key thing here is that when using flags we should only have 2 dq bins, one for when the flag was off and one for when it was on. So getting the trigger rate in the last dq bin means getting the trigger rate during segments when the flag was on.

If this plotting executable ingests a trigger rate file that came from using a timeseries dq product instead of a flag, it will plot only a subset of the normalized trigger rates, coming from the times that got mapped into the bin of worst dq problems (as determined by idq or whatever dq product was ingested). 

The other key thing to note here is that this plotting executable is only used in the workflow if the config file specifies that the dq product is a flag. If the dq product is a timeseries, the workflow will instead use https://github.com/gwastro/pycbc/blob/master/bin/plotting/pycbc_plot_dq_percentiles

PR #4485, opened 2023-09-13
Add the code to generate the sky grid that will be used by pycbc_multi_inspiral

hoangstephanie:
This PR adds the code that generates the sky grid that will be used by `pycbc_multi_inspiral`. 

The sky grid is circular and the points are placed following the method described in Section V of https://arxiv.org/abs/1410.6042.

The grid is first generated centered on the North pole, then rotated to the center of the localization error region. The output of the code is an hdf file with two datasets : `ra` and `dec`, corresponding to the sky points `pycbc_multi_inspiral`will loop over.

### DIFF
--- a/bin/pycbc_make_sky_grid
+++ b/bin/pycbc_make_sky_grid
@@ -1,0 +1,102 @@
+#!/usr/bin/env python
+
+"""For a given GRB, generate a sky grid covering its error box. This sky grid will be used by pycbc_multi_inspiral to find multi-detector gravitational wave
+trigger and calculate the coherent SNRs and related statistics. 
+
+The grid is constructed following the method described in Section V of https://arxiv.org/abs/1410.6042"""
+
+import numpy as np
+from numpy import zeros
+import h5py
+import argparse
+from pycbc.detector import Detector
+import itertools
+import pycbc.coordinates
+from scipy.spatial.transform import Rotation as R
+
+def spher_to_cart(sky_points):
+    """Convert spherical coordinates to cartesian coordinates.
+    """
+    cart = np.zeros((len(sky_points), 3))
+    cart[:,0] = np.cos(sky_points[:,0]) * np.cos(sky_points[:,1])
+    cart[:,1] = np.sin(sky_points[:,0]) * np.cos(sky_points[:,1])
+    cart[:,2] = np.sin(sky_points[:,1])
+    return cart
+
+def cart_to_spher(sky_points):
+    """Convert cartesian coordinates to spherical coordinates.
+    """
+    spher = np.zeros((len(sky_points), 2))
+    spher[:,0] = np.arctan2(sky_points[:,1], sky_points[:,0])
+    spher[:,1] = np.arcsin(sky_points[:,2])
+    return spher
+
+parser = argparse.ArgumentParser()
+parser.add_argument('--ra', type=float, help="Right ascension (in rad) of the center of the GRB error box")
+parser.add_argument('--dec', type=float, help="Declination (in rad) of the center of the GRB error box")
+parser.add_argument('--instruments', nargs="+", type=str, required=True, help="List of instruments to analyze.")
+parser.add_argument('--sky-error', type=float, help="3-sigma confidence radius (in rad) of the GRB error box")
+parser.add_argument('--trigger-time', type=int, help="Time (in s) of the GRB")
+parser.add_argument('--timing-uncertainty', type=float, default=0.0001, help="Timing uncertainty (in s) we are willing to accept")
+parser.add_argument('--output', type=str, help="Name of the sky grid")
+
+args = parser.parse_args()
+
+if len(args.instruments) == 1:
+    parser.error('Can not make a sky grid for only one detector.')
+
+args.instruments.sort() # Put the ifos in alphabetical order
+detectors = args.instruments
+detectors = [Detector(d) for d in detectors]
+detector_pairs = list(itertools.combinations(detectors, 2))
+
+# Select the detector pair for which the GRB target location has smallest relative arrival time difference
+tds = [detector_pairs[i][0].time_delay_from_detector(detector_pairs[i][1], args.ra, args.dec, args.trigger_time) for i in range(len(detector_pairs))]
+smallest_td = np.argmin(tds)
+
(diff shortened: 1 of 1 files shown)


bin/pycbc_make_sky_grid
```diff
@@ -0,0 +1,102 @@
...
+
+def cart_to_spher(sky_points):
+    """Convert cartesian coordinates to spherical coordinates.
+    """
+    spher = np.zeros((len(sky_points), 2))
+    spher[:,0] = np.arctan2(sky_points[:,1], sky_points[:,0])
+    spher[:,1] = np.arcsin(sky_points[:,2])
+    return spher
+
+parser = argparse.ArgumentParser()
```
titodalcanton, line 34:
```suggestion
parser = argparse.ArgumentParser(description=__doc__)
```

bin/pycbc_make_sky_grid
```diff
@@ -0,0 +1,102 @@
...
+parser.add_argument('--ra', type=float, help="Right ascension (in rad) of the center of the GRB error box")
+parser.add_argument('--dec', type=float, help="Declination (in rad) of the center of the GRB error box")
+parser.add_argument('--instruments', nargs="+", type=str, required=True, help="List of instruments to analyze.")
+parser.add_argument('--sky-error', type=float, help="3-sigma confidence radius (in rad) of the GRB error box")
```
titodalcanton, line 38:
```suggestion
parser.add_argument('--sky-error', type=float, required=True, help="3-sigma confidence radius (in rad) of the GRB error box")
```

bin/pycbc_make_sky_grid
```diff
@@ -0,0 +1,102 @@
...
+parser.add_argument('--trigger-time', type=int, help="Time (in s) of the GRB")
```
titodalcanton, line 39:
```suggestion
parser.add_argument('--trigger-time', type=int, required=True, help="Time (in s) of the external trigger.")
```

hoangstephanie:
Here is an example of the sky grid generated by `pycbc_make_sky_grid` for GW170817, using a `--sky-error 0.523599` : 
image

bin/pycbc_make_sky_grid
```diff
@@ -0,0 +1,102 @@
...
+parser.add_argument('--timing-uncertainty', type=float, default=0.0001, help="Timing uncertainty (in s) we are willing to accept")
+parser.add_argument('--output', type=str, help="Name of the sky grid")
```
titodalcanton, line 41:
```suggestion
parser.add_argument('--output', type=str, required=True, help="Name of the sky grid")
```

bin/pycbc_make_sky_grid
```diff
@@ -0,0 +1,102 @@
...
+r = R.from_rotvec(u)
+rota = r.apply(cart)
+
+# Convert cartesian coordinates back to spherical coordinates
+spher = cart_to_spher(rota)
+
+hf = h5py.File(args.output + '.hdf', 'w')
+hf.create_dataset('ra', data=spher[:,0])
+hf.create_dataset('dec', data=spher[:,1])
+hf.close()
```
titodalcanton, line 102:
```suggestion
with h5py.File(args.output, 'w') as hf:
    hf['ra'] = spher[:,0]
    hf['dec'] = spher[:,1]
```

titodalcanton:
Thanks! See some minor comments on formatting and CLI above, but this looks like a good start.

bin/pycbc_make_sky_grid
```diff
@@ -0,0 +1,103 @@
...
+cart = spher_to_cart(sky_points)
+
+grb = np.zeros((1, 2))
+grb[0] = args.ra, args.dec
+grb_cart = spher_to_cart(grb)
+
+north_pole = [0, 0, 1]
+
+ort = np.cross(grb_cart, north_pole)
+norm = np.sqrt(ort[0][0]**2 + ort[0][1]**2 + ort[0][2]**2)
```
titodalcanton, line 89: I suggest `using numpy.linalg.norm()` here instead of calculating the norm explicitly.

bin/pycbc_make_sky_grid
```diff
@@ -0,0 +1,103 @@
...
+u = ort*n
+
+# Rotate the sky grid to the center of the external trigger error box
+r = R.from_rotvec(u)
+rota = r.apply(cart)
+
+# Convert cartesian coordinates back to spherical coordinates
+spher = cart_to_spher(rota)
+
+with h5py.File(args.output + '.hdf', 'w') as hf:
```
titodalcanton, line 101: Let the `.hdf` extension be given from the command line, do not add it explicitly. Sometimes people use `.h5` instead.

bin/pycbc_make_sky_grid
```diff
@@ -0,0 +1,103 @@
...
+
+The grid is constructed following the method described in Section V of https://arxiv.org/abs/1410.6042"""
+
+import numpy as np
+from numpy import zeros
+import h5py
+import argparse
+from pycbc.detector import Detector
+import itertools
+import pycbc.coordinates
```
titodalcanton, line 16: You can remove the `zeros` and `pycbc.coordinates` imports as they are not used.

bin/pycbc_make_sky_grid
```diff
@@ -0,0 +1,112 @@
...
+
+# Calculate the light travel time between the two detectors
+t = detector_pairs[smallest_td][0].light_travel_time_to_detector(detector_pairs[smallest_td][1])
+
+# Calculate the required angular spacing of the sky points
+angular_spacing = (2*args.timing_uncertainty) / np.sqrt(t**2 - tds[smallest_td]**2)
+
+sky_points = np.zeros((1, 2))
+
+number_of_rings = int(args.sky_error / angular_spacing)
```
titodalcanton, line 65: This should use `math.ceil()` instead of `int()`. If the sky error is very very small (e.g. a Swift/BAT GRB with a localization much more precise than what is possible with LVK) then you still want at least one sky point in the grid.

titodalcanton: Ah but you use `number_of_rings + 1` below, so `int` is fine.

pannarale (approved):

PR #4490, opened 2023-09-15
Tweak default significance 'bins'

tdent:
As events may be considered releasable at IFAR ~0.5 days, extend the default sig bins to cover this.

Also tweak the FAP bins a bit


GarethCabournDavies (approved):
Looks sensible to me

PR #4491, opened 2023-09-15
Trigger timeseries ranking

GarethCabournDavies:
Allow all rankings from `pycbc/events/ranking` in `pycbc_plot_trigger_timeseries`

### DIFF
--- a/bin/minifollowups/pycbc_plot_trigger_timeseries
+++ b/bin/minifollowups/pycbc_plot_trigger_timeseries
@@ -72,7 +72,7 @@ for ifo in args.single_trigger_files.keys():
     logging.info("Keeping %d triggers in the window", sum(keep))
 
 
-    data_dict = {k: data[ifo][k][keep] for k in data[ifo].keys()
+    data_dict = {k: data[ifo][k][keep] for k in data[ifo]
                  if isinstance(data[ifo][k], h5py.Dataset)
                  and data[ifo][k].size == keep.size}
 
--- a/bin/minifollowups/pycbc_plot_trigger_timeseries
+++ b/bin/minifollowups/pycbc_plot_trigger_timeseries
@@ -42,7 +42,7 @@ parser.add_argument('--special-trigger-ids', nargs='+', type=int,
     help="The set of special trigger ids to plot a star at")
 parser.add_argument('--plot-type',
     choices=ranking.sngls_ranking_function_dict, default='snr',
-    help="Which plot to make; an 'snr' or a newsnr' plot.")
+    help="Which single-detector ranking statistic to plot.")
 parser.add_argument('--output-file')
 parser.add_argument('--log-y-axis', action='store_true')
 
--- a/bin/minifollowups/pycbc_plot_trigger_timeseries
+++ b/bin/minifollowups/pycbc_plot_trigger_timeseries
@@ -120,9 +120,11 @@ pylab.xlim(xmin=-args.window, xmax=args.window)
 pylab.legend()
 pylab.grid()
 
+logging.info("Saving figure")
 pycbc.results.save_fig_with_metadata(fig, args.output_file,
             cmd = ' '.join(sys.argv),
             title = 'Single Detector Trigger Timeseries (%s)' % args.plot_type,
             caption = 'Time series showing the single detector triggers'
                       ' centered around the time of the trigger of interest.',
          )
+logging.info("Done!")
--- a/bin/minifollowups/pycbc_plot_trigger_timeseries
+++ b/bin/minifollowups/pycbc_plot_trigger_timeseries
@@ -74,7 +74,7 @@ for ifo in args.single_trigger_files.keys():
 
     data_dict = {k: data[ifo][k][keep] for k in data[ifo].keys()
                  if isinstance(data[ifo][k], h5py.Dataset)
-                 and data[ifo][k].size == keep.size} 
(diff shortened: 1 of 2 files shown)


bin/minifollowups/pycbc_plot_trigger_timeseries
```diff
@@ -39,7 +40,8 @@ parser.add_argument('--times', nargs='+', type=float,
 parser.add_argument('--special-trigger-ids', nargs='+', type=int,
     action=MultiDetOptionAction, metavar="IFO:GPS_TIME",
     help="The set of special trigger ids to plot a star at")
-parser.add_argument('--plot-type', choices=['snr', 'newsnr'], default='snr',
+parser.add_argument('--plot-type',
+    choices=ranking.sngls_ranking_function_dict, default='snr',
```
titodalcanton, line 44: I suggest updating the help string as well here, maybe "Which single-detector ranking statistic to plot".

bin/minifollowups/pycbc_plot_trigger_timeseries
```diff
@@ -49,70 +51,70 @@ pycbc.init_logging(args.verbose)
...
+        # No triggers in this window, add to the legend and continue
+        pylab.scatter(0, 0, color=pycbc.results.ifo_color(ifo), marker='x',
+                  label=ifo)
+        continue
+
+    any_data = True
+    logging.info("Keeping %d triggers in the window", sum(keep))
+
+
+    data_dict = {k: data[ifo][k][keep] for k in data[ifo].keys()
```
titodalcanton, line 75:
```suggestion
    data_dict = {k: data[ifo][k][keep] for k in data[ifo]
```

bin/minifollowups/pycbc_plot_trigger_timeseries
```diff
@@ -49,70 +51,70 @@ pycbc.init_logging(args.verbose)
...
+                 and data[ifo][k].size == keep.size}
+
+    logging.info("Getting %s", args.plot_type)
+    rank = ranking.get_sngls_ranking_from_trigs(data_dict, args.plot_type)
+
+    if all(rank == 1):
+        # This is the default value to say "downranked beyond consideration"
+        # so skip these events
+        pylab.scatter(0, 0, color=pycbc.results.ifo_color(ifo), marker='x',
+                  label=ifo)
```
titodalcanton, line 86: Same comment as above.

bin/minifollowups/pycbc_plot_trigger_timeseries
```diff
@@ -49,70 +51,70 @@ pycbc.init_logging(args.verbose)
...
-
-    if type(val) in [numpy.float32, numpy.float64] or len(val) > 0:
-        any_data = True
-
-    pylab.scatter(times, val, color=pycbc.results.ifo_color(ifo), marker='x',
+    keep = abs(data[ifo + '/end_time'][:] - t) < args.window
+    if not any(keep):
+        # No triggers in this window, add to the legend and continue
+        pylab.scatter(0, 0, color=pycbc.results.ifo_color(ifo), marker='x',
+                  label=ifo)
```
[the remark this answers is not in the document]
GarethCabournDavies: the min_rank value _should_ mean that nothing with rank == 1 or lower should appear on the plot, so this is safe, but I will add the x value shift as well

bin/minifollowups/pycbc_plot_trigger_timeseries
```diff
@@ -49,70 +51,70 @@ pycbc.init_logging(args.verbose)
...
-try:
-    if len(val) > 0:
-        pylab.ylim(ymin=val.min())
-except TypeError:
-    pylab.ylim(ymin=val)
-
 pylab.xlim(xmin=-args.window, xmax=args.window)
 pylab.legend()
 pylab.grid()
+
```
titodalcanton, line 117: Can we add a `logging.info("Saving figure")` here and a `logging.info("Done")` at the very end, just to debug cases where things get unexplicably slow?

titodalcanton:
Hmm, as far as I remember, the yellow stars showed the two coincident triggers in the two detectors? How can they be more than 10 ms apart?

GarethCabournDavies:
> Hmm, as far as I remember, the yellow stars showed the two coincident triggers in the two detectors? How can they be more than 10 ms apart?

They are simply some trigger IDs which are selected to highlight in the command line. They were set to zero time for some reason. I chose random ones to test this though

tdent:
Should we add something to the caption to clarify that triggers mapped to the 'default terrible value' 1 are deliberately not plotted? 
(Should we actually plot them, though?)

GarethCabournDavies:
I agree that it may be worthwhile to plot them actually (so that we can see that they exist)

bin/minifollowups/pycbc_plot_trigger_timeseries
```diff
@@ -49,73 +51,80 @@ pycbc.init_logging(args.verbose)
...
+    rank = ranking.get_sngls_ranking_from_trigs(data_dict, args.plot_type)
+
+    if all(rank == 1):
+        # This is the default value to say "downranked beyond consideration"
+        # so skip these events
+        pylab.scatter(-2 * args.window, 0,
+                      color=pycbc.results.ifo_color(ifo),
+                      marker='x',
+                      label=ifo)
+        continue
```
GarethCabournDavies, line 92: I think we can just cut out this block, and lines 98-100 below

PR #4495, opened 2023-09-18
Remove certain events from sngls_statmap exclusive background

GarethCabournDavies:
There were no events being removed from the sngls_statmap exclusive background, unless they were signals in other detector combinations (through exclude_zerolag).

This change means that events are removed which have ifar > the foreground time of the search.

This corresponds to a FAP of `1 - exp(-1) = 0.63`, so may be a little on the conservative side, but will correctly blind us to any loud single-detector signals in the closed-box analysis

There are some additional changes which came from the fact that there are not any events with `timeslide_id != 0`, so having `fore_locs` and `back_locs` was superfluous.

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_statmap
+++ b/bin/all_sky_search/pycbc_coinc_statmap
@@ -48,6 +48,12 @@ parser.add_argument('--cluster-window', type=float, default=10,
 parser.add_argument('--veto-window', type=float, default=.1,
                     help='Time around each zerolag trigger to window out '
                          '[default=.1s]')
+parser.add_argument('--hierarchical-removal-ifar-threshold',
+                    type=float, default=0.5,
+                    help="Threshold to remove foreground triggers with "
+                         "IFAR (years) above this value from the exclusive "
+                         "background, and for hierarchical removal "
+                         "[default=0.5yr]")
 parser.add_argument('--hierarchical-removal-window', type=float, default=1.0,
                     help='Time around each trigger to window out for a very '
                          'louder trigger in the hierarchical removal '
@@ -307,21 +313,21 @@ orig_fore_stat = fore_stat
 if args.max_hierarchical_removal != 0:
     # If user wants to remove against inclusive background.
     if args.hierarchical_removal_against == 'inclusive':
-        ifar_foreground = ifar
+        ifar_foreground_yrs = conv.sec_to_year(ifar)
     # Otherwise user wants to remove against exclusive background
     else :
-        ifar_foreground = ifar_exc
+        ifar_foreground_yrs = conv.sec_to_year(ifar_exc)
 else :
-    # It doesn't matter if you choose ifar_foreground = ifar
+    # It doesn't matter if you choose ifar_foreground_yrs is from ifar
     # or ifar_exc, the while loop below will break
     # straight away. But this avoids a NameError
-    ifar_foreground = ifar
+    ifar_foreground_yrs = ifar
 
 # Step 2 : Loop until we don't have to hierarchically remove anymore. This
-#          will happen when ifar_foreground has no elements greater than
+#          will happen when ifar_foreground_yrs has no elements greater than
 #          the background time.
 
-while numpy.any(ifar_foreground >= background_time):
+while numpy.any(ifar_foreground_yrs >= args.hierarchical_removal_ifar_threshold):
     # If the user wants to stop doing hierarchical removals after a set
     # number of iterations then break when that happens.
     if (h_iterations == args.max_hierarchical_removal):
@@ -421,13 +427,13 @@ while numpy.any(ifar_foreground >= background_time):
         return_counts=True,
         **significance_dict[ifo_combo])
 
-    # Update the ifar_foreground criteria depending on whether foreground
+    # Update the ifar_foreground_yrs criteria depending on whether foreground
     # triggers are being removed via inclusive or exclusive background.   
     if args.hierarchical_removal_against == 'inclusive':
-        ifar_foreground = 1. / fg_far
+        ifar_foreground_yrs = 1. / fg_far
 
     # Exclusive background doesn't change when removing foreground triggers.
-    # So we don't have to take background ifar, just repopulate ifar_foreground
+    # So we don't have to take background ifar, just repopulate ifar_foreground_yrs
     else :
         _, fg_far_exc = significance.get_far(
             exc_zero_trigs.stat,
@@ -435,8 +441,8 @@ while numpy.any(ifar_foreground >= background_time):
             exc_zero_trigs.decimation_factor,
             background_time_exc,
             **significance_dict[ifo_combo])
-        ifar_foreground = 1. / fg_far_exc
-    # ifar_foreground has been updated and the code can continue.
+        ifar_foreground_yrs = 1. / fg_far_exc
+    # ifar_foreground_yrs has been updated and the code can continue.
 
     logging.info("Calculating ifar/fap values")
     f['background_h%s/ifar' % h_iterations] = conv.sec_to_year(1. / bg_far)
--- a/bin/all_sky_search/pycbc_coinc_statmap
+++ b/bin/all_sky_search/pycbc_coinc_statmap
@@ -48,12 +48,6 @@ parser.add_argument('--cluster-window', type=float, default=10,
 parser.add_argument('--veto-window', type=float, default=.1,
                     help='Time around each zerolag trigger to window out '
                          '[default=.1s]')
-parser.add_argument('--hierarchical-removal-ifar-threshold',
-                    type=float, default=0.5,
-                    help="Threshold to remove foreground triggers with "
-                         "IFAR (years) above this value from the exclusive "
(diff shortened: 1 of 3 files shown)


GarethCabournDavies:
Results page: https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/o4/pycbc/dev/o4_may_2023_CENSOR_SINGLES/

bin/all_sky_search/pycbc_sngls_statmap
```diff
@@ -128,19 +122,18 @@ for key in all_trigs.seg.keys():
...
-back_stat_exc = all_trigs.stat[back_locs]
+fore_stat = back_stat = all_trigs.stat
+# Need to make a copy as this will have items removed from it, and dont want
+# to overwrite the original
+back_stat_exc = copy.deepcopy(back_stat)
 
-bkg_dec_facs = all_trigs.decimation_factor[back_locs]
-bkg_exc_dec_facs = all_trigs.decimation_factor[back_locs]
+bkg_dec_facs = all_trigs.decimation_factor
+bkg_exc_dec_facs = all_trigs.decimation_factor
```
tdent, line 136: probably clearer to do this via deepcopy as well - ie if you've used it once for a specific reason which requires a comment to explain, you can use it again

bin/all_sky_search/pycbc_sngls_statmap
```diff
@@ -153,29 +146,66 @@ bg_far, fg_far = significance.get_far(
...
 f['background/ifar'] = conv.sec_to_year(bg_ifar)
-f['background_exc/ifar'] = conv.sec_to_year(bg_ifar_exc)
+
 f.attrs['background_time'] = fg_time
 f.attrs['foreground_time'] = fg_time
+
+# Find foreground triggers with IFAR > 1 per livetime and remove from
+# the exclusive background (not equal, as that could give an infinite
+# loop if using count_n_louder)
+fg_ifar_exc = copy.deepcopy(fg_ifar)
```
tdent, line 163: Why is this copying done here but other copying was done earlier? 
Would be easier to understand the whole thing if all the 'exc' copying is done in one place, IMO.  Also according to the principle of don't create a variable yet if the first place you use it is 40 lines later (back_stat_exc etc).

GarethCabournDavies: Added now

bin/all_sky_search/pycbc_sngls_statmap
GarethCabournDavies: Done

bin/all_sky_search/pycbc_sngls_statmap
```diff
@@ -153,29 +140,71 @@ bg_far, fg_far = significance.get_far(
...
+fg_time_exc = copy.deepcopy(fg_time)
+fg_ifar_exc = copy.deepcopy(fg_ifar)
+bg_ifar_exc = copy.deepcopy(bg_ifar)
+back_stat_exc = copy.deepcopy(back_stat)
+bkg_exc_dec_facs = copy.deepcopy(bkg_dec_facs)
+
+# Record indices into all_trigs for the exclusive background
+back_exc_locs = numpy.arange(len(all_trigs.stat))
+
+while any(bg_ifar_exc > fg_time):
```
tdent, line 169: if this is being done recursively, the probability that we end up removing several random noise triggers is too high
Hence replace by an input option eg 'exclude-ifar-threshold' (in years) with a sensible default value

tdent:
Should the change to minifollowups trigger time series be in this PR?

bin/all_sky_search/pycbc_sngls_statmap
tdent: Or stick with 1/livetime default and don't exclude recursively ..

GarethCabournDavies:
> Should the change to minifollowups trigger time series be in this PR?

No, and weirdly I am a bit confused that it is showing up, as it is on master?

bin/all_sky_search/pycbc_coinc_statmap
```diff
@@ -307,21 +313,21 @@ orig_fore_stat = fore_stat
...
     if args.hierarchical_removal_against == 'inclusive':
-        ifar_foreground = ifar
+        ifar_foreground_yrs = conv.sec_to_year(ifar)
     # Otherwise user wants to remove against exclusive background
     else :
-        ifar_foreground = ifar_exc
+        ifar_foreground_yrs = conv.sec_to_year(ifar_exc)
 else :
-    # It doesn't matter if you choose ifar_foreground = ifar
+    # It doesn't matter if you choose ifar_foreground_yrs is from ifar
```
tdent, line 321: minor typo 'choose (..) is from' ? 
```suggestion
    # It doesn't matter if ifar_foreground_yrs is derived from ifar
```

bin/all_sky_search/pycbc_coinc_statmap
```diff
@@ -307,21 +313,21 @@ orig_fore_stat = fore_stat
 if args.max_hierarchical_removal != 0:
     # If user wants to remove against inclusive background.
     if args.hierarchical_removal_against == 'inclusive':
-        ifar_foreground = ifar
+        ifar_foreground_yrs = conv.sec_to_year(ifar)
```
tdent, line 316: being very picky I prefer 'yr' as unit - once we resolve other stuff I can just do a find-replace, so no action immediately

bin/all_sky_search/pycbc_sngls_statmap
```diff
@@ -153,6 +146,42 @@ bg_far, fg_far = significance.get_far(
...
+
+f.attrs['background_time'] = fg_time
+f.attrs['foreground_time'] = fg_time
+
+# Find foreground triggers with IFAR > the set limit and remove from
+# the exclusive background
+
+# Need to make copies for use as exclusive triggers as these will have
+# items removed from them, and don't want to overwrite the original
+fg_time_exc = copy.deepcopy(fg_time)
```
tdent, line 165: is this actually a float rather than a list/array?

bin/all_sky_search/pycbc_sngls_statmap
```diff
@@ -153,6 +146,42 @@ bg_far, fg_far = significance.get_far(
...
+fg_ifar_exc = copy.deepcopy(fg_ifar)
+bg_ifar_exc = copy.deepcopy(bg_ifar)
+back_stat_exc = copy.deepcopy(back_stat)
+bkg_exc_dec_facs = copy.deepcopy(bkg_dec_facs)
+
+# Record indices into all_trigs for the exclusive background
+back_exc_locs = numpy.arange(len(all_trigs.stat))
+
+# Find which fg events should be removed from exc background
```
tdent, line 174:
```suggestion
# Remove trigs from 'exclusive' background if their IFAR is > livetime
```

bin/all_sky_search/pycbc_sngls_statmap
GarethCabournDavies: yes it is, so doesnt need the copy
```suggestion
fg_time_exc = fg_time
```

bin/all_sky_search/pycbc_sngls_statmap
```diff
@@ -49,6 +49,12 @@ parser.add_argument('--veto-window', type=float, default=.1,
                     help='Time around each zerolag trigger to window out '
                          '[default=.1s]')
 significance.insert_significance_option_group(parser)
+parser.add_argument('--hierarchical-removal-ifar-threshold',
+                    type=float, default=0.5,
+                    help="Threshold to remove foreground triggers with "
+                         "IFAR (years) above this value from the exclusive "
```
tdent, line 55: Presumably this help is to be fixed to just be for hier removal, not exclusive bg

GarethCabournDavies: Good point, missed this

bin/all_sky_search/pycbc_sngls_statmap
```diff
@@ -49,6 +49,12 @@ parser.add_argument('--veto-window', type=float, default=.1,
...
+                         "background, and for hierarchical removal "
+                         "[default=0.5yr]")
```
GarethCabournDavies, line 57:
```suggestion
                    help="Threshold to remove foreground triggers with "
                         "IFAR (years) above this value during "
                         "hierarchical removal [default=0.5yr]")
```

tdent:
TBH it might be preferable to keep everything in this code in seconds/Hz and just convert the threshold option value into seconds immediately (eg `hier_ifar_thresh_s`) when parsing options, at least it might make the code simpler overall

bin/all_sky_search/pycbc_sngls_statmap
tdent: -> 'Threshold to hierarchically remove foreground triggers with IFAR (years) above this value'

tdent (approved):

GarethCabournDavies:
Full run has completed and results page is [here](https://ldas-jobs.ligo.caltech.edu/~ian.harry/o4/pycbc/dev/o4_may_2023_TEST4/)

PR #4499, opened 2023-09-21
rank_stat_single was not using KDE or BBH weighting - bugfix

GarethCabournDavies:
The mass weighting for the BBH statistic, and the KDE weighting in the KDE statistic were not being used in the rank_stat_single.

This is as the logsignalrate function, to which these were added, is appropriate for coinc events only as it is where the PhaseTD stuff is calculated.

The sensitive volume term is directly calculated in the rank_stat_single function for ExpFitFgBgNormStatistic, so that is okay

This does show how making the statistics a bit more modular will definitely help prevent oversights like this, but that is more invasive and a long-term project

### DIFF
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -1733,6 +1733,31 @@ class ExpFitFgBgNormBBHStatistic(ExpFitFgBgNormStatistic):
         logr_s += numpy.log((self.curr_mchirp / 20.) ** (11. / 3.))
         return logr_s
 
+    def rank_stat_single(self, single_info,
+                         **kwargs): # pylint:disable=unused-argument
+        """
+        Calculate the statistic for a single detector candidate
+
+        This calls back to the Parent class and then applies the chirp mass
+        weighting factor.
+
+        Parameters
+        ----------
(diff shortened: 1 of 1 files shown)


pycbc/events/stat.py
```diff
@@ -1847,6 +1873,16 @@ def assign_kdes(self, kname):
         with h5py.File(self.files[kname + '-kde_file'], 'r') as kde_file:
             self.kde_by_tid[kname + '_kdevals'] = kde_file['data_kde'][:]
 
+    def kde_logsignalrate(self):
```
tdent, line 1876: I would rather call this kde_ratio .. it is not a 'signal rate' particularly

One could think of it as a rate of signals-per-template, but basically it's just the ratio of the kdes evaluated at the templates

tdent (changes requested):
Seems OK, the usual clunkiness due to the IDK stat not just inheriting from KDE stat, but fine if it works modulo the small change of method name

tdent (approved):

tdent:
Approved, assuming that the fix does work

PR #4502, opened 2023-09-25
PyGRB: Sky-error plot fix

jakeb245:
This is a separate PR to fix a specific bug when generating plots of injection sky-errors in the new PyGRB workflow. There is some preliminary discussion in #4443. 

This comes down to a tricky naming issue. Injection trigger files now contain the injection information AND the associated found triggers. In the `load_sky_error_data` function, this file is referred to as `input_file_handle`. Separately, the trigger file containing background triggers is used for other calculations. This one is referred to as `trig_data`. Previously the function was trying to compare data from two different types of trigger files (i.e. comparing found injections to background triggers). This went undetected until now because in early tests, one of these arrays ended up being empty. I believe this was because our test configuration had a very limited number of injections, and none of them were being found. If you run examples with more injections that can be found, you end up with two arrays of different size (with neither being empty), and an exception is thrown.

I've put together an example on CIT here: `/home/jacob.buchanan/tests/postproc_tests/skyerror`. Using two environments you can see the code running with the fix, and failing without it.

With fix: `/home/jacob.buchanan/.conda/envs/skyerror_fix`
Without fix: `/home/jacob.buchanan/.conda/envs/skyerror_nofix`

I can delve into this more if requested.

### DIFF
--- a/bin/pygrb/pycbc_pygrb_plot_injs_results
+++ b/bin/pygrb/pycbc_pygrb_plot_injs_results
@@ -100,7 +100,7 @@ def load_mass_data(input_file_handle, key, tag):
     return data
 
 
-def load_sky_error_data(input_file_handle, key, tag, trig_data):
+def load_sky_error_data(input_file_handle, key, tag):
     """Extract data related to sky_error from raw injection and trigger data"""
     if tag == 'missed':
         # Missed injections are assigned null values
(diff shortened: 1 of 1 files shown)


titodalcanton (approved):
I just went through the entire `pycbc_pygrb_plot_injs_results` to try and make sense of this PR. I think the change is justified; the code certainly did not make much sense before. @jakeb245 I have not seen before/after tests, but I trust you did. In any case we are developing this, so I think this change is ready to go.

PR #4505, opened 2023-09-27
add JointPrimaryMarginalizedModel likelihood model

WuShichao:
@ahnitz This PR implements a likelihood model for multiband PE.


pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +602,273 @@ def _loglikelihood(self):
             for stat in self.extra_stats_map[lbl]:
                 setattr(self._current_stats, stat, mstats[stat.subname])
         return logl
+
+
+class MultibandRelativeTimeDom(HierarchicalModel):
```
ahnitz, line 607: It's not really time to the RelativeTimeDom, right? It's just tied to a primary model / secondary model scenario, no?

WuShichao: I named it this way because the primary model uses `RelativeTimeDom`, and also emphasizes the multiband purpose. I know you want a more general likelihood model though.

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +610,301 @@ def _loglikelihood(self):
...
+        hh_others = numpy.zeros(nums)
+
+        for _, other_model in enumerate(self.other_models):
+            current_params_other = other_model.current_params.copy()
+            print("current_params_other: ", current_params_other)
+            # TODO: run this for-loop in parallel
+            for i in range(nums):
+                current_params_other.update(
+                    {key: value[i] for key, value in margin_params.items()})
+                if 'distance' in self.primary_model.static_params:
```
ahnitz, line 710: Don't hardcode specific things related to distance. Handle static parameters generally

ahnitz, line 710: See if you can make this more general

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +610,301 @@ def _loglikelihood(self):
...
+                # doesn't need marginalization
+                for p in wfparam_map[lbl]:
+                    submodel.static_params.pop(p.subname)
+            submodels[lbl] = submodel
+            logging.info("")
+
+        # remove all marginalized parameters from the top-level model's
+        # `variable_params` and `prior` sections
+        primary_model = '%s__model' % kwargs['primary_lbl'][0]
+        marginalized_params = cp.get(primary_model,
```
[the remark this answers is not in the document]
WuShichao: I have made this part more general. Please check.

pycbc/inference/models/hierarchical.py
WuShichao: I have made this part more general. In order not to pass the primary model's `f_lower` to LISA, I need to add it into that `if` check.

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +603,298 @@ def _loglikelihood(self):
...
+        sampattrs = fp.getattrs(group=fp.samples_group)
+        # if a group is specified, prepend the lognl names with it
+        if group is None or group == '/':
+            prefix = ''
+        else:
+            prefix = group.replace('/', '__')
+            if not prefix.endswith('__'):
+                prefix += '__'
+        try:
+            for lbl, model in self.submodels.items():
```
WuShichao, line 656: @ahnitz I also modified this `write_metadata` method, previously I just copied it from that `MultiSignalModel`, but I noticed there is a main difference between my model and it: doesn't like `MultiSignalModel`, my submodels don't share the same detector data, so I have to save all `lognl`. Is this part correct?

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +603,356 @@ def _loglikelihood(self):
...
+                sampattrs['{}lognl'.format(prefix + '%s__' % lbl)
```
ahnitz, line 657: You should be able to avoid duplicating this code.

pycbc/inference/models/hierarchical.py
ahnitz: @WuShichao I don't think this is tied to the relative model, so I think the name still needs to be fixed. (it is tied to models that do marginalization).

WuShichao: @ahnitz Can you suggest a name? This multiband likelihood relies on the primary model's `RelativeTimeDom`, and has specific code to do the marginalization for LISA. So it's not a likelihood model for general usage, right?

ahnitz: It doesn't is just relies on it doing vector maginalization. Nothing explicitly requires that either be a LISA model or even be specific models, just that the primary use vector marginalization at some point. 

It could be used for general useage if for example you want to do joint EM + GW analysis. In that case, sky marginalization would still be useful and one could just use your code with modified configuration files.

WuShichao: @ahnitz OK, then let's name it `JointPrimaryModelMarginalized`, do you think it's better?

pycbc/inference/models/hierarchical.py
```diff
@@ -113,17 +115,18 @@ def __init__(self, variable_params, submodels, **kwargs):
...
-                                 "in a hierarchical analysis, these are "
-                                 "handled by the hiearchical model"
-                                 .format(lbl))
-            if model.waveform_transforms is not None:
-                raise ValueError("Model {} has waveform transforms set; "
-                                 "in a hierarchical analysis, these are "
-                                 "handled by the hiearchical model"
-                                 .format(lbl))
+            # model, except for `joint_primary_marginalized` model
+            if self.name != "joint_primary_marginalized":
```
ahnitz, line 119: Why is this here? Ideally anything different only exists in each subclass. Parent classes shouldn't have know about subclasses really.

pycbc/inference/models/tools.py
```diff
@@ -766,7 +770,12 @@ def get_loglr():
             self.reconstruct_phase = False
 
         rec['loglr'] = loglr[xl]
-        rec['loglikelihood'] = self.lognl + rec['loglr']
+        if set_others_lognl is None:
```
ahnitz, line 773: This could probably be handled in your model since lognl doesn't affect the drawing of anything (it's a constant in your model), no?

ahnitz: That would mean you wouldn not need to pass this in, just update it in your method.

pycbc/inference/models/hierarchical.py
WuShichao: @ahnitz In the standard `hierarchical` model, all the transforms will be handled on the top level, but in our case, remember that we need to allow the transform with the prefix to be handled by submodels. Currently, I don't have a better idea. We can fix this in later patches.

WuShichao: @ahnitz In the standard `hierarchical` model, all the transforms will be handled on the top level, but in our case, remember that we need to allow the transform with the prefix to be handled by submodels. Currently, I don't have a better idea. We can fix this in later patches.

WuShichao: @ahnitz In the standard `hierarchical` model, all the transforms will be handled on the top level, but in our case, remember that we need to allow the transform with the prefix to be handled by submodels. Currently, I don't have a better idea. We can fix this in later patches.

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +604,316 @@ def _loglikelihood(self):
...
+        sh_others = numpy.full(nums, 0 + 0.0j)
+        hh_others = numpy.zeros(nums)
+
+        for _, other_model in enumerate(self.other_models):
+            current_params_other = other_model.current_params.copy()
+            for i in range(nums):
+                current_params_other.update(
+                    {key: value[i] for key, value in margin_params.items()})
+                for p in self.primary_model.static_params.keys():
+                    # don't pass primary model's 'f_lower' to LISA
```
ahnitz, line 707: Why are you copying any static params to LISA? Shouldn't the normal hierarchical handling already handle static params?

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +604,316 @@ def _loglikelihood(self):
...
+        logging.info("Loading joint_primary_marginalized model")
+        return super(HierarchicalModel, cls).from_config(
+                cp, submodels=submodels, **kwargs)
+
+    def reconstruct(self, seed=None):
+        """ Reconstruct marginalized parameters by using the primary
+        model's reconstruct method, total_loglr and others_lognl.
+        """
+        rec = self.primary_model.reconstruct(
+                seed=seed, set_loglr=self.total_loglr())
```
ahnitz, line 916: How is this supposed to work? You are just calling self.total_loglr and then passing in the return? Don't you need to pass in a *function* that can be called in the reconstruct method?

pycbc/inference/models/hierarchical.py
WuShichao: @ahnitz If I remember correctly, I did this because I wanted to avoid some of 3G's static parameters, such `f_lower` to be used by LISA, so just overwrite it here to make sure.

pycbc/inference/models/tools.py
```diff
@@ -728,11 +728,14 @@ def reconstruct(self, rec=None, seed=None):
         if rec is None:
             rec = {}
 
-        def get_loglr():
-            p = self.current_params.copy()
-            p.update(rec)
-            self.update(**p)
-            return self.loglr
+        if set_loglr is None:
```
ahnitz, line 731: @WuShichao The flags used here interact with the output of the loglr function. You should make sure you undersand how that works to determine how the function you pass in should behave. Will it behave correctly in these cases? You know the primary model on its own will, but you seem to pass your models loglr function. Does it respect these flags properly? I'm not sure your solution has through through this yet.

pycbc/inference/models/hierarchical.py
WuShichao: Now I almost exactly follow the structure of the original `recntruct`, define a `get_loglr` function to call my `total_loglr()`, my `self.total_loglr()` corresponding to `self.loglr`.

pycbc/inference/models/tools.py
WuShichao: Now I pass a function to it.

pycbc/inference/models/hierarchical.py
WuShichao: I forgot the difference between `set_loglr=get_loglr` and `set_loglr=get_loglr()`, now I know we need to use  `set_loglr=get_loglr` if we want to pass a _function_ to it.

pycbc/inference/models/hierarchical.py
```diff
@@ -599,3 +604,349 @@ def _loglikelihood(self):
...
+
+        # note that for SOBHB signals, ground-based detectors dominant SNR
+        # and accuracy of (tc, ra, dec)
+        self.primary_model.return_sh_hh = True
+        sh_primary, hh_primary = self.primary_model.loglr
+        self.primary_model.return_sh_hh = False
+
+        margin_names = list(
+            self.primary_model.marginalize_vector_params.keys())
+        margin_names.remove('logw_partial')
```
ahnitz, line 695: Do you need this removal? What happens if you don't do this?

ahnitz (approved):
@WuShichao If you are now confident this works in your tests, you may merge. Please see the few remaining comments to see if they can be easily addressed, otherwise go ahead.

pycbc/inference/models/hierarchical.py
WuShichao: @ahnitz Those are related to updating parameters in the LISA model, so there is no need for `logw_partial`  in the LISA model.

pycbc/inference/models/hierarchical.py
WuShichao: @ahnitz Just checked, the numbers saved in the final posterior file makes sense:
image
image

ahnitz (approved):
@WuShichao Add a comment as we discussed and then merge once the tests pass

PR #4506, opened 2023-09-27
Update pycbc_brute_bank to include option to cut wavelength and save …

kkacanja:
…the lower frequency


bin/bank/pycbc_brute_bank
```diff
@@ -270,16 +271,37 @@ class GenUniformWaveform(object):
...
-            hp = ws[0]
-            hc = ws[1]
-            if 'fratio' in kwds:
-                hp = hc * kwds['fratio'] + hp * (1 - kwds['fratio'])
+                hp = ws[0]
+                hc = ws[1]
+                if 'fratio' in kwds:
+                    hp = hc * kwds['fratio'] + hp * (1 - kwds['fratio'])
         else:
+            logging.info("failed at 2")
```
kkacanja: Remove this line

bin/bank/pycbc_brute_bank
```diff
@@ -270,15 +271,35 @@ class GenUniformWaveform(object):
...
         self.md2 = q._data[0:100]
 
-    def generate(self, **kwds):
+    def generate(self, **kwds):  
         kwds.update(fdict)
-        if kwds['approximant'] in pycbc.waveform.fd_approximants():
-            ws = pycbc.waveform.get_fd_waveform(delta_f=self.delta_f,
+        if kwds['approximant'] in pycbc.waveform.fd_approximants():   
+            if args.cut_wavelength:
+                import numpy
```
ahnitz, line 278: Shouldn't need to reimport numpy

ahnitz: Move the determination of the waveform length and corresponding flow to top of this function so the result can be used by the remainder with less code duplication.

bin/bank/pycbc_brute_bank
```diff
@@ -270,15 +271,35 @@ class GenUniformWaveform(object):
...
+                flow = numpy.arange(self.f_lower, 100, .1)[::-1]
+                maxlen = 512
+                length = spa_length_in_time(mass1=kwds['mass1'], mass2=kwds['mass2'], f_lower=flow, phase_order=-1)
+
+                x = numpy.searchsorted(length, maxlen) - 1
+                l = length[x]
+                f = flow[x]
+                hp, hc = pycbc.waveform.get_fd_waveform(delta_f=self.delta_f, 
+                                                f_lower=f, f_ref=10.0, **kwds)
+                kwds['flow'] = f
```
ahnitz, line 289: Should be named 'f_lower'

ahnitz (changes requested):

bin/bank/pycbc_brute_bank
```diff
@@ -46,7 +46,9 @@ parser.add_argument('--minimal-match', default=0.97, type=float)
 parser.add_argument('--buffer-length', default=2, type=float,
     help='size of waveform buffer in seconds')
 parser.add_argument('--cut-wavelength', type=bool, 
-                    help="Option to cut the wavelength")
+                    help="When enabled, the program will adjust the wavelength to fit the specified length in max-signal-length ")
```
ahnitz, line 49: We could probably combine these two options, so if max-signal-length isn't given, it doesn't impose any cut.

ahnitz:
@kkacanja Let me know when you want me to review this.

bin/bank/pycbc_brute_bank
```diff
@@ -270,15 +272,28 @@ class GenUniformWaveform(object):
...
 
-    def generate(self, **kwds):
+    def generate(self, **kwds):  
         kwds.update(fdict)
-        if kwds['approximant'] in pycbc.waveform.fd_approximants():
-            ws = pycbc.waveform.get_fd_waveform(delta_f=self.delta_f,
-                                                f_lower=self.f_lower, **kwds)
-            hp = ws[0]
-            hc = ws[1]
+        flow = numpy.arange(self.f_lower, 100, .1)[::-1]
```
ahnitz, line 277: This line and the next should probably be within the if statement.

ahnitz (approved):
@kkacanja Besides for the minor comment which should be straightforward to address, I think this is ready to merge.

kkacanja:
> @kkacanja In general, you should use a "rebase" rather than merging master into your code.

Thank you, I am unfamiliar with rebasing so I will do that now.

ahnitz:
@kkacanja Some of the tests are fragile to network outages. So for example, if data can't be retrieved when it tries. You have to take a look at the message to see if it looks like this is the case when you see a failure (or if the error seems related to your PR). If it's the former, there should be a button you can click to restart any failed jobs and have them try again. If not, we can walk through this together tomorrow and make sure you have the right permissions set up.

PR #4507, opened 2023-09-27
Dq plotting

maxtrevor:
Addresses https://github.com/gwastro/pycbc/issues/4484

### DIFF
--- a/bin/plotting/pycbc_plot_dq_flag_likelihood
+++ b/bin/plotting/pycbc_plot_dq_flag_likelihood
@@ -31,7 +31,8 @@ bin_names = logrates_grp.keys()
 # when using a flag, should have exactly 2 dq bins (off and on)
 # times when the flag was on go to the last dq bin
 for b in bin_names:
-    assert len(logrates_grp[b])==2, 'Flag should have exactly 2 dq bins'
+    num_bins = len(logrates_grp[b])
+    assert num_bins == 2, f'Flag should have exactly 2 dq bins, found {num_bins}'
 logrates = [logrates_grp[b][-1] for b in bin_names]
 x = numpy.arange(len(logrates))
 
@@ -42,9 +43,8 @@ fig = pyplot.figure(0)
 ax = fig.add_subplot(111)
 ax2 = ax.twinx()
 
-ax2.bar(x, height=logrates, label=f'{ifo} DQ Flag Trigger Rates',
-       color=color)
-ax2.set_ylabel('Data Quality Log Likelihood')
+ax2.bar(x, height=logrates, label=ifo, color=color)
+ax2.set_ylabel('DQ Log Likelihood')
 ax2.legend(loc='upper left', markerscale=5)
 ax2.set_ylim(0, ymax)
 ax2.grid()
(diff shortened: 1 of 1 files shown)


bin/plotting/pycbc_plot_dq_flag_likelihood
```diff
@@ -28,7 +28,10 @@ dq_name = f.attrs['stat'].split('-')[1]
 logrates_grp = f[ifo + '/dq_percentiles']
 bin_names = logrates_grp.keys()
 
-# when using a flag, times when the flag was on go to the last dq bin
+# when using a flag, should have exactly 2 dq bins (off and on)
+# times when the flag was on go to the last dq bin
+for b in bin_names:
+    assert len(logrates_grp[b])==2, 'Flag should have exactly 2 dq bins'
```
tdent, line 34: More informative to tell the user how many bins were in the file, ie 'should have exactly 2 dq bins, input file has {n}'

bin/plotting/pycbc_plot_dq_flag_likelihood
```diff
@@ -37,24 +40,25 @@ color = pycbc.results.ifo_color(ifo)
 
 fig = pyplot.figure(0)
 ax = fig.add_subplot(111)
-ax.bar(x, height=logrates, label=f'{ifo} DQ Flag Trigger Rates',
+ax2 = ax.twinx()
+
+ax2.bar(x, height=logrates, label=f'{ifo} DQ Flag Trigger Rates',
```
tdent, line 45: I think having just the ifo name as label would be fine, the y-axes already explain what the values mean, which in fact is two different things.  (Actually, how about allowing the plot to show multiple ifos at once?  Just an idea ..)

bin/plotting/pycbc_plot_dq_flag_likelihood
```diff
@@ -37,24 +40,25 @@ color = pycbc.results.ifo_color(ifo)
...
-ax2.set_ylim(1, ax2_ymax)
-new_ticks = range(0, int(numpy.ceil(numpy.log10(ax2_ymax))))
-ax2.set_yticks([10**t for t in new_ticks])
-ax2.set_ylabel('Relative Trigger Rate')
-ax2.set_yscale('log')
+ax_ymax = numpy.exp(ymax)
+ax.set_ylim(1, ax_ymax)
+new_ticks = range(0, int(numpy.ceil(numpy.log10(ax_ymax))))
+ax.set_yticks([10**t for t in new_ticks])
+ax.set_ylabel('(Trigger rate during DQ Flag)/(Mean trigger rate)')
```
tdent, line 58: A bit verbose, & would like to incorporate the fact this is truncated to be >=1 somehow.  
```suggestion
ax.set_ylabel('(Rate during DQ Flag)/(Mean Rate) [min 1]')
```

bin/plotting/pycbc_plot_dq_flag_likelihood
tdent:
```suggestion
ax2.bar(x, height=logrates, label=f'{ifo} Triggers',
```

tdent (changes requested):
Minor changes as per comments

bin/plotting/pycbc_plot_dq_flag_likelihood
```diff
@@ -37,24 +41,24 @@ color = pycbc.results.ifo_color(ifo)
...
-ax.set_xlabel('Template Bin Number')
-ax.set_ylabel('Data Quality Log Likelihood')
-ax.set_xticks(x)
-ax.legend(loc='upper left', markerscale=5)
-ax.set_ylim(0, ymax)
-ax.grid()
+ax2 = ax.twinx()
+
+ax2.bar(x, height=logrates, label=ifo, color=color)
+ax2.set_ylabel('DQ Log Likelihood')
```
tdent, line 47: Actually, this is a log likelihood _penalty_, right? I.e. it is a number that gets subtracted from the ranking stat? 
If so, I think it would be preferable to put that in the caption

maxtrevor: Ok I have added the word 'Penalty' to the end of this label

tdent (approved):

PR #4512, opened 2023-10-02
Close the file after writing events.

bhooshan-gadre:
I have not bothered to add a context manager as the class is used only here.


spxiwh (approved):
This seems a logical change to me (There's definitely some weirdness in needing to do `f.f.close()` and maybe we more generally want `HFile['variable'] = <array>` to do the things done here, but considering a PyCBCHFile object is much broader than this patch).

PR #4437, opened 2023-07-14
comments on the multi_inspiral executable

sebastiangomezlopez:
Hi, 

As discussed with @pannarale and @titodalcanton. I worked on commenting the pycbc_multi_inspiral executable in order to make clearer what the executable does, highlight where certain computations are happening, and mention a few possible fixes/optimizations that current developers could work on in the future.

PR #4513, opened 2023-10-02
Dev comments for the pycbc_multi_inspiral executable

sebastiangomezlopez:


pannarale:
Don't forget @titodalcanton's comments in the old version of this PR https://github.com/gwastro/pycbc/pull/4437

pannarale (changes requested):

sebastiangomezlopez:
> Don't forget @titodalcanton's comments in the old version of this PR #4437

Done, thanks for the comments @titodalcanton.

pannarale (approved):

bin/pycbc_multi_inspiral
```diff
@@ -136,20 +144,79 @@ opt.insert_optimization_option_group(parser)
 inject.insert_injfilterrejector_option_group_multi_ifo(parser)
 sgchisq.SingleDetSGChisq.insert_option_group(parser)
 args = parser.parse_args()
-init_logging(args.verbose)
+init_logging(args.verbose) # takes the --verbose option -> (bool)
 # Set GRB time variable for convenience
-t_gps = args.trigger_time
+t_gps = args.trigger_time # takes the --trigger-time option -> (float)
+# Put the ifos in alphabetical order so they are always called in
+# the same order. Takes the --instruments option -> (list of str)
```
titodalcanton, line 151: Something went wrong here, the comment is duplicated?

bin/pycbc_multi_inspiral
```diff
@@ -256,10 +338,13 @@ with ctx:
     autochisq = vetoes.SingleDetAutoChisq(
         args.autochi_stride, args.autochi_number_points,
         onesided=args.autochi_onesided)
+    # Here we take all frequency-domain segments and whiten them
+    # dividing by the psd estimate
```
titodalcanton, line 342:
```suggestion
    # Here we take all frequency-domain segments and over-whiten them
    # dividing by the PSD estimate
```

bin/pycbc_multi_inspiral
```diff
@@ -310,13 +395,16 @@ with ctx:
...
+
+    # Template bank: filtering and thinning
     logging.info("Read in template bank")
     bank = waveform.FilterBank(
         args.bank_file, flen, delta_f, complex64, low_frequency_cutoff=flow,
         phase_order=args.order, taper=args.taper_template,
         approximant=args.approximant, out=template_mem)
     # Use injfilterrejector to reduce the bank to only those templates that
     # might actually find something
+    # if inj_filter_rejector[args.instruments[0]].enabled ****
```
titodalcanton, line 407: Is this comment intended? Remove?

bin/pycbc_multi_inspiral
```diff
@@ -359,45 +447,52 @@ with ctx:
...
+		# Normalized trigger snr list for each ifo
                 snr[ifo] = snrv * norm
-            
-            # FIXME: wrong comment?
+
             # Move onto next segment if there are no triggers.
-            if len(ifo_list)==0: continue
+	    n_trigs = [len(snrv_dict[ifo]) for ifo in args.instruments]
+	    if np.all(n_trigs==np.zeros_like(args.instruments[:]), axis=0):
+		continue
```
titodalcanton, line 494: This is too complicated, how about this:
```suggestion
            n_trigs = [len(snrv_dict[ifo]) for ifo in args.instruments]
            if not any(n_trigs):
                continue
```
Also, be careful with tabs vs spaces. Generally we do not want tabs in Python code as they can have confusing side effects and misalign the indentation. You can configure your text editor to always use (four) spaces instead of a tab.

bin/pycbc_multi_inspiral
```diff
@@ -136,20 +144,79 @@ opt.insert_optimization_option_group(parser)
...
+                                args, parser, args.instruments)
+# The pycbc.scheme CLI verifier:
+#       these options are used below by the context manager "ctx"
+#       to properly manage resources without memory leaks
+# This method takes as input the option
+# --processing-scheme as name:N -> (str:int)
+# where name can be either mkl, cpu.
+# N is the number of threads(mkl) or CPUs(cpu)
+# NOTE:--processing-scheme can also be set as cuda for GPU usage.
+# However, the cuda scheme breaks the interaction with lalframe.
```
titodalcanton, line 185: I am not sure how useful it is to spell out the various CLI options as comments here (and in the similar comments just below). What if someone changes one option, will they remember to come here and update the comment as well? Also, the argparse code should be fairly self-explanatory to anyone who has a basic understanding of Python and is willing to dig a bit into the PyCBC code base to see what the various `verify_*options()` functions do. If I wanted to comment this part of the code, I would probably write something like "check the CLI options pertaining to computing optimizations", "check the CLI options controlling how FFTs are executed", etc. But again, this is not really where comments are important.

bin/pycbc_multi_inspiral
```diff
@@ -194,28 +269,32 @@ with ctx:
...
-    # Calculate time delays to each detector for each sky position and apply time slide shifts
+    # Calculate time delays to each detector and store time slide shifts
+    # in a dedicated dictionary
     slide_ids = np.arange(1 + args.num_slides)
     time_slides = {
         ifo: args.slide_shift * slide_ids * n_ifo
         for n_ifo, ifo in enumerate(args.instruments)}
+    # Given the time delays in time_slides, the following dictionary returns the
+    # time delay indices evaluated wrt the geocenter:
+    # (time delay from geocenter + timeslide)*sampling_rate
```
titodalcanton, line 297:
```suggestion
    # Given the time delays in time_slides, the following dictionary keeps the
    # time delays evaluated wrt the geocenter, in units of samples, i.e.
    # (time delay from geocenter + timeslide)*sampling_rate
```

bin/pycbc_multi_inspiral
titodalcanton: Ah, in fact this seems to be causing the tests to fail:
```
[…]
TabError: inconsistent use of tabs and spaces in indentation
```

bin/pycbc_multi_inspiral
sebastiangomezlopez: I agree. Following that line of thoughts, I think two things are yet to be done:

sebastiangomezlopez: I also noticed that the current version does not have parser.add_argument("--processing-scheme"), Then I wonder how is this being parsed... 

Such parser argument should have the following in the help kwarg:

help = " takes as input the processing method "
           " as name:N.  Where name can be either"
           " mkl or cpu, and N is the number of threads"
           " It can be set to cuda for GPU processing"

sebastiangomezlopez: Moreover, I noticed a few weeks ago that certain combinations of  --trig-end-time,  --trig-start-time, --segment-length,  --segment-start-pad, --segment-end-pad, may make the code complain at the pycbc.strain level. I will open an Issue soon, because we may need to avoid this with assert statements or something similar. 

To be more explicit, with the current version of the multi_inspiral example in https://github.com/gwastro/pycbc/blob/master/examples/multi_inspiral/run.sh , setting the option --segment-length 128 breaks the example. 

I've checked why this is happens and turns out that certain combinations of --segment-start-pad, --segment-end-pad & --segment-length cause the segment_width computed in:
master/pycbc/strain/strain.py to have values <=0

sebastiangomezlopez:
Hi @pannarale  and @titodalcanton, this PR is ready, could you have a final look at it again? Thanks in advance.

bin/pycbc_multi_inspiral
titodalcanton: @sebastiangomezlopez the tab/space issue seems to have been addressed, but my main point here still holds: the expression `np.all(n_trigs==np.zeros_like(args.instruments[:]), axis=0)` seems unnecessarily complicated. It seems that it is equivalent to `not any(n_trigs)`, right? If so, please use the simpler way.

bin/pycbc_multi_inspiral
titodalcanton: Part of my suggestion has been addressed, but please do include the rest as well - I think the comment becomes much clearer.

bin/pycbc_multi_inspiral
titodalcanton: @sebastiangomezlopez it seems you have addressed my comment here, and you opened separated issues for your other comments (https://github.com/gwastro/pycbc/issues/4553 and https://github.com/gwastro/pycbc/issues/4554), so I am marking this as resolved.

titodalcanton:
@sebastiangomezlopez see above, a couple of my comments are yet to be fully addressed. There is also a conflict now, which will have to be resolved before merging. Do not hesitate to ask if help is needed.

titodalcanton (changes requested):
@sebastiangomezlopez see above, a couple of my comments are yet to be fully addressed. There is also a conflict now, which will have to be resolved before merging. Do not hesitate to ask if help is needed.

sebastiangomezlopez:
@titodalcanton, since I don't have the right to resolve the conflicts. Could you ? 
It is complaining about the comment in [line 360](https://github.com/sebastiangomezlopez/pycbc/blob/a08f28dc495f6d2ed91e17ef1b996268200ebbe2/bin/pycbc_multi_inspiral#L360)

titodalcanton (approved):
Conflict resolved and all looks good now.

PR #4516, opened 2023-10-02
added support for mode by mode decomposition of NRHybSur3dq8 waveform model

lpathak97:


pycbc/waveform/waveform_modes.py
```diff
@@ -96,7 +96,11 @@ def default_modes(approximant):
         ma += [(l, -m) for l, m in ma]
     elif approximant.startswith('NRSur7dq4'):
         # according to arXiv:1905.09300
-        ma = [(l, m) for l in [2, 3, 4] for m in range(-l, l+1)]
+        ma = [(l, m) for l in [2, 3, 4] for m in range(-l, l+1)]        
+    elif approximant.startswith('NRHybSur3dq8'):
+        # according to arXiv:1812.07865
+        ma = [(2, 0), (2, 1), (2, 2), (3, 0), (3, 1), (3, 2),
```
ahnitz, line 102: What about negative m modes? Should this be similar to above but extending up above but adding in a lone 5, 5 mode?

lpathak97: There are no negative m supported for this waveform model.

lpathak97: Here is the link: [https://lscsoft.docs.ligo.org/lalsuite/lalsimulation/_l_a_l_sim_i_m_r_n_r_hyb_sur3dq8_8c.html](url)

ahnitz:
@lpathak97 I just triggered the unittests to run. Can you fix the remaining CC issues though? You can ignore the code duplication one for now, but the others can be fixed.

lpathak97:
I have fixed all the issues except duplication.

ahnitz:
@lpathak97 You have committed several files that shouldn't be in a repo again. Can you please remove these?

PR #4518, opened 2023-10-05
[pycbc live] Don't add snr options to command if they don't exist

ArthurTolley:
By chance I discovered this bug when a candidate event came into the MDC while I was running things.

Bug:
`setup_optimize_snr` adds the command line commands used to run the snr optimization to an object so the snr optimization can be re-run at a later time (I think). If `run_snr_optimization` is not given to `pycbc_live` (you don't want snr optimization) then presumably you won't have set snr optimization options and therefore it can't add `snr_opt_options` to the command and will crash.

Fix:
- Set the `snr_opt_options` to None if `run_snr_optimization` isn't set
- Check whether the `snr_opt_options` is not None before trying to add it to command.

Just need to do a quick test to see if this actually fixes the issue.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -134,6 +134,8 @@ class LiveEventManager(object):
                 self.fu_cores = 1
             # Convert SNR optimizer options into a string
             self.snr_opt_options = snr_optimizer.args_to_string(args)
+        else:
+            self.snr_opt_options = None
 
         if args.enable_embright_has_massgap:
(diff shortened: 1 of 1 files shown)


ArthurTolley:
Testing done, the fix works and the Live example in `examples/live` now works when all the snr related options are removed.

GarethCabournDavies:
You'll need to rebase to master to get the mac checks fix in

GarethCabournDavies (approved):

PR #4519, opened 2023-10-05
Apply some needed efficiency savings to pycbc_bin_trigger_rates_dq

GarethCabournDavies:
Using this meant that I was able to actually complete a run of this code using realistic analysis files!

Basically use the mechanism described in https://github.com/gwastro/pycbc/pull/4510 to pre-mask the triggers before loading, also move some repetition of the same calculation out of the loops it was repeated in.

Some increased logging as well

### DIFF
--- a/bin/all_sky_search/pycbc_bin_trigger_rates_dq
+++ b/bin/all_sky_search/pycbc_bin_trigger_rates_dq
@@ -54,22 +54,13 @@ ifo = args.ifo
 # This works as a pre-filter as SNR is always greater than or equal
 # to sngl_ranking, except in the psdvar case, where it could increase.
 
-def check_snr_psdvar(snr, psd_var_val):
-    """
-    Convenience function to calculate the maximum possible ranking statistic
-    value given SNR and psd variation statistic
-    """
-    rw_snr = snr / psd_var_val ** 0.5
-    rw_snr[psd_var_val == 0] = 0
-    return rw_snr >= args.stat_threshold
-
 with HFile(args.trig_file, 'r') as trig_file:
     n_triggers_orig = trig_file[f'{ifo}/snr'].size
     logging.info("Trigger file has %d triggers", n_triggers_orig)
     logging.info('Generating trigger mask')
     if f'{ifo}/psd_var_val' in trig_file:
         idx, _, _ = trig_file.select(
-            check_snr_psdvar,
+            lambda snr, psdvar: snr / psdvar ** 0.5 >= args.stat_threshold,
             f'{ifo}/snr',
             f'{ifo}/psd_var_val',
             return_indices=True
--- a/bin/all_sky_search/pycbc_bin_trigger_rates_dq
+++ b/bin/all_sky_search/pycbc_bin_trigger_rates_dq
@@ -48,25 +48,6 @@ pycbc.init_logging(args.verbose)
 
 logging.info('Start')
 
-with h5.File(args.bank_file, 'r') as bank:
-    if args.background_bins:
-        logging.info('Sorting bank into bins')
-        data = {
-            'mass1': bank['mass1'][:],
-            'mass2': bank['mass2'][:],
-            'spin1z': bank['spin1z'][:],
-            'spin2z': bank['spin2z'][:],
-            'f_lower': np.ones_like(bank['mass1'][:]) * args.f_lower
-            }
-        locs_dict = pycbc.events.background_bin_from_string(
-                                                   args.background_bins, data)
-        del data
-        locs_names = [b.split(':')[0] for b in args.background_bins]
-    else:
-        locs_dict = {'all_bin': np.arange(0, len(bank['mass1'][:]), 1)}
-        locs_names = ['all_bin']
-
-
 ifo = args.ifo
 # Setup a data mask based on SNR - the SNR will always be greater than
 # or equal to sngl-ranking by definition
@@ -74,10 +55,10 @@ with HFile(args.trig_file, 'r') as trig_file:
     n_triggers_orig = trig_file[f'{ifo}/snr'].size
     logging.info("Trigger file has %d triggers", n_triggers_orig)
     logging.info('Generating trigger mask')
-    idx, _ = trig_file.select(
+    idx = trig_file.select(
         lambda snr: snr > args.stat_threshold,
         f'{ifo}/snr',
-        return_indices=True
+        indices_only=True
     )
     data_mask = np.zeros(n_triggers_orig, dtype=bool)
     data_mask[idx] = True
@@ -100,7 +81,7 @@ trig_times = trigs.end_time
 trig_times_int = trig_times.astype('int')
 
 n_triggers = tmplt_ids.size
-logging.info("%d triggers are have been loaded after pre-cuts", n_triggers)
+logging.info("%d triggers have been loaded after pre-cuts", n_triggers)
 
 logging.info("Applying %s > %.3f cut", args.sngl_ranking,
              args.stat_threshold)
@@ -143,6 +124,24 @@ del dq_logl
 # create a dict to look up dq percentile at any time
 dq_percentiles_time = dict(zip(dq_times, seconds_bin/n_bins))
 del dq_times
+
+with h5.File(args.bank_file, 'r') as bank:
+    if args.background_bins:
+        logging.info('Sorting bank into bins')
+        data = {
+            'mass1': bank['mass1'][:],
+            'mass2': bank['mass2'][:],
+            'spin1z': bank['spin1z'][:],
+            'spin2z': bank['spin2z'][:],
+            'f_lower': np.ones_like(bank['mass1'][:]) * args.f_lower
+            }
+        locs_dict = pycbc.events.background_bin_from_string(
+                                                   args.background_bins, data)
+        del data
+        locs_names = [b.split(':')[0] for b in args.background_bins]
+    else:
+        locs_dict = {'all_bin': np.arange(0, len(bank['mass1'][:]), 1)}
+        locs_names = ['all_bin']
 
 logging.info("Placing triggers into bins")
 bin_triggers = {}
--- a/bin/all_sky_search/pycbc_bin_trigger_rates_dq
+++ b/bin/all_sky_search/pycbc_bin_trigger_rates_dq
@@ -52,15 +52,16 @@ ifo = args.ifo
 
 # Setup a data mask to remove any triggers with SNR below threshold
 # This works as a pre-filter as SNR is always greater than or equal
-# to sngl_ranking
+# to sngl_ranking, except in the psdvar case, where it could increase.
 
 with HFile(args.trig_file, 'r') as trig_file:
     n_triggers_orig = trig_file[f'{ifo}/snr'].size
     logging.info("Trigger file has %d triggers", n_triggers_orig)
     logging.info('Generating trigger mask')
-    idx, _ = trig_file.select(
-        lambda snr: snr >= args.stat_threshold,
+    idx, _, _ = trig_file.select(
+        lambda snr, psd_var_val: (snr / psd_var_val ** 0.5) >= args.stat_threshold,
         f'{ifo}/snr',
+        f'{ifo}psd_var_val',
         return_indices=True
     )
     data_mask = np.zeros(n_triggers_orig, dtype=bool)
--- a/bin/all_sky_search/pycbc_bin_trigger_rates_dq
+++ b/bin/all_sky_search/pycbc_bin_trigger_rates_dq
@@ -59,7 +59,7 @@ with HFile(args.trig_file, 'r') as trig_file:
     logging.info("Trigger file has %d triggers", n_triggers_orig)
     logging.info('Generating trigger mask')
     idx, _ = trig_file.select(
-        lambda snr: snr > args.stat_threshold,
+        lambda snr: snr >= args.stat_threshold,
         f'{ifo}/snr',
         return_indices=True
     )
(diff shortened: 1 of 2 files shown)


spxiwh:
Can you also move the trigger file parsing *above* the background bin computation. Otherwise it takes an hour to compute tons of SEOBNRv5 durations, and then the job gets evicted when the memory usage spikes. Do the high-memory-using-thing first.

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -63,38 +66,63 @@ with h5.File(args.bank_file, 'r') as bank:
...
-            del stat
 
+ifo = args.ifo
+# Setup a data mask based on SNR - the SNR will always be greater than
+# or equal to sngl-ranking by definition
+with HFile(args.trig_file, 'r') as trig_file:
+    n_triggers_orig = trig_file[f'{ifo}/snr'].size
+    logging.info("Trigger file has %d triggers", n_triggers_orig)
+    logging.info('Generating trigger mask')
+    idx, _ = trig_file.select(
```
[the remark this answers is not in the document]
GarethCabournDavies: I have changed to use ranking straight away - there's a bit of jiggery-pokery to get the trigs into the right format, but seems okay at the moment.

There is a bit of inefficiency with unused datasets if they arent a part of the ranking, but this is done in a couple of minutes anyway, and is nothing compared to the background_bins.

I've also updated the select() function to allow an indices_only return as suggested

spxiwh:
Thanks for taking the next steps with this, but for this PR, could we bring this back to just doing this by SNR (and moving the trigger file reading up). That change has been tested and should be good to go.

The other parts to this: having a `select` method that only returns indices, and being able to run `select` with the ranking only, affects more codes than just this one and so I'd prefer to pull that off to a separate (and broader) PR (with less urgency). I'd also like to consider this a bit further, as if we have ranking = snr (for instance) this code would be 4x slower than it needs to me while it reads a bunch of arrays that it won't use. I'd also like the "mask_to_loudest_n_..." method to use this functionality.

GarethCabournDavies:
I have de-scoped and added notes in https://github.com/gwastro/pycbc/issues/4524

GarethCabournDavies:
I ran a comparison of the output file to the one from the workflow. (Every attribute, every dataset; values and dtypes)

The files do differ slightly after this change, as the 'binX/rates' datasets are float64 in the old run, and float32 in the new test.

Which I _think_ is due to [this line](https://github.com/gwastro/pycbc/blob/master/bin/all_sky_search/pycbc_bin_trigger_rates_dq#L109), not specifying the exact float dtype, and they were run on different hardware

The datasets are numerically equal to one another though

bin/all_sky_search/pycbc_bin_trigger_rates_dq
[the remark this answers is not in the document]
tdent: @spxiwh seems like you want an issue for future development opened?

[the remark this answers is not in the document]
tdent: Oh I see, Gareth already set something up.

spxiwh:
`np.array([1,2,3,4], dtype=np.float32).astype('float').dtype` returns `float32`

GarethCabournDavies:
> `np.array([1,2,3,4], dtype=np.float32).astype('float').dtype` returns `float64` _not_ `float32`

Will that _always_ be the case though? I can't find source code to check that it will be the same whatever hardware is used.

I think changing the code here to use `.astype(np.float64)` is safest for this specific use-case (and also the other `astype`s in the executable)

spxiwh:
I tried print statements in the code in question, and indeed this ends up as a `float32` but I've no idea why. This shouldn't be hardware dependent etc. etc. as `float` means `float64` in python. (I know there's some weirdness around things like longdouble, but the built-in python float should be clear).

Anyway, changing `astype('float')` to `astype(np.float64)` at least solves the discrepancy, and should let us proceed if we've verified that files are identical. (This seems to store `int`s anyway, so not really sure why it needs to be float64 ... but let's stick with "doesn't change output files")

GarethCabournDavies:
> Anyway, changing `astype('float')` to `astype(np.float64)` at least solves the discrepancy, and should let us proceed if we've verified that files are identical.

Weirdly, this seems to make some of the datasets numerically different now!

spxiwh:
Interesting observation:
```
import numpy as np
print(np.array([1,2,3,4], dtype=np.float32).astype('float').dtype)
import logging
import argparse
import pycbc
import pycbc.events
from pycbc.events import stat as pystat
from pycbc.types.timeseries import load_timeseries
import numpy as np
import h5py as h5
from pycbc.io.hdf import HFile, SingleDetTriggers
from pycbc.version import git_verbose_msg as version
print(np.array([1,2,3,4], dtype=np.float32).astype('float').dtype)
```
returns different dtypes for the two prints! ... So something in `io.hdf` (which seems to trigger this) seems to be messing with how this works.

spxiwh:
Here's a small example that demonstrates why this is happening:
```
import numpy as np
print(np.array([1,2,3,4], dtype=np.float32).astype('float').dtype)
from ligo.lw import types as ligolw_types
print([(k,np.sctypeDict[k],ligolw_types.ToNumPyType[k]) for k in ligolw_types.ToNumPyType if k in np.sctypeDict])
# This line is the one that causes the change! It's in io/record.py
np.sctypeDict.update(ligolw_types.ToNumPyType)

print(np.array([1,2,3,4], dtype=np.float32).astype('float').dtype)
```

It also shows what will change. It does mean that the behaviour of `np.astype('int')` or `np.astype('float`)` (and perhaps `np.astype('double')` will change.

However, if we replace all np.astype('int') with `np.astype(np.int64)` and `np.astype('float')` with `np.astype('np.float64')` previous behaviour would be restored. I'll check with @ahnitz  in pycbc-code if this behaviour was ever intended (I assume we just wanted the weird things like 'int_2s' added, not the existing things overwritten).

spxiwh:
@GarethCabournDavies Can you see if we get identical output if we change *all* the `astype` commands in this cpde to match previous behaviour?? That *should* work, and then we can get this merged. The bigger question of "is this intended" can be considered outside of this PR.

GarethCabournDavies:
The output is _not_ identical after changing the `astype('float')` to `astype(np.float64)` - I think this is a case of later calculations being messed up, as the direct comparison after the change does not fail.

I don't know if we need to go through this code line-by-line and check that types are appropriate, but that is not part of this PR

If we leave things as they are (using the string types), then output is identical apart from dtype in the datasets

maxtrevor:
My only remaining concern is the numerical changes resulting from changing dtypes. Once that is udnerstood / resolved this should be good to go.

spxiwh:
So to avoid confusion on this point: This patch *should not* change any of the dtypes. Because of the issue being resolved in https://github.com/gwastro/pycbc/pull/4525 some arrays (with this patch as is) would use reduced precision. Once https://github.com/gwastro/pycbc/pull/4525 is merged this will no longer be the case. I would also recommend explicit type casting, rather than using `astype('int')` in any case, so I propose adding this to the patch:

```
diff --git a/bin/all_sky_search/pycbc_bin_trigger_rates_dq b/bin/all_sky_search/pycbc_bin_trigger_rates_dq
index 9457b61971..eb2bed4c14 100644
--- a/bin/all_sky_search/pycbc_bin_trigger_rates_dq
+++ b/bin/all_sky_search/pycbc_bin_trigger_rates_dq
@@ -82,7 +82,7 @@ with h5.File(args.trig_file, 'r') as trig_file:
         if args.prune_number < 1:
             del stat
 
-trig_times_int = trig_times.astype('int')
+trig_times_int = trig_times.astype(np.int64)
 
 dq_times = np.array([])
 dq_logl = np.array([])
@@ -102,12 +102,12 @@ dq_percentiles = np.percentile(dq_logl, percentiles)[1:]
 
 # seconds bin tells what bin each second ends up
 seconds_bin = np.array([n_bins - np.sum(dq_percentiles >= dq_ll)
-                        for dq_ll in dq_logl]).astype('int')
+                        for dq_ll in dq_logl]).astype(np.int64)
 del dq_percentiles
 
 # bin times tells how much time ends up in each bin
 bin_times = np.array([np.sum(seconds_bin == i)
-                      for i in range(n_bins)]).astype('float')
+                      for i in range(n_bins)]).astype(np.float64)
 full_time = float(len(seconds_bin))
 times_nz = (bin_times > 0)
 del dq_logl
@@ -148,7 +148,7 @@ with h5.File(args.output_file, 'w') as f:
         logging.info('Processing %d triggers...', len(trig_percentile))
 
         (counts, bins) = np.histogram(trig_percentile, bins=(percentiles)/100)
-        counts = counts.astype('float')
+        counts = counts.astype(np.float64)
         rates = np.zeros_like(bin_times)
         rates[times_nz] = counts[times_nz]/bin_times[times_nz]
         mean_rate = len(trig_percentile) / full_time
```

with that patch in place there should not be *any* difference due to dtypes, because the dtypes of arrays will not change (regardless of whether or not https://github.com/gwastro/pycbc/pull/4525 is applied).

The remaining question is why has Gareth observed changes in output files before/after this patch? I'm beginning to suspect that this may completely unrelated to dtype changes (and possibly the dtype change led Gareth to believe that files were otherwise the same, when in fact they were not). I haven't yet reproduced this though ...

spxiwh:
Here's the rest of the proposed change:

```
(pycbc_python3_throwawaytest4) [ian.harry@ldas-grid 01:13 am gareth_test]$ diff ~/virtualenvs/pycbc_python3_throwawaytest4/bin/pycbc_bin_trigger_rates_dq ./pycbc_bin_trigger_rates_dq_NEW_fixed
61,62c61,62
<     idx, _ = trig_file.select(
<         lambda snr: snr > args.stat_threshold,
---
>     idx, _, _ = trig_file.select(
>         lambda snr, psd_var_val: (snr/psd_var_val**0.5) >= args.stat_threshold,
63a64
>         f'{ifo}/psd_var_val',
94c95,96
< keep = stat > args.stat_threshold
---
> keep = stat >= args.stat_threshold
> 
```

In particular the `keep` line on line 94 becomes `>=` from `>`. Also the `select` becomes a bit hacky and explicitly includes `psd_var_val`:

```
with HFile(args.trig_file, 'r') as trig_file:
    n_triggers_orig = trig_file[f'{ifo}/snr'].size
    logging.info("Trigger file has %d triggers", n_triggers_orig)
    logging.info('Generating trigger mask')
    idx, _, _ = trig_file.select(
        lambda snr, psd_var_val: (snr/psd_var_val**0.5) >= args.stat_threshold,
        f'{ifo}/snr',
        f'{ifo}/psd_var_val',
        return_indices=True
    )
    data_mask = np.zeros(n_triggers_orig, dtype=bool)
    data_mask[idx] = True
```

Perhaps there should be some logic around this so that if `psd_var_val` is not computed (or worse, if it's 0!) this doesn't get included??

#4524 would help fix this properly, but needs some thought (and a bit more time) over how to do it properly.

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -54,14 +54,23 @@ ifo = args.ifo
 # This works as a pre-filter as SNR is always greater than or equal
 # to sngl_ranking, except in the psdvar case, where it could increase.
 
+def check_snr_psdvar(snr, psd_var_val):
+    """
+    Convenience function to calculate the maximum possible ranking statistic
+    value given SNR and psd variation statistic
+    """
+    rw_snr = snr / psd_var_val ** 0.5
+    rw_snr[psd_var_val == 0] = 0
```
tdent, line 63: What does it mean for psd var to be =0?  Ie what could cause that?

What value do we get if it is not computed in the first place?

spxiwh: `pycbc_inspiral` does not store it if not computed. PyCBC Live has not yet defined behaviour in this case.

.... If this will be fixed "in a better way" soon anyway, we probably don't need the "is it equal to 0" check (but it won't harm). Note that in the ranking computation, psd_var rescaling is not applied if the value is smaller than 0.65.

spxiwh: Noting that the current code would fail if psd_var_val is not computed.

GarethCabournDavies: Do you want me to prevent that failure? It would be a case of minor changes in the select function, where if the dataset isn't present but is asked for, it could be set to zero?

spxiwh: For now (avoiding things that should be in #4524), I would limit changes to this executable, and use a if statement to use the old call to select if psd_var is not present, and the new one if it is.

tdent: Failure is good if the alternative is continuing while producing a plausible but wrong answer .. 
Just noting that substituting a 'silent' default value might be rather dangerous

GarethCabournDavies:
Files now differ at a binary level, but all attributes and datasets or numerically equal

I haven't been able to test the psd_var_val missing check active, as finding a run with dq but without psd_var_val is not obvious. However the function is the same as previously tested

bin/all_sky_search/pycbc_bin_trigger_rates_dq
GarethCabournDavies: I think wrapping in a if statement would work here. We aren't giving an option being ignored, we are simply working around a dataset which may or may not have been saved

tdent: Can we then remove `rw_snr[psd_var_val == 0] = 0` ? If psd var is calculated the psd var val is 0, there is a big problem and I think we want the code to produce an error rather than ignore it ..

spxiwh: Please remove this line

spxiwh (approved):
One line needs removing, otherwise approved.

PR #4520, opened 2023-10-05
added support for mode-by-mode decomposition of NRHybSur3dq8 waveform…

lpathak97:


ahnitz:
@lpathak97 You don't need to close and reopen pull request to change what is in them. The PR reflects the latest state of the branch, so whatever you change in the branch will be reflected automatically. So if you accidently commit a file, you can just revert that or go back to the old version and then push the new version to github without the extra commits.

lpathak97:
Only two issues remain, which are related to duplication.

pycbc/waveform/waveform_modes.py
```diff
@@ -97,6 +97,10 @@ def default_modes(approximant):
     elif approximant.startswith('NRSur7dq4'):
         # according to arXiv:1905.09300
         ma = [(l, m) for l in [2, 3, 4] for m in range(-l, l+1)]
+    elif approximant.startswith('NRHybSur3dq8'):
+        # according to arXiv:1812.07865
+        ma = [(2, 0), (2, 1), (2, 2), (3, 0), (3, 1), (3, 2),
```
ahnitz, line 102: What about -m modes?

ahnitz (approved):
@lpathak97 Can you rebase from master and check the comment about the -m modes for the model. If you verify this then works, please go ahead and merge.

pycbc/waveform/waveform_modes.py
lalit-pathak: This model only supports +m modes.

ahnitz:
@lalit-pathak I think you may have rebased tbe wrong way. You want to use gwastro/master as the base and have your commits on top, not the other way around. (also avoid merge commits)

PR #4522, opened 2023-10-06
Update mac-test.yml

ahnitz:
This is a test


spxiwh (approved):
If it works, all good!

PR #4525, opened 2023-10-10
Don't overwrite default numpy bindings from liglow in record.io

cdcapano:
Currently, if you import the record module in `pycbc.io`, it will cause numpy to start casting `float` and `int` to `float32` and `int32` respectively. The default in numpy is to cast `float` and `int` to 64 bit scalars. The cause of this is due to the numpy `scTypeDict` being updated with ligolw bindings in `record.py`. In ligolw, `float` is cast to `float32` and `int` to `int32`.  This patch fixes that by not including the `float` and `int` mappings from ligolw. This will cause single precision values to be cast to double when reading a ligolw xml file, but that should only result in some meaningless extra digits being added to values.

### DIFF
--- a/pycbc/io/record.py
+++ b/pycbc/io/record.py
@@ -46,7 +46,11 @@ _numpy_function_lib = {_x: _y for _x,_y in numpy.__dict__.items()
 # =============================================================================
 #
 # add ligolw_types to numpy sctypeDict
-numpy.sctypeDict.update(ligolw_types.ToNumPyType)
+# but don't including float -> float32 and int -> int32 bindings, so as not
+# to change numpy defaults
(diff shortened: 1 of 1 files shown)


spxiwh (approved):
Looks good to me. Only stylistic comment is why to use variable names preceded by an underscore for the internal variables in the dict-comprehension?

cdcapano:
@spxiwh I was thinking these variables might be left over in the `io.record` namespace, so preceded them with `_` to make them hidden. But you're right, because they're in a dict comprehension they don't survive outside of it, so that wasn't necessary. It doesn't hurt though, and doesn't violate PEP8, so I'm just going to leave it as is.

PR #4527, opened 2023-10-12
[pycbc live] Allowing the use of template fits in the pycbc live ranking statistic

ArthurTolley:
This PR contains the changes required in pycbc to allow the use of the trigger fit files in the pycbc live search. This does not contain the scripts necessary to create these trigger fits files.

Any comments and questions are very welcome and I can share the files I have created for testing these changes.


pycbc/events/stat.py
```diff
@@ -1779,8 +1793,16 @@ def single(self, trigs):
...
-                                                   trigs.param['mass2'])
+        try:
+            self.curr_mchirp = mchirp_from_mass1_mass2(trigs.param['mass1'],
+                                                       trigs.param['mass2'])
+        except AttributeError:
+            try:
+                self.curr_mchirp = mchirp_from_mass1_mass2(trigs['mass1'],
+                                                           trigs['mass2'])
+            except KeyError:
+                logging.info("Triggers not of the correct format.")
```
GarethCabournDavies, line 1804: Should this be a warning / error?

pycbc/events/stat.py
```diff
@@ -1555,6 +1563,12 @@ def rank_stat_coinc(self, s, slide, step, to_shift,
 
         # First get signal PDF logr_s
         stat = {ifo: st for ifo, st in s}
+
+        try:
+            self.curr_mchirp = kwargs['mchirp']
+        except KeyError:
+            logging.info("Kwargs doesn't contain mchirp")
```
GarethCabournDavies, line 1570: again, this seems like it should be an actual error

GarethCabournDavies: (and the message doesnt seem that useful)

bin/pycbc_live
```diff
@@ -968,6 +968,9 @@ parser.add_argument('--gracedb-labels', metavar='LABEL', nargs='+',
 parser.add_argument('--ifar-upload-threshold', type=float, required=True,
                     help='Inverse-FAR threshold for uploading candidate '
                          'triggers to GraceDB, in years.')
+parser.add_argument('--coinc-threshold', type=float,
+                    help="The time window allowed between coincident triggers",
+                    default=0.002)
```
GarethCabournDavies, line 973: This must have had an at least implied value before - how was that set?

pycbc/events/stat.py
ArthurTolley: Whoops yes it should be, will fix it in the next commit

pycbc/events/stat.py
ArthurTolley: Will change it to an actual raised error. The error message theoretically shouldn't ever be met so I can try and think of a more useful message to display

bin/pycbc_live
ArthurTolley: Very very good question, the only other place I can remember off the top of my head is here: [link](https://github.com/gwastro/pycbc/blob/master/pycbc/events/coinc.py#L241) but I don't know if this is exactly where it is use for the time window for coincidences in the other ranking statistics.

pycbc/events/stat.py
```diff
@@ -1476,7 +1479,12 @@ def single(self, trigs):
             # exists for SingleDetTriggers
             self.curr_tnum = trigs['template_id']
             # Should only be one ifo fit file provided
-            assert len(self.ifos) == 1
+
+            # Arthur Comment: I can understand the need for the assertion
```
ArthurTolley, line 1483: @GarethCabournDavies This comment is a question I have right now. The assertion exists in the `single` method so I can understand why you would assert that you need only 1 ifo. But this code is only called when a coincident detection is made (I believe). Therefore if it is _not_ commented out I get this error: 

`  File "/home/arthur.tolley/.conda/envs/live_rank_stat/lib/python3.9/site-packages/pycbc/events/stat.py", line 1486, in single
    assert len(self.ifos) == 1
AssertionError`

and when it is commented out the code runs through. Should I remove this code? I don't know what the harm would be in removing it.

GarethCabournDavies: This seems to be a little bit of legacy - if calling through `rank_method.single()` when using SingleDetTriggers, this is used, but I don't see anywhere in the code (after a quick browse) which uses this, as most places seem to call direct to get_sngl_ranking.

I think it should be okay to update the comment a line above and remove this assertion.

ArthurTolley: Okay cool, I have removed both the comment and assertion (unless you wanted me to keep the comment for something else?)

GarethCabournDavies: my 'update the comment' referred to the `exists for SingleDetTriggers` comment above. If you could add to that to say the type that you're using now as well, that would be good

bin/pycbc_live
titodalcanton: Seems like you are adding this option but not using it anywhere, is this intended? Or is it being picked up "indirectly" within some deeper module?

pycbc/events/stat.py
```diff
@@ -1555,6 +1557,12 @@ def rank_stat_coinc(self, s, slide, step, to_shift,
 
         # First get signal PDF logr_s
         stat = {ifo: st for ifo, st in s}
+
+        try:
+            self.curr_mchirp = kwargs['mchirp']
+        except KeyError as e:
+            print(e)
```
titodalcanton, line 1564: Either provide a more useful error message, or avoid catching the exception.

pycbc/events/stat.py
```diff
@@ -1779,8 +1787,16 @@ def single(self, trigs):
...
-                                                   trigs.param['mass2'])
+        try:
+            self.curr_mchirp = mchirp_from_mass1_mass2(trigs.param['mass1'],
+                                                       trigs.param['mass2'])
+        except AttributeError:
+            try:
+                self.curr_mchirp = mchirp_from_mass1_mass2(trigs['mass1'],
+                                                           trigs['mass2'])
+            except KeyError as e:
+                print(e)
```
titodalcanton, line 1798: Separate the operations that can actually raise the expected exception from those which do not:
```suggestion
        try:
            mass1 = trigs.param['mass1']
            mass2 = trigs.param['mass2']
        except AttributeError:
            mass1 = trigs['mass1']
            mass2 = trigs['mass2']
        self.curr_mchirp = mchirp_from_mass1_mass2(mass1, mass2)
```

pycbc/events/coinc.py
```diff
@@ -1170,6 +1175,9 @@ def _find_coincs(self, results, valid_ifos):
                 trig_stat = trigs['stat'][i]
                 trig_time = trigs['end_time'][i]
                 template = trigs['template_id'][i]
+                mass1 = trigs['mass1'][i]
+                mass2 = trigs['mass2'][i]
+                mchirp = mchirp_from_mass1_mass2(mass1, mass2)
```
titodalcanton, line 1180: Doing this conversion one trigger at a time may be very very slow, maybe do them all using array operations first?

bin/pycbc_live
ArthurTolley: It is fed through in the `args` to `coinc.py` on [L1130](https://github.com/ArthurTolley/pycbc/blob/template_fits/bin/pycbc_live#L1130) to fulfil the class argument: [L968](https://github.com/ArthurTolley/pycbc/blob/template_fits/pycbc/events/coinc.py#L968) (in coinc.py).

pycbc/events/stat.py
ArthurTolley: Because there is no alternative to the 'mchirp' kwarg not existing, you are suggesting just having `self.curr_mchirp = kwargs['mchirp']` without the try except and then if it fails it fails (just like it would with the except anyway)?

pycbc/events/stat.py
ArthurTolley: That is a much better solution, especially for code readability. Thank you, I will implement it.

pycbc/events/coinc.py
ArthurTolley: I've made this change but I need to update the `test_live_coinc_compare.py` test case to include both 'mass1' and 'mass2' in the trigs dict. They are present in Live (need to quadruple check) but not in the test case.

pycbc/events/stat.py
```diff
@@ -1561,6 +1563,12 @@ def rank_stat_coinc(self, s, slide, step, to_shift,
 
         # First get signal PDF logr_s
         stat = {ifo: st for ifo, st in s}
+
+        try:
+            self.curr_mchirp = kwargs['mchirp']
+        except KeyError:
+            pass
```
ArthurTolley, line 1570: The setting of `self.curr_mchirp` is needed for the stat to run in Live but this code is also called by `pycbc_coinc_findtrigs` where the object being called doesn't have the keyword argument 'mchirp'.

Therefore I've used a try except with a simple `pass` because the error will be expected.

ArthurTolley:
All of the changes made which reference the parameter `mchirp` or `curr_mchirp` are required for the `ExpFitFgBgNormBBHStatistic` statistic. I have moved the setting of `self.curr_mchirp` out of the other stat and into a new method `rank_stat_coinc` which explicitly sets this parameter and then runs the previous `rank_stat_coinc` from the non-BBH fits statistic. Thanks @GarethCabournDavies for helping me unpick that knot!

test/test_live_coinc_compare.py
```diff
@@ -47,7 +47,9 @@ def get_trigs(self):
                     0,
                     self.num_templates,
                     size=self.num_trigs
-                ).astype(np.int32)
+                ).astype(np.int32),
+                "mass1": np.random.uniform(2.0, 100.0, size=self.num_trigs).astype(np.float32),
+                "mass2": np.random.uniform(2.0, 100.0, size=self.num_trigs).astype(np.float32)
```
titodalcanton, line 52: I think `uniform()` takes a `dtype` kwarg, which makes the code shorter, clearer and slightly more efficient.

bin/pycbc_live
titodalcanton: The help string is not correct. The docstring of the class claims that `coinc_threshold` is the

> Amount of time allowed to form a coincidence in addition to the time of flight in seconds.

I would rather call this quantity (and the CLI option) `coinc_window_pad`, `coinc_window_extra` or something like that. Maybe first check with Alex where the word "threshold" came from though.

pycbc/events/stat.py
```diff
@@ -1741,6 +1743,39 @@ def logsignalrate(self, stats, shift, to_shift):
...
+            mass1 = trigs.param['mass1']
+            mass2 = trigs.param['mass2']
+        except AttributeError:
+            mass1 = trigs['mass1']
+            mass2 = trigs['mass2']
+        self.curr_mchirp = mchirp_from_mass1_mass2(mass1, mass2)
+
+        if self.mcm is not None:
+            # Careful - input might be a str, so cast to float
+            self.curr_mchirp = min(self.curr_mchirp, float(self.mcm))
```
titodalcanton, line 1776: This method might be time-critical. Can the cast to float be done during initialization instead?

pycbc/events/stat.py
```diff
@@ -1767,32 +1802,37 @@ def rank_stat_single(self, single_info,
...
-                                                   trigs.param['mass2'])
-        if self.mcm is not None:
-            # Careful - input might be a str, so cast to float
-            self.curr_mchirp = min(self.curr_mchirp, float(self.mcm))
-        return ExpFitFgBgNormStatistic.single(self, trigs)
+
+        try:
+            self.curr_mchirp = kwargs['mchirp']
+        except KeyError:
+            pass
```
titodalcanton, line 1828:
```suggestion
        if 'mchirp' in kwargs:
            self.curr_mchirp = kwargs['mchirp']
```

pycbc/events/coinc.py
titodalcanton: Resolved?

pycbc/events/stat.py
titodalcanton: Ok, see my other suggestion on this though - a simple `if 'mchirp' in kwargs` is simpler (and might even be faster) than raising and catching an exception.

pycbc/events/stat.py
titodalcanton: More generally, I would consider it a bug if `self.mcm` has a string type. A chirp mass is conceptually a positive real number, so a variable representing a chirp mass should always have a real numeric type. What is causing `self.mcm` to be a string?

test/test_live_coinc_compare.py
ArthurTolley: I did follow the example set by the other parameters above, looks like it doesn't have `dtype` which is odd:

Code:
`"mass1": np.random.uniform(2.0, 100.0, size=self.num_trigs, dtype=np.float32)`

Error:
`TypeError: uniform() got an unexpected keyword argument 'dtype'`

pycbc/events/stat.py
ArthurTolley: Re-ordering the methods in the class and creating `rank_stat_coinc` has caused this to appear as though i've created the `single` method whereas I have just re-ordered them, unsure how to get that to appear properly.

The `if self.mcm is not None:` was added in this PR: https://github.com/gwastro/pycbc/pull/3009

So it looks like `self.mcm` might be a str type in offline?

test/test_live_coinc_compare.py
titodalcanton: Bummer, that is unexpected. Nevermind then.

pycbc/events/stat.py
titodalcanton: I would check this with Tom/Ian. I do not understand #3009, it was not properly documented and it looks like a mistake on a cursory look.

ArthurTolley: I will copy and paste @spxiwh's comment from my asking him:
"My hunch is that this is stored in the config file as a string for offline, and nothing outside of the stat module would, or should, know that, and so the top-level code would just look for something like an “additional kwargs” section and simply pass the entire contents of that to the stat class. The stat class would then need to parse that (which in this case would mean translating “65" -> 65). Live works quite differently, and doesn’t use config files. If it’s having to hardcode any possible statistic key-word argument on the command-line (perhaps not ideal) then it would enforce type at that stage."

I agree with you that the float type casting doesn't need to be done every single time this method is called so will move it to the initialisation.

titodalcanton: Ok I see. Well, if this turns out to be a problem (either performance-wise, or type incompatibility) we will find out pretty soon anyway.

ArthurTolley: `self.mcm` will be None in pycbc live so this if statement will execute, I'm prepared to leave it be if that's okay

pycbc/events/coinc.py
```diff
@@ -1208,12 +1218,18 @@ def _find_coincs(self, results, valid_ifos):
...
+                    c = self.stat_calculator.rank_stat_coinc(
+                        sngls_list,
+                        slide,
+                        self.timeslide_interval,
+                        shift_vec,
+                        time_addition=self.coinc_threshold,
+                        mchirp=mchirp
+                    )
+                else:
+                    c = numpy.array([])
```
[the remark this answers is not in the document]
ArthurTolley: This was definitely throwing errors a few months ago when first making these changes. I've just spent a few hours trying to re-create those errors by removing the length check but they don't re-appear so I'm happy to remove this if else block and leave the `rank_stat_coinc` standalone.

bin/pycbc_live
ArthurTolley: `coinc_window_pad` definitely makes more sense if this quantity is in addition to the
 light travel time. I will happily change this unless @ahnitz has an objection?

ArthurTolley:
The test: `run small search using pegasus + condor` was failing repeatedly so I've trawled the logs to find out why.

```
- Failure:
    - pegasus-plan_gw-main
    - gw-main.dag.rescue001:
    - Nodes that failed: 3
        - pegasus-plan_ID0000073
            - H1-SINGLES_MINIFOLLOWUP_FULL_DATA_VETOED-1186740100-3400
                - page_snglinfo_ID0_ID0000001
                    - ERROR:  Expected local file does not exist: /var/lib/condor/execute/dir_74309/pegasus.jACIXUYPs/H1-PAGE_SNGLINFO_FULL_DATA_VETOED_0-1186740100-3400.html
                    - 
        - pegasus-plan_ID0000079
            - L1-SINGLES_MINIFOLLOWUP_FULL_DATA_VETOED-1186740100-3400
                - page_snglinfo_ID0_ID000000
        - pegasus-plan_ID0000085
            - V1-SINGLES_MINIFOLLOWUP_FULL_DATA_VETOED-1186740100-3400
                - page_snglinfo_ID0_ID0000001

          File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/pycbc/events/stat.py", line 853, in find_fits
            ifo = trigs.ifo
        AttributeError: 'dict' object has no attribute 'ifo'
        
        During handling of the above exception, another exception occurred:
        
          File "/opt/hostedtoolcache/Python/3.8.18/x64/bin/pycbc_page_snglinfo", line 104, in <module>
            sds = rank_method.single(sngl_file.trig_dict())
          File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/pycbc/events/stat.py", line 1474, in single
            sngl_stat = self.lognoiserate(trigs)
          File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/pycbc/events/stat.py", line 1443, in lognoiserate
            alphai, ratei, thresh = self.find_fits(trigs)
          File "/opt/hostedtoolcache/Python/3.8.18/x64/lib/python3.8/site-packages/pycbc/events/stat.py", line 855, in find_fits
            ifo = trigs['ifo']
        KeyError: 'ifo'
```

Originally the code has:
```
try:
    tnum = trigs.template_num  # exists if accessed via coinc_findtrigs
    ifo = trigs.ifo
except AttributeError:
    tnum = trigs['template_id']  # exists for SingleDetTriggers
    assert len(self.ifos) == 1
    # Should be exactly one ifo provided
    ifo = self.ifos[0]
```

I've replaced it with:
```
try:
    tnum = trigs.template_num
except AttributeError:
    tnum = trigs['template_id']

try:
    ifo = trigs.ifo
except AttributeError:
    ifo = trigs.get('ifo', None)
    if ifo is None:
        ifo = self.ifos[0]
    assert ifo in self.ifos
```

So if trigs doesn't have the `ifo` attribute or key then it falls back to the original way the code has identified the ifo currently used. There is now a missing `assert len(self.ifos) == 1` but we shall see if the test runs through now before adding it back.

pycbc/events/coinc.py
```diff
@@ -879,8 +881,12 @@ def __init__(self, num_templates, analysis_block, background_statistic,
         self.lookback_time = (ifar_limit * lal.YRJUL_SI * timeslide_interval) ** 0.5
         self.buffer_size = int(numpy.ceil(self.lookback_time / analysis_block))
 
-        det0, det1 = Detector(ifos[0]), Detector(ifos[1])
-        self.time_window = det0.light_travel_time_to_detector(det1) + coinc_threshold
+        self.dets = {}
+        for ifo in ifos:
+            self.dets[ifo] = Detector(ifo)
```
titodalcanton, line 886:
```suggestion
        self.dets = {ifo: Detector(ifo) for ifo in ifos}
```

pycbc/events/coinc.py
```diff
@@ -1167,11 +1176,16 @@ def _find_coincs(self, results, valid_ifos):
                 continue
             # Find newly added triggers in fixed_ifo
             trigs = results[fixed_ifo]
+            # Calculate mchirp as a vectorized operation
+            mass1s = trigs['mass1']
+            mass2s = trigs['mass2']
+            mchirps = mchirp_from_mass1_mass2(mass1s, mass2s)
```
titodalcanton, line 1182: `mass1s` and `mass2s` are not needed past these few lines, so I would do
```suggestion
        mchirps = mchirp_from_mass1_mass2(
            trigs['mass1'], trigs['mass2']
        )
```

pycbc/events/coinc_rate.py
```diff
@@ -131,16 +136,20 @@ def multiifo_noise_coincident_area(ifos, slop):
...
         area in units of seconds^(n_ifos-1) that coincident values can fall in
     """
     # set up detector objects
-    dets = {}
-    for ifo in ifos:
-        dets[ifo] = pycbc.detector.Detector(ifo)
+    if dets is None:
+        dets = {}
+        for ifo in ifos:
+            dets[ifo] = pycbc.detector.Detector(ifo)
```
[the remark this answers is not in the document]
titodalcanton: Same suggestion here for using a one-line dict comprehension instead of 3 lines.

pycbc/events/stat.py
```diff
@@ -845,13 +845,18 @@ def find_fits(self, trigs):
...
             ifo = trigs.ifo
         except AttributeError:
-            tnum = trigs['template_id']  # exists for SingleDetTriggers
-            assert len(self.ifos) == 1
-            # Should be exactly one ifo provided
-            ifo = self.ifos[0]
+            ifo = trigs.get('ifo', None)
+            if ifo is None:
+                ifo = self.ifos[0]
+            assert ifo in self.ifos
```
titodalcanton, line 858: I cannot judge if this is correct, I do not understand what this bit is doing just by looking at it. Seems strange that the assertion is inside the except block though, is that intentional?

pycbc/events/stat.py
```diff
@@ -1542,8 +1546,23 @@ def rank_stat_coinc(self, s, slide, step, to_shift,
...
-        ln_noise_rate = coinc_rate.combination_noise_lograte(
-                                  sngl_rates, kwargs['time_addition'])
+        if 'dets' in kwargs:
+            ln_noise_rate = coinc_rate.combination_noise_lograte(
+                                    sngl_rates, kwargs['time_addition'],
+                                    kwargs['dets'])
+
+            # Find total volume of phase-time-amplitude space occupied by
+            # noise coincs
+            # Extent of time-difference space occupied
```
titodalcanton, line 1556: This comment seems to apply to both branches of the if statement, so maybe move it outside?

pycbc/events/stat.py
```diff
@@ -1741,6 +1755,39 @@ def logsignalrate(self, stats, shift, to_shift):
         logr_s += numpy.log((self.curr_mchirp / 20.) ** (11. / 3.))
         return logr_s
 
+    def single(self, trigs):
+        """
+        Calculate the necessary single detector information
+
+        In this case the ranking rescaled (see the lognoiserate method here)
+        with the phase, end time, sigma, SNR, template_id and the
+        benchmark_logvol values added in. This also stored the current chirp
```
titodalcanton, line 1764: This part of the comment does not make sense to me. We are talking about a single-detector ranking stat, so why are the phase and end time mentioned?

titodalcanton: I see there has been some code movement between `single()` and `rank_stat_coinc()`, perhaps that explains the confusion?

pycbc/events/stat.py
```diff
@@ -1741,6 +1755,39 @@ def logsignalrate(self, stats, shift, to_shift):
...
+            mass1 = trigs.param['mass1']
+            mass2 = trigs.param['mass2']
+        except AttributeError:
+            mass1 = trigs['mass1']
+            mass2 = trigs['mass2']
+        self.curr_mchirp = mchirp_from_mass1_mass2(mass1, mass2)
+
+        if self.mcm is not None:
+            # Careful - input might be a str, so cast to float
+            self.curr_mchirp = min(self.curr_mchirp, float(self.mcm))
```
titodalcanton, line 1788: I should have noticed this earlier, but `self.curr_mchirp` is an array, right? How does `min()` work here?

pycbc/events/stat.py
maxtrevor: This method does not just calculate the single detector ranking stat. It is also responsible for making sure that triggers have all the information attached to them that is needed for the coinc stat calculation, including phase and end time. In any case, Arthur has not actually edited the comment, the git diff is being thrown off by him reordering the methods.

pycbc/events/stat.py
maxtrevor: Like the above, this is not new. I don't quite understand how the `min()` works here, but Arthur said in the previous review thread `self.mcm` is None in PyCBC Live, and the offline case is unchanged.

pycbc/events/coinc_rate.py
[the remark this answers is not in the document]
GarethCabournDavies:
```suggestion
    if dets is None:
        self.dets = {ifo: Detector(ifo) for ifo in ifos}
```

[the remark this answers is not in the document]
GarethCabournDavies: (Arthur is on holiday, so I'm committing changes while he's away)

pycbc/events/stat.py
GarethCabournDavies: I echo that this needs comments.

If trigs is dict-like you'll get to the except, then if `ifo` is not a key in the trigs object, then it could be met.

I am fairly sure that `self.ifos[0]` is _not_ the right thing to get here though, as that will always get the same ifo, even when you're using the triggers from a different detector.

pycbc/events/stat.py
GarethCabournDavies, line 1556:
```suggestion
        # Find total volume of phase-time-amplitude space occupied by
        # noise coincs
        if 'dets' in kwargs:
            ln_noise_rate = coinc_rate.combination_noise_lograte(
                                    sngl_rates, kwargs['time_addition'],
                                    kwargs['dets'])
                                    
            # Extent of time-difference space occupied
```

pycbc/events/stat.py
GarethCabournDavies: I'm confused why this works as well - it shouldn't.

I think this should just use np.minimum instead, but I'm not sure if we ever depend on the fact that ReadByTemplate has the same mchirp for all triggers when this is called

pycbc/events/stat.py
GarethCabournDavies: Suggested change in another thread fixes this

pycbc/events/stat.py
```diff
@@ -1481,10 +1486,9 @@ def single(self, trigs):
             # exists if accessed via coinc_findtrigs
             self.curr_tnum = trigs.template_num
         except AttributeError:
-            # exists for SingleDetTriggers
+            # exists for SingleDetTriggers & pycbc_live get_coinc
             self.curr_tnum = trigs['template_id']
-            # Should only be one ifo fit file provided
-            assert len(self.ifos) == 1
```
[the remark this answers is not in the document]
GarethCabournDavies: Max is right, this is just removing a block which shouldn't be one anymore

Plus according to [python docs](https://wiki.python.org/moin/UsingAssertionsEffectively) we should really only use assert for debugging our internal logic
```
Assertions should *not* be used to test for failure cases that can occur because of bad user input 
```

GarethCabournDavies:
I've just cherry-picked a commit from master which should allow checks to progress more than they have, maybe that will fix things

pycbc/events/stat.py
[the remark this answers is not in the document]
titodalcanton: Yes, `assert` is basically a way to explicitly state what you are assuming at that point.

pycbc/events/stat.py
maxtrevor: @GarethCabournDavies when do we expect trigs to be dict-like vs not? I've lost track

pycbc/events/stat.py
maxtrevor: Since this is not a change that Arthur made (its being shown as a change due to Arthur moving code around) I support just leaving it as-is.

pycbc/events/stat.py
GarethCabournDavies: Dict-like will be when it is called using:
 - a dictionary keyed on the dataset names like `pycbc_fit_sngls_by_template`
 - a SingleDetTriggers object, as in `pycbc_page_snglinfo` and `pycbc_sngl_minifollowup`
 - the h5py file directly as in `pycbc_fit_sngls_binned` (it shouldn't really do this, and should use SingleDetTriggers)

Not dictlike will be when using ReadByTemplate, which is done in `pycbc_coinc_findtrigs` and `pycbc_sngls_findtrigs`

pycbc/events/stat.py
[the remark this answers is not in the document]
GarethCabournDavies: I think we're happy with this then

pycbc/events/stat.py
GarethCabournDavies: I think Arthur has introduced a use-case for it though (mchirp kwargs being passed through to rank_stat_coinc)

I will have a chat with Arthur and see what we can do to solve this, but it will be next week

maxtrevor: Arthur said here: https://github.com/gwastro/pycbc/pull/4527/files/640a68b622f1267a82960ca4d56f51b214f8ff94#r1476198713 that self.mcm will be None in PyCBC Live

GarethCabournDavies: True, it wont be used anytime soon I don't think

maxtrevor: Between that statement and the lack of any actual code change, I wouldn't let this bit of code delay the review as a whole

pycbc/events/stat.py
maxtrevor: I was hoping there was some clean division like always dict-like when in PyCBC live and always not in offline, but I guess thats not the case. Might be nice to make that more consistent in the future, but not a task for this MR. From what I can see, I do think its the case that Live always uses the dict-like path?

GarethCabournDavies: I think it is only used in live in the `single_stat = self.stat_calculator.single(trigsc)` line above, and `trigsc` is dictlike, so it seems this is safe

GarethCabournDavies (approved):
I think the CI test failure is unrelated, so I will approve, and see if anyone provides a fix over the weekend

maxtrevor:
I see all the required checks passed, happy to merge this

ArthurTolley:
Just wanted to say thank you to everyone on this commit for getting this over the line while I was on holiday. Nice to see it finally merged!

PR #4534, opened 2023-10-16
Scripts for dealing with preparation of offline gracedb upload files

GarethCabournDavies:
The PR #4438 was getting too long and complicated to review, and there was an obvious split in order to make this more simple.

Here I separate off the changes for new executables being used.

These are for:
 - The creation of the XML for the specific event to be uploaded 
 - The script for uploading the single event
 - A wrapper script around bayestar

The first two of these are basically a split of the existing 'upload many events from this XML' script so that the files can be prepared in advance, and then uploaded later.

The wrapper around bayestar is so that we can specify the output filename and use them in the workflow

The changes to existing files are mainly so that we can reuse existing code without repetition, and a minor change so that the XML uploads from offline match the convention used in online

### DIFF
--- a/bin/all_sky_search/pycbc_make_bayestar_skymap
+++ b/bin/all_sky_search/pycbc_make_bayestar_skymap
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 """
-Simple wraper around bayestar-localize-coincs to get sky localization for a
+Simple wrapper around bayestar-localize-coincs to get sky localization for a
 specific compact binary merger event.
 
 Uses an XML containing SNR timeseries and PSD and calls BAYESTAR to produce a
--- a/bin/all_sky_search/pycbc_make_bayestar_skymap
+++ b/bin/all_sky_search/pycbc_make_bayestar_skymap
@@ -1,0 +1,85 @@
+#!/usr/bin/env python
+
+"""
+Simple wraper around bayestar-localize-coincs to get sky localization for a
+specific compact binary merger event.
+
+Uses an XML containing SNR timeseries and PSD and calls BAYESTAR to produce a
+sky localization.
+
+The wrapper needs to exist as the XML format doesnt allow us to indicate
+the low frequency cutoff or waveform used. We also want to be able to control
+the output filename which is not possible in bayestar-localize-coincs
+
+Unrecognised options will be passed straight to the bayestar subprocess
+"""
+
+import argparse
+import subprocess
+import shutil
+import logging
+import os
+import glob
+import sys
+import random
+import tempfile
+import pycbc.version
+from pycbc import init_logging
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument('--version', action="version",
+                    version=pycbc.version.git_verbose_msg)
+parser.add_argument('--verbose', action='count',
+                    help="Make the logging increasingly more verbose")
+parser.add_argument('--bayestar-executable',
+                    help="The bayestar-localize-coinc executable to be run. "
+                         "If not given, will use whatever is available in "
+                         "the current environment.")
+parser.add_argument('--event-xml', required=True,
+                    help="XML file containing event information, SNR "
+                         "timeseries and PSD to pass to bayestar")
+parser.add_argument('--waveform', default='TaylorF2',
+                    help="Waveform used in the matched-filtering "
+                         "to generate the SNR timeseries.")
+parser.add_argument('--low-frequency-cutoff', type=float, default=20,
+                    help="Low-frequency cutoff used in the matched-filtering "
+                         "to generate the SNR timeseries")
+parser.add_argument('--output-file', required=True,
+                    help="Filename to output the fits file to.")
+args, unknown = parser.parse_known_args()
+
+# Default logging is set higher than normal for this job
+logging_level = args.verbose + 1 if args.verbose else None
+init_logging(logging_level)
+logging.info("Starting")
+
+bayestar_exe = args.bayestar_executable or 'bayestar-localize-coincs'
+
+tmpdir = tempfile.mkdtemp()
+
+# Set up the command to pass to bayestar.
+# The XML file path being passed twice is a legacy requirement, not a mistake.
+cmd = [bayestar_exe,
+       args.event_xml,
+       args.event_xml,
+       '--waveform', args.waveform,
+       '--f-low', str(args.low_frequency_cutoff),
+       '-o', tmpdir]
+
+# Pass any unknown options straight to the subprocess
+cmd += unknown
+
+logging.info("Running %s", ' '.join(cmd))
+subprocess.check_output(cmd)
+
+# Find the fits file in the temporary directory:
+# It would be nice to do this better! - maybe use the input xml
+# file to find the number which is going to be used?
+fits_filenames = glob.glob(os.path.join(tmpdir, '*.fits*'))
+if len(fits_filenames) != 1:
+    raise ValueError(f'argh, got {len(fits_filenames)} FITS files after running BAYESTAR, I want one!')
+
+logging.info("Moving output to %s", args.output_file)
+shutil.move(fits_filenames[0], args.output_file)
+
+logging.info("Done")
--- a/bin/all_sky_search/pycbc_prepare_xml_for_gracedb
+++ b/bin/all_sky_search/pycbc_prepare_xml_for_gracedb
@@ -1,0 +1,227 @@
+#!/usr/bin/env python
+
+# Copyright (C) 2015-2023 Ian Harry, Gareth Cabourn Davies
+#
+# This program is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License as published by the
+# Free Software Foundation; either version 3 of the License, or (at your
+# option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation, Inc.,
+# 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+
+"""
+Take a coinc xml file containing multiple events and prepare one of them
+for upload to gracedb.
+"""
+
+import argparse
+import logging
+import numpy as np
+import h5py
+import matplotlib
+matplotlib.use('agg')
+
+import pycbc
+import lal
+import lal.series
+from ligo.lw import lsctables
+from ligo.lw import utils as ligolw_utils
+from ligo.segments import segment, segmentlist
+from pycbc.io.ligolw import (
+    LIGOLWContentHandler,
+    snr_series_to_xml,
+)
+from pycbc.psd import interpolate
+from pycbc.types import FrequencySeries, TimeSeries, load_timeseries
+from pycbc.types import MultiDetOptionAction
+from pycbc.results import generate_asd_plot, generate_snr_plot
+
+parser = argparse.ArgumentParser(description=__doc__)
+parser.add_argument("--verbose", action='count',
+                    help="Increase logging level, default=info")
+parser.add_argument("--psd-files", nargs='+', required=True,
+                    help='HDF file(s) containing the PSDs to upload')
+parser.add_argument("--snr-timeseries", nargs='+', required=True,
+                    help='HDF file(s) containing the SNR timeseries to upload')
+parser.add_argument('--input-file', required=True, type=str,
+                    help='Input LIGOLW XML file of coincidences.')
+parser.add_argument('--event-id', type=int, required=True,
+                    help='ID for the event being prepared.')
+parser.add_argument('--output-file',
+                    help="Output filename for the locally stored XML. ")
+parser.add_argument('--snr-timeseries-plot',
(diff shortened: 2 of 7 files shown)


GarethCabournDavies:
See #4535 for associated workflow changes

GarethCabournDavies:
@titodalcanton note that these changes are the ones already reviewed in #4438

bin/all_sky_search/pycbc_make_bayestar_skymap
```diff
@@ -0,0 +1,85 @@
+#!/usr/bin/env python
+
+"""
+Simple wraper around bayestar-localize-coincs to get sky localization for a
```
titodalcanton, line 4:
```suggestion
Simple wrapper around bayestar-localize-coincs to get sky localization for a
```

pycbc/io/live.py
```diff
@@ -345,10 +344,6 @@ def upload(self, fname, gracedb_server=None, testing=True,
         labels: list
             Optional list of labels to tag the new event with.
         """
-        import matplotlib
-        matplotlib.use('Agg')
```
titodalcanton, line 349: I am afraid this must remain here, otherwise the matplotlib calls made by `generate_snr_plot()` will not (necessarily) be made with the Agg backend.

Maybe the backend should rather be set in `pycbc_live` straight at initialization, rather than deep inside a module like this…

pycbc/io/live.py
```diff
@@ -446,25 +441,15 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
+            triggers = {
+                ifo: (self.coinc_results[f'foreground/{ifo}/end_time']
+                      + self.time_offset,
+                      self.coinc_results[f'foreground/{ifo}/snr'])
+                for ifo in self.et_ifos
+                }
             ref_time = int(self.merger_time)
-            for ifo in sorted(self.snr_series):
-                curr_snrs = self.snr_series[ifo]
-                curr_snrs.save(snr_series_fname, group='%s/snr' % ifo)
```
titodalcanton, line 453: After this change, PyCBC Live will no longer save the SNR timeseries to an HDF file. Is this intended? Maybe it is, as I do not remember actually using such files ever. However, if this is a conscious decision, then we can probably drop the `snr_series_fname` variable?

titodalcanton: …and do not forget that the HDF file is actually uploaded to GraceDB a few lines below! I thought we had removed this…

pycbc/io/live.py
GarethCabournDavies: It would be better in pycbc_live I think, just in case the module is being called elsewhere

GarethCabournDavies: (which is the case I think?)

pycbc/io/live.py
titodalcanton: Nope, it is still there!

I am not super happy with a PR for the offline search modifying the behavior of the online search. If we want to drop the HDF file, can we at least do so in a separate mini-PR, for a simpler history?

pycbc/io/live.py
titodalcanton: Ah, `pycbc_live` does already set the Agg backend in fact. However, `pycbc_optimize_snr` does not.

titodalcanton: I am opening a separate PR to address this, as it is more a PyCBC Live thing than offline.

pycbc/io/live.py
GarethCabournDavies: Re-instated this for this PR, we can think about dropping this somewhere else

GarethCabournDavies: (This was a mistake more than an intentional change!)

pycbc/io/live.py
GarethCabournDavies: I will rebase once that is in then

pycbc/io/live.py
```diff
@@ -508,6 +497,10 @@ def upload(self, fname, gracedb_server=None, testing=True,
                          if v != 0.0}
             labels, sizes = zip(*prob_plot.items())
             colors = [source_color(label) for label in labels]
+            import matplotlib
+            matplotlib.use('Agg')
```
titodalcanton, line 501: Please rebase on master and remove these two lines.

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,227 @@
...
+    def __init__(self, psd, *args):
+        self.psd = psd
+
+logging.info("Reading PSDs")
+psds = {}
+for psd_file in args.psd_files:
+    (ifo, group), = h5py.File(psd_file, "r").items()
+    psd = [group["psds"][str(i)] for i in range(len(group["psds"].keys()))]
+    psds[ifo] = segmentlist(psd_segment(*segargs) for segargs in zip(
+        psd, group["start_time"], group["end_time"]))
```
titodalcanton, line 96: This is the kind of code block that might benefit from being turned into a short function, as it pollutes the global namespace with a few very generic variable names (`ifo`, `psd`, `group`) which have a high probability of producing variable-reuse bugs later. How about this:
```python
def read_psds(psd_files):
    logging.info("Reading PSDs")
    psds = {}
    for psd_file in args.psd_files:
        (ifo, group), = h5py.File(psd_file, "r").items()
        psd = [group["psds"][str(i)] for i in range(len(group["psds"].keys()))]
        psds[ifo] = segmentlist(psd_segment(*segargs) for segargs in zip(
            psd, group["start_time"], group["end_time"]))
    return psds


psds = read_psds(args.psd_files)
```

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,227 @@
...
+
+# Get information from the single_template snr timeseries files:
+logging.info("Reading SNR timeseries")
+snr_timeseries = {}
+for snr_filename in args.snr_timeseries:
+    with h5py.File(snr_filename, 'r') as snr_f:
+        ifo = snr_f.attrs['ifo']
+        time = snr_f.attrs['event_time']
+    snr_timeseries[(ifo, time)] = load_timeseries(snr_filename, 'snr')
```
titodalcanton, line 105: Another case for a short function here.
```python
def read_snr_timeseries(snr_timeseries_files):
    """Get information from the single_template snr timeseries files.
    """
    logging.info("Reading SNR timeseries")
    snr_timeseries = {}
    for snr_filename in snr_timeseries_files:
        with h5py.File(snr_filename, 'r') as snr_f:
            ifo = snr_f.attrs['ifo']
            time = snr_f.attrs['event_time']
        snr_timeseries[(ifo, time)] = load_timeseries(snr_filename, 'snr')
    return snr_timeseries


snr_timeseries = read_snr_timeseries(args.snr_timeseries)
```

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
titodalcanton: Though I volunteer to do this in a subsequent PR…

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
```diff
@@ -0,0 +1,227 @@
...
+
+if args.psd_plot:
+    logging.info("Saving ASD plot %s", args.psd_plot)
+    generate_asd_plot(psddict, args.psd_plot)
+
+triggers = {sngl.ifo: (sngl.end_time + sngl.end_time_ns * 1e-9, sngl.snr)
+            for sngl in sngl_inspiral_table_curr}
+base_time = int(np.floor(time))
+if args.snr_timeseries_plot:
+    logging.info("Saving SNR plot %s", args.snr_timeseries_plot)
```
titodalcanton, line 219:
```suggestion
if args.snr_timeseries_plot:
    triggers = {sngl.ifo: (sngl.end_time + sngl.end_time_ns * 1e-9, sngl.snr)
                for sngl in sngl_inspiral_table_curr}
    base_time = int(np.floor(time))
    logging.info("Saving SNR plot %s", args.snr_timeseries_plot)
```

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,195 @@
...
+                    help='Add a log entry to each upload with the given message')
+parser.add_argument('--testing', action="store_true", default=False,
+                    help="Upload event to the TEST group of gracedb.")
+parser.add_argument('--production-server', action="store_true", default=False,
+                    help="Upload event to production graceDB. If not given "
+                         "events will be uploaded to playground server.")
+parser.add_argument('--force-overwrite', action='store_true', default=False,
+                    help="GraceDb instance will be checked for if an event "
+                         "with the same event time and FAR already exist. "
+                         "If so, event will not be uploaded")
```
titodalcanton, line 47: The name of the option seems inconsistent with the help string.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,195 @@
...
+                    help="If not using --force-overwrite, add a string to "
+                         "gracedb query to further filter events. "
+                         "Default='pycbc'")
+parser.add_argument('--search-id-string', default='AllSky',
+                    help="Using T050017-v1 naming convention for "
+                         "output XML filename. This is a string for "
+                         "search identifier in filename, e.g. 'AllSky' "
+                         "would give H1L1V1-PYCBC_AllSky-1234567890-1.xml. "
+                         "See https://dcc.ligo.org/LIGO-T050017/public. "
+                         "Default: 'AllSky'")
```
titodalcanton, line 58: Is the help text really correct here? Is this option not specifying the search tag for GraceDB?

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,195 @@
...
+                          "event.")
+parser.add_argument('--labels', nargs='+',
+                    help="Labels to add to the event in GraceDB")
+
+args = parser.parse_args()
+
+pycbc.init_logging(logging.INFO)
+# Make the scitokens logger a little quieter
+# (it is called through GraceDB)
+logging.getLogger('scitokens').setLevel(logging.root.level + 10)
```
titodalcanton, line 81: Does the scitoken logger always exist, and if not, does this cause the code to crash?

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -0,0 +1,195 @@
...
+
+# document the absolute path to the input file
+input_file_str = 'Candidate uploaded from ' \
+    + os.path.abspath(args.xml_file_for_upload)
+gracedb.write_log(r['graceid'], input_file_str)
+
+# document the command line used in the event log
+command_str = 'Upload command: ' \
+    + ' '.join(sys.argv)
+gracedb.write_log(r['graceid'], command_str)
```
titodalcanton, line 121: Maybe we also want the `analyst_comments` tag here. And in the spirit of minimizing the calls to GraceDB, is it better to collate all strings into a single longer comment?

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
titodalcanton: …wait, this option does not seem to be used in the code.

titodalcanton: Same for --query-string. Are those inherited from old code?

GarethCabournDavies: Inherited - I will remove

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
GarethCabournDavies: I've tested and this works with a generic string, i.e.:
```
In [1]: import logging                                                                               

In [2]: logging.getLogger('something')                                                               
Out[2]: <Logger something (WARNING)>                                                                                             

In [3]: logging.getLogger('something').setLevel(40)                                                  

In [4]: logging.getLogger('something')                                                               
Out[4]: <Logger something (ERROR)>
```

So it won't crash

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
GarethCabournDavies: That was a holdover from previous usage

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
GarethCabournDavies: Done

bin/all_sky_search/pycbc_prepare_xml_for_gracedb
GarethCabournDavies: Done

pycbc/io/live.py
GarethCabournDavies: Done

titodalcanton (approved):
No further comments.

pycbc/io/live.py
```diff
@@ -506,6 +499,8 @@ def upload(self, fname, gracedb_server=None, testing=True,
                          if v != 0.0}
             labels, sizes = zip(*prob_plot.items())
             colors = [source_color(label) for label in labels]
+            import pylab as pl
```
titodalcanton, line 502: As codeclimate noted, this is not needed, pylab is imported at the beginning of this (insanely long) method.

PR #4538, opened 2023-10-19
Modify stats_dist to utilize hdf

ETVincent:
Remove glue dependency and modify file to use hdf instead of xml for pygrb stats_dist.


ETVincent:
This strongly depends on some updates from Jacob that may not be in any PRs right now.

pannarale (approved):

PR #4544, opened 2023-10-26
Plotting order for KDE scatter

PRAVEEN-mnl:


tdent:
I guess there are 3 possible ways to deal with the sorting: first, you leave the values in the order they are in the file; second, you sort in ascending order (highest on top); third, you sort in descending order (lowest on top).  Also I think 'normal' doesn't really explain what the option means. 

So I would suggest having three possible options, ie 'file', 'increasing', 'decreasing', with the default being 'file'

bin/all_sky_search/pycbc_plot_kde_vals
```diff
@@ -47,12 +45,16 @@ kde_arrays = {
...
     idx = numpy.argsort(kde_values)
-    param0 = param0[idx]
-    if len(args.param) > 1:
-        param1 = param1[idx]
-    kde_values = kde_values[idx]
+elif args.sort_type == 'decreasing':
+    idx = numpy.argsort(kde_values)[::-1]
+else:
+    idx = numpy.arange(len(kde_values))
+param_arrays = [param[idx] for param in param_arrays]
```
tdent, line 54: Maybe put a comment before this line, eg `Sort each of the parameter arrays`

PRAVEEN-mnl:
I did `git fetch origin` and `git rebase origin/master` on my master branch before creating a new branch. 

This branch was created just to introduce the sorting but the code became repetitive.
For example:  `--plot-type` has 2 choices: kde_vs_param & param_vs_param. In both, ax.scatter() line gets repeated 6 times overall in the code. The new code uses it twice.
Also, during sorting, the parameters need to be called again.
The new code produces the same plots as before. I have tried plotting all the combinations of params along with  `--which-kde`

PRAVEEN-mnl:
> I don't understand why there are 6 commits here and why there are more changes in the plotting code than just introducing the sorting. When you want to make a new change, update your local copy of the main (/master) branch to get changes from the remote repo then make a new feature branch for the pull request.

tdent:
OK, I thought there had been a git problem because these extra changes were not explained by the title or the description of the PR so I did not know why they were there.  That is why you should have a description in the PR unless the changes are really simple and self-explanatory, so that the reviewer knows why the changes are being made.

bin/all_sky_search/pycbc_plot_kde_vals
```diff
@@ -34,20 +36,30 @@ if args.signal_file:
...
-if len(args.param) > 1:
-    param1 = template_data[args.param[1]][:]
+param_arrays = [template_data[param][:] for param in args.param]
+
+kde_arrays = {
+    'signal_kde': signal_kde,
+    'template_kde': template_kde,
+    'ratio_kde': signal_kde / template_kde,
+}
+kde_values = kde_arrays[args.which_kde]
```
tdent, line 46: As you don't use the variable `kde_arrays` anywhere else, can you just do this instead? 
```
kde_values = {
    'signal_kde': signal_kde,
    'template_kde': template_kde,
    'ratio_kde': signal_kde / template_kde,
}[args.which_kde]
```

tdent (changes requested):
Minor change to use of dictionary, otherwise looks OK

tdent (approved):

PR #4545, opened 2023-10-26
Some changes to HFile.select() and SingleDetTriggers

GarethCabournDavies:
This is an attempt to sort #4524

These changes allow:
 - any of the rankings from ranking.py to be used in the `HFile.select()` function
 - A pre-sort directly within SingleDetectorTriggers, so that the `HFile.select()` functionality can used for simple things easily
 - A pre-cut in `mask_to_n_loudest_clustered_events` in cases where this is appropriate

So far as I can test, this has not broken previous functionality except for one usage, which is the filter string option in `pycbc_plot_singles_vs_params`


pycbc/io/hdf.py
```diff
@@ -56,52 +67,89 @@ def select(self, fcn, *args, **kwds):
...
             r = i + chunksize if i + chunksize < size else size
 
-            # Read each chunk's worth of data and find where it passes the function
-            partial = [refs[arg][i:r] for arg in args]
+            # Read each chunk's worth of data and find where it passes
+            # the function
+            partial_data = {arg: refs[arg][i:r][mask[i:r]]
+                            for arg in req_dsets}
+            partial = [partial_data[a] if a in partial_data
+                       else ranking.get_sngls_ranking_from_trigs(
```
ahnitz, line 128: @GarethCabournDavies This seems like a odd shoehorn into this class to me. The idea is that this class should be general and only really build upon h5py / numpy and so can be used for general purposes where more careful access of file informatoin is needed.  Is there a way to avoid explicitly calling a module from ranking.py? Perhaps by extending the options to this function?

ahnitz (changes requested):

pycbc/io/hdf.py
ahnitz: @GarethCabournDavies Ok, I think one approach would be to add something like a 'derived' keyword where on can provide a dictionary -> function mapping (could b lambda function to allow more complicated function interface like you might have with the ranking one. 

So something like the following input. 

derived = {'derived_name': function} 

And in this case, the 'required' keys should just be expanded with the ones you know your functions need. I see you added this data to ranking.py, so I think where you are calling the select method from should have this information.

GarethCabournDavies: That makes sense - I think this will still require the dictionary setup I used for passing to the derived (ranking) functions, as otherwise there could be duplicated datasets which we should avoid, but this does add a layer of generality

GarethCabournDavies: The select() function has been updated to be agnistic of any other module.

SingleDetTriggers still calls through for defined ranking statistics - as SingleDetTriggers already has this happening for certain attributes, I thought this should be okay still

GarethCabournDavies:
Testing on the updates to the executables which use these are ongoing. I will point to the success (or make changes) when they are complete

pycbc/io/hdf.py
ahnitz: SingleDetTriggers is specialized so I think it makes sense there.

GarethCabournDavies:
I am going to try and un-break the filter string option as the next step in this

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -50,42 +50,11 @@ logging.info('Start')
...
 trigs = SingleDetTriggers(
     args.trig_file,
-    None,
-    None,
-    None,
-    None,
     ifo,
-    premask=data_mask
+    filter_rank=args.sngl_ranking,
+    filter_threshold=args.stat_threshold,
```
GarethCabournDavies, line 57: Note that this example is pretty much the reasoning behind this change - to clean up codes like this

bin/minifollowups/pycbc_injection_minifollowup
```diff
@@ -305,19 +306,31 @@ for num_event in range(num_events):
                             use_exact_inj_params=True)
 
     for curr_ifo in args.single_detector_triggers:
+        # Finding loudest template in this detector near to the injection:
```
GarethCabournDavies, line 309: The changes here were possible before, but this is just to implement the select() function and make things quite a bit faster

bin/plotting/pycbc_plot_singles_vs_params
```diff
@@ -123,6 +139,29 @@ if opts.max_y is not None:
...
+    pycbc.results.save_fig_with_metadata(
+        fig,
+        opts.output_file,
+        title=title,
+        caption=fig_caption,
+        cmd=' '.join(sys.argv),
+        fig_kwds={'dpi': 200}
+    )
+    sys.exit()
+
```
GarethCabournDavies, line 164: This can be removed if wanted - it is basically a safety for if there aren't many triggers - this is a consequence of applying the `--min-z`

examples/gw150914/PyCBCInspiral.ipynb
```diff
@@ -909,14 +909,14 @@
...
+    "    'H1'\n",
+    "    'bank_file=H1L1-GW150914_BANK-1126051217-3331800.hdf',\n",
     ")\n",
     "\n",
     "l1_triggers = pycbc.io.hdf.SingleDetTriggers(\n",
     "    'L1-INSPIRAL_FULL_DATA_JOB0-1126258302-1591.hdf',\n",
-    "    'H1L1-GW150914_BANK-1126051217-3331800.hdf',\n",
-    "    None, None, None, 'L1'\n",
+    "    'L1'\n",
+    "    'bank_file=H1L1-GW150914_BANK-1126051217-3331800.hdf',\n",
```
GarethCabournDavies, line 919: I'm not actually sure this will work to directly change this like this

GarethCabournDavies:
Okay, I think this is ready to review now

pycbc/io/hdf.py
```diff
@@ -500,54 +618,115 @@ def get_param_names(cls):
...
-        be considered."""
+        single detector events as ranked by ranking statistic.
+
+        Events are clustered so that no more than 1 event within +/-
+        cluster_window will be considered. Can apply a threshold on the
+        ranking using ranking_threshold
+        """
 
-        # If this becomes memory intensive we can optimize
         sds = rank_method.single(self.trig_dict())
```
spxiwh, line 676: Could/should these lines be computed using `trigs_f.select` as well? Would be good to avoid reading in all trigger's information if we don't need to in this method.

spxiwh (commented):
Thanks Gareth. There's some comments/suggestions from me here, but I would like to have @ahnitz 's input before merging this. In general no big worries here, just some aesthetics, and a couple of ideas to help future proof.

bin/plotting/pycbc_plot_singles_vs_params
```diff
@@ -84,30 +84,46 @@ opts = parser.parse_args()
 
 logging.basicConfig(format='%(asctime)s %(message)s', level=logging.INFO)
 
-data_mask = None
-if opts.min_snr > 0:
-    with pycbc.io.HFile(opts.single_trig_file, 'r') as trig_file:
+if opts.min_snr:
```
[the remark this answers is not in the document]
GarethCabournDavies: I left this in as I could see the situation where applying a minimum SNR would be different to applying a minimum z_var value (e.g. when psdvar is in play) so thought it best to leave both in.

I can simply get rid of min_snr if that is wanted

pycbc/io/hdf.py
```diff
@@ -59,49 +76,99 @@ def select(self, fcn, *args, **kwds):
...
+        # remove any duplicates from req_dsets
+        dsets = list(set(dsets))
+
+        # Get the pointers to the h5py Datasets,
+        # check they can all be used together
+        refs = {}
+        size = None
+        group = kwds.get('group', '')
+        for ds in dsets:
+            refs[ds] = self[group + '/' + ds]
```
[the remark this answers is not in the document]
GarethCabournDavies: It doesn't, this itself was a way to avoid a problem which came up if using `self[group][ds]`

pycbc/io/hdf.py
```diff
@@ -45,9 +45,26 @@ def select(self, fcn, *args, **kwds):
         chunksize : {1e6, int}, optional
             Number of elements to read and process at a time.
 
+        derived : dictionary
```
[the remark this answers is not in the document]
GarethCabournDavies: I have specified the arguments

pycbc/io/hdf.py
```diff
@@ -59,49 +76,99 @@ def select(self, fcn, *args, **kwds):
...
 
+        return_tuple = tuple()
         # Combine the partial results into full arrays
-        if len(args) == 1:
-            res = np.concatenate(data[args[0]])
-            if return_indices:
-                return indices.astype(np.uint64), res
-            else:
-                return res
+        if indices_only or return_indices:
```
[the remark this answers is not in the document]
GarethCabournDavies: Okay, so in order to get this to consistently return the same number of outputs, I think that there will be two outputs: the indices, and the tuple of datasets, which could be empty if return_data is False

So I will remove the return_indices option

pycbc/io/hdf.py
```diff
@@ -59,49 +76,99 @@ def select(self, fcn, *args, **kwds):
...
+            new_mask = np.zeros(size, dtype=bool)
+            new_mask[mask] = True
+            mask = new_mask
+
+        if not mask.size == size:
+            raise RuntimeError(f"Using premask of size {mask.size} which "
+                               f"does not match the input datasets ({size}).")
+
+        # This will be the outputs:
+        return_indices = kwds.get('return_indices', False)
```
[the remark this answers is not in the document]
GarethCabournDavies: See the below comment

bin/plotting/pycbc_plot_singles_vs_params
```diff
@@ -123,6 +139,29 @@ if opts.max_y is not None:
 x = x[mask]
 y = y[mask]
 
+title = '%s of %s triggers over %s and %s' % (opts.z_var.title(),
+                         opts.detector, opts.x_var.title(), opts.y_var.title())
```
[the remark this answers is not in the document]
GarethCabournDavies: Fixed, (as is line below)

pycbc/io/hdf.py
```diff
@@ -402,47 +469,84 @@ class SingleDetTriggers(object):
...
+        # them in the order which cuts most things earliest.
+        self.mask = premask
+
+        if filter_rank:
+            assert filter_threshold is not None
+            logging.info("Applying threshold of %.3f on %s",
+                         filter_threshold, filter_rank)
+            idx = self.trigs_f.select(
+                 lambda rank: rank > filter_threshold,
+                 derived={ranking.sngls_ranking_function_dict[filter_rank]:
```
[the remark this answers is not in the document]
GarethCabournDavies: That would lose some functionality compared to what is in there at the moment. As it would then restrict us to the ranking functions only. We could use a tuple as the value, which defines the function and the datasets required, so we are not hard-coded only to those

e.g.
```
indices, tuple_out = trig_f.select(
    a_function_which_uses_end_time_and_newsnr_sgveto,
    'end_time',
    'newsnr_sgveto'
    group=opts.detector
    derived={'newsnr_sgveto': (ranking.sngls_ranking_function_dict['newsnr_sgveto'], 
                               ranking.required_datasets['newsnr_sgveto'])
)
```
which would return a tuple with 'end_time' and 'newsnr_sgveto'

pycbc/io/hdf.py
```diff
@@ -402,47 +469,84 @@ class SingleDetTriggers(object):
...
-
-            idx = np.flatnonzero(self.mask)[self.veto_mask]
-            self.mask[:] = False
-            self.mask[idx] = True
-            logging.info('%i triggers remain after vetoes',
-                          len(self.veto_mask))
-
-        # FIXME this should use the hfile select interface to avoid
-        # memory and processing limitations.
         if filter_func:
```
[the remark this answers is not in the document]
GarethCabournDavies: `pycbc_plot_singles_vs_params`, `pycbc_process_sngls` and `pycbc_fit_sngl_trigs` all use the `filter_func` option (from git grepping), I don't know whether all these are still needing to use it though

I think this could be sent to the select() function, but wanted to sort other parts first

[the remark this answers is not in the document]
GarethCabournDavies: One complication here is where the filter_func uses derived datasets, or template bank parameters

examples/gw150914/PyCBCInspiral.ipynb
GarethCabournDavies: I had a look and it was wrong - however the notebook instructions are to use a specific version of pycbc (v1.7.5), and the notebook doesn't work with the current master for other reasons.

So I think it may be best for me to completely remove any changes to this file and not worry about it

pycbc/events/ranking.py
```diff
@@ -299,6 +299,21 @@ def get_newsnr_sgveto_psdvar_scaled_threshold(trigs):
     'newsnr_sgveto_psdvar_scaled_threshold': get_newsnr_sgveto_psdvar_scaled_threshold,
 }
 
+# Lists of datasets required in the trigs object for each function
+required_datasets = {}
```
ahnitz, line 303: Does it make this linked to the function themselves? I'm just wondering if this is really maintainable.

ahnitz:
@GarethCabournDavies I don't see too much to comment on anymore,, except the one thing. Also can you comment or resolve the issues above? It's not quite clear what is still be working on and what the choice made was in response.

GarethCabournDavies:
> @GarethCabournDavies I don't see too much to comment on anymore,, except the one thing. Also can you comment or resolve the issues above? It's not quite clear what is still be working on and what the choice made was in response.

I was leaving them open until I have finished testing and committed, but will resolve as soon as possible

pycbc/events/ranking.py
GarethCabournDavies: These are not linked, the dictionary will need updating as more functions are added into this module.

One way to add this would be if we define the dictionary before the functions, and all functions have an assert that the required_datasets are in trigs. Then if another function is made (likely copied and adapted), then people will naturally add the list into this dictionary. This doesn't make it more maintainable, but does add a psychological incentive to add the list

GarethCabournDavies: I've added the check, so anyone using `get_sngls_ranking_from_trigs` without adding the required datasets will have a problem

pycbc/io/hdf.py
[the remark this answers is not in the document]
GarethCabournDavies: I have made this change

bin/plotting/pycbc_plot_singles_vs_params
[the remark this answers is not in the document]
GarethCabournDavies: removed

GarethCabournDavies:
Noticed an error in the way pre-masks are handled alongside the new filter_rank in SingleDetTriggers (it's not actually used anywhere in the codebase, but just to be safe), I think I have the fix ready, but this is not to merge before then

pycbc/events/ranking.py
GarethCabournDavies: I've had to abandon this idea, as the `get_sngls_ranking_from_trigs` function is used by quite a few different input classes for `trigs`, so making the check general enough was over-complicated (in my mind at least)

pycbc/io/hdf.py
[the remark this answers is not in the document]
GarethCabournDavies: I'll look at this after this has been merged - there has already been quite a bit of scope creep of this PR

GarethCabournDavies:
Fix to the above is in, so please approve if you are happy

tdent:
Where are we with this, looks like just waiting for review?

GarethCabournDavies:
> Where are we with this, looks like just waiting for review?

Yup - the last update was in the day or so before christmas break, so not many working days to have had a look at this

GarethCabournDavies:
Poke for review on this

spxiwh (approved):
I'm happy with the changes now. I think this would also require sign-off from @ahnitz  before it could be merged.

spxiwh:
... And there's now conflicts.

ahnitz (approved):
Sorry for the late reply, yes, I think this has addressed the issues I originally raised, and I don't see anything too major. Always cleanup can be done, but I think this leaves it in a better state than it was.

PR #4546, opened 2023-10-30
Fix error in aligned_stoch_bank

titodalcanton:
The current version of `pycbc_aligned_stoch_bank` fails when using a variable upper frequency cutoff. This seems to be due to a minor typo in variable names (`m1` instead of `mass1`, etc). This PR fixes the typo and includes some minor cleanup as well.


spxiwh (approved):
Am I right in saying that the bugfix here is changing 's1z' -> 'spin1z'? If so, this all looks fine, but want to make it clear in the history what the bugfix actually was!

PR #4547, opened 2023-10-30
Read DQ flags with finer time resolution

maxtrevor:
The performance of DQ flags is much better if we don't have to pad them to start/end on integer seconds. Previously PyCBC only checked if the start of each second was flagged, with no support for better time resolution.

### DIFF
--- a/bin/all_sky_search/pycbc_calculate_dqflag
+++ b/bin/all_sky_search/pycbc_calculate_dqflag
@@ -26,8 +26,8 @@ parser.add_argument('--flag', type=str, required=True,
                     help="name of the data quality flag")
 parser.add_argument('--ifo', type=str, required=True,
                     help="interferometer to analyze")
-parser.add_argument('--sample-frequency', required=False,
-                    default=128, help="timeseries sampling frequency")
+parser.add_argument('--sample-frequency',
+                    help="timeseries sampling frequency in Hz")
 parser.add_argument("--output-file", required=True,
                     help="name of output file")
 parser.add_argument("--output-channel", required=False,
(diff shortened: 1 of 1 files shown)


bin/all_sky_search/pycbc_calculate_dqflag
```diff
@@ -26,6 +26,8 @@ parser.add_argument('--flag', type=str, required=True,
                     help="name of the data quality flag")
 parser.add_argument('--ifo', type=str, required=True,
                     help="interferometer to analyze")
+parser.add_argument('--sample-frequency', required=False,
+                    default=128, help="timeseries sampling frequency")
```
tdent, line 30: Why 128 as default? Seems a rather large value compared to 1 as used before. Do you need a default anyway?

Also suggest adding units to the help ie `(Hz)`

tdent (changes requested):
Seems simple, one comment/suggestion

bin/all_sky_search/pycbc_calculate_dqflag
maxtrevor: Many of the vetoes I want to use have durations as little as 1/4 of a second. The idq timeseries itself has a sample rate of 1024 so theoretically vetoes can start/end on any 1/1024th of a second. 128 felt like a reasonable setting to pick up on a flag starting quickly without actually checking every 1024th of a second. But I can easily make this configurable.

tdent (approved):
Approved, assuming that it works

maxtrevor:
Removing the default option exposed a bug when the sample frequency was interpreted as a string by default. Had to add a type specification to the command line option

PR #4548, opened 2023-10-30
Non-coinc ifos in minifollowup trig plots

tdent:
Up to now non-coinc ifos have been ignored in terms of actually plotting their triggers around an event. Also there was a strange bug which led to a yellow star actually being plotted for these ifos (I guess with the SNR value of the last trigger in the file, given that the ID passed was -1). Use the mean coinc time to determine triggers in window and don't try to plot any special coinc trigger.


spxiwh (approved):
Looks good to me. Thanks.

PR #4549, opened 2023-10-31
Use gating windows in sngl_minifollowup and allow 'triggers within vetoes' option

GarethCabournDavies:
When we have lots of things in the autogating vetoes, these can dominate the singles minifollowups.

This allows us to provide the gating vetoes to this job as elsewhere in order to remove them.

However they are still informative (e.g. we may have an autogated loud signal!). So we also have a minifollowup with `--vetoed-time-only` option.

The main changes here are:
 - Add the gating vetoes options and implement
 - Allow a `--vetoed-time-only` option
 - Allow passing of both veto file and foreground-censor file, this is as we do not want foreground-censored things to show up in a 'vetoed time only' section
    - Update the offline workflow to use this specification as appropriate

### DIFF
--- a/bin/minifollowups/pycbc_page_snglinfo
+++ b/bin/minifollowups/pycbc_page_snglinfo
@@ -138,8 +138,11 @@ else:
 
 if args.ranking_statistic in ["quadsum", "single_ranking_only"]:
     stat_name = sngl_stat_name
+    stat_name_long = sngl_stat_name
 else:
-    stat_name = '_with_'.join(ranking_statistic, sngl_ranking)
+    # Name would be too long - just call it ranking statistic
+    stat_name = 'Ranking Statistic'
+    stat_name_long = ' with '.join([args.ranking_statistic, args.sngl_ranking])
 
 headers.append(stat_name)
 
@@ -201,7 +204,7 @@ html = pycbc.results.dq.redirect_javascript + \
 if args.n_loudest:
     title = 'Parameters of single-detector event ranked %s' \
         % (args.n_loudest + 1)
-    caption = 'Parameters of the single-detector event ranked number %s by %s. The figures below show the mini-followup data for this event.' % (args.n_loudest + 1, stat_name)
+    caption = 'Parameters of the single-detector event ranked number %s by %s. The figures below show the mini-followup data for this event.' % (args.n_loudest + 1, stat_name_long)
 else:
     title = 'Parameters of single-detector event'
     caption = 'Parameters of the single-detector event. The figures below show the mini-followup data for this event.'
--- a/bin/minifollowups/pycbc_sngl_minifollowup
+++ b/bin/minifollowups/pycbc_sngl_minifollowup
@@ -107,6 +107,12 @@ sngl_file = resolve_url_to_file(
     attrs={'ifos': args.instrument}
 )
 
+# Flatten the statistic_files option:
+statfiles = []
+for f in sum(args.statistic_files, []):
+    statfiles.append(resolve_url_to_file(os.path.abspath(f)))
+statfiles = wf.FileList(statfiles) if statfiles is not [] else None
+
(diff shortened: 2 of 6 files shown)


GarethCabournDavies:
(I've added the v23 release branch label, but this is not vital, more a helpful addition)

GarethCabournDavies:
An example (behind LVK paywall) with the O4a chunk 2 test is [here](https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/veto_sngl_minifollowup/a2_TEST/4._single_triggers/4.11_H1_loudest_vetoed_triggers/)

maxtrevor:
This looks good to me, approving

tdent:
It looks like it works, but I am not quite convinced that this will be what we want to do, ie will it allow us to catch a moderately loud signal whose merger happens to lie in a gating veto. 
The 'loudest vetoed' sections in the example (in both ifos) list things which even at event number 10 are still pretty loud, and they _all_ have really high masses.  If there was (say) a moderately loud _low_ mass signal with a merger 1-2s after a gating window, it might not make it to this list.

tdent:
Maybe what would mitigate this would be to apply some more intelligent statistic ranking (ie loudest _by what statistic_ - IMO it should ideally be the one used for detecting singles) to the 'loudest vetoed' section

GarethCabournDavies:
That's a fair point, and I think having the ranking-statistic option with `quadsum` as default (i.e. pass sngl-ranking straight through), and calculating the single-detector statistic should work for that. I've removed the release branch tag and will remove the request on the cherry-pick PR as well

tdent:
OK, so your example used `newsnr_sgveto_psdvar_threshold` ranking .. maybe activating the actual sngls detection statistic needs a bit more code plumbing then.

tdent:
FWIW, I don't think there is a reason to not merge this now, unless you want to mark it as draft and add more actual code changes later.

GarethCabournDavies:
I am already very close to getting the ranking-statistic changes in, so it may be worth waiting, though I will mark as draft for now

GarethCabournDavies:
So I was completely done when I said 'very close' before. But it turned out there was a bug when calling `rank_stat_single` with the quadsum statistic, this took me longer than I care to admit to find, so the tests were all failing. Workflow running now, and I will link it and mark as ready when it is done

GarethCabournDavies:
A couple of minor associated fixes, and we have results again:

https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/veto_sngl_minifollowup/a2_TEST_2/

GarethCabournDavies:
See https://github.com/GarethCabournDavies/pycbc/compare/c53e31d...sngls_minifollow_gating_veto#diff-19e7b672a4bc59c6b8bd7cbebc7c72b7424a3c9ee3a884739ea23cadbbb1ecd4R200 for changes since the last review

GarethCabournDavies:
I think this is ready now - I'll be away next week, so please merge if we are happy with it, and apply the release branch tag if we are going to use it in O4 production runs

GarethCabournDavies:
Any comments since last review? If not I will merge

GarethCabournDavies:
No comments against this, so I will merge

tdent:
So did we agree that this is now not a risk for missing loudish signals with merger inside a gating veto window?

GarethCabournDavies:
These would show up in the 'within vetoes' followup page, so long as that is specified in the config files

tdent:
OK, is a MR required for that?

GarethCabournDavies:
Added one for the CI here: #4597

There is also one for O4 production runs in the repo for that

tdent:
Ah, I see one there (currently in draft state I think)

tdent:
Ultimately we should replace gating veto with re-ranking around a gated time, ie gating DQ segment .. that is under development

PR #4550, opened 2023-10-31
PyGRB: Update postprocessing functions

jakeb245:
This PR is another step in resolving #4419. It should finish off converting the backend parts of the old XML functionality into HDF. The functionality is intended to be the same, but there could be some cases in here where it should be changed.


pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -460,36 +460,36 @@ def sort_trigs(trial_dict, trigs, slide_dict, seg_dict):
...
-    # New seems pretty slow, so run it once and then use deepcopy
-    tmp_table = glsctables.New(glsctables.MultiInspiralTable)
     for slide_id in slide_dict:
-        sorted_trigs[slide_id] = copy.deepcopy(tmp_table)
-    for trig in trigs:
-        sorted_trigs[int(trig.time_slide_id)].append(trig)
+        sorted_trigs[slide_id] = []
+        for i in range(trigs['network/event_id'].len()):
+            slide_id = trigs['network/slide_id'][i]
+            sorted_trigs[slide_id].append(i)
```
pannarale, line 467: Using `slide_id` to run over `slide_dict` and assign `trigs['network/slide_id'][i]` makes these lines of code hard to parse.

```suggestion
        for i, _ in enumerate(trigs['network/event_id']):
            idx = trigs['network/slide_id'][i]
            sorted_trigs[idx].append(i)
```

pannarale (dismissed):

bin/pycbc_multi_inspiral
```diff
@@ -629,6 +633,8 @@ with ctx:
                         network_out_vals['dec'] = [sky_positions[1][position_index]] * num_events
                         network_out_vals['ra'] = [sky_positions[0][position_index]] * num_events
                         network_out_vals['slide_id'] = [slide] * num_events
+                        network_out_vals['mass1'] = [template.params['mass1']] * num_events
+                        network_out_vals['mass2'] = [template.params['mass2']] * num_events
```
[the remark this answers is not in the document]
jakeb245: The template ID is stored already. Storing the masses is handy for the chirp mass calculation which happens down the line. The old code stored the template ID as well as the masses. I believe it also wrote the spins for each trigger, which I did not find the need for. This might be easier than adding extra dependencies on the template bank file.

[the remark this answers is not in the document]
pannarale: I don't see the template ID listed in `network_out_vals`.  If we want to avoid the dependency on the template bank, then we should copy all `template.params`.  Otherwise this will be outdated if the template parameters are expanded (e.g., say we have a template bank with tidal deformabilities or precession: then we have to go back and expand this to grab lambda's and spin components).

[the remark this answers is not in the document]
jakeb245: The template ID is written elsewhere (looks like it's done [here](https://github.com/gwastro/pycbc/blob/master/pycbc/events/eventmgr.py#L653)). I've confirmed it shows up in the output files as well.

titodalcanton:
@jakeb245 yes, I would use a similar approach used by `pycbc_inspiral` in the all-sky all-time search, which indeed consists in storing just the template hash for each trigger after matched filtering. To be precise, the following happens there:

1. `pycbc_inspiral` does matched filtering for each bank split separately, and the corresponding triggers store the template hashes only.
2. `pycbc_coinc_mergetrigs` then takes all the triggers from all the bank splits from step 1, merges them together, and uses the template hashes to recover the template IDs into the full bank.

Have a look at this documentation page for more info: https://pycbc.org/pycbc/latest/html/formats/hdf_format.html. You can also see an example of this by looking at the various files from one of the all-sky all-time O4 chunks, in particular you want to look at one of the `full_data/*-INSPIRAL_BANK*.hdf` files for step 1, and `full_data/*-HDF_TRIGGER_MERGE_FULL_DATA*.hdf` for step 2.

Ping me on Slack if more is needed.

bin/pygrb/pycbc_grb_inj_finder
```diff
@@ -329,12 +337,21 @@ with h5py.File(outfilename, "w") as h5out:
     for x in out_trigs:
         xgroup = h5out.create_group(x)
         for col in out_trigs[x]:
+            if "template_id" in col:
```
titodalcanton, line 340: Should this not be `if col == "template_id":`? Or does `col` contain other characters in addition?

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -207,6 +210,11 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
                     else:
                         ifo_event_id_pos = None
 
+                    # Template ID becomes unhelpful in the merged file.
+                    # This will be handled separately.
+                    if "template_id" in dset:
```
titodalcanton, line 215: Same comment as before, should this not be an exact match?

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -411,7 +411,7 @@ if post_proc_method == "PYGRB_OFFLINE":
     # The content and structure of pp_files are described in the definition
     # of _workflow.setup_pygrb_pp_workflow
     pp_files = _workflow.setup_pygrb_pp_workflow(wflow, pp_dir, seg_dir,
-                                      sciSegs[ifo][0], inspiral_files,
+                                      sciSegs[ifo][0], bank_files[0], inspiral_files,
```
titodalcanton, line 414: Is the first item of `bank_files` always guaranteed to be the full bank? Or is `bank_files` always going to contain a single item? This is not obvious from the code. A comment would be useful, or maybe a more self-explanatory choice of variables? Do not bother if too complicated though.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -461,84 +462,64 @@ def sort_trigs(trial_dict, trigs, slide_dict, seg_dict):
...
 
     # Begin by sorting the triggers into each slide
-    # New seems pretty slow, so run it once and then use deepcopy
-    tmp_table = glsctables.New(glsctables.MultiInspiralTable)
     for slide_id in slide_dict:
-        sorted_trigs[slide_id] = copy.deepcopy(tmp_table)
-    for trig in trigs:
-        sorted_trigs[int(trig.time_slide_id)].append(trig)
+        sorted_trigs[slide_id] = []
+    for i, _ in enumerate(trigs['network/event_id']):
```
titodalcanton, line 467: Could this be instead
```python
for i in range(len(trigs['network/event_id'])):
```

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -461,84 +462,64 @@ def sort_trigs(trial_dict, trigs, slide_dict, seg_dict):
...
-        trig_bestnr[slide_id] = get_bestnrs(slide_trigs,
-                                            q=chisq_index,
-                                            n=chisq_nhigh,
-                                            null_thresh=null_thresh,
-                                            snr_threshold=snr_thresh,
-                                            sngl_snr_threshold=sngl_snr_thresh,
-                                            chisq_threshold=new_snr_thresh,
-                                            null_grad_thresh=null_grad_thresh,
-                                            null_grad_val=null_grad_val)
+        trig_bestnr[slide_id] = trigs['network/reweighted_snr'][slide_trigs]
```
titodalcanton, line 522: Can you confirm that the BestNR calculation happens somewhere else (earlier) in the new workflow? I.e. if I wanted to change `opts.newsnr_threshold`, where would that be applied?

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -804,3 +785,24 @@ def get_coinc_snr(trigs_or_injs, ifos):
...
+    This function converts the template hashes from a trigger file
+    into 'template_id's that represent indices of the
+    templates within the bank.
+    Parameters
+    ----------
+    trigger_file: h5py File object for trigger file
+    bank_file: filepath for template bank
+    """
+    bank = h5py.File(bank_path, "r")
+    hashes = bank['template_hash'][:]
```
titodalcanton, line 801:
```suggestion
    with h5py.File(bank_path, "r") as bank:
        hashes = bank['template_hash'][:]
```
so you close the bank file as soon as you are done with it.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -804,3 +785,24 @@ def get_coinc_snr(trigs_or_injs, ifos):
...
+    ifos = [k for k in trigger_file.keys() if k != 'network']
+    trig_hashes = trigger_file[f'{ifos[0]}/template_hash'][:]
```
titodalcanton, line 803: Why is the `template_hash` dataset stored inside the ifo-specific groups, instead of the `network` group? Is that not unnecessarily duplicating the same dataset for each ifo?

pycbc/results/pygrb_postprocessing_utils.py
titodalcanton: I am a bit puzzled by this code though. Imagine that `trigs['network/event_id']` contains [2,3,0,1]. This would mean that `i` would go through the list [0,1,2,3] instead. Is that the correct thing to do? Somewhat confusing that this code does not appear to actually use the values in `event_id`.

titodalcanton:
@jakeb245 I had a close look at this PR. As you know I am a bit removed from the development of this code, so I left a bunch of questions above to fill some gaps in my understanding, but I do not see anything too concerning here. I am happy to approve this once you have answered the comments.

bin/pygrb/pycbc_grb_inj_finder
jakeb245: `col` is the full column header, so it includes the IFO/network tags.

bin/pygrb/pycbc_make_offline_grb_workflow
jakeb245: Good point. If we were to run with two bank files this would become an issue. If there's only one bank then the `bank_files` object should just the single file. I suppose it would be worth having the workflow generator allow for multiple bank files. I don't think it would be complicated to do that, but that case would cause problems with the template hash/ID mapping.

pycbc/results/pygrb_postprocessing_utils.py
jakeb245: Yes, currently the reweighted SNR is calculated in `pycbc_multi_inspiral` at around [here](https://github.com/gwastro/pycbc/blob/e2c35eff8367fa4700588a60f80737949d8fd254/bin/pycbc_multi_inspiral#L613). The `newsnr_threshold` option is also set in that executable.

pycbc/results/pygrb_postprocessing_utils.py
jakeb245: I'm not sure. The writing to file is done [here](https://github.com/gwastro/pycbc/blob/e2c35eff8367fa4700588a60f80737949d8fd254/pycbc/events/eventmgr.py#L787). The line that makes me unsure is [the definition of the index](https://github.com/gwastro/pycbc/blob/e2c35eff8367fa4700588a60f80737949d8fd254/pycbc/events/eventmgr.py#L719). It seems dependent on the ifo, but one would expect them to be the same.

pycbc/results/pygrb_postprocessing_utils.py
jakeb245: I think working with direct indices would be faster rather than having to do something like `trigs['network/reweighted_snr']['network/event_id' == event_id]`. I'm thinking it would be more intuitive to have this function output the `event_id` though, so in this case I think I'll change that.

pycbc/results/pygrb_postprocessing_utils.py
titodalcanton: Yeah, what I see in `eventmgr.py` does not look correct, or at least not efficient. All template properties should be identical across interferometers. Maybe we do not need to worry about this in this PR, but it should be addressed shortly after this is merged. Please bring this up at the development call.

bin/pygrb/pycbc_grb_inj_finder
titodalcanton: Are you sure? In the code above I see
```python
for x in triggers:
    ij = x.split('/')
    out_trigs[ij[0]][ij[1]].append(triggers[x][eid])
```
so it seems to me that `out_trigs` is a dict-of-dict type object, where the outer key is the interferometer or "network", and the inner key is the field name.

bin/pygrb/pycbc_make_offline_grb_workflow
titodalcanton: I do not see what use case we might want to support where there are two banks in a single workflow. Not in the immediate future, at least.

pycbc/results/pygrb_postprocessing_utils.py
titodalcanton: Not sure I follow. Presumably the code using the actual `event_id`s would be as simple as
```python
for i in trigs['network/event_id']:
```
Is that not the case?

bin/pygrb/pycbc_make_offline_grb_workflow
titodalcanton: I would do the "Multiple banks not supported" check early on, immediately after the definition of `bank_files` in `pycbc_make_offline_grb_workflow`. Then define `full_bank_file = bank_files[0]`. In that way you make it obvious from the code that there is a single full bank file.

pycbc/results/pygrb_postprocessing_utils.py
jakeb245: Good point. The only time the index gets used is to pull the `slide_id`. This operation is likely faster, but probably not worth using `enumerate` for.

bin/pygrb/pycbc_grb_inj_finder
jakeb245: Oh right, I was thinking about `pycbc_grb_trig_combiner`. You're correct!

pycbc/results/pygrb_postprocessing_utils.py
titodalcanton: I looked a bit more at the `eventmrg.py` code and the situation is more confusing than I thought, so I opened https://github.com/gwastro/pycbc/issues/4632 to track that issue.

pycbc/results/pygrb_postprocessing_utils.py
titodalcanton: Ah, I think I understand the goal now. Each trigger has an entry in the `event_id` and `slide_id` datasets (at the same index in the two) and you want to group the triggers by `slide_id`. So what you have now is probably very slow, but how about this:
```python
    for slide_id, event_id in zip(trigs['network/slide_id'], trigs['network/event_id']):
        sorted_trigs[slide_id].append(event_id)
```
Also, note that in a couple places you do `'network/event_id' == event_id` which does not make sense, you probably meant `trigs['network/event_id'] == event_id`.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -461,84 +462,67 @@ def sort_trigs(trial_dict, trigs, slide_dict, seg_dict):
...
     for slide_id in slide_dict:
         # These can only *reduce* the analysis time
         curr_seg_list = seg_dict[slide_id]
 
         # Check the triggers are all in the analysed segment lists
-        for trig in sorted_trigs[slide_id]:
-            if trig.end_time not in curr_seg_list:
+        for event_id in sorted_trigs[slide_id]:
+            end_time = trigs['network/end_time_gc'][
+                'network/event_id'][0] == event_id
```
titodalcanton, line 478: Not sure this will work, I think what you need is
```suggestion
            index = numpy.flatnonzero(trigs['network/event_id'] == event_id)[0]
            end_time = trigs['network/end_time_gc'][index]
```

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -461,84 +462,67 @@ def sort_trigs(trial_dict, trigs, slide_dict, seg_dict):
...
 
-        # The below line works like the inverse of .veto and only returns trigs
-        # that are within the segment specified by trial_dict[slide_id]
-        sorted_trigs[slide_id] = \
-            sorted_trigs[slide_id].vetoed(trial_dict[slide_id])
+        # Keep triggers that are in trial_dict
+        sorted_trigs[slide_id] = [event_id for event_id in
+                                  sorted_trigs[slide_id]
+                                  if trigs['network/end_time_gc'][
+                                      'network/event_id'][0] == event_id
```
titodalcanton, line 496: Same as above, if I understand the involved data structures correctly, this should produce an error. This should be what you want:
```suggestion
                                  if trigs['network/end_time_gc'][
                                      trigs['network/event_id'] == event_id][0]
```
The idea is that `trigs['network/event_id'] == event_id` will give you a boolean array of the same shape as `trigs['network/event_id']`, with a single True at the position of the right event. You then use that array to index `trigs['network/end_time_gc']`, which will give you a single-entry array of the end time. Then `[0]` picks out the only entry as a scalar.

titodalcanton:
I think there are a couple things to fix (noted above), then I think this can go in.

pycbc/results/pygrb_postprocessing_utils.py
jakeb245: This is definitely what I was intending to do. Thanks for catching that!

titodalcanton (approved):

titodalcanton:
I am still not super convinced that I understand the logic of the slide_id/event_id thing here, so whoever is developing that functionality should keep it in mind. But let's merge this and continue developing.

Issue #4555, opened 2023-11-02
SNR opt differential evolution init and verbosity


tdent:
The init arrays should be actual parameter values (randomized within bounds), not 0-1 random variables.  The bad consequence of feeding in 0-1 variables is that for almost all events the chirp mass is stuck to the lower bound (due to the scipy init function's clipping) for the entire minimization, i.e. the output is forced to use a bad chirp mass value.  (Also the number of independent eta and chi_1,2 points is greatly reduced, but this is less critical.) 

@PRAVEEN-mnl and I found this by printing out the parameter values and SNR at every call to the network SNR function, and I think there should also be a 'debug' level of verbosity to do this (for every choice of optimizer) as the SNR optimizer as a whole seems quite liable to breakage.

GarethCabournDavies:
Is this solved by the merged MR?

tdent:
Looks like it.

PR #4561, opened 2023-11-03
added scaling of initial pop in snr_optimizer

PRAVEEN-mnl:
We added the init population based on actual parameter values and randomized within bounds earlier it was 0-1 random variables. #4555

### DIFF
--- a/pycbc/live/snr_optimizer.py
+++ b/pycbc/live/snr_optimizer.py
@@ -166,7 +166,7 @@ def compute_minus_network_snr(v, *argv):
     if len(argv) == 1:
         argv = argv[0]
     nsnr, _ = compute_network_snr_core(v, *argv)
-    logging.info('snr: %s', nsnr)
+    logging.debug('snr: %s', nsnr)
     return -nsnr
 
 
--- a/pycbc/live/snr_optimizer.py
+++ b/pycbc/live/snr_optimizer.py
@@ -181,6 +181,23 @@ def normalize_initial_point(initial_point, bounds):
     return (initial_point - bounds[:,0]) / (bounds[:,1] - bounds[:,0])
 
 
+def init_population_array(init, parameter_count):
+    # Make sure you're using a float array
+    popn = numpy.asarray(init, dtype=numpy.float64)
+
+    if (numpy.size(popn, 0) < 5 or
+            popn.shape[1] != parameter_count or
+            len(popn.shape) != 2):
+        raise ValueError("The population supplied needs to have shape (S, len(x)), where S > 4.")
+
+    # Scale values to be between 0 and 1
+    population = (popn - popn.min(axis=0)) / (popn.max(axis=0) - popn.min(axis=0))
+    num_population_members = numpy.size(population, 0)
+    population_shape = (num_population_members, parameter_count)
+
+    return population, num_population_members, population_shape
+
+
 def optimize_di(bounds, cli_args, extra_args, initial_point):
     bounds = numpy.array([
         bounds['mchirp'],
@@ -190,9 +207,9 @@ def optimize_di(bounds, cli_args, extra_args, initial_point):
     ])
 
     # Currently only implemented for random seed initial array
-    rng = numpy.random.mtrand._rand
-    population_shape = (int(cli_args.snr_opt_di_popsize), 4)
-    population = rng.uniform(size=population_shape)
+    init = numpy.random.rand(cli_args.snr_opt_di_popsize, 4)
+    population, num_population_members, population_shape = init_population_array(init, 4)
+
     if cli_args.snr_opt_include_candidate:
         # Re-normalize the initial point into the correct range
         point_init = normalize_initial_point(initial_point, bounds)
--- a/pycbc/live/snr_optimizer.py
+++ b/pycbc/live/snr_optimizer.py
@@ -166,6 +166,7 @@ def compute_minus_network_snr(v, *argv):
     if len(argv) == 1:
         argv = argv[0]
     nsnr, _ = compute_network_snr_core(v, *argv)
+    logging.info('snr: %s', nsnr)
     return -nsnr
 
 
@@ -177,27 +178,6 @@ def compute_minus_network_snr_pso(v, *argv, **kwargs):
     return -nsnr_array
 
 
-def normalize_initial_point(initial_point, bounds):
-    return (initial_point - bounds[:,0]) / (bounds[:,1] - bounds[:,0])
-
-
-def init_population_array(init, parameter_count):
-    # Make sure you're using a float array
-    popn = numpy.asarray(init, dtype=numpy.float64)
-
-    if (numpy.size(popn, 0) < 5 or
-            popn.shape[1] != parameter_count or
-            len(popn.shape) != 2):
-        raise ValueError("The population supplied needs to have shape (S, len(x)), where S > 4.")
-
-    # Scale values to be between 0 and 1
-    population = (popn - popn.min(axis=0)) / (popn.max(axis=0) - popn.min(axis=0))
-    num_population_members = numpy.size(population, 0)
-    population_shape = (num_population_members, parameter_count)
-
-    return population, num_population_members, population_shape
-
-
 def optimize_di(bounds, cli_args, extra_args, initial_point):
     bounds = numpy.array([
         bounds['mchirp'],
@@ -206,15 +186,15 @@ def optimize_di(bounds, cli_args, extra_args, initial_point):
         bounds['spin2z']
     ])
 
-    # Currently only implemented for random seed initial array
-    init = numpy.random.rand(cli_args.snr_opt_di_popsize, 4)
-    population, num_population_members, population_shape = init_population_array(init, 4)
-
+    # Initialize the population with random values within specified bounds
+    population = numpy.random.uniform(bounds[:, 0], bounds[:, 1],
+        size=(int(cli_args.snr_opt_di_popsize), len(bounds)))
+ 
     if cli_args.snr_opt_include_candidate:
-        # Re-normalize the initial point into the correct range
-        point_init = normalize_initial_point(initial_point, bounds)
         # add the initial point to the population
-        population = numpy.concatenate((population[:-1], point_init))
+        population = numpy.concatenate((population[:-1],
+                                        initial_point))
+    logging.info('population: %s', population)
 
     results = differential_evolution(
         compute_minus_network_snr,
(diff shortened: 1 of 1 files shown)


pycbc/live/snr_optimizer.py
```diff
@@ -181,6 +181,23 @@ def normalize_initial_point(initial_point, bounds):
     return (initial_point - bounds[:,0]) / (bounds[:,1] - bounds[:,0])
 
 
+def init_population_array(init, parameter_count):
```
tdent, line 184: Why are you defining a function here and only using it once?  If the same basic method is used for both the pso optimizer (optimize_pso) and for di, can't we have a function that is used by both of them?

pycbc/live/snr_optimizer.py
```diff
@@ -181,6 +181,23 @@ def normalize_initial_point(initial_point, bounds):
...
+    # Make sure you're using a float array
+    popn = numpy.asarray(init, dtype=numpy.float64)
+
+    if (numpy.size(popn, 0) < 5 or
```
tdent, line 188: I don't know why you need all these checks when the function is only used once or twice.

pycbc/live/snr_optimizer.py
```diff
@@ -181,6 +181,23 @@ def normalize_initial_point(initial_point, bounds):
...
+            popn.shape[1] != parameter_count or
+            len(popn.shape) != 2):
+        raise ValueError("The population supplied needs to have shape (S, len(x)), where S > 4.")
+
+    # Scale values to be between 0 and 1
```
tdent, line 193: Why are you doing this?

pycbc/live/snr_optimizer.py
tdent: Also, the psd optimizer already has a function called `normalize_population`, can you not use that?

pycbc/live/snr_optimizer.py
```diff
@@ -181,6 +181,23 @@ def normalize_initial_point(initial_point, bounds):
...
+    population = (popn - popn.min(axis=0)) / (popn.max(axis=0) - popn.min(axis=0))
+    num_population_members = numpy.size(population, 0)
+    population_shape = (num_population_members, parameter_count)
+
+    return population, num_population_members, population_shape
```
tdent, line 198: why is this returning num_population_members and population_shape when these are not needed in the rest of the code?

pycbc/live/snr_optimizer.py
```diff
@@ -190,9 +207,9 @@ def optimize_di(bounds, cli_args, extra_args, initial_point):
...
     # Currently only implemented for random seed initial array
-    rng = numpy.random.mtrand._rand
-    population_shape = (int(cli_args.snr_opt_di_popsize), 4)
-    population = rng.uniform(size=population_shape)
+    init = numpy.random.rand(cli_args.snr_opt_di_popsize, 4)
+    population, num_population_members, population_shape = init_population_array(init, 4)
+
     if cli_args.snr_opt_include_candidate:
         # Re-normalize the initial point into the correct range
         point_init = normalize_initial_point(initial_point, bounds)
```
tdent, line 215: Is this correct in fact?

tdent:
Basically, I think it should be possible to reuse the PSO population init code here rather than writing something separate and more complicated for di. 

Also, I think we would like to have debug statements as I noted in the issue https://github.com/gwastro/pycbc/issues/4555 (ie logging.debug) for the parameter values and SNRs being calculated during the optimization.

pycbc/live/snr_optimizer.py
```diff
@@ -166,6 +166,7 @@ def compute_minus_network_snr(v, *argv):
     if len(argv) == 1:
         argv = argv[0]
     nsnr, _ = compute_network_snr_core(v, *argv)
+    logging.info('snr: %s', nsnr)
```
tdent, line 169: I think we prefer to use logging.debug (for an additional level beyond info)

pycbc/live/snr_optimizer.py
```diff
@@ -256,17 +248,15 @@ def optimize_pso(bounds, cli_args, extra_args, initial_point):
...
-    population = normalize_population(population, min_bounds, max_bounds)
+    # Initialize the population with random values within specified bounds
+    population = numpy.random.uniform(min_bounds, max_bounds,
+        size=(int(cli_args.snr_opt_di_popsize), len(bounds)))
 
     if cli_args.snr_opt_include_candidate:
         # add the initial point to the population
         population = numpy.concatenate((population[:-1],
                                         initial_point))
+    logging.info('population: %s', population)
```
tdent, line 259:
```suggestion
    logging.debug('Initial population: %s', population)
```

pycbc/live/snr_optimizer.py
```diff
@@ -177,10 +178,6 @@ def compute_minus_network_snr_pso(v, *argv, **kwargs):
     return -nsnr_array
 
 
-def normalize_initial_point(initial_point, bounds):
-    return (initial_point - bounds[:,0]) / (bounds[:,1] - bounds[:,0])
-
-
 def optimize_di(bounds, cli_args, extra_args, initial_point):
     bounds = numpy.array([
```
tdent, line 182: Suggest to add a comment here saying why we need this - eg `Convert from dict to array with parameters in a given order`

pycbc/live/snr_optimizer.py
```diff
@@ -189,15 +186,15 @@ def optimize_di(bounds, cli_args, extra_args, initial_point):
...
+        size=(int(cli_args.snr_opt_di_popsize), len(bounds)))
+ 
     if cli_args.snr_opt_include_candidate:
-        # Re-normalize the initial point into the correct range
-        point_init = normalize_initial_point(initial_point, bounds)
         # add the initial point to the population
-        population = numpy.concatenate((population[:-1], point_init))
+        population = numpy.concatenate((population[:-1],
+                                        initial_point))
+    logging.info('population: %s', population)
```
tdent, line 197:
```suggestion
    logging.debug('Initial population: %s', population)
```

pycbc/live/snr_optimizer.py
```diff
@@ -189,15 +186,15 @@ def optimize_di(bounds, cli_args, extra_args, initial_point):
...
     ])
 
-    # Currently only implemented for random seed initial array
-    rng = numpy.random.mtrand._rand
-    population_shape = (int(cli_args.snr_opt_di_popsize), 4)
-    population = rng.uniform(size=population_shape)
+    # Initialize the population with random values within specified bounds
+    population = numpy.random.uniform(bounds[:, 0], bounds[:, 1],
+        size=(int(cli_args.snr_opt_di_popsize), len(bounds)))
+
```
tdent, line 192:
```suggestion

```

pycbc/live/snr_optimizer.py
```diff
@@ -189,15 +186,15 @@ def optimize_di(bounds, cli_args, extra_args, initial_point):
         bounds['spin2z']
     ])
 
-    # Currently only implemented for random seed initial array
-    rng = numpy.random.mtrand._rand
-    population_shape = (int(cli_args.snr_opt_di_popsize), 4)
-    population = rng.uniform(size=population_shape)
+    # Initialize the population with random values within specified bounds
+    population = numpy.random.uniform(bounds[:, 0], bounds[:, 1],
+        size=(int(cli_args.snr_opt_di_popsize), len(bounds)))
```
tdent, line 191:
```suggestion
    population = numpy.random.uniform(
        bounds[:, 0],
        bounds[:, 1],
        size=(int(cli_args.snr_opt_di_popsize), len(bounds))
    )
```

pycbc/live/snr_optimizer.py
```diff
@@ -256,17 +248,15 @@ def optimize_pso(bounds, cli_args, extra_args, initial_point):
...
 
-    # Manually generate the initial points, this is the same as the default
-    # method, but allows us to make some modifications
-    population = numpy.random.uniform(
-        low=0.0, high=1.0, size=(int(cli_args.snr_opt_pso_particles), 4)
-    )
-    population = normalize_population(population, min_bounds, max_bounds)
+    # Initialize the population with random values within specified bounds
+    population = numpy.random.uniform(min_bounds, max_bounds,
+        size=(int(cli_args.snr_opt_di_popsize), len(bounds)))
```
tdent, line 253:
```suggestion
    population = numpy.random.uniform(
        bounds[:, 0],
        bounds[:, 1],
        size=(int(cli_args.snr_opt_di_popsize), len(bounds))
    )
```

pycbc/live/snr_optimizer.py
```diff
@@ -256,17 +248,15 @@ def optimize_pso(bounds, cli_args, extra_args, initial_point):
         bounds['spin2z'][1]
     ])
 
-    # Manually generate the initial points, this is the same as the default
-    # method, but allows us to make some modifications
-    population = numpy.random.uniform(
-        low=0.0, high=1.0, size=(int(cli_args.snr_opt_pso_particles), 4)
```
tdent, line 262: I made a suggestion for the formatting here, but you also have introduced a bug because the size here was controlled by the command line argument for PSO, but you have broken that.

tdent:
See some comments/ suggestions (some of which should deal with codeclimate issues). 
The use of logging.debug will not work at the moment but with https://github.com/gwastro/pycbc/pull/4564 it should be possible to make it work, so we may as well use it now.

Has this change been tested to work with both di and PSO optimizers?

PRAVEEN-mnl:
Yes, the change has been tested to work with both di and PSO optimizers.

tdent:
@ArthurTolley is there anything else you need on this atm?

ArthurTolley:
I will do the review most likely on Monday, if I need anything else then I will let you know 😁

With regards to the description, it is still a bit sparse. Is there a reason why this change was needed? I've seen talk about some snr optimizer bugs but not sure if this is related + how this fixes it. That is the kind of information I would've liked to see in the description @PRAVEEN-mnl.

tdent:
@ArthurTolley evidence of a SNR optimizer bug using di is in the presentations on MDC results from a couple of PyCBC telecons ago. It doesn't make much sense to link LVK internal material here though IMO.

tdent (approved):

ArthurTolley (approved):
All looks good to me, code comments are clean and the change to actual bound values for the random points in the population makes sense.

As long as this has been fully tested with both differential evolution and particle swarm optimization then I am happy with the changes.

(also, sorry for the delay in reviewing this)

PR #4562, opened 2023-11-03
Update PyGRB efficiency script for HDF files

jakeb245:
This contains the changes to `pycbc_pygrb_efficiency` necessary for it to run with HDF files. It relies on (currently open) #4542 and #4550. This PR is part of resolving #4419.


bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -239,18 +241,12 @@ if onsource_file:
...
-                                           snr_threshold=snr_thresh,
-                                           sngl_snr_threshold=sngl_snr_thresh,
-                                           chisq_threshold=new_snr_thresh,
-                                           null_grad_thresh=null_grad_thresh,
-                                           null_grad_val=null_grad_val)
+        on_trigs_bestnrs = on_trigs['network/reweighted_snr'][:]
         loud_on_bestnr_trigs, loud_on_bestnr = \
             np.asarray([[x, y] for y, x in sorted(zip(on_trigs_bestnrs, on_trigs),
                                                   key=lambda elem: elem[0],
                                                   reverse=True)])[0]
```
pannarale, line 248: I don't understand how this would work. You are zipping an array of reweighed SNRs with `on_trigs` which is an object returned by opening and hdf5 file, so it looks like `<KeysViewHDF5 ['H1', 'L1', 'V1', 'network']>`. What you want is to sort the content of each key (['H1', 'L1', 'V1', 'network']) on the basis of the reweighed SNRs.

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -287,35 +283,34 @@ if do_injections:
...
-    found_inj['ra'] = np.asarray(found_injs.get_column('longitude'))
-    found_inj['dec'] = np.asarray(found_injs.get_column('latitude'))
-    found_inj['time'] = np.asarray(found_injs.get_column('geocent_end_time')) +\
-                        np.asarray(found_injs.get_column('geocent_end_time_ns')*\
-                                   10**-9)
-    found_inj['dist'] = np.asarray(found_injs.get_column('distance'))
+    found_inj['ra'] = inj_file['found/ra'][:]
+    found_inj['dec'] = inj_file['found/dec'][:]
+    found_inj['time'] = inj_file['found/tc'][:]
+    found_inj['dist'] = inj_file['found/distance'][:]
```
pannarale, line 295: There is no need to extract columns with hdf5 files.  See comment above.

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -287,35 +283,34 @@ if do_injections:
     sites = [ifo[0] for ifo in ifos]
 
     # Triggers and injections recovered in some form: discard ones at vetoed times
-    found_trigs = ppu.load_injections(found_file, vetoes)
-    found_injs = ppu.load_injections(found_file, vetoes, sim_table=True)
+    inj_file = ppu.load_triggers(found_missed_file, vetoes)
```
pannarale, line 286: This is the hdf5 file object, so
```suggestion
    injs = ppu.load_triggers(found_missed_file, vetoes)
```
is more intuitive as a name and once you have this, there is no need to extract columns, you simply navigate the hdf5 object, when needed (e.g., `injs[found/ra][:]` instead of `found_inj['ra'][:]`).

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -287,35 +283,34 @@ if do_injections:
...
-    found_trig['mchirp'] = np.asarray(found_trigs.get_column('mchirp'))
-    found_trig['bestnr'] = ppu.get_bestnrs(found_trigs, q=chisq_index,
-                                           n=chisq_nhigh,
-                                           null_thresh=null_thresh,
-                                           snr_threshold=snr_thresh,
-                                           sngl_snr_threshold=sngl_snr_thresh,
-                                           chisq_threshold=new_snr_thresh,
-                                           null_grad_thresh=null_grad_thresh,
-                                           null_grad_val=null_grad_val)
+    found_trig['mchirp'] = mchirp_from_mass1_mass2(
```
pannarale, line 299: Here you would then have two solutions.
1. opening the file as `r+`, creating the new dataset `'found/mchirp'`, and using it (and perhaps deleting it to restore the file as it was), or
2. storing this as an ad-hoc `found_injs_mchirp` to avoid modifying the files

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -368,30 +365,27 @@ if do_injections:
...
-    logging.info("%d found injections analysed.", len(found_injs))
-
-    # Missed injections (ones not recovered at all)
-    missed_injs = ppu.load_injections(found_missed_file, vetoes, sim_table=True)
+    logging.info("%d found injections analysed.", len(found_inj['time']))
 
     # Process missed injections 'missed_inj'
     missed_inj = {}
-    missed_inj['dist'] = np.asarray(missed_injs.get_column('distance'))
+    missed_inj['dist'] = inj_file['missed/distance'][:]
```
pannarale, line 372: Don't extract these either.

pannarale (dismissed):

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -226,11 +227,12 @@ max_bestnr, _, full_time_veto_max_bestnr =\
 if onsource_file:
 
     # Get trigs
-    on_trigs = ppu.load_xml_table(onsource_file, "multi_inspiral")
-    logging.info("%d onsource triggers loaded.", len(on_trigs))
+    on_trigs = ppu.load_triggers(onsource_file, vetoes)
+    logging.info("%d onsource triggers loaded.", on_trigs['network/event_id'].len())
 
     # Separate off chirp mass column
```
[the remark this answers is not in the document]
jakeb245: Is there a change to be made here? If not, I'll resolve this one

bin/pygrb/pycbc_pygrb_efficiency
jakeb245: I also considered adding this to the file within `pycbc_multi_inspiral` but decided to only save the masses there and calculate mchirp here. Other scripts might be using these files at the same time, so writing to them here might be a problem. And if other scripts need this value, it may be better to move the calculation earlier in the workflow.

bin/pygrb/pycbc_pygrb_efficiency
jakeb245: Ah I see, I'll have to look at this one a bit more.

bin/pygrb/pycbc_pygrb_efficiency
jakeb245: Now that I'm going about this, I think doing this calculation earlier would be best (which means less work for this PR and more work for another). I'm willing to hear other opinions of course.

jakeb245:
This needs a test run and might also depend on #4550, particularly when getting trigger masses. This may involve opening the full template bank and reading the mass based on a template ID.

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -226,55 +234,48 @@ max_bestnr, _, full_time_veto_max_bestnr =\
...
-    on_trigs = ppu.load_xml_table(onsource_file, "multi_inspiral")
-    logging.info("%d onsource triggers loaded.", len(on_trigs))
+    on_trigs = ppu.load_triggers(onsource_file, vetoes)
+    logging.info("%d onsource triggers loaded.", on_trigs['network/event_id'].len())
 
-    # Separate off chirp mass column
-    on_mchirp = on_trigs.get_column('mchirp')
+    # Calculate chirp mass values
+    on_mchirp = mchirp_from_mass1_mass2(bank_file['mass1'][:][on_trigs['network/template_id']],
+                                        bank_file['mass2'][:][on_trigs['network/template_id']])
```
titodalcanton, line 242: You have very similar code in a couple places, so I suggest creating a short helper function that lets you to do
```python
on_mchirp = get_trig_mchirp_from_bank(bank_file, on_trigs)
```

titodalcanton: Alternatively, since the only use of the bank file is to get the chirp masses, create the array of template chirp masses first with
```python
logging.info('Reading template chirp masses')
with h5py.File(opts.bank_file, 'r') as bank_file:
    template_mchirps = mchirp_from_mass1_mass2(
        bank_file['mass1'][:],
        bank_file['mass2'][:]
    )
```
and then just carry `template_mchirps` around. This lets you get rid of the bank file right away (better not keep open file handlers around).

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -428,48 +417,48 @@ if do_injections:
...
     long_inj['dist_mc'] = ppu.mc_cal_wf_errs(num_mc_injs, long_inj['dist'],
                                              cal_error, wav_err, max_dc_cal_error)
 
-    logging.info("MC injection set distributed with %d iterations.",\
+    logging.info("MC injection set distributed with %d iterations.",
                  num_mc_injs)
 
     # Check injections against on source
-    more_sig_than_onsource = np.ndarray(len(found_injs))
+    more_sig_than_onsource = np.ndarray(len(injs['found/distance']))
```
titodalcanton, line 431: You can actually remove the whole line, as `more_sig_than_onsource` will be redefined anyway on lines 433 or 435.

titodalcanton (approved):
Went through this as well and this seems to make sense. Just a small comment above but happy to approve.

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -219,62 +227,65 @@ offsource_trigs.reverse()
...
-                                           sngl_snr_threshold=sngl_snr_thresh,
-                                           chisq_threshold=new_snr_thresh,
-                                           null_grad_thresh=null_grad_thresh,
-                                           null_grad_val=null_grad_val)
-        loud_on_bestnr_trigs, loud_on_bestnr = \
-            np.asarray([[x, y] for y, x in sorted(zip(on_trigs_bestnrs, on_trigs),
-                                                  key=lambda elem: elem[0],
-                                                  reverse=True)])[0]
+        loud_on_bestnr_idx = np.argsort(on_trigs['network/reweighted_snr'])
+        loud_on_bestnr_idx = np.flip(loud_on_bestnr_idx)
```
pannarale, line 261: I'd prefer storing only the first element only of the right hand side.  While it is irrelevant, `loud_on_bestnr` stores a single value and it would be cleaner to do the same with `loud_on_bestnr_idx`

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -428,48 +417,48 @@ if do_injections:
     # NOTE: the loop on num_mc_injs would fill up the *_inj['dist_mc']'s at the
     # same time, so filling them up sequentially will vary the numbers a little
     # (this is an MC, order of operations matters!)
-    found_inj['dist_mc'] = ppu.mc_cal_wf_errs(num_mc_injs, found_inj['dist'],
+    found_inj_dist_mc = ppu.mc_cal_wf_errs(num_mc_injs, injs['found/distance'],
```
pannarale, line 420: Is there a reason for using `found_inj_dist_mc` rather than ` found_inj['dist_mc']` (cleaner)?  A similar question applies to `missed_inj_dist_mc`

pannarale (changes requested):

pannarale (approved):

PR #4564, opened 2023-11-08
Update __init__.py

ahnitz:
This would modify init_logging in pycbc to work with a 'count' argument. This means if you give --verbose, the logging level is 'INFO', iv you give it twice (or -vv), it will use 'DEBUG'. 

If you keep giving (verbose) the logging level will decrease in increments in 10, but those don't have canonical names, and would be up to the user to use at their volition.

### DIFF
--- a/pycbc/__init__.py
+++ b/pycbc/__init__.py
@@ -97,9 +97,9 @@ def init_logging(verbose=False, format='%(asctime)s %(message)s'):
 
     # See https://docs.python.org/3/library/logging.html#levels
     # for log level definitions
-    initial_level = logging.WARNING - int(verbose) * 10
     logger = logging.getLogger()
-    logger.setLevel(initial_level)
+    verbose_int = 0 if verbose is None else int(verbose)
+    logger.setLevel(logging.WARNING - verbose_int * 10)  # Initial setting
     sh = logging.StreamHandler()
     logger.addHandler(sh)
     sh.setFormatter(LogFormatter(fmt=format))
--- a/pycbc/__init__.py
+++ b/pycbc/__init__.py
@@ -94,6 +94,9 @@ def init_logging(verbose=False, format='%(asctime)s %(message)s'):
         logger.setLevel(log_level)
 
     signal.signal(signal.SIGUSR1, sig_handler)
+
+    # See https://docs.python.org/3/library/logging.html#levels
+    # for log level definitions
     initial_level = logging.WARNING - int(verbose) * 10
     logger = logging.getLogger()
     logger.setLevel(initial_level)
--- a/pycbc/__init__.py
+++ b/pycbc/__init__.py
@@ -78,8 +78,7 @@ def init_logging(verbose=False, format='%(asctime)s %(message)s'):
         What level to set the verbosity level to. Accepts either a boolean
         or an integer representing the level to set. If True/False will set to
         ``logging.INFO``/``logging.WARN``. For higher logging levels, pass
-        an integer representing the level to set (see the ``logging`` module
-        for details). Default is ``False`` (``logging.WARN``).
+        an integer representing the level to set. (1 = INFO, 2 = DEBUG).
     format : str, optional
         The format to use for logging messages.
     """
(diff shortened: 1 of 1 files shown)


pycbc/__init__.py
```diff
@@ -96,12 +95,9 @@ def sig_handler(signum, frame):
 
     signal.signal(signal.SIGUSR1, sig_handler)
 
-    if not verbose:
-        initial_level = logging.WARN
-    elif int(verbose) == 1:
-        initial_level = logging.INFO
-    else:
-        initial_level = int(verbose)
+    initial_level = logging.WARN
```
tdent, line 98: Seems OK, except that the code is now a bit cryptic. Suggest to add a comment making people consult https://docs.python.org/3/library/logging.html#levels to explain what is happening. 
Also, `logging.WARN` seems to be deprecated in favor of `logging.WARNING` so can we have that.

tdent, line 98:
```suggestion
    initial_level = logging.WARNING
```

pycbc/__init__.py
```diff
@@ -96,12 +95,9 @@ def sig_handler(signum, frame):
...
+    if verbose:
+        initial_level -= int(verbose) * 10
```
tdent, line 100: Actually I think it could all be done in a single line:
`initial_level = logging.WARNING - 10 * int(verbose)`
since False maps to 30 - 0 = WARNING, True or 1 maps to INFO and 2 maps to DEBUG

pycbc/__init__.py
```diff
@@ -95,14 +94,7 @@ def sig_handler(signum, frame):
...
     signal.signal(signal.SIGUSR1, sig_handler)
-
-    if not verbose:
-        initial_level = logging.WARN
-    elif int(verbose) == 1:
-        initial_level = logging.INFO
-    else:
-        initial_level = int(verbose)
-
+    initial_level = logging.WARNING - int(verbose) * 10
```
tdent, line 100:
```suggestion

    # See https://docs.python.org/3/library/logging.html#levels for why this works
    initial_level = logging.WARNING - int(verbose) * 10
```

tdent (approved):

tdent:
FWIW, we might want to check if this works with a specific executable before merging - I will see if we can do that with the SNR optimizer fix that Praveen is working on

pycbc/__init__.py
```diff
@@ -96,13 +95,9 @@ def sig_handler(signum, frame):
 
     signal.signal(signal.SIGUSR1, sig_handler)
 
-    if not verbose:
```
GarethCabournDavies, line 99: I think this doesn't take into account if `--verbose` uses `action='store_true'` and ends up as `False`, which worked before.

I suggest to keep the `if not verbose: elif`, or find every instance which uses `init_logging` and verbose as `store_true`.

The second is probably better but time consuming, the first would be easier to implement.

I would like to standardise logging and versioning throughout the code, which would involve the first option as it has become quite inconsistent, but that is a medium-term objective

GarethCabournDavies: Wait - ignore that, `int(False)` is zero

tdent: ie one of the laundry list items in https://github.com/gwastro/pycbc/issues/4483 .. 
Resolvable, I guess

PRAVEEN-mnl:
We found if we don't supply --verbose, we get error of Nonetype object. So we made a change in the code to fix it.

tdent (changes requested):
Removing my approval as this has a bug for the non-verbose case ..

pycbc/__init__.py
```diff
@@ -96,13 +95,9 @@ def sig_handler(signum, frame):
...
-    elif int(verbose) == 1:
-        initial_level = logging.INFO
-    else:
-        initial_level = int(verbose)
-
+    # See https://docs.python.org/3/library/logging.html#levels
+    # for log level definitions
+    initial_level = logging.WARNING - int(verbose) * 10
     logger = logging.getLogger()
     logger.setLevel(initial_level)
```
tdent, line 102:
```suggestion
    logger = logging.getLogger()
    verbose_int = 0 if verbose is None else int(verbose)
    logger.setLevel(logging.WARNING - verbose_int * 10)  # Initial setting
```

tdent (approved):
Approve (assuming tests pass)

PR #4569, opened 2023-11-23
Fix over-verbose offline workflow generator after init_logging change

GarethCabournDavies:
After #4564, the offline workflow generator became super-verbose with nothing set.

This fixes it back to be info by default

### DIFF
--- a/bin/workflows/pycbc_make_offline_search_workflow
+++ b/bin/workflows/pycbc_make_offline_search_workflow
@@ -66,7 +66,7 @@ wf.add_workflow_settings_cli(parser)
 args = parser.parse_args()
 
 # By default, we do logging.info, each --verbose adds a level of verbosity
-logging_level = args.verbose + 1 if args.verbose else logging.INFO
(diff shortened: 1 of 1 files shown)


GarethCabournDavies:
Forgot it's thanksgiving - adding reviewers

tdent:
Ah, because 'logging.INFO' is 20, which is a lot of verbose ;)

Does this affect any other executables which are passing logging.LEVEL rather than a small integer?

GarethCabournDavies:
> Ah, because 'logging.INFO' is 20, which is a lot of verbose ;)
> 
> Does this affect any other executables which are passing logging.LEVEL rather than a small integer?

This was the one I saw it in. I'll check some others, but this one is affected due to https://github.com/gwastro/pycbc/pull/4517

GarethCabournDavies:
From a `git grep init_logging`, it looks that all other codes either use the `args.verbose` object directly, or set it to `True`

tdent (approved):

PR #4570, opened 2023-11-26
quickstart example for making a simple pycbc workflow

ahnitz:


GarethCabournDavies:
I just happened across this and it looks really cool! - maybe adding an option to the option_exe which shows how standalone arguments can be passed through the config would be nice, e.g. a string to append to the file

ahnitz:
@GarethCabournDavies Good advice. I'll do that. Otherwise, I want to keep this example pretty trivial. We should add  a set of examples that demonstrate different features specifically so as not to overwhelm.

ahnitz:
There are two changes to the core  workflow code as part of this PR. Both are minor, but remove something that appears odd if one just wants to do a generic workflow. 

(1) the analysis output directory is always created for you so you don't have to manually do so
(2) an ifo is not required anywhere.No change if one is provided

ahnitz:
Given this requires a small code change, let's put off adding a doc version of this until the next release.

GarethCabournDavies (approved):

PR #4571, opened 2023-11-28
Fix bug with numpy import in pycbc_coinc_findtrigs

ArthurTolley:
Numpy is imported simply as `import numpy` in `pycbc_coinc_findtrigs`. There is a single call of `np.array` which will cause a name error:

`NameError: name 'np' is not defined`.

This is a simple fix to replace `np.array` with `numpy.array`.


GarethCabournDavies (approved):
LGTM, I suppose we just never run without --randomize-template-order!

ArthurTolley:
Mustn't have done, I was using it too. I only noticed the bug from the VSCode line highlight! Thanks :)

PR #4574, opened 2023-11-30
Fix cases where ifar limit was not being applied

GarethCabournDavies:
There are some places where the FAR limit was not being applied. I'll try to find if there are any more, but these are the ones which we noticed

### DIFF
--- a/bin/all_sky_search/pycbc_add_statmap
+++ b/bin/all_sky_search/pycbc_add_statmap
@@ -605,6 +605,11 @@ while True:
             sep_bg_data[key].data['decimation_factor'],
             bg_t_y,
             **significance_dict[ifo_combo_key])
+        fg_far = significance.apply_far_limit(
+            fg_far,
+            significance_dict,
+            combo=key,
+        )
         sep_bg_data[key].data['ifar'] = 1. / bg_far
         sep_fg_data[key].data['ifar'] = 1. / fg_far
         sep_fg_data[key].data['fap'] = 1 - \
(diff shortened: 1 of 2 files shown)


GarethCabournDavies:
This looks to be all of them, by doing a grep for 'get_far' and checking that all outputs get the limits applied

tdent:
Changes look OK, what level of testing do we want?

GarethCabournDavies:
> Changes look OK, what level of testing do we want?

Up to you - I've rerun the case we found, and confirmed that the IFAR is limited properly.

Injections take a different code path to zerolag, so can't really use those to check this kind of oversight (and they are exclusive-only).

The only suggestion I have for large-scale testing is to rerun postprocessing on a chunk and ensure there are no singles with IFAR above the limit, or set different limits for different coinc types, and check them each

GarethCabournDavies:
Found one more place through the extensive testing I mentioned above - there may be another but I am not 100% sure

tdent:
Hopefully you can get this pinned down then..

tdent:
Should I be seeing more commits or are you waiting for testing?

GarethCabournDavies:
Waiting for testing still

GarethCabournDavies:
> @GarethCabournDavies can you explain more about the significance dict 'unrelated bug' ? Did it have any consequences for the current search outputs?

Basically, the IFOs were in the wrong order when supplied as arguments to the code. So when checking for a 'L1H1' key, there wasn't one, as it would be 'H1L1' in all other codes

This has no consequences to results unless applying a significance limit to multi-detector events (which we don't do anyway).

PR #4576, opened 2023-12-01
Standardise logging: bin/all_sky_search and pycbc/events

GarethCabournDavies:
Some changes to how logging is handled, ~this will grow to a bigger change over the coming weeks as I start to add more executables and modules into this~. This MR contains executables in `bin/all_sky_search` and modules in `pycbc/events`:
1. The first change is to include the logging level in the default logging format
1. I plan to use `init_logging` in the executables as much as possible / where it is appropriate in order to keep logging the same throughoput the code
1. The last change is one I wanted to check, see below

I have made this PR early as I thought it best to check that my plan was agreeable to others before going through all parts of the code.

One of the things I was going to change is that each module gets its own named logger ([using this standard](https://docs.python.org/2/howto/logging-cookbook.html#using-logging-in-multiple-modules), one can then specify that logger to be more or less verbose when the module is called - this is useful for people downstream of us and is fairly standard practice from what I can see (people can even just set 'pycbc' to set things for all from pycbc which is nice). I have given an example from the cuts and significance modules.

### DIFF
--- a/bin/all_sky_search/pycbc_add_statmap
+++ b/bin/all_sky_search/pycbc_add_statmap
@@ -25,9 +25,9 @@ def get_ifo_string(fi):
     return istring
 
 parser = argparse.ArgumentParser()
+pycbc.common_options(parser)
 parser.add_argument('--version', action="version",
                     version=pycbc.version.git_verbose_msg)
-parser.add_argument('--verbose', action='store_true')
 parser.add_argument('--statmap-files', nargs='+',
     help="List of coinc files to be combined")
 parser.add_argument('--background-files', nargs='+', default=None,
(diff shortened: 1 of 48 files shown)


GarethCabournDavies:
Tagging @ahnitz @spxiwh @titodalcanton @cdcapano @tdent in case anyone has opinions on this before I start properly

titodalcanton:
This will heavily affect some things in PyCBC Live, like the lag plotter. Need not forget this!

GarethCabournDavies:
> This will heavily affect some things in PyCBC Live, like the lag plotter. Need not forget this!

That's exactly the situation I thought of with "where it is appropriate" - if I leave the `pycbc_live` executable alone, that should be okay

GarethCabournDavies:
As the verbose option was being set differently by different codes, I've also added a `common_options` function to `__init__.py`, which is called by each executable to set certain options as standard

It's named purposely vaguely so that we can add e.g. the version information to this later.

GarethCabournDavies:
Poke on this

PR #4577, opened 2023-12-05
split/bin parameters in pycbc_fit_sngls_split_binned

GarethCabournDavies:
Previously only a few parameters were allowed to be used as binning / splitting parameters in `pycbc_fit_sngls_split_binned`; this change makes many more parameters available for use

This also makes the labelling a bit more straightforward

I'd like to look over this code and make it a lot nicer to work with at some point - mainly as this was written fairly early-on in my use of python so is quite badly written

### DIFF
--- a/bin/all_sky_search/pycbc_fit_sngls_split_binned
+++ b/bin/all_sky_search/pycbc_fit_sngls_split_binned
@@ -127,8 +127,7 @@ extparams = [par for par in usedparams if not par in params]
 for ex_p in extparams:
     if ex_p == 'template_duration':
         logging.info('Reading duration from trigger file')
-        # List comprehension loops over templates and returns the first
-        # duration value from that region.
+        # Loop over template regions, return the first duration value from each region.
         dur = []
         for ref in trigf[args.ifo + '/template_duration_template'][:]:
             dur.append(trigf[args.ifo + '/template_duration'][ref][0])
--- a/bin/all_sky_search/pycbc_fit_sngls_split_binned
+++ b/bin/all_sky_search/pycbc_fit_sngls_split_binned
@@ -117,8 +117,8 @@ logging.info('Getting template bank parameters')
 usedparams = [args.split_param_two, args.split_param_one, args.bin_param]
 
 params = {}
-for ex_p in usedparams:
-    if ex_p == 'template_duration':
+for par in usedparams:
+    if par == 'template_duration':
         logging.info('Reading duration from trigger file')
         # Loop over template regions, return the first duration value from each region.
         dur = []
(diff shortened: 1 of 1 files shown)


bin/all_sky_search/pycbc_fit_sngls_split_binned
```diff
@@ -121,7 +122,7 @@ bank.close()
 
 # Only calculate params if needed
 usedparams = [args.split_param_two, args.split_param_one, args.bin_param]
-extparams = [par for par in ['mchirp', 'eta', 'chi_eff', 'template_duration']
+extparams = [par for par in bank_conversions.conversion_options
```
tdent, line 125: Looking at the bank conversions code, mass1,2 and spin1,2z are in the options list .. I am not sure I get the thinking here.  'extparams' was extra parameters _beyond_ the ones already in the bank, no? 
Maybe it should be `if (par in usedparams and not par in params)` ?

bin/all_sky_search/pycbc_fit_sngls_split_binned
```diff
@@ -136,28 +137,21 @@ for ex_p in extparams:
             ]
```
tdent, line 137: Consider rewriting this as a simple for loop rather than list comprehension? 
```
durations = []
for ref in trigf[args.ifo + '/template_duration_template'][:]:
    dur.append(trigf[args.ifo + '/template_duration'][ref][0])
params['template_duration'] = np.array(durations)
```
- this is 4 lines. 
Also, I don't understand what the comment about the quirk of h5py implies here - the code doesn't seem to do anything special for the case with no triggers

bin/all_sky_search/pycbc_fit_sngls_split_binned
GarethCabournDavies: extparams is basically a list of parameters which will need calculation, I think what you've said makes a lot more sense than the loop here, I was shoehorning into what had been done before, but your way makes more sense

bin/all_sky_search/pycbc_fit_sngls_split_binned
```diff
@@ -110,60 +111,46 @@ trigf = h5py.File(args.trigger_file, 'r')
...
 
-# Only calculate params if needed
+# Calculate params needed
 usedparams = [args.split_param_two, args.split_param_one, args.bin_param]
-extparams = [par for par in ['mchirp', 'eta', 'chi_eff', 'template_duration']
-             if par in usedparams]
 
-for ex_p in extparams:
+params = {}
+for ex_p in usedparams:
```
tdent, line 120: much simpler now - just rename the variable to something less obscure, eg `for par in usedparams`

GarethCabournDavies:
Tested and output plots are [here](https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/split_binnedhist_in_ci/search/?C=M;O=D) - note the lack of 'OLD' mtotal plots as that wasn't possible before

I've also set come 'choices' in the argparse, as the previous failure message was fairly obtuse

PR #4579, opened 2023-12-05
remove removal of downranked triggers

GarethCabournDavies:
Looking at the plots of missed loud injections in results pages, it was confusing that things had triggers with `snr` but not with `newsnr_bells_and_whistles`

This adds back in the things which were removed as they had sngl-ranking of 1

Old output:
image

New output:
image

SNR for good measure:
image

I've also added a brief explanation to the caption about why there are some triggers with ranking 1


tdent (approved):

tdent:
In slightly longer term we may want to define a module level macro, eg ranking.MINSNGLSTAT, which can be used instead of hardcoding 1 in several places

PR #4580, opened 2023-12-06
update lalsimulation cvmfs path

GarethCabournDavies:
Some places were missed in #4440

The basic method behind finding the changes needed was:

```git grep cvmfs | grep lalsimulation | grep openscience```

followed by
```
:%s:/cvmfs/oasis.opensciencegrid.org/ligo/sw/pycbc/lalsuite-extra/current/share/lalsimulation:/cvmfs/
software.igwn.org/pycbc/lalsuite-extra/current/share/lalsimulation:g
```
(VI command) in each file it was found in to make the change

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN dnf -y install https://ecsft.cern.ch/dist/cvmfs/cvmfs-release/cvmfs-release-
 
 # set up environment
 RUN cd / && \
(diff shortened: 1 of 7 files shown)


GarethCabournDavies:
This hasn't worked so solve the failed docker build, so I will attempt to find what else the problem could be. The change is still needed though

GarethCabournDavies:
docker build passes! - was just that I missed a couple of places that also needed updated

spxiwh (approved):

PR #4581, opened 2023-12-07
Added parameter try-index to allow for more iterations for placing points

kkacanja:


bin/bank/pycbc_brute_bank
```diff
@@ -45,13 +46,17 @@ parser.add_argument('--approximant',  required=True,
...
+parser.add_argument('--max-signal-length', type= float, 
+                    help="When specified, it cuts the maximum length of the waveform model to the lengh provided")
 parser.add_argument('--sample-rate', default=2048, type=float,
     help='sample rate in seconds')
 parser.add_argument('--low-frequency-cutoff', default=20.0, type=float)
 parser.add_argument('--enable-sigma-bound', action='store_true')
 parser.add_argument('--tau0-threshold', type=float)
 parser.add_argument('--permissive', action='store_true',
     help='Allow waveform generator to fail.')
+parser.add_argument('--try-index', default=1000, type=int,
```
ahnitz, line 58: Is there a better name than index? Maybe name something like '--placement-iterations'? Would that be clearer? Suggestions?

ahnitz (approved):
@kkacanja When / if this passes the unittests, I believe this can be merged.

ahnitz:
For future reference here, the aim of this PR is to be able to control how quickly the placement code gives up when it is having trouble finding any points that belong within the bounds of the parameter space. For very narrow regions, underplacement can occur with the standard settings we have found, so this option gives some control.

PR #4582, opened 2023-12-08
Adding likelihood to multi_signal model for inference of LISA signals.

ConWea:


WuShichao:
@ConWea Thanks! I will review this PR tomorrow.

examples/inference/lisa_smbhb_inj/lisa_smbhb_relbin.ini
```diff
@@ -50,6 +50,7 @@ spin2z = 0.36905807298613247
 distance = 17758.367941273442
 inclination = 1.5970175301911231
 t_obs_start = 31536000
+delta_f = 3.1709791983764586e-08
```
ahnitz, line 53: You should never be setting this as it is not your choice but the analysis which must determine the delta_f it wants to use.

examples/inference/lisa_smbhb_ldc/lisa_smbhb_relbin.ini
```diff
@@ -41,6 +41,7 @@ spin2z = 0.36905807298613247
 distance = 17758.367941273442
 inclination = 1.5970175301911231
 t_obs_start = 31536000
+delta_f = 3.1709791983764586e-08
```
ahnitz, line 44: See previous comment, this must be removed.

pycbc/waveform/waveform.py
```diff
@@ -547,8 +548,8 @@ def get_fd_det_waveform_sequence(template=None, **kwds):
         channels, values are FrequencySeries.
     """
     input_params = props(template, **kwds)
-    input_params['delta_f'] = -1
-    input_params['f_lower'] = -1
+    if input_params['f_lower'] is None:
```
ahnitz, line 551: It should be an error if f_lower is not provided, no?

ahnitz (changes requested):

examples/inference/lisa_smbhb_inj/lisa_smbhb_relbin.ini
ConWea: I see. So this should really be set and checked like the f_lower check in relbin.py instead of a static parameter here?

ahnitz: @ConWea I don't think I follow. Maybe it would help to explain the motivation behind this suggestion.

WuShichao: @ahnitz It seems Connor also pushed changes from this PR https://github.com/gwastro/pycbc/pull/4584 to here.

ConWea: @ahnitz I was thinking that because BBHX plugin needs delta_f to be set (or you get the error that @WuShichao showed in the other PR), that it should be set in waveform.py like the f_lower check, and set it to -1. Otherwise it isn't set anywhere and throws an error when trying to produce the waveform without delta_f.

pycbc/waveform/waveform.py
ahnitz: @ConWea I believe this change is actually in error. Where you are probably confused is between the get_fd_waveform and the get_fd_waveform_sequence interfaces. The sequency interface requires you to provide the frequency samples so flow / df are not supplied, whereas for the other they should be. 

I think this change is why the code examples failed.

examples/inference/lisa_smbhb_inj/lisa_smbhb_relbin.ini
ahnitz: @ConWea See comments on that section.

pycbc/waveform/waveform.py
ConWea: @ahnitz Ahh I see. Okay I will revert these changes!

pycbc/waveform/waveform.py
```diff
@@ -736,7 +736,7 @@ def get_fd_det_waveform(template=None, **kwargs):
     """
     input_params = props(template, **kwargs)
     if 'f_lower' not in input_params:
-        input_params['f_lower'] = -1
+        raise Exception("Must set f_lower in parameters")
```
ahnitz, line 739: You shouldn't have to add an exception, just take out setting it to -1. There will be an error already if it is not set in that case.

pycbc/inference/models/relbin.py
```diff
@@ -490,19 +491,31 @@ def multi_loglikelihood(self, models):
...
-                                 a0, a1)
-                loglr += - h1h2.real # This is -0.5 * re(<h1|h2> + <h2|h1>)
+        if self.still_needs_det_response:
+            for m1, m2 in itertools.combinations(models, 2):
+                for det in self.data:
+                    a0, a1, fedge = self.hihj[(m1, m2)][det]
+                    dtc, channel, h00 = m1._current_wf_parts[det]
+                    dtc2, channel2, h002 = m2._current_wf_parts[det]
+                    c1c2 = self.mlik(fedge,
+                                    dtc, channel, h00,
```
WuShichao, line 501: Please keep the same indentation as the above line. Same as 502-503 line.

pycbc/inference/models/relbin.py
```diff
@@ -490,19 +491,31 @@ def multi_loglikelihood(self, models):
...
+            # finally add in the lognl term from this model
+            for m1, m2 in itertools.combinations(models, 2):
+                for det in self.data:
+                    a0, a1, fedge = self.hihj[(m1, m2)][det]
+
+                    fp, fc, dtc, hp, hc, h00 = m1._current_wf_parts[det]
+                    fp2, fc2, dtc2, hp2, hc2, h002 = m2._current_wf_parts[det]
+
+                    h1h2 = self.mlik(fedge,
+                                    fp, fc, dtc, hp, hc, h00,
```
WuShichao, line 515: Please keep the same indentation as the above line. Same as 516-517 line.

WuShichao (changes requested):
Thanks for this PR. Please see my other two comments in the code, and also fix codeclimate issues. If you have time, can you also add a note about the math behind this changes?

pycbc/inference/models/relbin.py
ConWea: Just to check, do you mean like this?

```
for det in self.data:
                    a0, a1, fedge = self.hihj[(m1, m2)][det]

                    dtc, channel, h00 = m1._current_wf_parts[det]
                    dtc2, channel2, h002 = m2._current_wf_parts[det]

                    c1c2 = self.mlik(fedge,
                                    dtc, channel, h00,
                                    dtc2, channel2, h002,
                                    a0, a1)
                    loglr += - c1c2.real # This is -0.5 * re(<h1|h2> + <h2|h1>)
```

WuShichao: > Just to check, do you mean like this?
> 
> ```
> for det in self.data:
>         a0, a1, fedge = self.hihj[(m1, m2)][det]
>         
>         dtc, channel, h00 = m1._current_wf_parts[det]
>         dtc2, channel2, h002 = m2._current_wf_parts[det]
>         
>         c1c2 = self.mlik(fedge,
>                         dtc, channel, h00,
>                         dtc2, channel2, h002,
>                         a0, a1)
>         loglr += - c1c2.real # This is -0.5 * re(<h1|h2> + <h2|h1>)
> ```

I mean `fedge`, `dtc`, `dtc2`, and `a0` should be on the same column.

pycbc/inference/models/relbin.py
ConWea: Corrected in latest commit

pycbc/inference/models/relbin.py
ConWea: Corrected in latest commit

pycbc/inference/models/relbin.py
```diff
@@ -541,10 +556,17 @@ def _loglr(self):
             # with detector response. Otherwise, skip detector response.
 
             if self.still_needs_det_response:
+
+                # TODO: Check that p['tc'] is the appropriate variable here
```
WuShichao, line 560: Why is there a TODO? Do you want to implement it in the future?

ConWea: I forgot to remove it :D Will edit now. 

Am changing the required params in the next push as well (tested and is working)

WuShichao: @ConWea Your new modification looks good.

ahnitz (approved):
@ConWea I think you've resolved the comments i had before.

WuShichao (approved):
Looks good.

PR #4584, opened 2023-12-11
If f_lower not set by user, set to default value of -1.

ConWea:


WuShichao:
@ConWea Thanks, I see an error https://github.com/gwastro/pycbc/actions/runs/7168689439/job/19517425259?pr=4584#step:6:1380

ahnitz (changes requested):
Sorry, I guess I comment on these in the other PR already. You should not be setting delta_f at all here.

PR #4592, opened 2023-12-20
Do not set matplotlib's backend in internal modules

titodalcanton:
Discussion in #4534 highlighted another case of matplotlib's backend being set deep into a module, which can lead to problems. This PR fixes that case by making sure the backend is set at the initialization of the executable scripts which use that module.

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -11,6 +11,12 @@ import logging
 import subprocess
 import tempfile
 import shutil
+
+# we will make plots on a likely headless machine, so make sure matplotlib's
+# backend is set appropriately
+from matplotlib import use as mpl_use_backend
+mpl_use_backend('agg')
+
 import numpy as np
 import h5py
 import pycbc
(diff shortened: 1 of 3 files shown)


GarethCabournDavies (approved):

titodalcanton:
Tagging as "low latency" because this affects the SNR optimizer, so we might want to pull this into the 2.1 release line.

tdent:
what do the 'headless machine' comments mean?

titodalcanton:
The use of `use()` (sorry for the pun) is discouraged in matplotlib's documentation, so the comments are meant to explain why we use it. Those scripts will most likely run on a computer without a graphical environment and will not use matplotlib's interactive features, so we explicitly force matplotlib to use a non-interactive backend. The reason for doing this is because we have seen problems with matplotlib's default backend in the past when a graphical environment is not present.

tdent:
Ah, I wasn't aware headless was an actual computing term 🐔
Then our head nodes are actually headless 🤯

PR #4593, opened 2023-12-20
bug in choosing the far calculation method

GarethCabournDavies:
The wrong IFO combo was being passed to the significance dictionary to decide the method used for calculating the FAR in each combination during hierarchical removal

This led to events using the exponential method of FAR calculation when they should have used the counting method, and having very extrapolated IFARs!

### DIFF
--- a/bin/all_sky_search/pycbc_add_statmap
+++ b/bin/all_sky_search/pycbc_add_statmap
@@ -345,6 +345,8 @@ else:
                 f_in.attrs['background_time_exc'],
                 **significance_dict[ifo_combo_key])
 
+del ifo_combo_key
+
 logging.info('Combining false alarm rates from all available backgrounds')
 
 # Convert dictionary of whether the ifo combination is available at trigger
--- a/bin/all_sky_search/pycbc_add_statmap
+++ b/bin/all_sky_search/pycbc_add_statmap
@@ -611,7 +611,7 @@ while True:
             sep_fg_data[key].data['stat'],
             sep_bg_data[key].data['decimation_factor'],
             bg_t_y,
-            **significance_dict[ifo_combo_key])
+            **significance_dict[key])
         fg_far = significance.apply_far_limit(
             fg_far,
             significance_dict,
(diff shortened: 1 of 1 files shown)


bin/all_sky_search/pycbc_add_statmap
```diff
@@ -611,7 +611,7 @@ while True:
             sep_fg_data[key].data['stat'],
             sep_bg_data[key].data['decimation_factor'],
             bg_t_y,
-            **significance_dict[ifo_combo_key])
+            **significance_dict[key])
```
tdent, line 614: This is definitely a bug, I wonder how it managed to work apparently reasonably well before

GarethCabournDavies: I think if the event is either not super-loud, or a single in single time, the numbers we get out will seem sensible.

This will only show up after the first round of hierarchical removal as well, so we need two very-loud things in a chunk for this to show up!

tdent:
I guess the bug would have been instantly caught if you had done `del ifo_combo_key` around line 327, as that variable name should not have been used again outside the loop. 
So I suggest doing that anyway (stable door, horse, etc) unless there is a better way to make sure the dummy loop variable doesn't continue lying around like a mat on a polished floor

tdent:
The slightly more elegant solution would be to make what happens from line 311 into a function, but that may be too much restructuring for just this bug

tdent (changes requested):
See comments, one minor change to a previous code block requested

titodalcanton:
I think the general idea is to avoid very long stretches of code that do many things and juggle many objects, breaking them down into a series of shorter functions instead.

tdent:
The bug was actually easy to spot if you know which block to look at.  The code as a whole does a lot of things so 'finding the right bit' is not necessarily easy. 
The bug was _possible_ because `__main__` contains several for-loops and a variable from one can get reused inside another. 

So we could prefer as a design choice for for-loops to occur only inside functions, etc.  Something to think about more broadly ...

GarethCabournDavies:
I think hierarchical removal _may_ be possible using a function shared across codes, but there is a lot which goes into it, so the sheer number of variables may make that not be sensible!

tdent (approved):

tdent:
Well using functions may be good for refactoring or to avoid having lots of global variables in `__main__` or both ..

tdent:
Not sure what will be up with the CI tests ..

titodalcanton:
See #pycbc-code… Certificate issue with dcc.ligo.org.

PR #4595, opened 2024-01-05
MNT: upgrade upload/download-artifact GitHub actions

neutrinoceros:
Hi, version 4 of upload-artifact and download-artifact were recently released.
Staying on old version is not viable long term because GitHub will sooner or later drop support for the old versions of node that these depend on.
Since migration to v4 is not as straightfoward as bumping the version number, I'm helping all astropy-related packages byt manually performing the necessary changes.

### DIFF
--- a/.github/workflows/basic-tests.yml
+++ b/.github/workflows/basic-tests.yml
@@ -51,7 +51,7 @@ jobs:
         tox -e py-inference
     - name: store documentation page
       if: matrix.test-type == 'docs' && matrix.python-version == '3.8'
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v4
       with:
         name: documentation-page
         path: _gh-pages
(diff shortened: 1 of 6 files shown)


.github/workflows/basic-tests.yml
```diff
@@ -51,19 +51,20 @@ jobs:
         tox -e py-inference
     - name: store documentation page
       if: matrix.test-type == 'docs' && matrix.python-version == '3.8'
-      uses: actions/upload-artifact@v2
+      uses: actions/upload-artifact@v4
       with:
-        name: documentation-page
+        name: documentation-page-${{ matrix.os }}-{{ matrix.python-version }}
```
GarethCabournDavies, line 56: This seems to be making/uploading the documentation for each OS and each python version to be deployed - is that right?

I'm not sure whether that is the model we are wanting

neutrinoceros: > This seems to be making/uploading the documentation for each OS and each python version to be deployed - is that right?

not quite, in fact it merely enables it. I mindlessly applied the most general solution (which is to make artifact names unique) while I worked on upgrading a couple dozen repos at once, but I see that it's not actually needed here since only one os is present in the matrix and this step only runs on Python 3.8
I'll revert this one line.

ahnitz (approved):

PR #4597, opened 2024-01-08
Use vetoed times followup in example

GarethCabournDavies:
#4549 allowed us to omit vetoed times from the singles minifollowup and separate them into vetoed vs not-vetoed times.

This change adds vetoed times minifollowup into the CI example


tdent (approved):
Looks OK, if tests pass

GarethCabournDavies:
Tests were failing due to the statistic files being inherited automatically from the sngls_minifollowup to page_snglinfo.

This may not be what we want moving forward, but to get the minifollowup into the CI, I have removed this option

PR #4600, opened 2024-01-09
Live: produce empty trigger fit plot for detectors with no triggers

GarethCabournDavies:
We noticed that fit plots weren't being made; this was as there were no triggers produced.

To show that nothing is broken, we should make an empty plot stating that there are no triggers


maxtrevor (approved):
I think this all looks good

Issue #4601, opened 2024-01-09
Some visible loud missed injections are not in injection minifollowups


tdent:
In the new search pipeline including single-ifo candidates and with gating veto windows, quite often we see loud (SNR>100) missed injections in the f/m plots that are not in the inj minifollowups. 

Possible explanations for this: 
 - These may be single-ifo-time injections that are missed and plotted but are not included in minifollowups
 - These may be injections inside gating veto windows that are missed and plotted but are not included in minifollowups 

In either case we would like to get the f/m plots and minifollowups consistent ; probably we do want to include gating-vetoed missed inj in any case.

titodalcanton:
This appears to be due to the minifollowup sorting the missed injections in order of decreasing *second-highest* optimal SNR. Given that many injections fall in single-detector time, the second-highest SNR is zero, and those injections drop to the bottom of the list.

However, we now do have the ability to detect signals that are loud in one detector only. So it would probably be better for the minifollowup to sort missed injections by *network* optimal SNR instead.

As I type this, I seem to recall that this was indeed discussed a while ago in another issue or pull request…

titodalcanton:
Possibly related to this: #4498

tdent:
Addressed by #4602

PR #4602, opened 2024-01-10
Add the ability to choose how to sort injections in minifollowups

titodalcanton:
This addresses #4601 by adding a `sort-by` option to the `[workflow-injection_minifollowups]` section of the offline search configuration. The supported values are `decisive_optimal_snr` (previous behavior), `network_optimal_snr` (this would address #4601) and `decisive_eff_dist` (previous behavior in the absence of optimal SNRs). The option is mandatory, I think it is better to have this set explicitly.

There is some refactoring/cleanup as well.

### DIFF
--- a/bin/minifollowups/pycbc_injection_minifollowup
+++ b/bin/minifollowups/pycbc_injection_minifollowup
@@ -36,7 +36,7 @@ from pycbc.io.hdf import SingleDetTriggers, HFile
 from pycbc.conversions import mchirp_from_mass1_mass2
 
 
-def sort_injections(args, inj_file, missed, sort_by):
+def sort_injections(args, inj_group, missed, sort_by):
     """Return an array of indices to sort the missed injections according to a
     metric of choice.
 
@@ -44,10 +44,10 @@ def sort_injections(args, inj_file, missed, sort_by):
     ----------
     args : object
         CLI arguments parsed by argparse.
-    inj_file : h5py.File instance
-        HDF5 file object containing the injection definition.
+    inj_group : h5py group object
+        HDF5 group object containing the injection definition.
     missed : array
-        Array of indices of missed injections.
+        Array of indices of missed injections into `inj_group`.
     sort_by : str
         Which metric to sort by.
 
@@ -56,8 +56,6 @@ def sort_injections(args, inj_file, missed, sort_by):
     sorting : array
         Array of indices that will sort the missed injections as requested.
     """
-    inj_group = inj_file['injections']
-
     if 'optimal_snr' in sort_by:
         optimal_snrs = [
             inj_group[dsn][:][missed] for dsn in inj_group.keys()
@@ -179,20 +177,21 @@ for ifo in args.single_detector_triggers:
     insp_analysed_seglists[ifo].coalesce()
 
 f = h5py.File(args.injection_file, 'r')
+inj_def = f['injections']
 missed = f['missed/after_vetoes'][:]
 if args.ifar_threshold is not None:
     try:  # injections may not have (inclusive) IFAR present
         ifars = f['found_after_vetoes']['ifar'][:]
     except KeyError:
         ifars = f['found_after_vetoes']['ifar_exc'][:]
-        logging.warn('Inclusive IFAR not found, using exclusive')
+        logging.warning('Inclusive IFAR not found, using exclusive')
     lgc_arr = ifars < args.ifar_threshold
     missed = numpy.append(missed,
                           f['found_after_vetoes']['injection_index'][lgc_arr])
 
 # Get the trigger SNRs and times
 # But only ones which are within a small window of the missed injection
-missed_inj_times = numpy.sort(f['injections/tc'][:][missed])
+missed_inj_times = numpy.sort(inj_def['tc'][:][missed])
 
 # Note: Adding Earth diameter in light seconds to the window here
 # to allow for different IFO's arrival times of the injection
@@ -249,7 +248,7 @@ sort_by = workflow.cp.get_opt_tags(
     'sort-by',
     ''
 )
-sorting = sort_injections(args, f, missed, sort_by)
+sorting = sort_injections(args, inj_def, missed, sort_by)
 sorted_missed_indexes = missed[sorting]
 
 # loop over sorted missed injections to be followed up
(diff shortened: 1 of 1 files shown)


bin/minifollowups/pycbc_injection_minifollowup
```diff
@@ -14,20 +14,96 @@
...
+        # descending order of decisive (2nd largest) optimal SNR
+        dec_snr = numpy.array([
+            sorted(snrs)[-2] for snrs in zip(*optimal_snrs)
+        ])
+        if args.maximum_decisive_snr is not None:
+            # By setting to 0, these injections will not be considered
+            dec_snr[dec_snr > args.maximum_decisive_snr] = 0
+        return dec_snr.argsort()[::-1]
+
+    if sort_by == 'network_optimal_snr':
```
GarethCabournDavies, line 77: Should this use `elif` rather than `if`? I suppose the return means you don't need the `else`

bin/minifollowups/pycbc_injection_minifollowup
```diff
@@ -14,20 +14,96 @@
...
 from pycbc.detector import Detector
-import pycbc.workflow.minifollowups as mini
 from pycbc.workflow.core import resolve_url_to_file
-import pycbc.version, pycbc.pnutils
 from pycbc.io.hdf import SingleDetTriggers, HFile
+from pycbc.conversions import mchirp_from_mass1_mass2
+
 
-parser = argparse.ArgumentParser(description=__doc__[1:])
+def sort_injections(args, inj_group, missed, sort_by):
```
GarethCabournDavies, line 39: Could this be a function in a module? At least the 'calculate sort variable' part. Then it can be shared, and use a single source of truth

bin/minifollowups/pycbc_injection_minifollowup
```diff
@@ -14,20 +14,96 @@
...
+            eff_dists.append(eff_dist)
+        dec_eff_dist = sorted(eff_dists)[-2]
+        mchirp = mchirp_from_mass1_mass2(
+            inj_group['mass1'][:][missed],
+            inj_group['mass2'][:][missed]
+        )
+        dec_chirp_dist = pycbc.pnutils.chirp_distance(dec_dist, mchirp)
+        return dec_chirp_dist.argsort()
+
+    raise ValueError(f'Invalid injection sorting method "{sort_by}"')
```
tdent, line 106: On the principle 'error as soon as possible' I would prefer to define a list of allowed sort-by options outside this function and check the sort by argument is in the list before doing anything else.  Then you can also print out the list in the error message ('Allowed methods are ...')

bin/minifollowups/pycbc_injection_minifollowup
```diff
@@ -188,23 +236,37 @@ for trig in single_triggers:
...
     num_events = len(missed)
 
-# loop over loudest events to be followed up
+# sort the injections
+sort_by = workflow.cp.get_opt_tags(
+    'workflow-injection_minifollowups',
+    'sort-by',
+    ''
+)
+sorted_missed_indexes = sort_injections(args, inj_def, missed, sort_by)
```
tdent, line 254: 'indices' ? or just eg 'sorted_missed_idx'

bin/minifollowups/pycbc_injection_minifollowup
```diff
@@ -14,20 +14,99 @@
...
+                inj_group['distance'][:][missed],
+                inj_group['ra'][:][missed],
+                inj_group['dec'][:][missed],
+                inj_group['polarization'][:][missed],
+                inj_group['tc'][:][missed],
+                inj_group['inclination'][:][missed]
+            )
+            eff_dists.append(eff_dist)
+        dec_eff_dist = sorted(eff_dists)[-2]
+        mchirp = mchirp_from_mass1_mass2(
```
tdent, line 98: As the eff chirp dist sorting option is mainly provided for legacy / backwards compatibility, how about only importing the conversions and pnutils functions inside this block?

bin/minifollowups/pycbc_injection_minifollowup
titodalcanton: Sure, any suggestion? `pycbc/workflow/injection.py`? `pycbc/workflow/minifollowups.py`? `pycbc/inject/inject.py`?

bin/minifollowups/pycbc_injection_minifollowup
titodalcanton: Yes, and codeclimate would complain about that.

bin/minifollowups/pycbc_injection_minifollowup
titodalcanton: In fact `missed` is fine, and keeps the namespace cleaner.

bin/minifollowups/pycbc_injection_minifollowup
titodalcanton: Ok

bin/minifollowups/pycbc_injection_minifollowup
GarethCabournDavies: It doesn't feel like a workflow thing as I think it could be used in the foundmissed plot as well

Would a new module in inject, e.g. `inject/injection_utils.py` make sense here?

titodalcanton: Ok, I think I found a plausible way to address this and Tom's other comment.

titodalcanton:
Hmm, scratch that, I do think I can come up with something better at the moment, and it is almost telecon time. I think this PR could be merged and improved on later, if anyone has more concrete ideas.

bin/minifollowups/pycbc_injection_minifollowup
tdent: I'd say a new module for just one thing is overkill to some extent .. it may become clearer how to refactor this as and when we do find a second use case for the functions.

tdent:
Do you have to put this option in the example workflow ini now?

GarethCabournDavies (approved):

bin/minifollowups/pycbc_injection_minifollowup
GarethCabournDavies: Yeah this can wait then

bin/minifollowups/pycbc_injection_minifollowup
GarethCabournDavies: supply as choices in the argparse then?

titodalcanton:
I changed the option name to `--distance-type` to be consistent with the CLI of `pycbc_page_foundmissed`. In this way we can use argument substitution in the workflow config, like I did in the example.

I agree that in the future we should make this code uniform in `pycbc_page_foundmissed` and `pycbc_injection_minifollowup`.

bin/minifollowups/pycbc_injection_minifollowup
tdent: This has not been addressed

titodalcanton: Fair enough, this requires some duplication with the present organization of the code, but we can straighten that in that refactor PR I was talking about earlier.

bin/minifollowups/pycbc_injection_minifollowup
```diff
@@ -102,7 +110,10 @@ def sort_injections(args, inj_group, missed):
         sorter = dec_chirp_dist.argsort()
         return missed[sorter]
 
-    raise ValueError(f'Invalid injection sorting method "{args.distance_type}"')
+    raise ValueError(
```
tdent, line 113: The real suggestion was to check this (`if not args.hasattr('distance type') or args.distance type not in legal_distance_types`) before trying to do anything else, on the principle 'error as soon as possible'.  I guess it doesn't make much difference overall, but putting the input check first would avoid eg weird errors from `if 'optimal_snr' in args.distance_type`

tdent (approved):

PR #4603, opened 2024-01-10
O4b idq offline update

maxtrevor:
This is a rework of the entire offline dq workflow
The goals of this rework are to
- Simplify the code and logical structure of the workflow
- Allow for reweighting of triggers in a window around autogated times
- Re-implement the dq ranking statistic in a way that simplifies future use in low-latency


tdent:
Some comments/requests before really starting review ..

Concerning the content of changes I am dubious whether 'future use in low latency' belongs in this PR. Are you developing that in parallel? If so could the changes related to low latency idq be put in that development branch, not this one?  It is preferable for a single PR to deal with no more than a single conceptual change or improvement, possibly in addition to restructuring, this one seems to contain one and a half improvements or more (add gating segments, restructuring, plus some fraction of low latency idq development). 

The diff to stat.py is very large and much of it is whitespace before `# pylint` comments.  This is unnecessarily hard to read and review. Can you remove all whitespace changes from stat.py please so that we can focus on the functional changes.  Also, in some places it seems you have removed inline explanatory comments from stat.py code (eg the exp fit stat class), I think this is undesirable - can the comments be reinstated or edited appropriately please.

tdent:
Ah, I guess I'd also like to know if these changes include the ability to use idq segments shorter than 1s as we have discussed a few times

maxtrevor:
> Ah, I guess I'd also like to know if these changes include the ability to use idq segments shorter than 1s as we have discussed a few times

That feature is included in this change. I knew there was a feature I forgot to mention in the description.

maxtrevor:
> Some comments/requests before really starting review ..
> 
> The diff to stat.py is very large and much of it is whitespace before `# pylint` comments. This is unnecessarily hard to read and review. Can you remove all whitespace changes from stat.py please so that we can focus on the functional changes. Also, in some places it seems you have removed inline explanatory comments from stat.py code (eg the exp fit stat class), I think this is undesirable - can the comments be reinstated or edited appropriately please.

These changes have been reverted

tdent:
Ah, good, you can edit the description to also mention the sub-second aspect. 

Looking at where the description mentions low latency in more detail, I guess it's OK to have a change to dq stat implementation in this PR if it is just reworking / rearranging of the previous calculation and does not _add_ functionality that is specific to low latency (ie not used in offline).  But it does look like you have put in a bit of that, eg allowing for different hdf file formats? 
If it's not a huge disruption of this PR, could the changes in functionality be restricted to what is needed for offline?

maxtrevor:
> Ah, good, you can edit the description to also mention the sub-second aspect.
> 
> Looking at where the description mentions low latency in more detail, I guess it's OK to have a change to dq stat implementation in this PR if it is just reworking / rearranging of the previous calculation and does not _add_ functionality that is specific to low latency (ie not used in offline). But it does look like you have put in a bit of that, eg allowing for different hdf file formats? If it's not a huge disruption of this PR, could the changes in functionality be restricted to what is needed for offline? (Maybe you already did that but I haven't re-read the diffs in detail.)

I have edited the description.

The LL specific features have been removed, to be added in a future PR with some other LL machinery. The remaining changes to the dq stat are for compatibility with the changes in the workflow, but create a statistic that should be mathematically equivalent to the previous version.

spxiwh (commented):
So I've looked over the changes and I'm happy that there's nothing here that would affect any code other than the DQ code, that the changes seem stylistically good, and the checks etc. pass. Some minor comments are left.

However, there's a lot of DQ code being altered, and I hadn't followed what it did prior to this. What checks have you done to convince yourself that these changes haven't affected anything and that this code is now doing what you want?

maxtrevor:
> However, there's a lot of DQ code being altered, and I hadn't followed what it did prior to this. What checks have you done to convince yourself that these changes haven't affected anything and that this code is now doing what you want?

I have an offline run that has been testing the changes. All of the dq workflow jobs, and coinc and sngl jobs have compelted, but the followups failed out over the weekend. I restarted the workflow this morning so hopefully they will complete this time.

I'm happy to make the other changes requested later today

pycbc/workflow/dq.py
```diff
@@ -22,183 +22,122 @@
...
+    output_labels = []
+    if not tags:
+        tags = []
+
+    dq_labels = workflow.cp.get_subsections('workflow-data_quality')
+    dq_labels = numpy.array(dq_labels)
+
+    dq_types = numpy.array([workflow.cp.get_opt_tags(
+            'workflow-data_quality', 'dq-type', [dq_label])
+            for dq_label in dq_labels])
```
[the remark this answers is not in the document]
maxtrevor: A few lines later I use a boolean array as a mask to extract dq files associates with each ifo. It needs to be an array for a boolean mask to work

pycbc/workflow/dq.py
```diff
@@ -22,183 +22,122 @@
...
+            'workflow-data_quality', 'dq-name', [dq_label])
+            for dq_label in dq_labels])
+
+    ifos = numpy.unique(dq_ifos)
+
+    for ifo in ifos:
+        # FIXME : make this able to take multiple dq files per ifo
+
+        # get the dq label, type, and name for this ifo
+        mask = numpy.array([ifo == dq_ifo for dq_ifo in dq_ifos])
```
tdent, line 91: does `mask = (dq_ifos == ifo)` not work?

pycbc/workflow/dq.py
```diff
@@ -22,183 +22,122 @@
...
     make_analysis_dir(output_dir)
-    output = FileList()
-    if tags:
+    output_files = FileList()
+    output_labels = []
+    if not tags:
+        tags = []
+
+    dq_labels = workflow.cp.get_subsections('workflow-data_quality')
+    dq_labels = numpy.array(dq_labels)
```
tdent, line 73: I can't help thinking that the block of code from here up to line 96 or so would be easier to read and manipulate by using a dict keyed on dq_labels, and by looping over keys, rather than doing several list comprehensions over the same array (`... for dq_label in dq_labels`).

pycbc/workflow/dq.py
maxtrevor: This works

pycbc/workflow/dq.py
```diff
@@ -22,183 +22,122 @@
...
+
+    ifos = numpy.unique(dq_ifos)
+
+    for ifo in ifos:
+        # FIXME : make this able to take multiple dq files per ifo
+
+        # get the dq label, type, and name for this ifo
+        mask = numpy.array([ifo == dq_ifo for dq_ifo in dq_ifos])
+        assert numpy.sum(mask) > 0, f"Received no dq files for {ifo}"
+        assert numpy.sum(mask) < 2, f"Received more than one dq file for {ifo}"
```
tdent, line 93: FWIW if your mask has exactly 1 true value it's equivalent to an index, python `index` or numpy.where will give it

pycbc/workflow/dq.py
```diff
@@ -22,183 +22,122 @@
...
+    ifos = set(dq_ifos)
+
+    for ifo in ifos:
+        # FIXME : make this able to take multiple dq files per ifo
+
+        # get the dq label, type, and name for this ifo
+        mask = (dq_ifos == ifo)
+        assert numpy.sum(mask) > 0, f"Received no dq files for {ifo}"
+        assert numpy.sum(mask) < 2, f"Received more than one dq file for {ifo}"
+        dq_label = str(dq_labels[mask][0])
```
tdent, line 94: comment on why `str` is needed ?

pycbc/workflow/dq.py
maxtrevor: So that is true for now. I do plan to expand this code in the future to support using multiple dq files per ifo and was trying to plan for that, but I'm happy to make this just use an index for now

pycbc/workflow/dq.py
maxtrevor: Numpy converts it to some kind of numpy specific string format that pegasus doesn't like. If I don't use arrays, this won't be necessary

pycbc/workflow/dq.py
```diff
@@ -22,183 +22,118 @@
...
+
+    ifos = set(dq_ifos.values())
+
+    for ifo in ifos:
+        # get the dq label, type, and name for this ifo
+        ifo_dq_labels = [dql for dql in dq_labels if (dq_ifos[dql] == ifo)]
+        assert len(ifo_dq_labels) < 2, f"Received multiple dq files for {ifo}"
+        dq_label = ifo_dq_labels[0]
+        dq_name = dq_names[dq_label]
+        dq_type = dq_types[dq_label]
```
maxtrevor, line 92: @tdent @spxiwh I redid this section of code to avoid using numpy arrays

bin/all_sky_search/pycbc_bin_templates
```diff
@@ -0,0 +1,54 @@
...
+        'spin2z': bank['spin2z'][:],
+        'f_lower': np.ones_like(bank['mass1'][:]) * args.f_lower
+        }
+
+    bin_dict = background_bin_from_string(args.background_bins, data)
+    bin_names = [b.split(':')[0] for b in args.background_bins]
+
+logging.info('Writing bin template ids to file')
+with h5.File(args.output_file, 'w') as f:
+    ifo_grp = f.create_group(args.ifo)
```
tdent, line 48: This is fine, maybe it would be an improvement to record the binning settings via the output file attrs? Or at least the f-lower and background-bin settings could be recorded

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -103,108 +109,103 @@ logging.info("Applying %s > %.3f cut", args.sngl_ranking,
...
+
+# get analysis segments
+analysis_segs = select_segments_by_definer(
+    args.analysis_segment_file,
+    segment_name=args.analysis_segment_name,
+    ifo=ifo)
+
+
+# utility function to compute segment list durations
+def segs_duration(segs):
```
tdent, line 131: I don't think this is needed, segs.abs() should do the same calculation - see https://git.ligo.org/lscsoft/ligo-segments/-/blob/master/ligo/segments/__init__.py#L579

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -103,108 +109,103 @@ logging.info("Applying %s > %.3f cut", args.sngl_ranking,
...
+        bin_grp['tids'] = bin_tids
+
+        # get the dq states of the triggers in this bin
+        inbin = np.isin(tmplt_ids, bin_tids)
+        trig_times_bin = trig_times[inbin]
+        trig_states = np.array([dq_state_at_time(t) for t in trig_times_bin])
+
+        # calculate the dq rates in this bin
+        dq_rates = np.zeros(3, dtype=np.float64)
+        for state in [0, 1, 2]:
```
tdent, line 195: I think it's nicer to do `for state, segs in dq_state_segs_dict.items()` and then `frac_dt = segs.abs() / livetime` as per my other comment

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -75,6 +71,19 @@ with HFile(args.trig_file, 'r') as trig_file:
     data_mask = np.zeros(n_triggers_orig, dtype=bool)
     data_mask[idx] = True
 
+    # also get the gated times while we have the file open
+    gate_times = []
+    if args.gating_windows:
+        logging.info('Getting gated times')
+        try:
+            gating_key = f'{ifo}/gating'
```
tdent, line 79: I'm not seeing why the variable `gating key` is needed, does something bad happen if you just go directly to `gating_types = trig_file[f'{ifo}/gating'].keys()` ?

tdent (changes requested):
Looks generally OK, see a few suggested improvements / simplifications

tdent:
For my education, and as a prelude to review, can you give me a run down of how the information about gating times and windows (or equivalently the set of gated segments) gets from the trigger files to the statistic calculation?

bin/all_sky_search/pycbc_bin_templates
maxtrevor: f-lower and background-bin arguments are now saved in the file attrs

maxtrevor:
@tdent I incorporated your newest round of comments, will write up the answer to your question in a bit

bin/all_sky_search/pycbc_bin_trigger_rates_dq
```diff
@@ -103,108 +108,95 @@ logging.info("Applying %s > %.3f cut", args.sngl_ranking,
...
+            dq_rates[state] = frac_eff / frac_dt
+        bin_grp['dq_rates'] = dq_rates
+
+    # save dq state segments
+    for dq_state, segs in dq_state_segs_dict.items():
+        name = f'dq_state_{dq_state}'
+        dq_grp = all_dq_grp.create_group(name)
+        starts, ends = segments_to_start_end(segs)
+        dq_grp['segment_starts'] = starts
+        dq_grp['segment_ends'] = ends
```
maxtrevor, line 198: Segments corresponding to each state are saved by this block of code

pycbc/events/stat.py
```diff
@@ -2061,38 +2055,70 @@ def assign_dq_val(self, key):
...
+            dq_state_segs_dict = {}
+            for k in ifo_grp['dq_segments'].keys():
+                seg_dict = {}
+                seg_dict['start'] = \
+                    ifo_grp[f'dq_segments/{k}/segment_starts'][:]
+                seg_dict['end'] = \
+                    ifo_grp[f'dq_segments/{k}/segment_ends'][:]
+                dq_state_segs_dict[k] = seg_dict
+
+        return dq_state_segs_dict
```
maxtrevor, line 2083: Segments are read in during stat initialization using this method

pycbc/events/stat.py
```diff
@@ -2107,12 +2133,28 @@ def lognoiserate(self, trigs):
...
+            # Should be exactly one ifo provided
+            ifo = unq[0]
+
+        dq_state = self.find_dq_state_by_time(ifo, trigs['end_time'][:])
+        dq_rate = self.find_dq_noise_rate(trigs, dq_state)
+
         logr_n = ExpFitFgBgNormStatistic.lognoiserate(
                     self, trigs)
-        logr_n += self.find_dq_val(trigs)
+        logr_n += numpy.log(dq_rate)
```
maxtrevor, line 2157: When calculating the log of the noiserate for a trigger, the code determines the dq state of each trigger (which set of segments it fell into out of clean, dq_flag, or gated) and then looks up the appropriate trigger rate adjustment, which depends on the triggers' ifo, template, and dq state

maxtrevor (commented):
@tdent
Here is short self-review to explain how segments are passed to the stat and applied to downweight triggers

GarethCabournDavies (commented):
A couple of minor points that i noticed about this

pycbc/events/stat.py
```diff
@@ -2107,12 +2133,29 @@ def lognoiserate(self, trigs):
...
+
+        try:
+            # works in offline
+            ifo = trigs.ifo
+        except AttributeError:
+            ifo = trigs['ifo']
+            unq = numpy.unique(ifo)
+            assert len(unq) == 1
+            # Should be exactly one ifo provided
+            ifo = unq[0]
```
tdent, line 2150: is this equivalent to `ifo = ifo[0]` ? 
If so, it is probably simpler not to introduce a new variable .. ie `assert len(numpy.unique[ifo]) == 1` then `ifo = ifo[0]`

maxtrevor:
> To satisfy my curiosity, what piece of code previously did the truncation of the DQ penalty below 0?
> 
> I see `dq_val[i] = max(dq_val[i], val)` in the old code but I'm not sure how to parse what this is (was) doing.

https://github.com/gwastro/pycbc/blob/master/bin/all_sky_search/pycbc_rerank_dq#L75 was responsible for applying the truncation previously. That whole executable was cut as part of simplifying the workflow.

The line you were looking at was handling the case of having multiple dq products for one ifo, it would loop over the dq products, find the dq trigger rate penalty for each and only keep the largest penalty. This was never actually used in production.

tdent (approved):
Seems OK from my point of view, not sure why tests would be failing

maxtrevor:
> Seems OK from my point of view, not sure why tests would be failing

Looks like the test suite is failing to get some files from gwosc that it needs

bin/workflows/pycbc_make_offline_search_workflow
```diff
@@ -560,7 +560,8 @@ for ifo in ifo_bins:
     background_bins = workflow.cp.get_opt_tags(
         'bin_templates', 'background-bins', tags=[ifo])
     wf.make_template_plot(workflow, hdfbank, rdir['data_quality'],
-                          bins=background_bins, tags=[ifo, 'dq_bins'])
+                          bins=background_bins,
+                          tags=[ifo, 'dq_bins']+bank_tags)
```
tdent, line 564:
```suggestion
                          tags=[ifo, 'dq_bins'] + bank_tags)
```
PEP 8

bin/plotting/pycbc_plot_dq_flag_likelihood
```diff
@@ -24,45 +26,60 @@ pycbc.init_logging(args.verbose)
...
+for n, dqstate_name in dq_states.items():
+    dq_rates = numpy.array(
+        [ifo_grp['bins'][b]['dq_rates'][n] for b in bin_names])
+    dq_rates = numpy.maximum(dq_rates, 1)
+    logrates = numpy.log(dq_rates)
+
+    ymax = max(ymax, numpy.max(dq_rates))
 
-yticks = ax.get_yticks()
+    offset = width*(n-1)
```
tdent, line 63:
```suggestion
    offset = width * (n - 1)
```
PEP8

PR #4607, opened 2024-01-24
REFERENCE: fixes to the some broken statistics

GarethCabournDavies:
THIS SHOULD NOT BE MERGED

This is a reference PR to show the changes needed to fix statistics which are broken.

These are fixed in another PR, but I wanted to show which changes are specifically for fixing the existing statistics.

### DIFF
--- a/bin/all_sky_search/pycbc_sngls_findtrigs
+++ b/bin/all_sky_search/pycbc_sngls_findtrigs
@@ -151,7 +151,7 @@ for tnum in template_ids:
     sds = rank_method.single(trigs)[trigger_keep_ids]
     stat_t = rank_method.rank_stat_single((ifo, sds),
                                           **extra_kwargs)
-    trigger_times = sds['end_time']
+    trigger_times = trigs['end_time'][:][trigger_keep_ids]
     if args.cluster_window:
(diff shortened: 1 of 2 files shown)


GarethCabournDavies:
This is a reference issue, simply to point to the fact that the comparisons in #4608 are not against master. These are minor fixes and rare use cases, which can be subsumed into the changes of that PR

 - The statistics which didn't use a `single_dtype` containing `'end_time'` (mainly snr-like ones) were broken for singls_findtrigs , which assumed they would - we never used this so it hasnt shown up until I was doing full comparisons
 - The logic was wrong on `coinc_lim_for_thresh` for the `PhaseTDstatistic` - again this statistic hasn't been used in offline searches so wasn't a problem before
 - `PhaseTDstatistic`'s `rank_stat_single` was not using the `snglstat` key for single info, which it should have
 - `ExpFitCombinedSNR`'s `rank_stat_single` was using the `snglstat` key for single info, which it shouldn't have
 - `ExpFitStatistic`, `PhaseTDExpFitStatistic` and `ExpFitBgRateStatistic` look to be ones which are used as parent classes only, and don't work directly, so I've commented these out. These are completely removed in #4608

Issue #4594, opened 2023-12-21
Statistic refactor


GarethCabournDavies:
I am planning to refactor the `pycbc/events/stat.py` file to basically use two base classes, `QuadratureSumStatistic` and `LogLikeStatistic` (names can be changed!) and everything else just gets defined in the statistic keywords, e.g.:

| Current name | Current class | New name | New class | New statistic keywords |
| --- | --- | --- | --- | --- |
| `quadsum` | `QuadratureSumStatistic` | `quadsum` | `QuadratureSumStatistic` | |
| `single_ranking_only` | `QuadratureSumStatistic` | `single_ranking_only` | `QuadratureSumStatistic` | |
| `phasetd` | `PhaseTDStatistic` | `quadsum` | `QuadratureSumStatistic` | `phasetd` |
| `exp_fit_csnr` | `ExpFitCombinedSNR` | `quadsum` | `QuadratureSumStatistic` | `exp_fit` |
| `exp_fit_stat` | `ExpFitStatistic` | `loglike` | `LogLikeStatistic` |  |
| `phasetd_exp_fit_stat` | `PhaseTDExpFitStatistic` | `loglike` | `LogLikeStatistic` | `phasetd` |
| `dq_phasetd_exp_fit_fgbg_norm` | `DQExpFitFgBgNormStatistic` | `loglike` | `LogLikeStatistic` | `phasetd dq` |
| `exp_fit_bg_rate` | `ExpFitBgRateStatistic` |  `loglike` | `LogLikeStatistic` |  |
| `phasetd_exp_fit_fgbg_norm` | `ExpFitFgBgNormStatistic`  |  `loglike` | `LogLikeStatistic` | `phasetd` |
| `phasetd_exp_fit_fgbg_bbh_norm` | `ExpFitFgBgNormBBHStatistic` |  `loglike` | `LogLikeStatistic` | `phasetd chirp_mass_weighting` |
| `phasetd_exp_fit_fgbg_kde` | `ExpFitFgBgKDEStatistic` |  `loglike` | `LogLikeStatistic` | `phasetd kde` |
| `dq_phasetd_exp_fit_fgbg_kde` | `DQExpFitFgBgKDEStatistic` |  `loglike` | `LogLikeStatistic` | `phasetd kde dq` |

There is one thing holding me up at the moment:
 - I need to understand the difference between `exp_fit_stat` and `exp_fit_bg_rate` - which seem to imply the same thing, but what the difference is, I can't see right now.

[...]

PR #4608, opened 2024-01-24
Update rates-based statistics to be modular

GarethCabournDavies:
## Overview of changes

The exponential fit statistics are all similar, with simple factors added or subtracted, as discussed in #4594

I have refactored the `ExpFitStatistic` to be able to use the different features using a `--statistic-features` option. These then all use `exp_fit` as the `--ranking-statistic` option

The available features are:
| Feature | Description | Notes |
| --- | --- | --- |
| `phasetd` | Apply a factor of how the phase, time and amplitude differences match up to what is expected for signals. | |
| `kde` | Apply a factor according to a kernel density estimate of the ratio of signal and noise distributions for each template. | |
| `dq` | Apply a factor according to any data quality channel information. | |
| `chirp_mass` | Apply a reweighting according to the chirp mass of the template. | |
| `sensitive_volume` | Apply a factor of the log of sensitive volume (compared to a median for the template). This means that the statistic takes into account any changes in the detector sensitivity and that, e.g., we expect to see more events in the HL network than HV coincidences. | |
| `normalize_fit_rate` | Normalize the rates fits by the analysis time. This is needed so that the statistic is comparable over chunks of different lengths. | This was done for all `exp_fit` statistics, but is implemented here explicitly so that the `exp_fit_csnr` statistic can reuse the `lognoiserate` function from the `ExpFitStatistic` |

I have also removed the different treatment of triggers with sngl_ranking below threshold; this is now required explicitly as `--statistic-keywords alpha_below_threshold:6`. Again this was so that the `exp_fit_csnr` statistic can reuse the `lognoiserate` function from the `ExpFitStatistic`

In addition, there are some minor changes as well to fix some of the statistics which didn't work at all on previous master. There is a (just for reference) PR at #4607 to show these.

## Testing

I have tested all existing (and working) statistics to check that given the appropriate features, the output remains identical.

The testing is done against the codes in #4607, so that we can test statistics against what they _should_ be, rather than what they are.

Initial tests with a very small fraction of the bank have shown that the SNR-like statistics output files have identical hashes. The exp_fit statistics outputs are all the same to within a numpy.isclose test, i.e. ~1e-6 difference for values O(1). But I will add the results of more stringent testing here.

| Statistic | New statistic | Features | Keywords | pycbc_sngls_findtrigs max stat difference | pycbc_coinc_findtrigs max stat difference |
| --- | --- | --- | --- | --- | --- |
| quadsum | quadsum | | | File hash the same | File hash the same |
| single_ranking_only | single_ranking_only | -- | -- | File hash the same |  File hash the same |
| phasetd | phasetd | -- | -- | File hash the same  |  File hash the same |
| exp_fit_csnr | exp_fit_csnr | -- | -- | File hash the same  |  File hash the same |
| phasetd_exp_fit_fgbg_norm | exp_fit | `phasetd sensitive_volume normalize_fit_rate` |  `alpha_below_thresh:6` | 9.5e-7 | 3.8e-6 |
| phasetd_exp_fit_fgbg_bbh_norm | exp_fit | `phasetd sensitive_volume normalize_fit_rate chirp_mass` | `alpha_below_thresh:6` | 1.9e-6 | 3.8e-6 |
| phasetd_exp_fit_fgbg_kde | exp_fit | `phasetd sensitive_volume normalize_fit_rate kde` | `alpha_below_thresh:6` | 1.9e-6 | 3.8e-6 |
| dq_phasetd_exp_fit_fgbg_norm | exp_fit | `phasetd sensitive_volume normalize_fit_rate dq` | `alpha_below_thresh:6` |  9.5e-7 |  3.8e-6 |
| dq_phasetd_exp_fit_fgbg_kde | exp_fit | `phasetd sensitive_volume normalize_fit_rate dq kde` | `alpha_below_thresh:6` | 1.9e-6 | 3.8e-6 |


pycbc/events/stat.py
```diff
@@ -2307,6 +1596,16 @@ def parse_statistic_keywords_opt(stat_kwarg_list):
                       "Received {}".format(' '.join(stat_kwarg_list))
             raise ValueError(err_txt)
 
+    # Check that the statistic keywords are allowed
+    for feature in opts.statistic_features:
+        if feature not in _allowed_statistic_features:
+            err_msg = f"--statistic-feature {feature} not recognised"
+            raise NotImplementedError(err_msg)
```
GarethCabournDavies, line 1604: This shouldn't actually happen due to the argparse choices, but safety is best

bin/all_sky_search/pycbc_sngls_findtrigs
```diff
@@ -149,9 +137,8 @@ for tnum in template_ids:
 
     # Stat class instance to calculate the ranking statistic
     sds = rank_method.single(trigs)[trigger_keep_ids]
-    stat_t = rank_method.rank_stat_single((ifo, sds),
-                                          **extra_kwargs)
-    trigger_times = sds['end_time']
+    stat_t = rank_method.rank_stat_single((ifo, sds))
+    trigger_times = trigs['end_time'][:][trigger_keep_ids]
```
GarethCabournDavies, line 141: Some singles objects don't have the end time included

pycbc/events/stat.py
```diff
@@ -33,9 +33,18 @@
 from .eventmgr_cython import logsignalrateinternals_computepsignalbins
 from .eventmgr_cython import logsignalrateinternals_compute2detrate
 
+_allowed_statistic_features = [
```
GarethCabournDavies, line 36: I'm not sure where is best to describe each feature here to be honest

pycbc/events/stat.py
```diff
@@ -709,20 +719,23 @@ def coinc_lim_for_thresh(self, sngls_list, thresh, limifo,
...
+        fixed_stat = [b['snglstat'] for a, b in sngls_list if a != limifo][0]
+        s1 = thresh ** 2. - fixed_stat ** 2.
         # Assume best case scenario and use maximum signal rate
         s1 -= 2. * self.hist_max
         s1[s1 < 0] = 0
         return s1 ** 0.5
 
 
-class ExpFitStatistic(QuadratureSumStatistic):
+class ExpFitStatistic(PhaseTDStatistic):
```
GarethCabournDavies, line 730: Subclassing PhaseTDStatistic here in order to get the phasetd stuff in __init__

GarethCabournDavies:
The numbers in the comparison table have been updated

I am sure it is not a coincidence that the error in the coincs is double that of the singles, but I think a ~4e-6 difference is not too important given the dynamic range of the statistic.

GarethCabournDavies:
I am adding a description of the statistics to the docs - I am writing it in the `pycbc_make_offline_search_workflow` documentation at the moment, but that can be moved if requested.

GarethCabournDavies:
Note that I found and fixed a bug in the way that  `sngl_ranking_` keywords are handled and passed through to the ranking module

maxtrevor:
Noting here that it was suggested on today's call that we want to wait to merge this until after creating the new PyCBC Live branch intended for the rest of O4

GarethCabournDavies:
I think this is ready for review now. All tests are passing

I haven't been able to test how this works with Live as well as I would like to; there seems to be a problem with my testing environment

GarethCabournDavies:
Poke @spxiwh post-winter break

spxiwh (changes requested):
Some comments below, but nothing significant. It's hard to review the bulk of changes here because of the amount of code being moved around, but the sanity checks look convincing. @ahnitz do you any concerns/comments?

pycbc/events/stat.py
```diff
@@ -167,12 +184,10 @@ def get_sngl_ranking(self, trigs):
             The array of single detector values
         """
         return ranking.get_sngls_ranking_from_trigs(
-            trigs,
-            self.sngl_ranking,
-            **self.sngl_ranking_kwargs
+            trigs, self.sngl_ranking, **self.sngl_ranking_kwargs
```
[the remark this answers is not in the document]
GarethCabournDavies: Reverted

pycbc/events/stat.py
```diff
@@ -384,29 +407,30 @@ def __init__(self, sngl_ranking, files=None, ifos=None,
...
+            ("end_time", numpy.float64),
+            ("sigmasq", numpy.float32),
+            ("snr", numpy.float32),
         ]
 
         # Assign attribute so that it can be replaced with other functions
         self.has_hist = False
         self.hist_ifos = None
-        self.ref_snr = 5.
+        self.ref_snr = 5.0
```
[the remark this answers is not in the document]
GarethCabournDavies: Changed back

pycbc/events/stat.py
```diff
@@ -608,7 +645,8 @@ def update_file(self, key):
 
     def logsignalrate(self, stats, shift, to_shift):
         """
-        Calculate the normalized log rate density of signals via lookup
+        Calculate the normalized log rate density of coincident signals
```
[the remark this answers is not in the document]
GarethCabournDavies: The line was too long when adding "coincident" - I've changed to "coinc" and it is okay now

docs/workflow/pycbc_make_offline_search_workflow.rst
```diff
@@ -306,9 +306,73 @@ Specify the name of the channel you want to run the inspiral analysis over for H
...
+   * - ``reference_ifos``
+     - If using the ``sensitive_volume`` feature, these are the detectors used to determine the benchmark value by which the sensitive volume is compared. We use the median sensitive volume in the network of detectors supplied. Default H1,L1.
+   * - ``max_chirp_mass``
+     - If using the ``chirp_mass`` feature, this chirp mass defines a maximum weighting which can be applied to the statistic.
+   * - ``sngl_ranking_*``
+     - This is used to provide the keyword arguments to functions in `the events.ranking module <https://pycbc.org/pycbc/latest/html/_modules/pycbc/events/ranking.html>`_. For example, to use a different psdvar threshold in the newsnr_sgveto_psdvar_threshold function, we would use ``sngl_ranking_psd_var_val_threshold:10``.
+
```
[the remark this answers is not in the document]
GarethCabournDavies: I've put it into a separate page within a "search" section of the docs page - the rest of this would obviously need building up more to be particularly useful!

pycbc/events/stat.py
GarethCabournDavies: This is a good point - I've added an `type(self).__name__ == "PhaseTD" or self.kwargs["phasetd"]` check so that this is only called during in `init()` if needed

GarethCabournDavies: though right now, `pregenerate` hist doesn't have any way to be set, so `get_hist()` is only called when the `coinc_lim_for_thresh` / `rank_stat_coinc` are called for the first time. so it doesn't have any code paths to try and get the histograms when files aren't set unless they are actually needed.

spxiwh (approved):
I think I'm happy with this now. Please do poke Alex to see if he has any thoughts before merging.

PR #4613, opened 2024-01-30
Remove M2Crypto dependency

facutuesca:
Looking at https://github.com/gwastro/pycbc/pull/3997, it seems like `M2Crypto` was added to the conda environment due to being a transitive dependency ([via `ciecplib`](https://github.com/gwastro/pycbc/pull/3997#issuecomment-1092842815)). Since `ciecplib` has removed its dependency to `M2Crypto` on version `0.7.0` [(src)](https://github.com/duncanmmacleod/ciecplib/blob/5ee4a251b347c41ede9f77060c72df4f8c34b6b9/debian/changelog#L4), this PR attempts to bump the minimum required version for `ciecplib` and remove the `M2Crypto` dependency.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -203,7 +203,7 @@ extras_require = {
         'scikit-cuda',
     ],
     'igwn': [
-        'ciecplib>=0.4.4',
+        'ciecplib>=0.7.0',
     ],
 }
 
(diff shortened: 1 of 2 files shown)


ahnitz (commented):
From my perspective this all looks good. I won't approve directly though as I think this is primarily an LDG-only feature atm. Best if @spxiwh or @GarethCabournDavies can approve.

spxiwh (approved):
Thanks for the PR @facutuesca . I'm very happy to remove the m2crypto dependency. I wonder if this also removes the requirement to pin openssl to 1.1, but this can be merged as is now.

PR #4615, opened 2024-02-01
fix thinko in option parsing for defaults

tdent:
When an option is not given at all getattr on the args object gives None, but we don't want to translate that into "--option-name None" on the subprocess command line.


ArthurTolley (approved):
Looks good to me. Thanks for fixing it.

titodalcanton:
Note that the same code will also break if we add an option that takes a list of values (as in `nargs='+'`). Perhaps worth adding a FIXME or WARNING comment?

tdent:
This is fairly clunky customized code to deal with the specific way the optimizer args are set up .. as long as it stays in the optimizer.py module I think it's OK to fix the multiple args case as and when it ever occurs (it may never do)

PR #4621, opened 2024-02-07
Standardize logging in bin/inference

mj-will:
Prompted by trying to debug an inference run and following the changes made in https://github.com/gwastro/pycbc/pull/4576, standardize the logging in `bin/inference` to use the new `pycbc.add_common_pycbc_options` function.

I've only changed executables that already had a `verbose` option.

### DIFF
--- a/bin/inference/pycbc_inference_plot_gelman_rubin
+++ b/bin/inference/pycbc_inference_plot_gelman_rubin
@@ -29,7 +29,9 @@ from pycbc.inference import (gelman_rubin, io, option_utils)
 # add options to command line
 parser = io.ResultsArgumentParser(skip_args=['walkers'])
 
-pycbc.add_common_pycbc_options(parser)
+# verbose option
+parser.add_argument("--verbose", action="store_true", default=False,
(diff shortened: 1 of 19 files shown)


bin/inference/pycbc_inference_plot_gelman_rubin
```diff
@@ -29,9 +29,7 @@ from pycbc.inference import (gelman_rubin, io, option_utils)
 # add options to command line
 parser = io.ResultsArgumentParser(skip_args=['walkers'])
 
-# verbose option
```
GarethCabournDavies, line 32: This script doesn't use init_logging (see approx line 56) and uses different definitions of each level (`--verbose` becomes DEBUG, not INFO). So we should either use init_logging or keep the current option

mj-will: For simplicity, I've reverted the change to that file in https://github.com/gwastro/pycbc/pull/4621/commits, we can always update it further down the line

GarethCabournDavies (approved):

PR #4623, opened 2024-02-07
fix for injection file created from pycbc_create_injections

yi-fan-wang:
After introducing the LVK format for injections, there are three kinds of injection files right now:  `*.xml`, `*.hdf` with LVK format and `*.hdf` with PyCBC format (which is generated from `pycbc_create_injections`). 

This PR adds an operation to deal with PyCBC format: don't do anything and just make an copy. 

I slightly change the order of the `if...else...` in the last few lines, as the current logic appears clearer for me. 

I tested this PR with a PyCBC format injection file, and it indeed got copied and seems to work fine.

### DIFF
--- a/bin/pycbc_convertinjfiletohdf
+++ b/bin/pycbc_convertinjfiletohdf
@@ -181,7 +181,7 @@ else:
     inj_file.close()
 
 if injclass is not None:
-    # Assume the injection is a XML file or HDF file (LVK format)
+    # The injection file is of known type
     data = injclass.pack_data_into_pycbc_format_input()
     samples = FieldArray.from_kwargs(**data)
     CBCHDFInjectionSet.write(args.output_file, samples)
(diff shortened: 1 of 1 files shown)


bin/pycbc_convertinjfiletohdf
```diff
@@ -179,14 +180,11 @@ else:
...
 
-if injclass is None:
-    # Assume this is a PyCBC format
-    injclass = InjectionSet(args.injection_file)
-    data = {}
-    for key in xinj.table[0].__slots__:
-        data[str(key)] = numpy.array([getattr(t, key) for t in xinj.table])
-else:
+if injclass is not None:
+    # Assume the injection is a XML file or HDF file (LVK format)
```
GarethCabournDavies, line 184: No assumption here, so you can remove the "Assume"

```suggestion
    # The injection file is of known type
```

GarethCabournDavies (changes requested):
Minor change in the comment, but then I will approve

bin/pycbc_convertinjfiletohdf
yi-fan-wang: @GarethCabournDavies Fixed as suggested. Thanks

GarethCabournDavies (approved):

PR #4628, opened 2024-02-09
[pycbc_live] Simplify and fix how CLI options are passed to the SNR optimizer

ArthurTolley:
As noticed by @maxtrevor , the current `args_to_string` in `snr_optimizer.py` doesn't pass through the snr optimizer independent options to the subprocess, notably `--snr-opt-include-candidate` but also '--snr-opt-seed'.

I have added these as simple if statement checks at the end of `args_to_string`. I haven't tested this but will look again on Monday morning.


ArthurTolley:
```
  if args.snr_opt_seed is not None:
      argstr += f'--snr-opt-seed {args.snr_opt_seed} '
```

I think because `args.snr_opt_seed` is given a default in `pycbc_live` this will always pass through the default to the subprocess which is okay because it would always pass it through anyway. We just want to make sure that if a `0` seed is passed through then it is actually passed and not treated like False. Hence the `is not None`.

ArthurTolley:
Putting this here following Tom's point about all the arguments sharing the prefix 'snr-opt':

```   
argstr = ''
for arg_name, arg_value in vars(args).items():
    if arg_name.startswith('snr_opt') and arg_value is not None:
        arg_name = arg_name.replace('_', '-')
        argstr += f'--{arg_name} {arg_value} '
```

This would be the simplest way to transfer these arguments but it is thwarted by the existence of: `snr-opt-timeout` and `snr-opt-label` ([link](https://github.com/gwastro/pycbc/blob/master/bin/pycbc_live#L986)). So either we add all and then remove specific ones or we have this 'extra args' as suggested by Tito:
`--snr-opt-extra-opts="--random-seed 42 --snr-opt-include-candidate --etc 1234`
Which I will code up and test now

ArthurTolley:
Hi @titodalcanton , is this the change you envisioned with your suggestion? It's how I interpreted your message but I might've interpreted it incorrectly. With this change, to use the snr optimizer within pycbc_live you have to give the new `snr_opt_extra_opts` command with a string for the extra opts.

If you want to use the snr optimizer by itself then you will need to give the two options separately.

ArthurTolley:
This is the `snr_optimize_command.txt` of a test with these changes:

```
timeout 400 /home/arthur.tolley/.conda/envs/live_test/bin/pycbc_optimize_snr
--snr-opt-method differential_evolution
--snr-opt-di-maxiter 50.0
--snr-opt-di-popsize 100.0
--snr-opt-seed 42
--snr-opt-include-candidate
--data-files V1:output/2020_05_06/candidate_1272790100.164_jjwl/coinc-1272790100.164_V1_data_overwhitened.hdf L1:output/2020_05_06/candidate_1272790100.164_jjwl/coinc-1272790100.164_L1_data_overwhitened.hdf H1:output/2020_05_06/candidate_1272790100.164_jjwl/coinc-1272790100.164_H1_data_overwhitened.hdf
--psd-files V1:output/2020_05_06/candidate_1272790100.164_jjwl/coinc-1272790100.164_V1_psd.hdf L1:output/2020_05_06/candidate_1272790100.164_jjwl/coinc-1272790100.164_L1_psd.hdf H1:output/2020_05_06/candidate_1272790100.164_jjwl/coinc-1272790100.164_H1_psd.hdf
--params-file output/2020_05_06/candidate_1272790100.164_jjwl/coinc-1272790100.164_attributes.hdf
--approximant SEOBNRv4_ROM
--gracedb-server None
--gracedb-search AllSky
--gracedb-labels SNR_OPTIMIZED
--verbose
--output-path output/2020_05_06/candidate_1272790100.164_jjwl/optimize_snr
--cores 76
--processing-scheme cpu:1 
```

You can see the extra options have now made it to be passed through to the optimizer.

pycbc/live/snr_optimizer.py
```diff
@@ -325,6 +325,10 @@ def insert_snr_optimizer_options(parser):
         help='SNR Optimizer choices: ' + ', '.join(optimizer_choices))
 
     # Add the generic options
+    opt_opt_group.add_argument('--snr-opt-extra-opts',
+        default=None,
```
titodalcanton, line 329: No need to say `default=None`… It is the default :)

examples/live/run.sh
```diff
@@ -195,10 +195,10 @@ python -m mpi4py `which pycbc_live` \
 --src-class-eff-to-lum-distance 0.74899 \
 --src-class-lum-distance-to-delta -0.51557 -0.32195 \
 --run-snr-optimization \
+--snr-opt-method differential_evolution \
 --snr-opt-di-maxiter 50 \
 --snr-opt-di-popsize 100 \
```
titodalcanton, line 200: Should these not also be passed via `--snr-opt-extra-opts`?

pycbc/live/snr_optimizer.py
```diff
@@ -401,4 +405,8 @@ def args_to_string(args):
         option_fullname = f'--snr-opt-{optimizer_name}-{opt}'
         argstr += f'{option_fullname} {option_value} '
 
+    # Add the extra opts
+    if args.snr_opt_extra_opts is not None:
+        argstr += args.snr_opt_extra_opts
+
```
titodalcanton, line 410: Yes this is what I suggested, but I am missing why you still need all the code above. Why does `--snr-opt-method` need special handling for example?

examples/live/run.sh
ArthurTolley: I can do, would make args_to_string a lot cleaner

titodalcanton:
@ArthurTolley yes I think your interpretation of my suggestion is correct, though I might not have conveyed how extreme my suggestion was :)

Also, `--gracedb-server None` showing up in your example above does not sound correct!

pycbc/live/snr_optimizer.py
ArthurTolley: It doesn't need any special handling if everything is passed through snr_opt_extra_opts 😁

ArthurTolley:
> Also, --gracedb-server None showing up in your example above does not sound correct!

pycbc_live adds gracedb_server to the command line given to the optimizer, when it's not given it must take the default (None) and give that. I can add this as a patch in this PR if you want, moving [these](https://github.com/gwastro/pycbc/blob/master/bin/pycbc_live#L386) into [this](https://github.com/gwastro/pycbc/blob/master/bin/pycbc_live#L402) if statement

titodalcanton:
But None should be interpreted as "do not pass the `--gracedb-server` option", not as "pass `--gracedb-server None`". None (usually) not a valid GraceDB server.

ArthurTolley:
I can add a quick `if self.gracedb_server is not None:` to pycbc_live so solve that

examples/live/run.sh
```diff
@@ -195,10 +195,12 @@ python -m mpi4py `which pycbc_live` \
...
 --run-snr-optimization \
---snr-opt-di-maxiter 50 \
---snr-opt-di-popsize 100 \
---snr-opt-include-candidate \
---snr-opt-seed 42 \
+--snr-opt-extra-opts \
+    "--snr-opt-method differential_evolution \
+    --snr-opt-di-maxiter 50 \
+    --snr-opt-di-popsize 100 \
+    --snr-opt-seed 42 \
```
tdent, line 202: can we remove this seed option?  Since we would like to check that the default seed works (ie it gives stable output)

examples/live/run.sh
```diff
@@ -210,11 +212,15 @@ python -m mpi4py `which pycbc_live` \
...
-# --snr-opt-pso-c2 2.0 \
-# --snr-opt-pso-w 0.01 \
+# --snr-opt-extra-opts \
+#   "--snr-opt-method pso \
+#   --snr-opt-pso-iters 5 \
+#   --snr-opt-pso-particles 250 \
+#   --snr-opt-pso-c1 0.5 \
+#   --snr-opt-pso-c2 2.0 \
+#   --snr-opt-pso-w 0.01 \
+#   --snr-opt-seed 42 \
```
tdent, line 222: prefer to remove, as per other comment

ArthurTolley:
@titodalcanton With regards to your comment on the call.

We're effectively removing the snr_opt option group right? I can add `--snr-opt-extra-opts` to the pycbc_live parser arguments ([here](https://github.com/gwastro/pycbc/blob/master/bin/pycbc_live#L988)) and remove the `snr_optimizer.insert_snr_optimizer_options(parser)` call completely.

Alongside this, I'm not sure if `check_snr_optimizer_options` is actually being used anywhere? I think this should be called somewhere before the snr optimizer in pycbc_live and it should check the string fed to the optimizer is appropriate. Also avoids the problem of finding out the snr_optimizer doesn't work after starting the search

ArthurTolley:
`check_snr_optimizer_options` was being used in `pycbc_optimize_snr` (i neglected to check that file 🤦🏻‍♂️).

I've removed the `snr_optimizer.insert_snr_optimizer_options(parser)` from pycbc_live so these are no longer taken by pycbc_live (and don't appear in `pycbc_live --help`). I've added `parser.add_argument('--snr-opt-extra-opts',` to pycbc_live because this is now an option that pycbc_live takes to pass these options through to the optimizer. There are also a couple of things removed that were using the previous inputs but didn't seem to be needed.

bin/pycbc_live
```diff
@@ -987,6 +986,11 @@ parser.add_argument('--snr-opt-timeout', type=int, default=400, metavar='SECONDS
                     help='Maximum allowed duration of followup process to maximize SNR')
 parser.add_argument('--snr-opt-label', default='SNR_OPTIMIZED',
                     help='Label to apply to snr-optimized GraceDB uploads')
```
titodalcanton, line 988: Any reason not to include `--snr-opt-label` into `--snr-opt-extra-opts`? As far as I remember `pycbc_live` just passes that straight through to the optimizer.

titodalcanton: And it has the correct default value for GWCelery, so if we just ignore it, everything will be fine.

pycbc/live/snr_optimizer.py
```diff
@@ -388,17 +388,8 @@ def args_to_string(args):
...
-        key_name = f'snr_opt_{optimizer_name}_{opt}'
-        option_value = getattr(args, key_name)
-        # If the option is not given, don't pass it and use default
-        if option_value is None:
-            continue
-        option_fullname = f'--snr-opt-{optimizer_name}-{opt}'
-        argstr += f'{option_fullname} {option_value} '
+    # Add the extra opts
+    if args.snr_opt_extra_opts is not None:
+        argstr = args.snr_opt_extra_opts
```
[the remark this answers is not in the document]
titodalcanton: I think the easiest thing to do is remove this function entirely, and then in [here](https://github.com/gwastro/pycbc/blob/master/bin/pycbc_live#L137) do:
```python
self.snr_opt_options = args.snr_opt_extra_opts
```

bin/pycbc_live
titodalcanton: Nevermind, I take this back. `--snr-opt-label` has to stay because `pycbc_live` adds its argument to whatever is already used as a label for the first upload.

pycbc/live/snr_optimizer.py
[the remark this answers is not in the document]
ArthurTolley: yeah that seems way more simple

ArthurTolley:
@titodalcanton I've done the changes you've requested, if everything looks good then I'm happy to merge. If you want me to add the if statement to stop the gracedb server being sent to the snr_optimizer I can do that quickly

titodalcanton (approved):
Thanks @ArthurTolley looks good. The GraceDB None thing is actually harmless at the moment, so let's fix that another time (I think there are a few other options that need the same fix but do not want to think about now).

PR #4631, opened 2024-02-13
Standardize logging: bin, bin/bank and pycbc/tmpltbank

GarethCabournDavies:
Following on from #4576

This is the executables directly in `bin`, and in `bin/bank`. We specifically ignore:
 - `pycbc_live` - so that the logging monitors don't break
 - `pycbc_get_ffinal` - this uses optparse rather than argparse, and seems complicated using a mix of config files and CLI, so I'm skipping it for now.
 - `pycbc_live_nagios_monitor` - this doesn't really want to be verbose anyway, as it just monitors the logging output of `pycbc_live`
 - `pycbc_make_banksim` and `pycbc_make_faithsim` don't have verbose options; logging is set to INFO by default. Though the logging is standardized by this change

Also adding in the modules in `pycbc/tmpltbank`

### DIFF
--- a/bin/bank/pycbc_aligned_bank_cat
+++ b/bin/bank/pycbc_aligned_bank_cat
@@ -48,9 +48,8 @@ __program__ = "pycbc_aligned_bank_cat"
 parser = argparse.ArgumentParser(description=__doc__,
            formatter_class=tmpltbank.IndentedHelpFormatterWithNL)
 
+pycbc.add_common_pycbc_options(parser)
 parser.add_argument("--version", action="version", version=__version__)
-parser.add_argument("--verbose", action="store_true", default=False,
-                    help="verbose output")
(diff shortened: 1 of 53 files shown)


GarethCabournDavies:
Also adding in the executables in `bin`, but specifically ignoring:
 - `pycbc_live` - so that the logging monitors don't break
 - `pycbc_get_ffinal` - this uses optparse rather than argparse, and seems complicated using a mix of config files and CLI, so I'm skipping it for now.
 - `pycbc_live_nagios_monitor` - this doesn't really want to be verbose anyway, as it just monitors the logging output of `pycbc_live`
 -  `pycbc_make_banksim` and `pycbc_make_faithsim` don't have verbose options, though the logging is standardized

And other files which are bash rather than python

spxiwh (approved):

PR #4633, opened 2024-02-13
Added tables to dq results pages

maxtrevor:
This MR adds some tables of information to the DQ results pages, as requested by LVK reviewers

### DIFF
--- a/bin/plotting/pycbc_page_dq_table
+++ b/bin/plotting/pycbc_page_dq_table
@@ -1,0 +1,51 @@
+#!/usr/bin/env python
+""" Make a table of dq state information
+"""
+import sys
+import argparse
+import h5py as h5
+import numpy as np
+
+import pycbc
+import pycbc.results
+from pycbc.version import git_verbose_msg as version
+
+parser = argparse.ArgumentParser()
+parser.add_argument("--version", action="version", version=version)
+parser.add_argument('--ifo', required=True)
+parser.add_argument('--dq-file', required=True)
+parser.add_argument('--verbose', action='count')
+parser.add_argument('--output-file')
+args = parser.parse_args()
+
+dq_states = {
+    'dq_state_0': 'Clean',
+    'dq_state_1': 'DQ Flag',
+    'dq_state_2': 'Autogating',
+}
+
+f = h5.File(args.dq_file, 'r')
+grp = f[args.ifo]['dq_segments']
+
+livetimes = []
+total_livetime = 0
+for dq_state in dq_states:
+    livetime = grp[dq_state]['livetime'][()]
+    livetimes.append(livetime)
+    total_livetime += livetime
+livetimes.append(total_livetime)
+
+frac_livetimes = [lt / total_livetime for lt in livetimes]
+state_names = list(dq_states.values()) + ['Total']
+columns = [state_names, livetimes, frac_livetimes]
+columns = [np.array(c) for c in columns]
+col_names = ['DQ State', 'Livetime', 'Fractional Livetime']
+
+html_table = pycbc.results.html_table(columns, col_names,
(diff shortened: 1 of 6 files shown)


bin/plotting/pycbc_page_template_bin_table
```diff
@@ -0,0 +1,67 @@
...
+trigger_rate = [n / livetime for n in num_triggers]
+
+col_names = ['Template Bin', 'Number of Templates', '% of Templates',
+             'Number of Loud Triggers', '% of Loud Triggers',
+             'Loud Trigger Rate (Hz)']
+columns = [bin_names, num_templates, frac_templates,
+           num_triggers, frac_triggers, trigger_rate]
+columns = [np.array(c) for c in columns]
+
+html_table = pycbc.results.html_table(columns, col_names,
```
tdent, line 59: Minor point: this seems to default to 3 decimal places on all floats.  This is too many s.f. for times in seconds (1,290.199 s?) but not enough for fractions of templates or triggers (% of templates displayed as 0 in one case).

tdent (changes requested):
It seems to work, so approve generally.  But a couple of points to address on percentages and 'loud triggers'

bin/plotting/pycbc_page_template_bin_table
```diff
@@ -0,0 +1,67 @@
...
+
+bin_names.append('Total')
+num_triggers.append(total_triggers)
+num_templates.append(total_templates)
+frac_triggers = [n / total_triggers for n in num_triggers]
+frac_templates = [n / total_templates for n in num_templates]
+trigger_rate = [n / livetime for n in num_triggers]
+
+col_names = ['Template Bin', 'Number of Templates', '% of Templates',
+             'Number of Loud Triggers', '% of Loud Triggers',
```
tdent, line 53: as noted via chat, better to have a quantitative definition of 'Loud' visible somewhere

bin/plotting/pycbc_page_template_bin_table
maxtrevor: Any idea how to change the precision? I agree that its currently not ideal.

tdent: Nothing beyond what you can see here .. https://pycbc.org/pycbc/latest/html/_modules/pycbc/results/table_utils.html#html_table

bin/plotting/pycbc_page_dq_table
```diff
@@ -39,10 +39,13 @@ frac_livetimes = [lt / total_livetime for lt in livetimes]
 state_names = list(dq_states.values()) + ['Total']
 columns = [state_names, livetimes, frac_livetimes]
 columns = [np.array(c) for c in columns]
-col_names = ['DQ State', 'Livetime', 'Fractional Livetime']
+col_names = ['DQ State', 'Livetime', '% of Livetime']
+
+format_strings = [None, '#.#', '#.##%']
```
tdent: If I understand correctly the '%' here automatically converts a fraction into a percentage (and prints '%' after it).

bin/plotting/pycbc_page_template_bin_table
```diff
@@ -56,11 +59,16 @@ columns = [bin_names, num_templates, frac_templates,
            num_triggers, frac_triggers, trigger_rate]
 columns = [np.array(c) for c in columns]
 
+format_strings = [None, '#', '#.###%', '#', '#.###%', '#.##E0']
```
tdent: Seems OK, although I think one decimal place for '% of loud triggers' should be sufficient (ie `'#.#%'`)

tdent (approved):
Approve modulo maybe reducing the number of d.p. on % of loud triggers

Issue #4632, opened 2024-02-13
Is `EventManagerCoherent.write_to_hdf()` doing what we want?


titodalcanton:
In #4550 we noticed that the `write_to_hdf()` method of `EventManagerCoherent` does something unexpected with some of the properties of the templates. Specifically, it writes some of the template properties into each detector's dataset, instead of the `network` dataset. [Here](https://github.com/gwastro/pycbc/blob/e2c35eff8367fa4700588a60f80737949d8fd254/pycbc/events/eventmgr.py#L767-L770) it does so for the template duration. [Here](https://github.com/gwastro/pycbc/blob/e2c35eff8367fa4700588a60f80737949d8fd254/pycbc/events/eventmgr.py#L787) it does it again for the template hash. Now that I look closer at that method, I think there is an indentation bug throughout it, combined with an incorrect use of the `f.prefix` attribute. This would cause the HDF file to contain datasets like `<detector>/search/time_slides` while (I think) what we want is just `search/time_slides`. Apparently a similar issue was noted in #2812, though not really fixed?!

titodalcanton:
Using the `multi_inspiral` example and looking at the output file with `h5ls`, I see the following datasets:
```
H1/bank_chisq               Dataset {336}
H1/bank_chisq_dof           Dataset {336}
H1/chisq                    Dataset {336}
H1/chisq_dof                Dataset {336}
H1/coa_phase                Dataset {336}
H1/cont_chisq               Dataset {336}
H1/end_time                 Dataset {336}
H1/event_id                 Dataset {336}
H1/search                   Group
H1/sigmasq                  Dataset {336}
H1/snr_h1                   Dataset {336}
H1/template_duration        Dataset {336}
H1/template_hash            Dataset {336}
H1/time_index               Dataset {336}

L1/bank_chisq               Dataset {336}
L1/bank_chisq_dof           Dataset {336}
L1/chisq                    Dataset {336}
L1/chisq_dof                Dataset {336}
L1/coa_phase                Dataset {336}
L1/cont_chisq               Dataset {336}
L1/end_time                 Dataset {336}
L1/event_id                 Dataset {336}
L1/search                   Group
L1/sigmasq                  Dataset {336}
L1/snr_l                    Dataset {336}
L1/template_duration        Dataset {336}
L1/template_hash            Dataset {336}
L1/time_index               Dataset {336}

V1/bank_chisq               Dataset {336}
V1/bank_chisq_dof           Dataset {336}
[...]

Issue #4416, opened 2023-06-28
snr_IFO labels in PyGRB


pannarale:
https://github.com/gwastro/pycbc/blob/master/pycbc/events/eventmgr.py#L710 causes individual detector SNR labels to end with '1' only when the IFO in question is LHO.  In turn, all pygrb post-processing that uses individual detector SNR information has to adapt to this convention: basically an `if ifo == 'H1'` ends up being needed in several places, which is not ideal.  Better switch to a consistent naming: 'snr_h1', 'snr_l1', 'snr_v1', 'snr_k1'.

pannarale:
Speaking of ugly PyGRB labels, here is more: https://github.com/gwastro/pycbc/blob/master/bin/pycbc_multi_inspiral#L269 and https://github.com/gwastro/pycbc/blob/master/bin/pycbc_multi_inspiral#L280 (affects `pycbc_pygrb_plot_chisq_veto`)

titodalcanton:
@pannarale I do not understand your second comment as I think the code has changed in the mean time. What were the other ugly labels?

titodalcanton:
The `snr_*` field names have now been fixed in #4639.

PR #4639, opened 2024-02-16
Various fixes to `EventManager`s

titodalcanton:
This addresses, at least in part, #4416 and #4632.

It mostly affects the HDF5 output of `pycbc_multi_inspiral`, where the single-detector `snr_*` datasets are renamed to just `snr`, and per-detector gating information is now stored correctly. However, this also drops the `template_duration` dataset from the output of both `pycbc_multi_inspiral` and `pycbc_inspiral`, the idea being that template parameters should always be taken from the bank file instead.


jakeb245 (approved):
Looks good to me. It will break things downstream in the PyGRB workflow, but it should be easy to fix and will simplify the code.

PR #4640, opened 2024-02-16
Clean up options and help for fit over multiparam

tdent:
Older code that this is copied from had the option of reading fit smoothing parameter values from an input fit file, rather than from the template bank (or calculating on the fly).  This is no longer done so remove the (useless) option and references to it in help for other options

### DIFF
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -174,10 +174,10 @@ parser = argparse.ArgumentParser(usage="",
 
 pycbc.add_common_pycbc_options(parser)
 parser.add_argument("--version", action=pycbc.version.Version)
-parser.add_argument("--template-fit-file",
+parser.add_argument("--template-fit-file", required=True,
                     help="hdf5 file containing fit coefficients for each"
                          " individual template. Required")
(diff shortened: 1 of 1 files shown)


GarethCabournDavies:
I think that the save_trig_param stuff in pycbc_fit_sngls_by_template had this same idea in mind. Though that again isn't looked at at all, so should probably also be removed at the same time

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -178,14 +178,10 @@ parser.add_argument("--template-fit-file",
                     help="hdf5 file containing fit coefficients for each"
                          " individual template. Required")
 parser.add_argument("--bank-file", default=None,
-                    help="hdf file containing template parameters. Required "
-                         "unless reading param from template fit file")
+                    help="hdf file containing template parameters. Required")
```
titodalcanton, line 181: Set `required=True` and remove `default=None`

tdent:
I don't know why f lower had a default of 0 as this would break the duration calculation (ie you get infinite duration ..).  Removed this.

tdent:
@GarethCabournDavies as Tito is probably rather busy with live stuff, any chance of review?

titodalcanton (approved):

PR #4641, opened 2024-02-16
Reapply: upgrade upload/download-artifact GitHub actions""

ahnitz:
This is to test reapplying the patch that was proposed to upgrade the artifact github actions.


titodalcanton:
I think this is superseded by #4867.

PR #4642, opened 2024-02-16
Move template fits functionality into functions ready for use in live

GarethCabournDavies:


GarethCabournDavies:
Note on testing - running the executables with the same code, checking that the hashes of the output file are the same. Working on all tests (non-exhaustive, standard use) so far

GarethCabournDavies:
The plan is not to do this for a while now, so I will close

PR #4645, opened 2024-02-20
Fix comment bugs in #4635

titodalcanton:
The GraceDB comments modified in #4635 were not correct because they ignored the observing state of the localization-only detectors. This should fix that, along with a separate minor formatting bug.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -542,6 +542,9 @@ class LiveEventManager(object):
         )
         logging.info('Coincident candidate! Saving as %s', fname)
 
+        # Which IFOs were active?
+        live_ifos = [ifo for ifo in sld if 'snr_series' in sld[ifo]]
+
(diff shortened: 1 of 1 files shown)


maxtrevor (approved):
Looks good, approving

bin/pycbc_live
```diff
@@ -553,9 +556,9 @@ class LiveEventManager(object):
         comment = comment.format(
             ppdets(coinc_ifos),
             args.ranking_statistic,
-            set(ifos) - self.skymap_only_ifos,
-            ppdets(followup_ifos),
-            ppdets(self.skymap_only_ifos)
+            ppdets(set(ifos) - self.skymap_only_ifos),
```
tdent, line 559: ok, so 'ifos' here is all ifos involved in some way in the event?  is this going to be different from live_ifos?

PR #4650, opened 2024-02-23
SNR optimizer option types

tdent:
Fix issue in non-default settings for some options that Max complained about

### DIFF
--- a/bin/pycbc_optimize_snr
+++ b/bin/pycbc_optimize_snr
@@ -45,12 +45,12 @@ parser.add_argument('--psd-files', type=str, nargs='+',
                          'by PyCBC Live.')
 parser.add_argument('--approximant', required=True,
                     help='Waveform approximant string.')
-parser.add_argument('--snr-threshold', default=4.0,
+parser.add_argument('--snr-threshold', type=float, default=4.0,
                     help='If the SNR in ifo X is below this threshold do not '
                          'consider it part of the coincidence. Not implemented')
-parser.add_argument('--chirp-time-f-lower', default=20.,
+parser.add_argument('--chirp-time-f-lower', type=float, default=20.,
                     help='Starting frequency for chirp time window (Hz).')
-parser.add_argument('--chirp-time-window', default=2.,
+parser.add_argument('--chirp-time-window', type=float, default=2.,
                     help='Chirp time window (s).')
 parser.add_argument('--gracedb-server', metavar='URL',
                     help='URL of GraceDB server API for uploading events. '
--- a/bin/pycbc_optimize_snr
+++ b/bin/pycbc_optimize_snr
@@ -27,9 +27,9 @@ from pycbc.live import snr_optimizer
 
 
 parser = argparse.ArgumentParser(description=__doc__)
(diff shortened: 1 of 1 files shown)


ArthurTolley (approved):
Just tested this change and it fixes the previous `TypeError: can't multiply sequence by non-int of type 'float'` error. Happy to approve

tdent:
Was that for the f-lower option?

ArthurTolley:
@tdent I can re-run the test again to find the exact error but I simply gave `--snr-opt-extra-opts \
"--chirp-time-f-lower 20.0 --chirp-time-window 2.0 " \` and the error was there before the fix and not there after the fix :), like I said, I will re-run this in a minute.

ArthurTolley:
This is the full error:
```
  File "/home/arthur.tolley/.conda/envs/live_rank_stat/bin/pycbc_optimize_snr", line 141, in <module>
    tau0 = cv.tau0_from_mass1_mass2(fp['mass1'][()], fp['mass2'][()], tau0flow)
  File "/home/arthur.tolley/.conda/envs/live_rank_stat/lib/python3.9/site-packages/pycbc/conversions.py", line 364, in tau0_from_mass1_mass2
    return tau0_from_mtotal_eta(mtotal, eta, f_lower)
  File "/home/arthur.tolley/.conda/envs/live_rank_stat/lib/python3.9/site-packages/pycbc/conversions.py", line 337, in tau0_from_mtotal_eta
    return _a0(f_lower) / (mtotal**(5./3.) * eta)
  File "/home/arthur.tolley/.conda/envs/live_rank_stat/lib/python3.9/site-packages/pycbc/conversions.py", line 322, in _a0
    return 5. / (256. * (numpy.pi * f_lower)**(8./3.))
TypeError: can't multiply sequence by non-int of type 'float'
```

So f_lower.

ArthurTolley:
For the CI, maybe related to the pegasus upgrade?

Just for completeness, when applying the fix to f_lower and not chirp_time_window we get another error which is also fixed by this explicit type :) 
```
  File "/home/arthur.tolley/.conda/envs/live_rank_stat/bin/pycbc_optimize_snr", line 143, in <module>
    mintau0 = max(tau0 - args.chirp_time_window,
numpy.core._exceptions._UFuncNoLoopError: ufunc 'subtract' did not contain a loop with signature matching types (dtype('float64'), dtype('<U3')) -> None
```

maxtrevor (changes requested):
While we are improving the command line options, can we please change the verbosity option to use https://github.com/gwastro/pycbc/blob/master/pycbc/__init__.py#L69 ?

tdent:
Max if you make a code suggestion (via the comment interface) we can accept it, although the verbose option is not related to this specific fix

maxtrevor:
made the requested change myself, and somehow it seems to have gone through without needing to be accepted?

maxtrevor (approved):

tdent:
Oh, yeah, github does let you edit someone else's PR just like that ..

titodalcanton:
Remember to apply the right label to the PR, otherwise this risks not being deployed in the 2.1 release line.

PR #4653, opened 2024-02-27
Update how live fitting deals with invalid values

GarethCabournDavies:
We noticed some plotting issues with when there are zero-live-time days for singles fit plots.

This PR:
 - removes invalid values from the `mean` count and fit coefficient calculation inputs, so that the count does not become infinite. This would effect results where we use the `mean` value - at the moment we use `conservative` so it doesn't matter
 - adds markers to the plot, so that we can see information where there is a day on its own
 - Allows the labels to be in scientific format below 1e-3, so that we don't simply state zero for the mean and conservative values of these

The cumulative effect of these can be seen in the before / after comparison below:

Before:

image

After:

image

### DIFF
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -161,9 +161,8 @@ for ifo in args.ifos:
     fout_ifo['separate_fits/start_time'] = trigger_file_starts[ad_order]
     fout_ifo['separate_fits/end_time'] = trigger_file_ends[ad_order]
 
-    for counter, a_c_u_l in enumerate(zip(alphas_bin[ifo],
-                                          counts_bin[ifo], bu, bl)):
-        a, c, u, l = a_c_u_l
+    for counter, (a, c) in enumerate(zip(alphas_bin[ifo],
+                                         counts_bin[ifo])):
         # Sort alpha and counts by date
         a = np.array(a)[ad_order]
         c = np.array(c)[ad_order]
--- a/bin/live/pycbc_live_combine_single_fits
+++ b/bin/live/pycbc_live_combine_single_fits
@@ -154,7 +154,8 @@ for ifo in args.ifos:
     logging.info(ifo)
     fout_ifo = fout.create_group(ifo)
     l_times = np.array(live_times[ifo])
-    fout_ifo.attrs['live_time'] = l_times.sum()
+    total_time = l_times.sum()
+    fout_ifo.attrs['live_time'] = total_time
 
     fout_ifo['separate_fits/live_times'] = l_times[ad_order]
     fout_ifo['separate_fits/start_time'] = trigger_file_starts[ad_order]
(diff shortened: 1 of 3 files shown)


bin/live/pycbc_live_plot_combined_single_fits
```diff
@@ -151,7 +152,7 @@ for ifo in ifos:
         bin_colour = plt.get_cmap(args.colormap)(bin_prop)
         bin_label = f"duration {bl:.2f}-{bu:.2f}"
         alpha_lines += ax_alpha.plot(separate_starts[ifo], alphas, c=bin_colour,
-                                     label=bin_label)
+                                     label=bin_label, marker='x')
```
titodalcanton, line 155: Could you also try with the `.` (period) marker? For some reason crosses make plots look more cluttered to me, and are less visible.

bin/live/pycbc_live_plot_combined_single_fits
```diff
@@ -140,7 +140,8 @@ for ifo in ifos:
             continue
 
         l_times = separate_times[ifo]
-        rate = counts / l_times
+        with np.errstate(divide='ignore'):
+            rate = counts / l_times
```
titodalcanton, line 144: Cool, I did not know this!

bin/live/pycbc_live_combine_single_fits
```diff
@@ -166,16 +167,20 @@ for ifo in args.ifos:
         # Sort alpha and counts by date
```
titodalcanton, line 167: I just noticed that `u` and `l` appear to be unused. It there any reason to include `bu` and `bl` in the zip? Could  you do instead
```python
for counter, (a, c) in enumerate(zip(alphas_bin[ifo], counts_bin[ifo])):
```

bin/live/pycbc_live_plot_combined_single_fits
GarethCabournDavies: I didn't either, first time I found it!

bin/live/pycbc_live_combine_single_fits
GarethCabournDavies: I didn't realise you could unpack like that either, that makes things much nicer

GarethCabournDavies:
image

I've added the dots, but made them bigger in order to be more visible

bin/live/pycbc_live_combine_single_fits
```diff
@@ -154,28 +154,32 @@ for ifo in args.ifos:
...
+                                         counts_bin[ifo])):
         # Sort alpha and counts by date
         a = np.array(a)[ad_order]
         c = np.array(c)[ad_order]
-        invalphan = c / a
-        mean_alpha = c.mean() / invalphan.mean()
-        cons_alpha = np.percentile(a, 100 - args.conservative_percentile)
+
+        # ignore anything with the 'invalid' salient values
+        valid = c > 0
```
titodalcanton, line 171: Hmm, what happens if all items of `valid`are False? Presumably the lines below will crash?

titodalcanton: This would be the case when the interferometer had unusable data for ~20 days, which is certainly unlikely but maybe we should decide upfront what we want the code to do in that situation?

bin/live/pycbc_live_combine_single_fits
```diff
@@ -154,28 +154,32 @@ for ifo in args.ifos:
...
+
+        invalphan = c[valid] / a[valid]
+        mean_alpha = c[valid].mean() / invalphan.mean()
+        cons_alpha = np.percentile(a[valid], 100 - args.conservative_percentile)
         cons_alphas_out[ifo][counter] = cons_alpha
         alphas_out[ifo][counter] = mean_alpha
         # To get the count values, we need to convert to rates and back again
-        r = c / l_times[ad_order]
+        r = c [valid]/ l_times[ad_order][valid]
```
titodalcanton, line 179: Remove space between `c` and `[valid]`.

bin/live/pycbc_live_plot_combined_single_fits
```diff
@@ -161,14 +163,22 @@ for ifo in ifos:
                                             label=alpha_lab))
 
         count_lines += ax_count.plot(separate_starts[ifo], rate, c=bin_colour,
-                                     label=bin_label)
+                                     label=bin_label, marker='.',
+                                     markersize=10)
+
+        if mr < 1e-3:
+            mlab = f"mean = {mr:.3e}"
+            clab = f"{conservative_percentile:d}th %ile = {cr:.2e}"
```
titodalcanton, line 171: Should we have 3 decimal digits for `cr` as well?

titodalcanton:
A couple minor comments, then I am ok with merging.

Rebase on master to get the checks to pass.

bin/live/pycbc_live_combine_single_fits
GarethCabournDavies: I've added in a 'if not any(valid)' check to use nan values if that is the case.

I'll need to update what happens in the case that the search itself gets a nan from this file though for that edge case

GarethCabournDavies: Added now

PR #4657, opened 2024-02-28
avoid using for loop in space psds; fix typo

HumphreyWang:
@WuShichao I removed all for loops from the space PSD functions to avoid unnecessary time costs. There are still duplicate lines though, not sure if you want to optimize them further.

### DIFF
--- a/pycbc/psd/analytical_space.py
+++ b/pycbc/psd/analytical_space.py
@@ -29,7 +29,7 @@ and paper <10.1088/1361-6382/ab1101>.
 """
 
 import numpy as np
-from astropy import constants
+from scipy.constants import c
 from pycbc.psd.read import from_numpy_arrays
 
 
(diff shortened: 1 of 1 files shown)


ahnitz:
@HumphreyWang Thank you for fixing this!

pycbc/psd/analytical_space.py
```diff
@@ -29,7 +29,7 @@
 """
 
 import numpy as np
-from astropy import constants
+from scipy.constants import c
```
ahnitz, line 32: In general we should preference astropy for constants. Is there a reason to prefer the scipy version?

HumphreyWang: When using `astropy.constants.c`, there will be a non-critical warning said "Cannot find reference 'c' in '__init__.py'", and also it needs to be converted to float type by additionally saying `c.value`. It's OK to change back, I'll do it later!

WuShichao:
@HumphreyWang Thanks for the optimization of the PSD module; I will review that soon. I will also add other space-borne detectors' PSD in a new PR soon. I remember there is a reason why I didn't forbid for loop here; I will test if all is functional.

WuShichao (approved):
I have tested all functions in this module:
image
image
Looks good, but I notice there are small but noticeable changes in the T channel curve in some tests, maybe due to different frequency bins.

PR #4658, opened 2024-02-28
update time_slice documentation, add additional option for the exact …

ahnitz:
…index identification threshold

### DIFF
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -175,31 +175,67 @@ class TimeSeries(Array):
     sample_rate = property(get_sample_rate,
                            doc="The sample rate of the time series.")
 
-    def time_slice(self, start, end, mode='floor'):
+    def time_slice(self, start, end,
+                   mode='floor',
+                   exact_indices_threshold=1e-3):
         """Return the slice of the time series that contains the time range
         in GPS seconds.
+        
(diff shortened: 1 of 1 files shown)


pycbc/types/timeseries.py
```diff
@@ -175,31 +175,67 @@ def get_sample_rate(self):
...
-            end_idx = round(end_idx)
-
-        if mode == 'floor':
+        # When dealing with time series of different size or transforming
+        # between different sample rates, floating point precision calculations
+        # can result in start / end times that should be at exact index locations
+        # to be some epsilon different in floating point value. 
+        # (2) identify if we intended the start / end to be at an exact sample
+        # based on identifying only a trivial difference.
+        if exact_indices_threshold:
```
ahnitz, line 219: Here's a further explainer why this is needed. One might ask, why not just use the 'nearest' mode. Is that not the same? This block is primarily useful when doing the 'floor' rounding. You don't want it to accidentally round down say a value of 3.9999999 to the previous sample if the difference was just floating point error. If you do then depending on the situation you might get a different length in the same situation. This is intended to make it more robust against this scenario.

spxiwh:
Beyond the bunch of whitespace that needs removing, I'm happy with this, but I think I'd prefer Tito review this as the request for this came from him.

PR #4660, opened 2024-03-06
Add TianQin and Taiji PSD

WuShichao:
@ahnitz This PR adds PSD functions for TianQin and Taiji. Similar to LISA, I also added confusion noise version PSD for Taiji, for TianQin, it will almost not be affected by DWD confusion noise, so I don't add that for TianQin. I need more tests.

Those detectors are TDI-based, so those functions look very similar. DECIGO is not based on TDI, so I won't add it now.

**UPDATE: I also add confusion noise into TianQin now.**

- [x] fix TianQin confusion noise fit
- [x] double-check the form of equations for TianQin and Taiji
- [x] current confusion noise PSD only valid for X channel, need to correct AE, T channel's implementation
- [x] TDI-2.0 support


titodalcanton:
@WuShichao please rebase on master to fix the checks.

WuShichao:
Several diagnostic plots:
image
image
image
image
image
image

HumphreyWang:
@WuShichao Thanks for adding more PSDs! As for the issues about TianQin confusion noise, I suggest removing the confusion noise functions:

- TianQin has a higher sensitive frequency range than LISA/Taiji, so the confusion noise offers minor effects for the PSDs.
- Although we provide coefficients for the polynomial fit for the confusion, results extrapolating from these points could be physically unreliable.
- Also these coefficients are truncated. Polynomial fit can be significantly sensitive to these coefficients and may cause problems.

If you want to keep these functions, I suggest removing the extrapolating parts to make them only available for given duration sets. Again it will have minor effects for TianQin.

WuShichao:
> @WuShichao Thanks for adding more PSDs! As for the issues about TianQin confusion noise, I suggest removing the confusion noise functions:
> 
> * TianQin has a higher sensitive frequency range than LISA/Taiji, so the confusion noise offers minor effects for the PSDs.
> * Although we provide coefficients for the polynomial fit for the confusion, results extrapolating from these points could be physically unreliable.
> * Also these coefficients are truncated. Polynomial fit can be significantly sensitive to these coefficients and may cause problems.
> 
> If you want to keep these functions, I suggest removing the extrapolating parts to make them only available for given duration sets. Again it will have minor effects for TianQin.

Actually, I used `4` years in the above plots, so no extrapolation here. Below is a comparison:
image
image
I believe I implemented this exactly as the TianQin paper described, but I couldn't reproduce the original figure.

HumphreyWang:
@WuShichao Please check the review comments I leave in the code. I have also opened a pull request in your branch for fixing the TianQin confusion noise, I hope that will answer your question and recommend you remove the confusion noise for TianQin.

As I said in the review comments, the major issue is that TianQin doesn't share the same expressions of acc & oms noise with LISA/Taiji.

WuShichao:
> @WuShichao Please check the review comments I leave in the code. I have also opened a pull request in your branch for fixing the TianQin confusion noise, I hope that will answer your question and recommend you remove the confusion noise for TianQin.
> 
> As I said in the review comments, the major issue is that TianQin doesn't share the same expressions of acc & oms noise with LISA/Taiji.

@HumphreyWang I see your PR to my branch, I remember I tried that before because, in the Taiji paper, they misused those two functions, let me try it again. I don't see your review comments (maybe you didn't save and submit it), I do know the form of equation os acc/oms noise is different, but I believe if those 3 detectors are TDI-based, they should share the same equations, just change the level of those noises. Do you know why?

WuShichao:
TianQin confusion issue has been fixed:
image
image
I also tried several extrapolation cases, which seem good enough:
image
image
But as @HumphreyWang mentioned, the current TianQin's curve is slightly different than papers, I do notice my curve is slightly higher. We need to know why.

HumphreyWang (changes requested):

pycbc/psd/analytical_space.py
```diff
@@ -58,8 +62,78 @@ def psd_lisa_acc_noise(f, acc_noise_level=3e-15):
...
+    Returns
+    -------
+    s_acc_nu : float or numpy.array
+        The PSD value or array for acceleration noise.
+    Notes
+    -----
+        Please see Table(1) in <10.1088/0264-9381/33/3/035010>
+        for more details.
+    """
+    s_acc_nu = _psd_acc_noise(f, acc_noise_level)
```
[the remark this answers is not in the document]
WuShichao: In Eq. (120), why `S_acc_nv` is related to `S_x`? Is this a typo? In the LISA case, it's `S_a`.

pycbc/psd/analytical_space.py
```diff
@@ -157,18 +349,19 @@ def analytical_psd_lisa_tdi_1p5_XYZ(length, delta_f, low_freq_cutoff,
     acc_noise_level = np.float64(acc_noise_level)
     oms_noise_level = np.float64(oms_noise_level)
     fr = np.linspace(low_freq_cutoff, (length-1)*2*delta_f, length)
-    s_acc_nu, s_oms_nu = lisa_psd_components(
-                          fr, acc_noise_level, oms_noise_level)
-    omega_len = omega_length(fr, len_arm)
+    s_acc_nu = _psd_acc_noise(fr, acc_noise_level)
+    s_oms_nu = _psd_oms_noise(fr, oms_noise_level)
+    omega_len = _omega_length(fr, len_arm)
     psd = 16*(np.sin(omega_len))**2 * (s_oms_nu+s_acc_nu*(3+np.cos(omega_len)))
```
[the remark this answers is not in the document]
WuShichao: Good spot, is fixed.

pycbc/psd/analytical_space.py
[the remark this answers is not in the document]
WuShichao: I also need to update TianQin's sensitivity curve form.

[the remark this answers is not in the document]
HumphreyWang: Thanks, that is a typo, we will update it in the next version!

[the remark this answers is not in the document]
WuShichao: @HumphreyWang  I have a question here, do they share the same sensitivity equation (this is from LDC manual):
image
Just different in the PSD of acc/oms noise, arm length, and <Fp+>?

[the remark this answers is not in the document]
HumphreyWang: @WuShichao Right, although they have different acc/oms noise, the convention between PSD and sensitivity follows the same.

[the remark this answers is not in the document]
WuShichao: @HumphreyWang, the current code has addressed all your questions. However, I find an inconsistency between the implementation of this PR and the previous TianQin paper. Let me show you:
image
I use this simplified equation for the dashed black curve:
image
Do you know the reason why my plot is higher than Fig. 2 in http://arxiv.org/abs/2005.07889, but consitent with Fig.3 in https://link.aps.org/doi/10.1103/PhysRevD.100.043003 ?

pycbc/psd/analytical_space.py
```diff
@@ -157,18 +355,20 @@ def analytical_psd_lisa_tdi_1p5_XYZ(length, delta_f, low_freq_cutoff,
     acc_noise_level = np.float64(acc_noise_level)
     oms_noise_level = np.float64(oms_noise_level)
     fr = np.linspace(low_freq_cutoff, (length-1)*2*delta_f, length)
-    s_acc_nu, s_oms_nu = lisa_psd_components(
-                          fr, acc_noise_level, oms_noise_level)
-    omega_len = omega_length(fr, len_arm)
-    psd = 16*(np.sin(omega_len))**2 * (s_oms_nu+s_acc_nu*(3+np.cos(omega_len)))
+    s_acc_nu = _psd_acc_noise(fr, acc_noise_level)
+    s_oms_nu = _psd_oms_noise(fr, oms_noise_level)
```
HumphreyWang, line 359: These lines indicate that psd_detector_acc/oms_noise functions will not be used, especially after you apply changes for TianQin components. This issue also shows in PSD functions for other channels.

pycbc/psd/analytical_space.py
```diff
@@ -711,3 +1902,99 @@ def analytical_psd_lisa_tdi_AE_confusion(length, delta_f, low_freq_cutoff,
     fseries = psd_AE + psd_X_confusion
```
HumphreyWang, line 1902: Could you show me where I can find this approximation (or assumption)?

HumphreyWang (changes requested):

pycbc/psd/analytical_space.py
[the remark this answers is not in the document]
HumphreyWang: @WuShichao The black one, as the equation you showed here, removed the sky-averaged factor 10/3. You can check Eq. (11) in 2005.07889 and the following descriptions for explanation.

So as far as I understand, for consistency, the confusion noise described in that paper should directly be added into non-averaged sensitivity, and then multiply the factor. (I'll double-check this later, thanks for the question!)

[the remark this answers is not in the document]
WuShichao: @HumphreyWang You can see Eq.(15) in https://link.aps.org/doi/10.1103/PhysRevD.100.043003. It doesn't remove 3/10. I prefer to keep it so that we can be consistent with the LDC convention, I'm sure the LISA one is correct, and TianQin cannot be below LISA's curve.

pycbc/psd/analytical_space.py
WuShichao: @HumphreyWang They will be used in LISA/Taiji case, because those two share the same frequency dependency in PSD of acc/oms noise. For TianQin, I'm not using those, I just code in TianQin's functions.

pycbc/psd/analytical_space.py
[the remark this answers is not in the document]
WuShichao: @HumphreyWang I can reproduce the confusion plot:
image

pycbc/psd/analytical_space.py
HumphreyWang: Right, but for funcs like `analytical_psd_tianqin_tdi_1p5_XYZ`, now you are still calling **this** function, so the acc/oms parameters will pass to `_psd_acc/oms_noise` instead of `psd_tianqin_acc/oms_noise` you edited in last commit.

pycbc/psd/analytical_space.py
WuShichao: @HumphreyWang By definition,
image
Those confusion noise PSDs are for X channel, so just add them, the CSD of X, Y channels `S_xy` is well below the PSD of X at low-frequency region of confusion noise, so we can ignore the effect of confusion noise on `S_xy`. For T channel, it contributes very little to SNR, so I also ignore modification of it.

pycbc/psd/analytical_space.py
WuShichao: OK. I will update them.

WuShichao: @HumphreyWang Fixed
image
image
image
image

pycbc/psd/analytical_space.py
WuShichao: This assumption has a good match with LDC data:
image

WuShichao: This assumption has a good match with LDC data:
image

WuShichao:
@HumphreyWang, I added a warning for TianQin/Taiji confusion noise fit when it tries to extrapolate. I also double-checked the fit equations, I believe those are the fit of the sensitivity curve of DWD, not the PSD of DWD.

pycbc/psd/analytical_space.py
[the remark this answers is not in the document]
HumphreyWang: @WuShichao Whether we should use/remove the factor of 10/3 depends on the situation, for here I agree to keep it. However, the confusion noise for TianQin is based on the non-averaged one, so it should also be multiplied by the factor.

WuShichao:
@HumphreyWang @ahnitz I have added TDI-2.0 support to LISA/TianQin/Taiji PSDs and sensitivity curves with or without DWD confusion noise. I also update the SMBHB PE example to include the `tdi` argument. Here are some plots:
image
image
image
image
image
image
image
image
image
image

HumphreyWang (approved):

pycbc/psd/analytical_space.py
```diff
@@ -184,32 +392,31 @@ def analytical_psd_lisa_tdi_2p0_XYZ(length, delta_f, low_freq_cutoff,
...
-    acc_noise_level = np.float64(acc_noise_level)
-    oms_noise_level = np.float64(oms_noise_level)
     fr = np.linspace(low_freq_cutoff, (length-1)*2*delta_f, length)
-    s_acc_nu, s_oms_nu = lisa_psd_components(
-                          fr, acc_noise_level, oms_noise_level)
-    omega_len = omega_length(fr, len_arm)
-    psd = (64*(np.sin(omega_len))**2 * (np.sin(2*omega_len))**2 *
-           (s_oms_nu+s_acc_nu*(3+np.cos(2*omega_len))))
+    psd_components = np.array(lisa_psd_components(
+        fr, acc_noise_level, oms_noise_level))
```
HumphreyWang, line 408: I have a minor question: what is the advantage of creating a numpy array and then unpacking them at, for example, line 361? My understanding is this can directly be transferred as a tuple, and then unpack like `s_acc_nu, s_oms_nu = psd_components`. (Or just set these two parameters as keyword arguments instead of `psd_components`)

pycbc/psd/analytical_space.py
```diff
@@ -139,36 +342,41 @@ def analytical_psd_lisa_tdi_1p5_XYZ(length, delta_f, low_freq_cutoff,
...
-    s_acc_nu, s_oms_nu = lisa_psd_components(
-                          fr, acc_noise_level, oms_noise_level)
+    s_acc_nu = psd_components[0]
+    s_oms_nu = psd_components[1]
     omega_len = omega_length(fr, len_arm)
-    psd = 16*(np.sin(omega_len))**2 * (s_oms_nu+s_acc_nu*(3+np.cos(omega_len)))
+    psd = 16*(np.sin(omega_len))**2 * (s_oms_nu +
+                                       s_acc_nu*(3+np.cos(2*omega_len)))
+    if str(tdi) not in ["1.5", "2.0"]:
+        raise Exception("The version of TDI, currently only for 1.5 or 2.0.")
```
HumphreyWang, line 367: I agree with Alex that make these Exceptions more specific, i.e. ValueError.

WuShichao: Thanks, I have replaced all `raise Exception` with `raise ValueError`.

pycbc/psd/analytical_space.py
WuShichao: @HumphreyWang Thanks, that's just my bad habit, I have replaced all of them.

Issue #4638, opened 2024-02-16
Get codeclimate to loook at executables


GarethCabournDavies:
There is a rule on extensions where codeclimate decides which engine to use based on the file extension:

> Furthermore, engines only run their analysis on files with an appropriate file extension for the language being analyzed (.rb for a Ruby engine, .py for a Python engine, etc). Therefore, in most cases you should not need to specify excludes for files in languages other than your target languages.

[https://docs.codeclimate.com/docs/excluding-files-and-folders#auto-generated-file-and-folder-exclusions](this page)

So if a file has no extension, it seems that nothing is done, but we want to use codeclimate to look at all the executables with the correct engine.

I don't know if we can use anything with the shebang to work this out?

GarethCabournDavies:
(I had an idea in #4637, but it didn't work)

titodalcanton:
Maybe worth posing the problem to the IGWN computing folks?

GarethCabournDavies:
> Maybe worth posing the problem to the IGWN computing folks?

So the consensus from that discussion (which I forgot to add at the time) was to update all executables to modules, and then use entry points.

That's quite a bit of work, but can be done one-at-a-time if wanted

GarethCabournDavies:
It may be worth just using flake8 and not relying on the GUI for results - codeclimate seems to have disappeared, and flake8 seems to be quite nice

PR #4665, opened 2024-03-19
Standardize logging - final places

GarethCabournDavies:
Final places where we need to standardise including the logging

This looks like more places than it is because I've added two things in order to simplify logging setup:
 - a 'default_level' kwarg to init_logging, which means we can easily state that logging should happen, but `--verbose` make it a higher level
 - a `to_file` kwarg, which sets up logging to a file instead of to the stderr

### DIFF
--- a/bin/inference/pycbc_inference_plot_gelman_rubin
+++ b/bin/inference/pycbc_inference_plot_gelman_rubin
@@ -22,16 +22,15 @@ import logging
 import matplotlib as mpl; mpl.use("Agg")
 import matplotlib.pyplot as plt
 import sys
-from pycbc import results
+
+from pycbc import results, init_logging, add_common_pycbc_options
 from pycbc import __version__ 
 from pycbc.inference import (gelman_rubin, io, option_utils)
 
 # add options to command line
 parser = io.ResultsArgumentParser(skip_args=['walkers'])
 
-# verbose option
-parser.add_argument("--verbose", action="store_true", default=False,
-                    help="Print logging info.")
+pycbc.add_common_pycbc_options(parser)
 parser.add_argument('--version', action='version', version=__version__,
                     help='show version number and exit')
 
@@ -55,11 +54,7 @@ parser.add_argument("--segment-step", type=int, required=True,
 opts = parser.parse_args()
 
 # setup log
-if opts.verbose:
-    log_level = logging.DEBUG
-else:
-    log_level = logging.WARN
-logging.basicConfig(format="%(asctime)s : %(message)s", level=log_level)
+pycbc.init_logging(opts.verbose)
 
 # enfore that this is not a single iteration
 if opts.iteration is not None:
--- a/bin/plotting/pycbc_banksim_plot_eff_fitting_factor
+++ b/bin/plotting/pycbc_banksim_plot_eff_fitting_factor
@@ -14,17 +14,21 @@
 # You should have received a copy of the GNU General Public License along
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
-""" Plot effective fitting factor from a bunch of point-source files.
+"""
+Plot effective fitting factor vs mass1 and mass2 from a various
+point-source files.
 """
 
 import sys
 import h5py
 import numpy
 import argparse
+import logging
 import matplotlib
 matplotlib.use('Agg')
 import matplotlib.pyplot as plt
 
+from pycbc import init_logging, add_common_pycbc_options
 import pycbc.version
 from pycbc import results
 
@@ -34,7 +38,8 @@ __date__    = pycbc.version.date
 __program__ = "pycbc_banksim_plot_eff_fitting_factor"
 
 parser = argparse.ArgumentParser(usage='',
-    description="Plot effective fitting factor vs mass1 and mass2.")
+    description=__doc__)
+add_common_pycbc_options(parser)
 parser.add_argument("--version", action="version", version=__version__)
 parser.add_argument('--input-files', nargs='+', default=None, required=True,
                     help="List of input files.")
@@ -56,6 +61,8 @@ parser.add_argument('--log-colorbar', action='store_true', default=False,
                          "showing 1 - fitting factor.")
 
 opt = parser.parse_args()
+
+init_logging(opt.verbose)
 
 m1 = []
 m2 = []
--- a/bin/plotting/pycbc_page_injtable
+++ b/bin/plotting/pycbc_page_injtable
@@ -2,20 +2,28 @@
 """ Make a table of found injection information
 """
 import argparse, h5py, numpy as np, pycbc.results, pycbc.detector, sys
+from itertools import combinations
+
+from pyccb import add_common_pycbc_options, init_logging 
 from pycbc.types import MultiDetOptionAction
 import pycbc.pnutils, pycbc.events
 import pycbc.version
-from itertools import combinations
 
 
-parser = argparse.ArgumentParser()
-parser.add_argument("--version", action="version", version=pycbc.version.git_verbose_msg)
-parser.add_argument('--injection-file', help='HDF File containing the matched injections')
-parser.add_argument('--single-trigger-files', nargs='*', help="HDF format single detector trigger files", action=MultiDetOptionAction)
-parser.add_argument('--verbose', action='count')
+parser = argparse.ArgumentParser(description=__doc__)
+add_common_pycbc_options(parser)
+parser.add_argument("--version", action="version",
+                    version=pycbc.version.git_verbose_msg)
+parser.add_argument('--injection-file',
+                    help='HDF File containing the matched injections')
+parser.add_argument('--single-trigger-files', nargs='*',
+                    action=MultiDetOptionAction,
+                    help="HDF format single detector trigger files")
 parser.add_argument('--show-missed', action='store_true')
 parser.add_argument('--output-file')
 args = parser.parse_args()
+
+init_logging(args.verbose)
 
 f = h5py.File(args.injection_file,'r')
 inj = f['injections']
--- a/bin/plotting/pycbc_plot_bank_corner
+++ b/bin/plotting/pycbc_plot_bank_corner
@@ -26,7 +26,7 @@ import h5py
 import numpy as np
 import argparse
 import logging
-from pycbc import init_logging
+
 import pycbc.version
 from pycbc import __version__
 from pycbc.results.plot import (add_style_opt_to_parser, set_style_from_cli)
@@ -41,13 +41,11 @@ conversion_options = bconv.conversion_options
 
 parser = argparse.ArgumentParser(usage='pycbc_plot_bank_corner [--options]',
                     description=__doc__)
+pycbc.add_common_pycbc_options(parser)
 parser.add_argument("--version",
     action="version",
     version=__version__,
     help="Prints version information.")
-parser.add_argument("--verbose",
-    action='count',
-    help="Output progress and information to stderr")
 parser.add_argument("--bank-file",
     required=True,
     help="The bank file to read in and plot")
@@ -112,7 +110,7 @@ parser.add_argument('--dpi',
 add_style_opt_to_parser(parser)
 args = parser.parse_args()
 
-init_logging(args.verbose)
+pycbc.init_logging(args.verbose)
 set_style_from_cli(args)
 
 mins, maxs = option_utils.plot_ranges_from_cli(args)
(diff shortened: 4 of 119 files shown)


GarethCabournDavies:
To check that all executables have the standardized logging, when we check the repo for files which _don't_ have `init_logging`, I get no output, i.e. (`grep -Le` means list of files which _don't_ contain the string):
```
$ grep -Le init_logging bin/**/pycbc_*
```
returns nothing

The common pycbc options added to the parser is slightly more complicated, as some of the pygrb codes use a common parser setup, so we see:
```
$ grep -Le common_pycbc bin/**/pycbc_*
bin/pygrb/pycbc_pygrb_efficiency
bin/pygrb/pycbc_pygrb_page_tables
bin/pygrb/pycbc_pygrb_plot_chisq_veto
bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
bin/pygrb/pycbc_pygrb_plot_injs_results
bin/pygrb/pycbc_pygrb_plot_null_stats
bin/pygrb/pycbc_pygrb_plot_skygrid
bin/pygrb/pycbc_pygrb_plot_snr_timeseries
bin/pygrb/pycbc_pygrb_plot_stats_distribution
```

but the common options are called within `pygrb_initialize_plot_parser`, so we can see the same group of files which have that option (`grep -le` means list of files which _do_ contain the string):
```
$ grep -le pygrb_initialize_plot_parser bin/**/pycbc_*
bin/pygrb/pycbc_pygrb_efficiency
bin/pygrb/pycbc_pygrb_page_tables
bin/pygrb/pycbc_pygrb_plot_chisq_veto
bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
bin/pygrb/pycbc_pygrb_plot_injs_results
bin/pygrb/pycbc_pygrb_plot_null_stats
bin/pygrb/pycbc_pygrb_plot_skygrid
bin/pygrb/pycbc_pygrb_plot_snr_timeseries
bin/pygrb/pycbc_pygrb_plot_stats_distribution
```

GarethCabournDavies:
Codeclimate failures are unrelated to these changes

GarethCabournDavies:
Nte i realised i skipped a couple of places for the init_logging check, and saw
```
grep -Le init_logging bin/pycbc_*
bin/pycbc_copy_output_map
bin/pycbc_live_nagios_monitor
bin/pycbc_splitbank
bin/pycbc_split_inspinj
bin/pycbc_stageout_failed_workflow
bin/pycbc_submit_dax
```

of these, `pycbc_copy_output_map`, `pycbc_stageout_failed_workflow` and `pycbc_submit_dax` are bash scripts, the others do not have any logging

.pylintrc
```diff
@@ -0,0 +1,3 @@
+[BASIC]
```
GarethCabournDavies, line 1: This silences the codeclimate "logger is not a valid name" complaints

GarethCabournDavies: The other variables are the pyint defaults

pycbc/__init__.py
```diff
@@ -86,7 +86,8 @@ def add_common_pycbc_options(parser):
                             'logging at the info level, but -vv or '
                             '--verbose --verbose provides debug logging.')
 
-def init_logging(verbose=False,
+
+def init_logging(verbose=False, default_level=0, to_file=None,
```
GarethCabournDavies, line 90: This is probably the best place to start the review - the more significant change is here:
 - We add a default_level kwarg so that the code can define the default level of logging to be higher, but --verbose still adds incremental logging levels
 - We add a to_file kwarg, so that we can define logging to a file through the init_logging interface. If wanting logging to more than one file, or to both stderr and the file, we can call init_logging twice with different to_file values

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -57,9 +61,8 @@ def pygrb_initialize_plot_parser(description=None, version=None):
     formatter_class = argparse.ArgumentDefaultsHelpFormatter
     parser = argparse.ArgumentParser(description=description,
                                      formatter_class=formatter_class)
+    add_common_pycbc_options(parser)
```
GarethCabournDavies, line 64: Outwith this change, all the changes in `pycbc` directory are:
 - adding the named logger to the module
 - updating `logging` --> `logger` to use the named logger
 - some reordering/reformatting of imports

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -66,9 +66,6 @@ def check_gracedb_for_event(gdb_handle, query, far):
     return False
 
 
-logging.basicConfig(format='%(asctime)s:%(levelname)s : %(message)s',
-                    level=logging.INFO)
-
```
ArthurTolley, line 71: I think based on level=INFO the init_logging should include default_level = 1

bin/pygrb/pycbc_grb_trig_cluster
ArthurTolley, line 1: This init_logging in this file still includes the formatting options, is that what we want?

bin/pygrb/pycbc_pygrb_minifollowups
```diff
@@ -122,7 +123,7 @@ wf.add_workflow_settings_cli(parser, include_subdax_opts=True)
 ppu.pygrb_add_bestnr_cut_opt(parser)
 args = parser.parse_args()
 
-init_logging(args.verbose, format="%(asctime)s: %(levelname)s: %(message)s")
+pycbc.init_logging(args.verbose, format="%(asctime)s: %(levelname)s: %(message)s")
```
ArthurTolley, line 126: Again, is the format what we want here? I notice these are pygrb files so they might be what is needed.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
```diff
@@ -33,7 +33,9 @@ from matplotlib import pyplot as plt
 from matplotlib import rc
 import numpy
 import scipy
+
 import pycbc.version
+from pycbc.detector import Detector
 from pycbc import init_logging
 from pycbc.detector import Detector
```
ArthurTolley, line 40: Duplicate import of pycbc.detector.Detector

bin/pygrb/pycbc_pygrb_pp_workflow
```diff
@@ -47,9 +47,8 @@ __program__ = "pycbc_pygrb_pp_workflow"
 # =============================================================================
 # Use the standard workflow command-line parsing routines.
 parser = argparse.ArgumentParser(description=__doc__[1:])
+pycbc.add_common_pycbc_options(parser)
```
ArthurTolley, line 50: `import pycbc`?

examples/cal/foton_filter_esd_saturation/pycbc_foton_filter
```diff
@@ -80,8 +83,7 @@ parser.add_argument("--sample-rate", type=int, required=True,
 opts = parser.parse_args()
 
 # setup log
-logging_level = logging.DEBUG
-logging.basicConfig(format='%(asctime)s : %(message)s', level=logging_level)
+pycbc.init_logging(args.verbose, default_level=1)
```
ArthurTolley, line 86: logging level was DEBUG so I think default level should be 2?

ArthurTolley (changes requested):
I'm sure that 99% of my changes (the most I've ever done by far) are related to simple changes or wanting to understand a bit more the 'standardised' part of things. I imagine I'm just misinformed on a few pythonic things, such as my `import pycbc` comment so if you answer it once then feel free to ignore the rest.

bin/plotting/pycbc_page_injtable
```diff
@@ -2,21 +2,29 @@
 """ Make a table of found injection information
 """
 import argparse, h5py, numpy as np, pycbc.results, pycbc.detector, sys
```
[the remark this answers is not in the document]
ArthurTolley: I understand now what it is saying (with the `as` statement in the middle) but I think to match the rest of the files this definitely needs to be split up :)

bin/pygrb/pycbc_pygrb_minifollowups
ArthurTolley: Checking `__init__.py` the default for the format option is 'format='%(asctime)s %(levelname)s : %(message)s'):' which when compared to this one is missing a colon after '(asctime)s '

bin/inference/pycbc_inference_plot_gelman_rubin
```diff
@@ -55,11 +55,7 @@ parser.add_argument("--segment-step", type=int, required=True,
 opts = parser.parse_args()
 
 # setup log
-if opts.verbose:
-    log_level = logging.DEBUG
-else:
-    log_level = logging.WARN
-logging.basicConfig(format="%(asctime)s : %(message)s", level=log_level)
```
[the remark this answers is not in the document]
GarethCabournDavies: If using the default level of zero without any `--verbose`, then you will get `logging.WARNING`.

The logic is in `__init__.py`, where the logging level boils down to logging.WARNING - (default_level + n_verbose) * 10.

logging.WARNING is numerically 30, 20 is info, 10 is debug.

bin/inference/pycbc_validate_test_posterior
```diff
@@ -5,15 +5,21 @@ test posterior model.
 import sys
 import numpy
 import argparse
-from matplotlib import use; use('Agg')
+from matplotlib import use
+use('Agg')
```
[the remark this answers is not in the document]
GarethCabournDavies: Again, this is dependent upon the style of the file, but I will look at standardising

bin/inference/pycbc_inference_plot_gelman_rubin
```diff
@@ -22,16 +22,16 @@ import logging
 import matplotlib as mpl; mpl.use("Agg")
```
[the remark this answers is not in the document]
GarethCabournDavies: This should be split, yes

That is the kind of thing that pylint will check once I get it to work on executables (#4638)

bin/plotting/pycbc_page_injtable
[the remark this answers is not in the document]
GarethCabournDavies: yup, and splitting makes it clearer, especially in the case youve just pointed out

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -28,7 +28,6 @@ import argparse
 import logging
 from textwrap import wrap
 
-from pycbc import init_logging
```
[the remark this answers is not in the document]
GarethCabournDavies: There is an implicit import (Otherwise this would fail CI tests), but I will add the explicit one back in

bin/pygrb/pycbc_grb_trig_cluster
GarethCabournDavies: I left it in so that the logging would remain as it was before. Im not sure if the grb search uses the outputs for monitoring in an automatic way, and this would break that

bin/pygrb/pycbc_pygrb_minifollowups
GarethCabournDavies: Yup, didnt want to break things, just in case

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
GarethCabournDavies: Thanks - again this would be solved once the above mentioned issue is solved

examples/live/check_results.py
```diff
@@ -195,7 +195,7 @@ def check_found_events(args):
 parser.add_argument('--detectors', type=str, required=True, nargs='+')
 args = parser.parse_args()
 
-log.basicConfig(level=log.INFO, format='%(asctime)s %(message)s')
+pycbc.init_logging(1)
```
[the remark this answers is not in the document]
GarethCabournDavies: Ive put the add_common_pycbc_options here so that the log level can be changed (It couldnt before, so that was why it was hard-set)

bin/inference/pycbc_inference_plot_gelman_rubin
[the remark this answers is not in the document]
ArthurTolley: So to get debugging for the logging here you would need to give verbose twice?

bin/plotting/pycbc_banksim_plot_eff_fitting_factor
```diff
@@ -15,7 +15,7 @@
 # with this program; if not, write to the Free Software Foundation, Inc.,
 # 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 """
-Plot effective fitting factor vs mass1 and mass2 from various
+Plot effective fitting factor vs mass1 and mass2 from a various
```
ArthurTolley, line 18: 'from a various point-source files.'

bin/inference/pycbc_inference_plot_gelman_rubin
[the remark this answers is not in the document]
GarethCabournDavies: that's right

ArthurTolley (approved):
lfg

PR #4666, opened 2024-03-19
improve consistency of example / defaults for variable_params

ahnitz:
@cdcapano This is just a small change to switch some of the examples to default to the more common convention now of static_params / variable_params. Also the default in one function that wasn't changed over is now done.

### DIFF
--- a/pycbc/distributions/__init__.py
+++ b/pycbc/distributions/__init__.py
@@ -110,8 +110,8 @@ def _convert_liststring_to_list(lstring):
 
 
 def read_params_from_config(cp, prior_section='prior',
-                            vargs_section='variable_args',
-                            sargs_section='static_args'):
+                            vargs_section='variable_params',
+                            sargs_section='static_params'):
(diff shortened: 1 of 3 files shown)


pycbc/distributions/__init__.py
```diff
@@ -122,7 +122,7 @@ def read_params_from_config(cp, prior_section='prior',
         Check that priors exist in the given section. Default is 'prior.'
     vargs_section : str, optional
         The section to get the parameters that will be varied/need priors
-        defined for them. Default is 'variable_args'.
+        defined for them. Default is 'variable_params'.
     sargs_section : str, optional
         The section to get the parameters that will remain fixed. Default is
         'static_args'.
```
titodalcanton, line 128: Looks like line 128 also needs updating.

cdcapano (approved):
Thanks for fixing this. Looks good. Too bad code climate failed, but the other tests look fine.

PR #4670, opened 2024-03-22
Allow pycbc_fit_sngls_over_multiparam to take in multiple files

GarethCabournDavies:
Update `pycbc_fit_sngls_over_multiparam` to take in multiple --template-fit-file inputs

#### Why?
This will be useful for e.g. if we generate `pycbc_fit_sngls_by_template` for part of the data, e.g. a day, and then want to smooth over many days at once, we could do this in a rolling window in the live search

#### How?
 - Firstly, read in the files and concatenate datasets together: `template_id`, median_sigma`, `count_above_thresh`, `count_in_template` and `fit_coeff`.
 - While doing this, check that all the attributes which should match (`ifo`, `sngl_stat`, `fit_threshold`), do. Otherwise, RuntimeError
 - For each of `nabove`, `ntotal` and `invalphan`, smooth identical template_ids over files before doing the smoothing over parameters as before. This works as these datasets are linearly added for smoothing.

#### Nuances
 - Here we take a mean of `median_sigma` over different files. `median_sigma` is meant to be a representative value for each template in that IFO, and taking the median previously removes any outliers. So taking the mean after this should be safe
 - The `fit_by_template` output group now contains the `fit_by_template` values' mean average over files - like with `median_sigma`, this group is later used to find particularly noisy, isolated templates, so this should be okay

#### Testing
I have run the new code with one input file against the old code, and tested the output files against one another. This was done for all different smoothing methods [see note on n_closest below]. Every dataset and (shared) attribute was compared between new/old outputs and are all identical in this case.

The files differ by more than just numerical noise in a couple of minor ways:
 - The attributes in the output file now include the `sngl_stat` and `fit_function` from the `fit_by_template` attributes
 - The `fit_by_template/count_above_thresh` and `fit_by_template/count_in_template` datasets change from int64 to float64 when more than one file is given, to allow for the smoothing over files to give a non-integer count values.

For taking in multiple files, I have tested providing the same file two / three times, these should result in exactly the same output, e.g. mean(0,1,2,3) = mean(0,1,2,3,0,1,2,3)

In these tests, all new/old datasets are the same to within a `numpy.isclose`; more stringent comparison shows that all new/old datasets are identical except there is variation in the `median_sigma` dataset. `median_sigma` values change by up to 9e-10 (values are in the thousands, and we take a log of this in the statistic so I don't think this matters too much).

#### Efficiency
The smoothing over files uses the similar methodology to the efficient 1d tophat smoothing, so this step takes less than a second for a full O4 offline bank

#### Other notes
 - I think this change means we could parallelise `pycbc_fit_sngls_by_template` in the offline search to work over different parts of the bank, but haven't thought too much about how to implement that or whether it would even be useful
 - I have additionally fixed a bug in selecting templates for the `n_closest` smoothing method. I did this for both the 'master' copy of the code, and the new version. The results after this fix were the same as for other smoothing methods.

### DIFF
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -292,7 +292,7 @@ for filename in args.template_fit_file:
                     logging.warning(
                         "Fit files correspond to different IFOs: %s and %s, "
                         "only %s is being used for output file attributes",
-                        attr_dict[k], fits.attrs[k]
+                        attr_dict[k], fits.attrs[k], attr_dict[k],
                     )
                 continue
             elif not attr_dict[k] == fits.attrs[k]:
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -265,8 +265,7 @@ for filename in args.template_fit_file:
         for k in fits.attrs.keys():
             if k == 'analysis_time':
                 # For this, we want the mean, not check that it is identical
-                analysis_time += fits.attrs['analysis_time'] \
-                    if 'analysis_time' in fits.attrs else 0
+                analysis_time += fits.attrs['analysis_time']
                 continue
             if k not in attr_dict:
                 # This is the first time this attribute
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -253,10 +253,6 @@ init_logging(args.verbose)
 analysis_time = 0
 attr_dict = {}
 
-# This will be used to check the fit files have the same number of templates
-# (therefore probably using the same bank)
-num_templates = None
-
 # These end up as n_files * n_templates arrays
 tid = numpy.array([], dtype=int)
 nabove = numpy.array([], dtype=int)
@@ -265,10 +261,14 @@ alpha = numpy.array([], dtype=float)
 median_sigma = numpy.array([], dtype=float)
 
 logging.info("Loading input template fits")
+
+# This will be used to check the fit files have the same number of templates
+# (therefore probably using the same bank)
+num_templates = None
+
 for filename in args.template_fit_file:
     with h5py.File(filename, 'r') as fits:
         if num_templates is None:
-            # First file; store number of templates
             num_templates = fits['template_id'].size
         elif not num_templates == fits['template_id'].size:
             raise RuntimeError(
@@ -285,7 +285,15 @@ for filename in args.template_fit_file:
                 # It's the first time we encounter this attribute
                 attr_dict[k] = fits.attrs[k]
             elif k == 'ifo':
-                # We don't mind if this attribute is different
+                # We don't mind if this attribute is different, however the
+                # output attributes will only correspond to the first file's
+                # IFO. Warn if different IFOs are being used.
+                if not attr_dict[k] == fits.attrs[k]:
+                    logging.warning(
+                        "Fit files correspond to different IFOs: %s and %s, "
+                        "only %s is being used for output file attributes",
+                        attr_dict[k], fits.attrs[k]
+                    )
                 continue
             elif not attr_dict[k] == fits.attrs[k]:
                 # Check that the files are consistent with one another
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -245,13 +245,29 @@ assert len(args.log_param) == len(args.fit_param) == len(args.smoothing_width)
 
 init_logging(args.verbose)
 
-fits = h5py.File(args.template_fit_file, 'r')
+with h5py.File(args.template_fit_file, 'r') as fits:
+    # get attributes from the template-level fit
+    ifo = fits.attrs['ifo']
+    stat_threshold = fits.attrs['stat_threshold']
+    analysis_time = fits.attrs['analysis_time'] if 'analysis_time' in fits.attrs else None
 
-# get the ifo from the template-level fit
-ifo = fits.attrs['ifo']
+    # get template id and template parameter values
+    tid = fits['template_id'][:]
+    nabove = fits['count_above_thresh'][:]
+    ntotal = fits['count_in_template'][:]
+    alpha = fits['fit_coeff'][:]
+    try:
+        median_sigma = fits['median_sigma'][:]
+    except KeyError:
+        logging.info('Median_sigma dataset not present in input file')
+        median_sigma = None
 
-# get template id and template parameter values
-tid = fits['template_id'][:]
+    if args.output_fits_by_template:
+        fbtdict = {}
+        for k in ['count_above_thresh', 'fit_coeff', 'count_in_template']:
+            fbtdict[k] = fits[k][:]
+
+
 
 logging.info('Calculating template parameter values')
 bank = h5py.File(args.bank_file, 'r')
@@ -273,11 +289,9 @@ for param, slog in zip(args.fit_param, args.log_param):
     else:
         raise ValueError("invalid log param argument, use 'true', or 'false'")
 
-nabove = fits['count_above_thresh'][:]
-ntotal = fits['count_in_template'][:]
 # For an exponential fit 1/alpha is linear in the trigger statistic values
 # so calculating weighted sums or averages of 1/alpha is appropriate
-invalpha = 1. / fits['fit_coeff'][:]
+invalpha = 1. / alpha
 invalphan = invalpha * nabove
 
 nabove_smoothed = []
@@ -377,10 +391,8 @@ outfile['template_id'] = tid
 outfile['count_above_thresh'] = nabove_smoothed
 outfile['fit_coeff'] = alpha_smoothed
 outfile['count_in_template'] = ntotal_smoothed
-try:
-    outfile['median_sigma'] = fits['median_sigma'][:]
-except KeyError:
-    logging.info('Median_sigma dataset not present in input file')
+if median_sigma is not None:
+    outfile['median_sigma'] = median_sigma
 
 for param, vals, slog in zip(args.fit_param, parvals, args.log_param):
     if slog in ['false', 'False', 'FALSE']:
@@ -391,13 +403,13 @@ for param, vals, slog in zip(args.fit_param, parvals, args.log_param):
 if args.output_fits_by_template:
     outfile.create_group('fit_by_template')
     for k in ['count_above_thresh', 'fit_coeff', 'count_in_template']:
-        outfile['fit_by_template'][k] = fits[k][:]
+        outfile['fit_by_template'][k] = fbtdict[k]
 
 # Add metadata, some is inherited from template level fit
 outfile.attrs['ifo'] = ifo
-outfile.attrs['stat_threshold'] = fits.attrs['stat_threshold']
-if 'analysis_time' in fits.attrs:
-    outfile.attrs['analysis_time'] = fits.attrs['analysis_time']
+outfile.attrs['stat_threshold'] = stat_threshold
+if analysis_time is not None:
+    outfile.attrs['analysis_time'] = analysis_time
 
 # Add a magic file attribute so that coinc_findtrigs can parse it
 outfile.attrs['stat'] = ifo + '-fit_coeffs'
--- a/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
+++ b/bin/all_sky_search/pycbc_fit_sngls_over_multiparam
@@ -177,8 +177,10 @@ parser = argparse.ArgumentParser(usage="",
 pycbc.add_common_pycbc_options(parser)
 parser.add_argument("--version", action=pycbc.version.Version)
 parser.add_argument("--template-fit-file", required=True, nargs='+',
-                    help="hdf5 file containing fit coefficients for each"
-                         " individual template. Required")
+                    help="hdf5 file(s) containing fit coefficients for each "
+                         "individual template. Can smooth over multiple "
+                         "files provided they correspond to the same bank "
+                         "and fitting settings. Required")
 parser.add_argument("--bank-file", required=True,
                     help="hdf file containing template parameters. Required")
 parser.add_argument("--output", required=True,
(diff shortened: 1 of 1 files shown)


bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -174,7 +174,7 @@ parser = argparse.ArgumentParser(usage="",
 
 pycbc.add_common_pycbc_options(parser)
 parser.add_argument("--version", action=pycbc.version.Version)
-parser.add_argument("--template-fit-file", required=True,
+parser.add_argument("--template-fit-file", required=True, nargs='+',
                     help="hdf5 file containing fit coefficients for each"
```
ArthurTolley, line 180: Can the description be updated to say this can take multiple files?

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -245,13 +245,98 @@ assert len(args.log_param) == len(args.fit_param) == len(args.smoothing_width)
...
+                # has been encountered
+                attr_dict[k] = fits.attrs[k]
+            elif not attr_dict[k] == fits.attrs[k]:
+                # Check that the files are consistent with one another
+                err_msg = f"Input files are not consistent, file {filename} "
+                err_msg += f"has attribute {k} of {fits.attrs[k]}, whereas "
+                err_msg += f"previous files have value {attr_dict[k]}"
+                raise RuntimeError(err_msg)
+
+        analysis_time += fits.attrs['analysis_time'] if 'analysis_time' in fits.attrs else 0
```
ArthurTolley, line 278: Could this be moved up to the if statement above: 'if k == 'analysis_time':'

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
GarethCabournDavies: Thought I'd done that but must have missed it. Added now

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -245,13 +245,98 @@ assert len(args.log_param) == len(args.fit_param) == len(args.smoothing_width)
...
+# For an exponential fit 1/alpha is linear in the trigger statistic values
+# so calculating weighted sums or averages of 1/alpha is appropriate
+invalpha = 1. / alpha
+invalphan = invalpha * nabove
+
+analysis_time /= len(args.template_fit_file)
+
+# If we are using multiple files, we should 'smooth' within each template
+# before smoothing over templates going forward. As the templates will
+# all be have identical template_ids and have the same number of values,
```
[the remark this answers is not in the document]
GarethCabournDavies: Sounds a bit austin powers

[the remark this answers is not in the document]
ArthurTolley: 😂 maybe we should have pycbc comment easter eggs

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -262,7 +264,9 @@ for filename in args.template_fit_file:
         # get attributes from the template-level fit
         for k in fits.attrs.keys():
             if k == 'analysis_time':
-                # We handle this differently
+                # For this, we want the mean, not check that it is identical
+                analysis_time += fits.attrs['analysis_time'] \
+                    if 'analysis_time' in fits.attrs else 0
```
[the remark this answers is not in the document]
GarethCabournDavies: :facepalm:

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -88,7 +88,11 @@ def smooth_tophat(nabove, invalphan, ntotal, dists):
                             idx_within_area)
 
 
-def smooth_n_closest(nabove, invalphan, ntotal, dists, total_trigs=500):
+_default_total_trigs = 500
```
tdent, line 91: add a comment here to explain what this value will be used for (the name is slightly cryptic)

tdent:
FWIW, I just noticed the indentation of the input checking line `err_txt = "--smoothing-keywords must take input` (etc) looks off .. 8 spaces not 4?

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -174,9 +176,11 @@ parser = argparse.ArgumentParser(usage="",
...
 pycbc.add_common_pycbc_options(parser)
 parser.add_argument("--version", action=pycbc.version.Version)
-parser.add_argument("--template-fit-file", required=True,
-                    help="hdf5 file containing fit coefficients for each"
-                         " individual template. Required")
+parser.add_argument("--template-fit-file", required=True, nargs='+',
+                    help="hdf5 file(s) containing fit coefficients for each "
+                         "individual template. Can smooth over multiple "
+                         "files provided they correspond to the same bank "
+                         "and fitting settings. Required")
```
tdent, line 183: What happens if you give multiple files with different banks?  

Same question, for multiple files with the same bank but different fitting methods?

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -245,13 +249,108 @@ assert len(args.log_param) == len(args.fit_param) == len(args.smoothing_width)
...
-ifo = fits.attrs['ifo']
+# For an exponential fit 1/alpha is linear in the trigger statistic values
+# so calculating weighted sums or averages of 1/alpha is appropriate
+invalpha = 1. / alpha
+invalphan = invalpha * nabove
+
+# convert the sum above into a mean
+analysis_time /= len(args.template_fit_file)
+
+# If we are using multiple files, we should 'smooth' within each template
```
tdent, line 300:
```suggestion
# If we are using multiple files, we should average within each template
```
I think 'smooth' is a bit confusing here ..

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -245,13 +249,108 @@ assert len(args.log_param) == len(args.fit_param) == len(args.smoothing_width)
...
+                # has been encountered
+                attr_dict[k] = fits.attrs[k]
+            elif not attr_dict[k] == fits.attrs[k]:
+                # Check that the files are consistent with one another
+                err_msg = f"Input files are not consistent, file {filename} "
+                err_msg += f"has attribute {k} of {fits.attrs[k]}, whereas "
+                err_msg += f"previous files have value {attr_dict[k]}"
+                raise RuntimeError(err_msg)
+
+        # get template id and template parameter values
```
tdent, line 281: I guess it will be clearer to see what happens in the next step if you say explicitly that these end up being big arrays each of which contains n_templates * n_files entries

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -245,13 +249,108 @@ assert len(args.log_param) == len(args.fit_param) == len(args.smoothing_width)
...
+# so calculating weighted sums or averages of 1/alpha is appropriate
+invalpha = 1. / alpha
+invalphan = invalpha * nabove
+
+# convert the sum above into a mean
+analysis_time /= len(args.template_fit_file)
+
+# If we are using multiple files, we should 'smooth' within each template
+# before smoothing over templates going forward. As the templates will
+# all have identical template_ids and have the same number of values,
```
tdent, line 302: This explanation is not quite clear, which part of this calculation is 'taking the mean of means'?  The discussion also seems to mix up the implementation aspect (getting entries with identical tids) with the correctness of what is implemented. 

What I think this is saying is that out of the long array of n_templates * n_files values, averaging within each template involves getting the n_files occurrences which have the same tid, that explains the first part about implementation. 

Then you want to argue that the averaging here does not bias the final result. I am not sure if that _necessarily_ requires the same number of per-file entries for each template, but the linearity argument basically takes care of it

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -245,13 +249,108 @@ assert len(args.log_param) == len(args.fit_param) == len(args.smoothing_width)
...
-# get template id and template parameter values
-tid = fits['template_id'][:]
+    tid = tid_unique
+    nabove = (nasum[right] - nasum[left]) / num
+    invalphan = (invsum[right] - invsum[left]) / num
+    ntotal = (ntsum[right] - ntsum[left]) / num
+    median_sigma = (mssum[right] - mssum[left]) / num
+
+if args.output_fits_by_template:
+    # Save fit_by_template values for saving later
```
tdent, line 329: too much 'saving' here .. 
```suggestion
    # Store fit_by_template values for output file
```

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -245,13 +249,108 @@ assert len(args.log_param) == len(args.fit_param) == len(args.smoothing_width)
...
+    # Note that these will be the values averaged over the same template in
+    # different files in the case that we have two or more input files
```
tdent, line 331:
```suggestion
    # For more than one input fit file, these values are averaged over the same template
    # in different files
```

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
tdent: I guess it's also true that all the files have to be from the same ifo?  (Is this necessary to be hard-coded as impossible, though?)

tdent (changes requested):
See multiple suggestions for rewriting/adding comments.  Otherwise, this seems reasonable

tdent:
I think the main question on implementation is why we necessarily should forbid trying to average fit results from different ifos and/or different fit settings.  Doing so would be messy in terms of recording this (as attrs or whatever), but basically the job of checking that the inputs are sane/consistent should be done outside this code unless it would make the calculation impossible or meaningless.

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -97,13 +101,11 @@ def smooth_n_closest(nabove, invalphan, ntotal, dists, total_trigs=500):
...
-    # start at closest
-    # total_trigs, if supplied on command line will be a str so convert to int
-    while n_triggers < int(total_trigs):
-        n_triggers += nabove[dist_sort[n_triggers]]
-        templates_required += 1
-    logging.debug("%d templates required to obtain %d triggers",
-                  templates_required, n_triggers)
+    # starting at closest
+    ntcs = nabove[dist_sort].cumsum()
+    templates_required = numpy.searchsorted(ntcs, total_trigs) + 1
```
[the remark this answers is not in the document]
GarethCabournDavies: Yes that's it. The problem before was that `dist_sort[n_triggers]` was basically nonsense, as it was indexing on the number of triggers rather than the template id. (No idea how it didn't error out though...).

The solution removes the manual cumulative sum and then uses searchsorted rather than a manual search.

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -245,13 +249,108 @@ assert len(args.log_param) == len(args.fit_param) == len(args.smoothing_width)
...
+alpha = numpy.array([], dtype=float)
+median_sigma = numpy.array([], dtype=float)
+
+logging.info("Loading input template fits")
+for filename in args.template_fit_file:
+    with h5py.File(filename, 'r') as fits:
+        # get attributes from the template-level fit
+        for k in fits.attrs.keys():
+            if k == 'analysis_time':
+                # For this, we want the mean, not check that it is identical
```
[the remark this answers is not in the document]
GarethCabournDavies:
```suggestion
                # For this attribute only, we want the mean
```

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
GarethCabournDavies: Different banks will not actually error - I thought that I had added that check, so I will add that.

As the bank file is needed to get the parameters, it doesn't make sense to use files from different banks, unless I make it so we can use multiple bank files as well, but that could become complicated.

Errors will be raised (during attribute checks) if ifo, fitting methods, or stat threshold are different. The latter two I think would make the averaging nonsensical, but different IFOs should be allowed, like you say the attributes will become difficult

GarethCabournDavies:
> I think the main question on implementation is why we necessarily should forbid trying to average fit results from different ifos and/or different fit settings. Doing so would be messy in terms of recording this (as attrs or whatever), but basically the job of checking that the inputs are sane/consistent should be done outside this code unless it would make the calculation impossible or meaningless.

Different IFOs should be okay to be used, though I'm not sure how attrs would be set in this case, and we would not be able to use this in the later statistic, so I think disallowing this case is reasonable.

I think different fit settings _would_ make the calculation meaningless; definitely `fit_function`. I could be persuaded on `stat_threshold`, but I think that estimator would be biased, and again the statistic calculation later would not make sense.

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -236,22 +241,133 @@ for inputstr in smooth_kwargs:
...
 
 init_logging(args.verbose)
 
-fits = h5py.File(args.template_fit_file, 'r')
+analysis_time = 0
+attr_dict = {}
+
+# This will be used to check the fit files have the same number of templates
+# (therefore probably using the same bank)
+num_templates = None
```
tdent, line 258: can this be defined closer to where it is used? eg directly before the loop over template fit files.  If so then the second line of comment can probably be removed as it will be visually close enough to the check itself

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
```diff
@@ -236,22 +241,133 @@ for inputstr in smooth_kwargs:
...
+        for k in fits.attrs.keys():
+            if k == 'analysis_time':
+                # For this attribute only, we want the mean
+                analysis_time += fits.attrs['analysis_time']
+                continue
+            if k not in attr_dict:
+                # It's the first time we encounter this attribute
+                attr_dict[k] = fits.attrs[k]
+            elif k == 'ifo':
+                # We don't mind if this attribute is different
```
tdent, line 288: note here also that the output file attrs will only be for one ifo even if more than one is input?

tdent (changes requested):
Minor changes on comments / line ordering

bin/all_sky_search/pycbc_fit_sngls_over_multiparam
GarethCabournDavies: I've added this into the comment, and also added a warning if files from multiple IFOs are used

tdent (approved):

spxiwh:
@GarethCabournDavies I note that this patch introduced the efficiency issue the #4957  fixed. Therefore there is no reason to merge #4957 onto the v23 release branch, because it does not include this patch.

PR #4671, opened 2024-03-23
Properly fix ASD plot minimum value

tdent:
Rather than a hard coded minimum use the minimum value we are going to plot.  Also put the calculation of the asd we are plotting on a single line rather than splitting over 2.

### DIFF
--- a/pycbc/results/psd.py
+++ b/pycbc/results/psd.py
@@ -51,15 +51,18 @@ def generate_asd_plot(psddict, output_filename):
     from matplotlib import pyplot as plt
     asd_fig, asd_ax = plt.subplots(1)
     asd_min = [3E-24]
+
     for ifo in sorted(psddict.keys()):
+        curr_psd = psddict[ifo]
         # Can't plot log(0) so start from point 1
-        asd_to_plot = psddict[ifo][1:] ** 0.5 / DYN_RANGE_FAC
+        asd_to_plot = curr_psd[1:] ** 0.5 / DYN_RANGE_FAC
         asd_min.append(min(asd_to_plot))
         asd_ax.loglog(curr_psd.sample_frequencies[1:],
                       asd_to_plot,
                       c=ifo_color(ifo),
                       label=ifo)
 
+    asd_ax.grid(True)
     asd_ax.legend()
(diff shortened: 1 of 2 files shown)


titodalcanton:
I actually prefer the fixed y range, I would just set the minimum to 1e-25 which should be good for the foreseeable future. Also, can you add a grid while you are at it?

titodalcanton:
There is also an actual error caught by the tests:
```
>           asd_ax.loglog(curr_psd.sample_frequencies[1:],
                          asd_to_plot,
                          c=ifo_color(ifo),
                          label=ifo)
E           NameError: name 'curr_psd' is not defined
```

titodalcanton:
> Why is a fixed y range preferable? If we have points that are 'off the plot' in terms of being too low (due to eg a calibration problem), making the plot scale to fit them in would definitely diagnose the problem.

I generally prefer the axes to stay fixed whenever there is a chance that I might want to compare two such plots side-by-side to spot small differences. If the axes change depending on the data, that comparison becomes harder.

If there is a problem that makes the ASD go below reasonable values, we would immediately notice the line drop below the (fixed) minimum y value, and look at more proper ASD/PSD plots elsewhere anyway.

I will not hold up your choice any further, but please at least lower the "default" minimum y value to 1e-24.

titodalcanton:
One reason *not* to set the y range based on the minimum of `asd_to_plot` is that the low-frequency items of that array are effectively garbage due to low-pass filters, inverse truncation etc. So you may want to actually take the minimum only above a certain "safe" index corresponding to the low-frequency cutoff.

tdent:
Oh, you mean the minimum for 'silly' frequencies below 10Hz ⬇️  .. 
Yeah, that could be tricky to keep under control

tdent:
Hopefully CI should pass now.

tdent:
@titodalcanton any more on this?  Tests are passing ...

titodalcanton (approved):
Looks good. Not sure why the BBHx change shows up in the diff here since that is already on master, but that should not be a problem.

tdent:
I guess github shows all commit changes on the branch (rather than changes relative to master)

PR #4672, opened 2024-03-25
specify an older branch of BBHx in tox.ini

WuShichao:
@ahnitz This PR specified a branch 2 weeks ago for BBHx. 
**Note:
`BBHx` updates its Python requirement to be >=3.12 due to some updates in Michael Katz's other LISA packages, but PyCBC's CI tests are for older versions of Python.**

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -25,7 +25,7 @@ deps =
     {[base]deps}
     pytest
     ; Needed for `BBHx` package to work with PyCBC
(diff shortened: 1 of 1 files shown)


ahnitz (approved):
@WuShichao Please add comment here to explain why we are pinning the version and to make it easier to revert in the future. 

Once you do that and the tests pass, please merge.

PR #407, opened 2015-10-01
fix limits on pycbc_page_ifar search for timeslides

ahnitz:
The limits of which timeslides to include in the ifar plot were wrong and in general allowed it to try to look for timeslides where none could exist.

PR #4675, opened 2024-03-26
frequencyseries.py modified to fix the bug while saving file in .xml format

NehaSinghGW:
Bug fix for saving .xml format
Line added at #407


spxiwh (approved):
Thank you @NehaSinghGW . This fixes an issue you had spotted and discussed with me where PSDs saved from `FSeries.save('name.xml')` would not work. PSDs saved from PyCBC Live's path use a lot of the same code, but does set this. So good to merge from me!

NehaSinghGW:
Thanks for finding the solution for the bug and approving the merge @spxiwh !

PR #4678, opened 2024-03-27
Add new extremely fast direct monte-carlo method

ahnitz:
This is the work in progress PR for a new extremely fast sampling algorithm. It requires some externally generated files (not provided yet). The code is in preview / toy stage at the moment.

Todo: 
  - simplify the pregenerated storage format and provide script + exapmle that makes the pregenerated input.
  - add options to the few tunable knobs
  - add multi-process / mpi support in
  - sanity checking of input
  - [longer term]
     - allow more flexible change of prior

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -431,8 +431,9 @@ def lambda_tilde(mass1, mass2, lambda1, lambda2):
     mask = m1 < m2
     ldiff[mask] = -ldiff[mask]
     eta = eta_from_mass1_mass2(m1, m2)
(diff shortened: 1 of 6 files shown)


ahnitz:
This is now feature complete and tested for an initial implementation. There is an example at https://github.com/gwastro/games-rapid-pe.

ahnitz:
The unittests should pass. The remaining code climate issues I am not planning to address as they relate to some aspects which have to be implemented this way.

WuShichao (approved):
Looks good to me. Thanks for making non-ML O(s) BNS parameter estimation possible!

PR #4679, opened 2024-03-27
Add efficiency plots to PyGRB workflow

jakeb245:
This code add the ability to run `pycbc_pygrb_efficiency` in the full PyGRB workflow. An example results page can be seen  [here](https://ldas-jobs.ligo.caltech.edu/~jacob.buchanan/LVC/pygrb_dev_webpage_fullinj/5._exclusion_distances/5.01_offtrial_1/). The Detection Efficiency plots there are using the trigger files in the wrong order, which I've since fixed in the code. The exclusion distances are written to the caption of the plots, although there is likely a more visually pleasing way to approach it.

In testing this, I've found an issue where some offtrial files are empty. This causes the efficiency code to break (and stops me from generating a fixed example webpage). I'm going to consider that a separate issue unless otherwise suggested.

There are also various fixes scattered throughout. One notable change is the removed of `pp_files`/`pp_outs`, which has been replaced by clearer variable names.

### DIFF
--- a/bin/pygrb/pycbc_grb_trig_combiner
+++ b/bin/pygrb/pycbc_grb_trig_combiner
@@ -507,10 +507,20 @@ numpy.random.seed(args.seed)
 offtrials = numpy.random.choice(range(sum(offsource_ntrials)), args.num_trials, replace=False)
 offtrials.sort()
 for i, j in enumerate(offtrials):
-    _ts = bins["OFFSOURCE"][0][0] if j < offsource_ntrials[0] else bins["OFFSOURCE"][1][0]
+    if j < offsource_ntrials[0]:
+        _ts = bins["OFFSOURCE"][0][0]
+    else:
+        _ts = bins["OFFSOURCE"][1][0]
+        # if the offtrial in the second offsource segment,
+        # subtract the number of trials in the first offsource segment.
+        # This avoids the offtrials overflowing out of the offsource segments.
+        j -= offsource_ntrials[0]
     _ts += j * trialtime
     _te = _ts + trialtime
     bins["OFFTRIAL_{}".format(i+1)] = seg = segments.segment(_ts, _te)
+    if seg not in bins["OFFSOURCE"]:
+        raise ValueError(f"off-trial {i+1} not in off-source segments\n"
+                         f"off-trial {i+1} : {seg}\n")
     vprint("          off-trial {} : {}".format(i+1, seg))
 
 # -- read triggers ------------------------------
--- a/bin/pygrb/pycbc_grb_trig_combiner
+++ b/bin/pygrb/pycbc_grb_trig_combiner
@@ -511,7 +511,7 @@ for i, j in enumerate(offtrials):
         _ts = bins["OFFSOURCE"][0][0]
     else:
         _ts = bins["OFFSOURCE"][1][0]
-        # if the offtrial in the second offsource segment,
+        # if the offtrial is in the second offsource segment,
         # subtract the number of trials in the first offsource segment.
         # This avoids the offtrials overflowing out of the offsource segments.
         j -= offsource_ntrials[0]
--- a/bin/pygrb/pycbc_grb_trig_combiner
+++ b/bin/pygrb/pycbc_grb_trig_combiner
@@ -226,6 +226,9 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
             while search_datasets:
                 dataset_names.remove(search_datasets.pop())
 
+        template_ids = template_hash_to_id(trigger_file=h5out, bank_path=args.bank_file)
+        h5out.create_dataset("network/template_id", data=template_ids)
+
     if verbose:
         print("Merged triggers written to {}".format(outputfile))
     return outputfile
@@ -525,11 +528,6 @@ else:
         verbose=args.verbose,
         **compression_kw
     )
-
-vprint("-- Mapping template hashes to ids")
-ids = template_hash_to_id(trigger_file=outfile, bank_file=args.bank_file)
-with h5py.File(outfile, "a") as h5f:
-    h5f["network/template_id"] = ids
 
 vprint("-- Binning events")
 bin_events(outfile, bins, args.output_dir, args.user_tag, verbose=args.verbose)
--- a/bin/pygrb/pycbc_grb_trig_combiner
+++ b/bin/pygrb/pycbc_grb_trig_combiner
@@ -227,7 +227,10 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
                 dataset_names.remove(search_datasets.pop())
 
         template_ids = template_hash_to_id(trigger_file=h5out, bank_path=args.bank_file)
-        h5out.create_dataset("network/template_id", data=template_ids)
+        h5out.create_dataset("network/template_id",
+                             data=template_ids,
+                             compression_kw="gzip",
+                             compression_opts=9)
 
     if verbose:
         print("Merged triggers written to {}".format(outputfile))
--- a/bin/pygrb/pycbc_grb_trig_combiner
+++ b/bin/pygrb/pycbc_grb_trig_combiner
@@ -229,7 +229,7 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
         template_ids = template_hash_to_id(trigger_file=h5out, bank_path=args.bank_file)
         h5out.create_dataset("network/template_id",
                              data=template_ids,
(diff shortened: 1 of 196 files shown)


pycbc/workflow/grb_utils.py
```diff
@@ -241,20 +241,33 @@ def setup_pygrb_pp_workflow(wf, pp_dir, seg_dir, segment, bank_file,
...
+    inj_insp_files : The list of inspiral files for injections
+    inj_tags : The list of injection tags
+
+    Returns
+    -------
+    trig_files : FileList
+        The list of combined trigger files
+        [ALL_TIMES, ONSOURCE, OFFSOURCE, OFFTRIAL_1, ..., OFFTRIAL_N]
+        FileList (N can be set by the user and is 6 by default)
+    cluster_files : FileList
```
pannarale, line 263: I would go for `clustered_files`: cluster makes me think of something that could be related to the machines we run on.

pycbc/workflow/grb_utils.py
```diff
@@ -241,20 +241,33 @@ def setup_pygrb_pp_workflow(wf, pp_dir, seg_dir, segment, bank_file,
...
+        CLUSTERED FileList, same order as trig_files
```
pannarale, line 264: add something along the lines of "containing triggers after clustering"

pycbc/workflow/grb_utils.py
```diff
@@ -241,20 +241,33 @@ def setup_pygrb_pp_workflow(wf, pp_dir, seg_dir, segment, bank_file,
...
+    inj_find_files : FileList
+        FOUNDMISSED FileList covering all injection sets
+    inj_comb_files : FileList
+        FOUNDMISSED-FILTERED FileList covering all injection sets
+                     in the same order as inj_find_files
```
pannarale, line 269: Here too I would try to specify how the 2 file families differ.

bin/pygrb/pycbc_pygrb_pp_workflow
```diff
@@ -50,12 +50,20 @@ parser = argparse.ArgumentParser(description=__doc__[1:])
 parser.add_argument("--version", action="version", version=__version__)
 parser.add_argument("-v", "--verbose", default=False, action="store_true",
                     help="Verbose output")
-parser.add_argument("-t", "--trig-file", action="store",
-                    default=None, required=True,
-                    help="The location of the trigger file")
+parser.add_argument("-t", "--trig-files", action="store",
+                    required=True, nargs="+",
+                    help="The location of the trigger files"
```
pannarale, line 55: locations

pannarale: and add that the assumed order is as follows

bin/pygrb/pycbc_pygrb_pp_workflow
```diff
@@ -85,19 +93,41 @@ rdir = layout.SectionNumber('webpage', ['offsource_triggers_vs_time',
...
+    if not os.path.isfile(trig_file):
+        raise ValueError("Trigger file %s does not exist" % trig_file)
+    if "ALL_TIMES" in trig_file:
+        all_times_file = os.path.join(start_rundir, trig_file)
+    if "ONSOURCE" in trig_file:
+        onsource_file = os.path.join(start_rundir, trig_file)
+    if "OFFSOURCE" in trig_file:
+        offsource_file = os.path.join(start_rundir, trig_file)
+    if "OFFTRIAL" in trig_file:
+        offtrial_files.append(os.path.join(start_rundir, trig_file))
```
pannarale, line 113: My request above is to tell the user we expect this order in the FileList, but we should not rely on file names. Rather:
```
all_times_file = os.path.join(start_rundir, args.trig_files[0])
onsource_file  = os.path.join(start_rundir, args.trig_files[1])
offsource_file = os.path.join(start_rundir, args.trig_files[2])
offtrial_files = [os.path.join(start_rundir, trig_file) for trig_file in args.trig_files[3:]]
```
and at each step inform the user with messages like "using such-and-such as all-times trigger file" etc.

bin/pygrb/pycbc_pygrb_pp_workflow
```diff
@@ -85,19 +93,41 @@ rdir = layout.SectionNumber('webpage', ['offsource_triggers_vs_time',
...
+offtrial_files.sort()
```
pannarale, line 114: At which point, this is not needed.

pannarale:
Thanks, @jakeb245.  While I review the code changes, and before I forget, it would be nice to track in the captions and/or titles what injection set tag the efficiency plot refers to.

pycbc/workflow/grb_utils.py
pannarale: Holds for `pycbc_make_offline_grb_workflow` as well.

pannarale (changes requested):

pycbc/workflow/grb_utils.py
```diff
@@ -260,15 +260,16 @@ def setup_pygrb_pp_workflow(wf, pp_dir, seg_dir, segment, bank_file,
...
         FileList (N can be set by the user and is 6 by default)
-    cluster_files : FileList
+    clustered_files : FileList
         CLUSTERED FileList, same order as trig_files
+        Contains triggers after clustering
     inj_find_files : FileList
         FOUNDMISSED FileList covering all injection sets
     inj_comb_files : FileList
         FOUNDMISSED-FILTERED FileList covering all injection sets
                      in the same order as inj_find_files
```
pannarale, line 270: Align `in` with `FOUNDMISSED-FILTERED` for consistency.

pannarale (changes requested):

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -507,10 +507,20 @@ numpy.random.seed(args.seed)
...
+        # if the offtrial in the second offsource segment,
+        # subtract the number of trials in the first offsource segment.
+        # This avoids the offtrials overflowing out of the offsource segments.
+        j -= offsource_ntrials[0]
     _ts += j * trialtime
     _te = _ts + trialtime
     bins["OFFTRIAL_{}".format(i+1)] = seg = segments.segment(_ts, _te)
+    if seg not in bins["OFFSOURCE"]:
+        raise ValueError(f"off-trial {i+1} not in off-source segments\n"
+                         f"off-trial {i+1} : {seg}\n")
```
pannarale, line 523: Has this happened to you, or is it there just as an extra layer of safety?

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -507,10 +507,20 @@ numpy.random.seed(args.seed)
 offtrials = numpy.random.choice(range(sum(offsource_ntrials)), args.num_trials, replace=False)
 offtrials.sort()
 for i, j in enumerate(offtrials):
-    _ts = bins["OFFSOURCE"][0][0] if j < offsource_ntrials[0] else bins["OFFSOURCE"][1][0]
+    if j < offsource_ntrials[0]:
+        _ts = bins["OFFSOURCE"][0][0]
+    else:
+        _ts = bins["OFFSOURCE"][1][0]
+        # if the offtrial in the second offsource segment,
```
pannarale, line 514:
```suggestion
        # if the offtrial is in the second offsource segment,
```

pannarale (changes requested):

bin/pygrb/pycbc_grb_trig_combiner
jakeb245: This was the cause of getting empty off-trial files. It should hopefully be fixed now, so it's more of a sanity check.

jakeb245:
It turns out the postprocessing files list is not in the expected order. It seems the on-source and off-source files were switched! See [this webpage](https://ldas-jobs.ligo.caltech.edu/~jacob.buchanan/LVC/pygrb_dev_webpage_fullinj_fix/1._offsource_triggers_vs_time/1.01_reweighted_snr_timeseries/). I'm not sure if this is consistent or not

jakeb245:
Another issue I should note here is that chunks of time padded in `pycbc_multi_inspiral` are not accounted for later on in the workflow. This causes some trials to contain no triggers, which can cause `pycbc_pygrb_efficiency` to fail if those trials are picked as off-trials. I'm not sure of the best way to handle this, or if it lies in the scope of this PR. It is making the generation of a new example webpage a game of chance for now...

bin/pygrb/pycbc_grb_trig_combiner
pannarale: Ok, then it's good to have this check here for safety.  Thanks

pannarale (approved):

PR #4683, opened 2024-04-03
Fix bug from scipy version changing

GarethCabournDavies:


spxiwh (approved):
This was deprecated in quite old versions of scipy, so should probably always have been this. While I think that our stated scipy requirement (> 0.16.0) is probably wrong, this would work even on 0.16.0

PR #4685, opened 2024-04-04
Check if sample_points is a pycbc Array and convert if not

adivijaykumar:
`pycbc.waveform.get_fd_waveform_sequence` requires `sample_points` to be a pycbc `Array`  instead of a numpy array. Since `get_fd_waveform` and `get_td_waveform` do not demand any inputs to be pycbc custom types (probably because most/all inputs are floats), it makes sense to add a check for `sample_points` being a numpy array and converting to an `Array` if that is the case. This PR does exactly that.

### DIFF
--- a/pycbc/waveform/waveform.py
+++ b/pycbc/waveform/waveform.py
@@ -526,6 +526,8 @@ def get_fd_waveform_sequence(template=None, **kwds):
         required = wav_gen.required
     else:
         required = parameters.fd_required
+    if not isinstance(input_params['sample_points'], Array):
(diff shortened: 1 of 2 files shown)


adivijaykumar:
Should I also add a test for this?

ahnitz:
@adivijaykumar If you are willing to add a unittest that would be great.

adivijaykumar:
Added a unittest, but not sure if it is sufficient...

ahnitz (approved):
This looks reasonable to me. If nothing shows up that is relevant in the CI runs then I think this can be merged.

PR #4686, opened 2024-04-05
Support detector frame waveforms for inference

mj-will:
This PR adds support for using detector-frame waveforms (e.g. those produced by `get_fd_det_waveform`) in the inference module.

**Summary of changes**

- Add a new generator class, `FDomainDirectDetFrameGenerator` 
- Add the `det_frame_waveform` option to the `GaussianNoise` class to which enables the use of the new generator.

I considered automatically determining if a waveform is implemented in the detector frame or not, but I wasn't sure if a waveform could support both, so opted for a boolean instead. I can change this if it is preferred. I'm happy to rename the class/variables.

**Example**

My use-case for this is for performing inference in LISA with BBHx, so I've tested this with a modified version of the existing SMBH injection example. Here's the result:

image

I can also add this version to the directory if desired. I will note though that this is quite slow; running with nessai and 8 process, it takes around 10 minutes.

**Other thoughts**

I have not tested how this interacts with any other likelihoods, I know it won't be valid for some of them.

### DIFF
--- a/pycbc/inference/models/gaussian_noise.py
+++ b/pycbc/inference/models/gaussian_noise.py
@@ -707,6 +707,15 @@ class GaussianNoise(BaseGaussianNoise):
         log likelihood. Default is to not include it.
     static_params : dict, optional
         A dictionary of parameter names -> values to keep fixed.
+    det_frame_waveform : bool
+        If True, the waveform will be generated directly in the detector frame
+        using the
+        :py:class:`~pycbc.waveform.generator.FDomainDirectDetFrameGenerator`.
+        This requires the approximant be implemented in
+        :py:func:`~pycbc.waveform.get_fd_det_waveform`.
+        If False, the
+        :py:class:`~pycbc.waveform.generator.FDomainDetFrameGenerator` will be
+        used instead. Defaults to :code:`False`.
     \**kwargs :
         All other keyword arguments are passed to ``BaseDataModel``.
 
--- a/pycbc/inference/models/gaussian_noise.py
+++ b/pycbc/inference/models/gaussian_noise.py
@@ -552,11 +552,18 @@ class BaseGaussianNoise(BaseDataModel, metaclass=ABCMeta):
             args['normalize'] = True
         if cp.has_option('model', 'ignore-failed-waveforms'):
             args['ignore_failed_waveforms'] = True
+        if cp.has_option('model', 'det-frame-waveform'):
+            args['det_frame_waveform'] = True
         if cp.has_option('model', 'no-save-data'):
             args['no_save_data'] = True
         # get any other keyword arguments provided in the model section
-        ignore_args = ['name', 'normalize',
-                       'ignore-failed-waveforms', 'no-save-data']
+        ignore_args = [
+            'name',
+            'normalize',
+            'ignore-failed-waveforms',
+            'no-save-data',
+            'det-frame-waveform'
+        ]
         for option in cp.options("model"):
             if option in ("low-frequency-cutoff", "high-frequency-cutoff"):
                 ignore_args.append(option)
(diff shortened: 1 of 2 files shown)


WuShichao:
@mj-will Thanks for this PR, can you share the config file that you modified here? So I can test it, too. You mentioned you used your `nessai` and 8 threads, so I guess `dynesty` will take an even longer time.

mj-will:
> @mj-will Thanks for this PR, can you share the config file that you modified here? So I can test it, too. You mentioned you used your `nessai` and 8 threads, so I guess `dynesty` will take an even longer time.

@WuShichao sure, here you are:

```
[data]
instruments = LISA_A LISA_E LISA_T
trigger-time = 4800021.15572853
analysis-start-time = -4800021
analysis-end-time = 26735979
pad-data = 0
sample-rate = 0.2
fake-strain = LISA_A:analytical_psd_lisa_tdi_1p5_AE LISA_E:analytical_psd_lisa_tdi_1p5_AE LISA_T:analytical_psd_lisa_tdi_1p5_T
; fake-strain-extra-args = LISA_A:len_arm:2.5e9 LISA_A:acc_noise_level:3e-15 LISA_A:oms_noise_level:15e-12 LISA_E:len_arm:2.5e9 LISA_E:acc_noise_level:3e-15 LISA_E:oms_noise_level:15e-12 LISA_T:len_arm:2.5e9 LISA_T:acc_noise_level:3e-15 LISA_T:oms_noise_level:15e-12
fake-strain-extra-args = len_arm:2.5e9 acc_noise_level:3e-15 oms_noise_level:15e-12
fake-strain-seed = LISA_A:100 LISA_E:150 LISA_T:200
fake-strain-flow = 0.0001
fake-strain-sample-rate = 0.2
fake-strain-filter-duration = 31536000
psd-estimation = median-mean
psd-inverse-length = 2678400
psd-segment-length = 2678400
psd-segment-stride = 1339200
psd-start-time = -4800021
psd-end-time = 26735979
channel-name = LISA_A:LISA_A LISA_E:LISA_E LISA_T:LISA_T
injection-file = /home/mjwill/git_repos/pycbc/examples/inference/lisa_smbhb_inj/injection_smbhb.hdf

[model]
name = gaussian_noise
low-frequency-cutoff = 0.0001
high-frequency-cutoff = 1e-2
det-frame-waveform =

[variable_params]
mchirp =
q =
tc =

[static_params]
; Change it to "ref_frame = SSB", if you use SSB frame in injection file.
ref_frame = LISA
approximant = BBHX_PhenomD
coa_phase = 4.275929308696054
eclipticlongitude = 5.4431083771985165
eclipticlatitude = -1.2734504596198182
polarization = 0.22558110042980073
spin1z = 0.597755394865021
spin2z = 0.36905807298613247
distance = 17758.367941273442
t_obs_start = 31536000
f_lower = 1e-4
; Put LISA behind the Earth by ~20 degrees.
t_offset = 7365189.431698299
; ifos = LISA_A LISA_E LISA_T
f_final = 0.1

[prior-mchirp]
name = uniform
min-mchirp = 703772.7245316936
max-mchirp = 860166.6633165143

[prior-q]
name = uniform
min-q = 1.1469802543574181
max-q = 1.401864755325733

[prior-tc]
name = uniform
min-tc = 4798221.15572853
max-tc = 4801821.15572853

[waveform_transforms-mass1+mass2]
name = mchirp_q_to_mass1_mass2

[sampler]
name = nessai
nlive = 100

; NOTE: While this example doesn't sample in polarization, if doing this we
; recommend the following transformation, and then sampling in this coordinate
;
; [waveform_transforms-polarization]
; name = custom
; inputs = better_pol, eclipticlongitude
; polarization = better_pol + eclipticlongitude

```

ahnitz:
@mj-will This is still listed as a draft PR, would you like us to review it at this point?

ahnitz (approved):
@mj-will I think this is a suitable straightforward implementation. Also thank you for adding errors for cases not implemented here. 

Just confirm that this works in practice and I think a quick final look by @cdcapano would suffice here in case there's potential issues with this.

I think having an explicit switch of the generator is OK and as you said on the call will result in fewer mistakes. However, this is a bit inconsistent with what we've done in the heterodyne likelihood. 

Should we try to make this consistent there now? E.g. there it 'automatically' detectors, but an explicit option would be just as good but would require updating the examples to reflect the option.

ahnitz:
@mj-will Barring the last question I posed and @cdcapano view on that, I think I am happy with this. I've also restarted the CI failed tests. I think (maybe) that the error was unrelated, but we'll see on rerun.

cdcapano (approved):
@mj-will This looks good.

@ahnitz Yes, the relative binning model should be updated to reflect the changes here. But that's going to require changes to several lines of code, along with updates to the examples. To keep the PRs short and manageable, I think it best to merge this now, then fix relative binning to be consistent with it in a separate PR.

WuShichao (approved):

PR #4687, opened 2024-04-05
Add live singles significance fitting supervision script to github repo

GarethCabournDavies:
This was a code we had been using to supervise the running of `pycbc_live_single_trigger_fits` and `pycbc_live_combine_single_fits`, but we now add to this repo (after some tidying up)

Note that one of the other significant changes here is to update the naming of the fitting codes in order to obviously separate them from the denty-bin-style fits.
```
renamed: pycbc_live_combine_single_fits -> pycbc_live_combine_single_significance_fits
renamed: pycbc_live_plot_combined_single_fits -> pycbc_live_plot_combined_single_significance_fits
renamed: pycbc_live_plot_single_trigger_fits -> pycbc_live_plot_single_significance_fits
renamed: pycbc_live_single_trigger_fits -> pycbc_live_single_significance_fits
```

(Diff removed for a later comment)


bin/live/pycbc_live_supervise_single_significance_fits
```diff
@@ -0,0 +1,516 @@
...
+
+pycbc.init_logging(True)
+
+if args.run_daily_at is not None and args.date is not None:
+    parser.error('Cannot take --run-daily-at and --date at the same time')
+
+if args.run_daily_at is not None:
+    # keep running and repeat the fitting every day at the given hour
+    if not re.match('[0-9][0-9]:[0-9][0-9]:[0-9][0-9]', args.run_daily_at):
+        parser.error('--run-daily-at takes a UTC time in the format HH:MM:SS')
```
titodalcanton, line 499: After the discussion we had a few days ago on Slack, can you add this just before the infinite loop:
```python
    logging.info('Starting in daily run mode')
```

titodalcanton:
@GarethCabournDavies I am unable to read diffs in that format, could you remake the diffs using the `-u` option? And I think you can indicate that you are pasting diff code into the Markdown to make it apply the proper syntax highlight, which makes it even easier to read.

GarethCabournDavies:
> @GarethCabournDavies I am unable to read diffs in that format, could you remake the diffs using the `-u` option? And I think you can indicate that you are pasting diff code into the Markdown to make it apply the proper syntax highlight, which makes it even easier to read.

Well that's much easier to read! - added in comment now

bin/live/pycbc_live_supervise_single_significance_fits
```diff
@@ -0,0 +1,517 @@
...
+            symlink(cpo, public_dir)
+
+    logging.info('Done')
+
+
+def wait_for_utc_time(target_str):
+    """Wait until the UTC time is as given by `target_str`, in HH:MM:SS format.
+    """
+    target_hour, target_minute, target_second = map(int, target_str.split(':'))
+    now = datetime.utcnow()
```
GarethCabournDavies, line 452: I think utcnow is being deprecated actually so this may need now()

titodalcanton: For context, here is a nice [writeup](https://blog.miguelgrinberg.com/post/it-s-time-for-a-change-datetime-utcnow-is-now-deprecated). I would not worry about this just yet (this is going in production soon, and it will work for the rest of the run) but maybe adding a comment or opening an issue is warranted to plan ahead.

PR #4688, opened 2024-04-07
add acceleration option to JointPrimaryMarginalizedModel likelihood

WuShichao:
@ahnitz This PR adds an acceleration option to JointPrimaryMarginalizedModel likelihood by assuming all extrinsic parameters can be fixed.


WuShichao:
@ahnitz As you suggested, I have made this PR a general one. The specific LISA multiband one will be in another PR.

pycbc/inference/models/hierarchical.py
```diff
@@ -723,24 +733,43 @@ def total_loglr(self):
             # not using self.primary_model.current_params, because others_model
             # may have its own static parameters
             current_params_other = other_model.current_params.copy()
-            for i in range(nums):
+            if not self.accelerate_loglr:
```
ahnitz, line 736: Can you choose a more descriptive name to what the option does (e.g. *how* does it make the likelihood faster than than just the fact that it does)

WuShichao: Is `self.static_margin_params_in_other_models` better?

WuShichao: @ahnitz In this PR, we don't apply any amplitude and phase correction, so I will not include these in the option name.

pycbc/inference/models/hierarchical.py
```diff
@@ -640,6 +626,11 @@ def __init__(self, variable_params, submodels, **kwargs):
         self.other_models.pop(kwargs['primary_lbl'][0])
         self.other_models = list(self.other_models.values())
 
+        # determine whether to accelerate total_loglr
+        from pycbc.inference.models.tools import str_to_bool
+        self.static_margin_params_in_other_models = \
```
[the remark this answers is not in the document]
WuShichao: @ahnitz OK, I have updated.

pycbc/inference/models/hierarchical.py
```diff
@@ -723,24 +735,43 @@ def total_loglr(self):
...
+            # the type of the original sh/hh_others are numpy.array,
+            # might not the same as sh/hh_primary during reconstruct,
+            # during reconstruct of distance, sh/hh_others need to be scalar
             sh_others = sh_others[0]
+            hh_others = hh_others[0]
         sh_total = sh_primary + sh_others
         hh_total = hh_primary + hh_others
 
         # calculate marginalize_vector_weights
         self.primary_model.marginalize_vector_weights = \
```
ahnitz, line 771: This line shouldn't be here.

WuShichao: OK, removed.

bin/inference/pycbc_inference
```diff
@@ -141,7 +141,12 @@ with ctx:
     if pool.is_main_process():
         for fn in [sampler.checkpoint_file, sampler.backup_file]:
             with loadfile(fn, 'a') as fp:
-                fp.write_config_file(cp)
+                # some models will interally modify original cp for sampling,
```
ahnitz, line 144: This shouldn't be an option as one should always save the original configuration. Why isn't your internal sampler modifying a copy? The version saved here then doesn't have to know (and really shouldn't) that you may have modified internally.

bin/inference/pycbc_inference_model_stats
```diff
@@ -96,7 +96,10 @@ model.sampling_transforms = None
 def callmodel(arg):
     iteration, paramvals = arg
     # calculate the logposterior to get all stats populated
-    model.update(**{p: paramvals[p] for p in model.variable_params})
+    try:
```
ahnitz, line 99: Again, this is not correct. Your top level model should just have an update method. That method should do whatever is needed to prepare for the log* methods to actually work. You shouldn't be requiring anyone know about a new method 'update_all_models'. It's fine if you want to have an new method for *internal* use, but not for the top-level api.

pycbc/inference/models/hierarchical.py
```diff
@@ -919,14 +972,13 @@ def from_config(cls, cp, **kwargs):
...
-            marginalized_params.pop('logw_partial')
-            marginalized_params = list(marginalized_params.keys())
-        else:
-            marginalized_params = []
-        # this may also include 'f_ref', 'f_lower', 'approximant',
-        # but doesn't matter
-        marginalized_params += list(primary_model.static_params.keys())
+        marginalized_params = list(marginalized_params.keys())
+        # add distance or phase if they are marginalized
+        if primary_model.distance_marginalization:
```
ahnitz, line 977: Why do you need to hardcode this? This will be brittle and break easily to changes to marginalization for example, which we don't want. If you need a list of marginalized parameters, why not add this to the class in inference/tools so that it is kept up to date?

pycbc/inference/models/relbin.py
```diff
@@ -596,10 +596,11 @@ def _loglr(self):
             filt += filter_i
             norm += norm_i
 
+        loglr = self.marginalize_loglr(filt, norm)
         if self.return_sh_hh:
```
ahnitz, line 600: Why do we need this if statement? Shouldn't the existing flag already used for demarginalization take care of this?  E.g. why not use the reconstruct_phase flag?

https://github.com/gwastro/pycbc/blob/master/pycbc/inference/models/tools.py#L241

pycbc/inference/models/hierarchical.py
```diff
@@ -686,25 +683,48 @@ def total_loglr(self):
...
+        else:
+            for key, value in self.primary_model.current_params.items():
+                # add marginalize_vector_params
+                if key in margin_names_vector:
+                    margin_params[key] = value
+                    if isinstance(value, numpy.ndarray):
+                        nums = len(value)
+                    else:
+                        nums = 1
         # add distance if it has been marginalized,
```
ahnitz, line 728: Again, avoid needed to know explicitly about distance here. Think instead about the format that you require. If the format differs how to generically convert.

pycbc/inference/models/hierarchical.py
```diff
@@ -723,24 +743,47 @@ def total_loglr(self):
...
-                        value for key, value in margin_params.items()})
+                    {key: value[0] if isinstance(value, numpy.ndarray)
+                     else value for key, value in margin_params.items()})
                 other_model.update(**current_params_other)
                 other_model.return_sh_hh = True
-                sh_others[i], hh_others[i] = other_model.loglr
+                sh_other, hh_other = other_model.loglr
                 other_model.return_sh_hh = False
+                # set logr, otherwise it will store (sh, hh)
+                setattr(other_model._current_stats, 'loglr',
```
ahnitz, line 770: Same as above. It's not necessarily a problem as it might be useful to have the separate loglrs, but it's not clear that it will always make sense.

bin/inference/pycbc_inference
WuShichao: @ahnitz My understanding is `cp` is saved when it runs `logging.info("Loading joint_primary_marginalized model")
        return super(HierarchicalModel, cls).from_config(
                cp, submodels=submodels, **kwargs)` https://github.com/WuShichao/pycbc/blob/accelerate_multiband/pycbc/inference/models/hierarchical.py#L1002 the `cp` here is the modified config for sampling, can't be used as the initial config. So how to let PyCBC save the original one here?

bin/inference/pycbc_inference_model_stats
WuShichao: So do I need to rename my `update_all_models` to be `update` (so overwrite the base one), or just use the original `update` in `callmodel`?

bin/inference/pycbc_inference
ahnitz: @WuShichao I don't see anything related to saving the configfile where you've pointed. The place is in pycbc_inference where I'm comment on. Take a look at my first comment, I say how to do it. Don't modify the configparser you are passed in-place. Make a copy so you aren't editting the original. Then you don't need to save a separate copy and this particular line will just work to begin with.

bin/inference/pycbc_inference_model_stats
ahnitz: @WuShichao A sampler is just using 'update' so why are you doing so differently here? Think through how to do this best, but it should be clear that making this change in this program seems very inconsistent. If it is required, that indicates something is wrong with your model, and if so fix that. Otherwise maybe you made this change in error.

pycbc/inference/models/hierarchical.py
```diff
@@ -723,24 +743,47 @@ def total_loglr(self):
...
+                        {key: value[i] if isinstance(value, numpy.ndarray)
+                         else value for key, value in margin_params.items()})
+                    other_model.update(**current_params_other)
+                    other_model.return_sh_hh = True
+                    sh_other, hh_other = other_model.loglr
+                    sh_others[i] += sh_other
+                    hh_others[i] += hh_other
+                    other_model.return_sh_hh = False
+                    # set logr, otherwise it will store (sh, hh)
+                    setattr(other_model._current_stats, 'loglr',
```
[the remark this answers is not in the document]
WuShichao: My understanding is that when `pycbc_inference_model_stats` check `for pi, p in enumerate(model.default_stats):` it will try to access submodel's `loglr`, no? If so, I need to store it. When I not rest it, I found `other_model._current_stats` also contain (sh, hh) for each point.

pycbc/inference/models/relbin.py
WuShichao: @ahnitz This is used here https://github.com/WuShichao/pycbc/blob/accelerate_multiband/pycbc/inference/models/hierarchical.py#L752 We need to get LISA's sh and hh.

pycbc/inference/models/hierarchical.py
WuShichao: OK, I will save it in the `tools` module and just use it here.

pycbc/inference/models/hierarchical.py
```diff
@@ -686,25 +678,48 @@ def total_loglr(self):
...
+            # waveform. Using SNR will cancel out the effect of amplitude.err
+            i_max_extrinsic = numpy.argmax(
+                numpy.abs(sh_primary) / hh_primary**0.5)
+            for p in margin_names_vector:
+                if isinstance(self.primary_model.current_params[p],
+                              numpy.ndarray):
+                    margin_params[p] = \
+                        self.primary_model.current_params[p][i_max_extrinsic]
+                    nums = len(self.primary_model.current_params[p])
+                else:
```
ahnitz, line 711: @WuShichao This logic should *already* take care of the distance case, except that later on you assume that if any parameter is a scalar they all are. That's the part you should stop assuming. Don't assume they are any particular mix of scalar or vector.

WuShichao:
Replace deprecated ConfigParser `readfp` with `read_file`:
https://github.com/python/cpython/pull/92503
https://issues.apache.org/jira/browse/SVN-4899

ahnitz (approved):

PR #4689, opened 2024-04-08
Ranking statistic for live singles

GarethCabournDavies:
This PR does a couple of things:
 - Allows the ranking statistic mechanism to be used for single-detector events in pycbc-live (as is done for coincs in #4527)
 - Allow the live single significance fits to use ranking statistic rather than sngl-ranking
 - Fixes rank_stat_single for PhaseTDStatistic (which didn't work, but was unused before this)

### DIFF
--- a/bin/live/pycbc_live_combine_single_significance_fits
+++ b/bin/live/pycbc_live_combine_single_significance_fits
@@ -101,7 +101,7 @@ for f in args.trfits_files:
         gps_last = 0
         gps_first = np.inf
         for ifo in args.ifos:
-            if ifo not in fits_f:
+            if ifo not in fits_f or 'triggers' not in fits_f[ifo]:
                 continue
             trig_times = fits_f[ifo]['triggers']['end_time'][:]
             gps_last = max(gps_last, trig_times.max())
--- a/bin/live/pycbc_live_combine_single_significance_fits
+++ b/bin/live/pycbc_live_combine_single_significance_fits
@@ -157,9 +157,8 @@ fout['fits_dates'] = ad + start_time_n
 for ifo in args.ifos:
     logging.info(ifo)
     fout_ifo = fout.create_group(ifo)
-    for ifo in args.ifos:
-        fout_ifo.attrs['fit_threshold'] = fit_thresh[ifo]
-        fout_ifo.attrs['fit_function'] = fit_func[ifo]
+    fout_ifo.attrs['fit_threshold'] = fit_thresh[ifo]
+    fout_ifo.attrs['fit_function'] = fit_func[ifo]
     l_times = np.array(live_times[ifo])
     total_time = l_times.sum()
     fout_ifo.attrs['live_time'] = total_time
--- a/bin/live/pycbc_live_combine_single_significance_fits
+++ b/bin/live/pycbc_live_combine_single_significance_fits
@@ -62,8 +62,8 @@ for f in args.trfits_files:
             bl = fit_f['bins_lower'][:]
             bu = fit_f['bins_upper'][:]
             sngl_rank = fit_f.attrs['sngl_ranking']
-            fit_thresh = fit_f.attrs['fit_threshold']
-            fit_func = fit_f.attrs['fit_function']
+            fit_thresh = {ifo: fit_f[ifo].attrs['fit_threshold'] for ifo in args.ifos}
+            fit_func = {ifo: fit_f[ifo].attrs['fit_function'] for ifo in args.ifos}
 
 # Now go back through the fit files and read the actual information.  Skip the
 # files that have fit parameters inconsistent with what we found earlier.
@@ -80,9 +80,11 @@ for f in args.trfits_files:
     with h5py.File(f, 'r') as fits_f:
         # Check that the file uses the same setup as file 0, to make sure
         # all coefficients are comparable
+        new_fit_func = {ifo: fits_f[ifo].attrs['fit_function'] for ifo in args.ifos}
+        new_fit_thresh = {ifo: fits_f[ifo].attrs['fit_threshold'] for ifo in args.ifos}
         same_conf = (fits_f.attrs['sngl_ranking'] == sngl_rank
-                     and fits_f.attrs['fit_threshold'] == fit_thresh
-                     and fits_f.attrs['fit_function'] == fit_func
+                     and new_fit_thresh == fit_thresh
+                     and new_fit_func == fit_func
                      and fits_f['bins_lower'].size == bl.size
                      and all(fits_f['bins_lower'][:] == bl)
                      and all(fits_f['bins_upper'][:] == bu))
@@ -143,7 +145,8 @@ cons_counts_out = {ifo: np.inf * np.ones(len(alphas_bin[ifo])) for ifo in args.i
 logging.info("Writing results")
 
 fout = h5py.File(args.output, 'w')
-fout.attrs['fit_threshold'] = fit_thresh
+for ifo in args.ifos:
+    fout.attrs[f'{ifo}_fit_threshold'] = fit_thresh[ifo]
 fout.attrs['conservative_percentile'] = args.conservative_percentile
 fout.attrs['ifos'] = args.ifos
 fout['bins_edges'] = list(bl) + [bu[-1]]
--- a/bin/live/pycbc_live_combine_single_significance_fits
+++ b/bin/live/pycbc_live_combine_single_significance_fits
@@ -149,9 +149,6 @@ cons_counts_out = {ifo: np.inf * np.ones(len(alphas_bin[ifo])) for ifo in args.i
 logging.info("Writing results")
 
 fout = h5py.File(args.output, 'w')
-for ifo in args.ifos:
-    fout.attrs[f'{ifo}_fit_threshold'] = fit_thresh[ifo]
-    fout.attrs[f'{ifo}_fit_function'] = fit_func[ifo]
 fout.attrs['conservative_percentile'] = args.conservative_percentile
 fout.attrs['ifos'] = args.ifos
 fout['bins_edges'] = list(bl) + [bu[-1]]
@@ -160,6 +157,9 @@ fout['fits_dates'] = ad + start_time_n
 for ifo in args.ifos:
     logging.info(ifo)
     fout_ifo = fout.create_group(ifo)
+    for ifo in args.ifos:
+        fout_ifo.attrs['fit_threshold'] = fit_thresh[ifo]
+        fout_ifo.attrs['fit_function'] = fit_func[ifo]
     l_times = np.array(live_times[ifo])
     total_time = l_times.sum()
     fout_ifo.attrs['live_time'] = total_time
--- a/bin/live/pycbc_live_combine_single_significance_fits
+++ b/bin/live/pycbc_live_combine_single_significance_fits
@@ -62,8 +62,10 @@ for f in args.trfits_files:
             bl = fit_f['bins_lower'][:]
             bu = fit_f['bins_upper'][:]
             sngl_rank = fit_f.attrs['sngl_ranking']
-            fit_thresh = {ifo: fit_f[ifo].attrs['fit_threshold'] for ifo in args.ifos}
-            fit_func = {ifo: fit_f[ifo].attrs['fit_function'] for ifo in args.ifos}
+            fit_thresh = {ifo: fit_f[ifo].attrs['fit_threshold']
+                          for ifo in args.ifos}
+            fit_func = {ifo: fit_f[ifo].attrs['fit_function']
+                        for ifo in args.ifos}
 
 # Now go back through the fit files and read the actual information.  Skip the
 # files that have fit parameters inconsistent with what we found earlier.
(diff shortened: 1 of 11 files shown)


bin/live/pycbc_live_plot_combined_single_significance_fits
```diff
@@ -139,7 +139,7 @@ for ifo in ifos:
             continue
 
         l_times = separate_times[ifo]
-        with np.errstate(divide='ignore'):
+        with np.errstate(divide='ignore', invalid='ignore'):
```
GarethCabournDavies, line 142: 0/0 uses the invalid rather than divide keyword

examples/live/make_singles_significance_fits.py
```diff
@@ -6,7 +6,7 @@
 import h5py
 import numpy as np
 
-f = h5py.File('single_trigger_fits.hdf','w')
+f = h5py.File('single_significance_fits.hdf','w')
```
GarethCabournDavies, line 9: Renaming to clarify the two types of single fits

pycbc/events/single.py
```diff
@@ -232,10 +291,14 @@ def calculate_ifar(self, sngl_ranking, duration):
...
+            [sngl_ranking],
+            coeff,
+            thresh
+        )[0]
 
         # apply a trials factor of the number of duration bins
         rate_louder *= len(rates)
 
-        return conv.sec_to_year(1. / rate_louder)
+        return min(conv.sec_to_year(1. / rate_louder), self.maximum_ifar)
```
GarethCabournDavies, line 304: This avoids stupid-loud IFAR for singles, especially when we can't get very high IFAR for backgrounds.

pycbc/events/stat.py
```diff
@@ -682,7 +682,7 @@ def rank_stat_single(self, single_info,
         numpy.ndarray
             The array of single detector statistics
         """
-        return single_info[1]
+        return single_info[1]['snglstat']
```
GarethCabournDavies, line 685: This allows the `phasetd` `rank_stat_single` to actually work!

GarethCabournDavies (commented):
Minor explanation comments

bin/live/pycbc_live_combine_single_significance_fits
```diff
@@ -62,8 +62,10 @@ for f in args.trfits_files:
             bl = fit_f['bins_lower'][:]
             bu = fit_f['bins_upper'][:]
             sngl_rank = fit_f.attrs['sngl_ranking']
-            fit_thresh = fit_f.attrs['fit_threshold']
-            fit_func = fit_f.attrs['fit_function']
+            fit_thresh = {ifo: fit_f[ifo].attrs['fit_threshold']
```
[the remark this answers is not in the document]
maxtrevor: Is it planned to actually use different values of `fit_func` and `fit_thresh` for different ifos in production?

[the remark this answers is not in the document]
GarethCabournDavies: `fit_func` - probably not, `fit_thresh` - possibly

The sensitive distance thing shifts the distribution of the noise events by the log of volume, so a factor of 1.1 in distance becomes a change of 1.33 in the statistic.

As the event rate at a given stat is distributed as an exponential, this will be a factor of ~3.8 is the number of events, which when dealing with low-number statistics as we are, can be important.

Its fairly easy to see this in the offline results, if you look at A18/A19 in results pages

In the offline configs we use a difference of 1.2 in threshold in the detectors, so a factor of 3.3 in the number of events

GarethCabournDavies:
BTW - I requested bith Arthur and Tito as I am aware that Arthur has the expertise in adding the statistic objects to Live, and Tito understands more of the significance fitting

examples/live/make_fit_coeffs.py
```diff
@@ -0,0 +1,21 @@
+import h5py
```
GarethCabournDavies, line 1: Makes files which can be used as the `fit_coeffs` statistic. These are not of any scientific use, but the code will accept them and run properly

pycbc/events/single.py
```diff
@@ -158,43 +198,62 @@ def check(self, trigs, data_reader):
         # Apply cuts to trigs before clustering
         # Cut on snr so that triggers which could not reach newsnr
         # threshold do not have newsnr calculated
+        if 'psd_var_val' in trigs:
```
GarethCabournDavies, line 201: See [here](https://github.com/gwastro/pycbc/blob/v2.1.3/pycbc/events/ranking.py#L64) for the prdvar conversions

pycbc/events/single.py
```diff
@@ -158,43 +198,62 @@ def check(self, trigs, data_reader):
...
-        nsnr_all = ranking.newsnr(cutdurchi_trigs['snr'],
-                                  cutdurchi_trigs['chisq'])
-        nsnr_idx = nsnr_all > self.thresholds['newsnr']
-        if not np.any(nsnr_idx):
+        # Convert back from the pycbc live convention of chisq always
+        # meaning the reduced chisq.
+        trigsc = copy.copy(cut_trigs)
+        trigsc['ifo'] = self.ifo
+        trigsc['chisq'] = cut_trigs['chisq'] * cut_trigs['chisq_dof']
+        trigsc['chisq_dof'] = (cut_trigs['chisq_dof'] + 2) / 2
```
GarethCabournDavies, line 226: This is the same as [this part of the coinc version](https://github.com/gwastro/pycbc/blob/v2.1.3/pycbc/events/coinc.py#L1086)

pycbc/events/stat.py
```diff
@@ -2435,7 +2436,10 @@ def get_statistic_from_opts(opts, ifos):
         opts.statistic_keywords = []
 
     # flatten the list of lists of filenames to a single list (may be empty)
-    opts.statistic_files = sum(opts.statistic_files, [])
+    # if needed (e.g. not calling get_statistic_from_opts in a loop)
```
GarethCabournDavies, line 2439: As this edits the input argparser namespace, we need to avoid redoing this step multiple times

pycbc/events/stat.py
```diff
@@ -711,8 +711,8 @@ def coinc_lim_for_thresh(self, sngls_list, thresh, limifo,
         if not self.has_hist:
             self.get_hist()
 
-        lim_stat = [b['snglstat'] for a, b in sngls_list if a == limifo][0]
-        s1 = thresh ** 2. - lim_stat ** 2.
+        fixed_stat = [b['snglstat'] for a, b in sngls_list if a != limifo][0]
```
ArthurTolley, line 714: Is this a change in logic (going from == to !=) or was this not working properly before?

ArthurTolley (approved):
Just a single question from me but happy to approve. Glad to see the fits being included in the live example too!

pycbc/events/stat.py
GarethCabournDavies: This wasn't working properly before - the statistic from the limifo was being used instead of the fixed IFO network

Looking again, I don't think this will actually work, so I will check it again

GarethCabournDavies:
Note: new commits do not affect the part of the code which Arthur has reviewed

GarethCabournDavies:
poke @titodalcanton to look at the supervision scripts here

bin/live/pycbc_live_combine_single_significance_fits
```diff
@@ -143,7 +149,9 @@ cons_counts_out = {ifo: np.inf * np.ones(len(alphas_bin[ifo])) for ifo in args.i
 logging.info("Writing results")
 
 fout = h5py.File(args.output, 'w')
-fout.attrs['fit_threshold'] = fit_thresh
+for ifo in args.ifos:
+    fout.attrs[f'{ifo}_fit_threshold'] = fit_thresh[ifo]
+    fout.attrs[f'{ifo}_fit_function'] = fit_func[ifo]
```
titodalcanton, line 154: Why not put these attributes inside the interferometer groups instead?

bin/live/pycbc_live_plot_single_significance_fits
```diff
@@ -71,17 +71,22 @@ with h5py.File(args.trigger_fits_file, 'r') as trfit_f:
     ifos = [k for k in trfit_f.keys() if not k.startswith('bins')]
 
     # Grab some info from the attributes
-    sngl_ranking = trfit_f.attrs['sngl_ranking']
-    fit_threshold = trfit_f.attrs['fit_threshold']
-    fit_function = trfit_f.attrs['fit_function']
+    if trfit_f.attrs['ranking_statistic'] in ['quadsum','single_ranking_only']:
+        x_label = trfit_f.attrs['sngl_ranking']
+    else:
+        x_label = "Statistic"
```
titodalcanton, line 77:
```suggestion
        x_label = "Ranking statistic"
```

bin/live/pycbc_live_plot_single_significance_fits
```diff
@@ -134,8 +142,15 @@ for ifo in all_ifos:
...
+        bin_prop = bin_num / len(tbins)
+        bin_colour = plt.get_cmap(args.colormap)(bin_prop)
+
         # Skip if there are no triggers in this bin in this IFO
-        if not any(inbin):
+        if not any(inbin) or alphas[ifo][bin_num] == -1:
+            ax.plot([],[], linewidth=2,
+                color=bin_colour, label=binlabel, alpha=0.6)
+            ax.plot([],[], "--", color=bin_colour,
+                label="No triggers")
```
titodalcanton, line 153: Could we format these lines as `black` would please?

bin/live/pycbc_live_single_significance_fits
```diff
@@ -396,16 +406,32 @@ for ifo in events:
...
         )
 
 logging.info("Writing results")
 with h5py.File(args.output, 'w') as fout:
-    for ifo in events:
+    for ifo in args.ifos:
+        if ifo not in events:
+            # There were no triggers, but we should still produce some
+            # information
+            fout_ifo = fout.create_group(ifo)
```
titodalcanton, line 425: This line is common to both branches of the `if` statement, move earlier.

titodalcanton: Same is true for the `fit_function` and `fit_threshold` attrs below.

bin/live/pycbc_live_single_significance_fits
```diff
@@ -427,8 +455,7 @@ with h5py.File(args.output, 'w') as fout:
     fout.attrs['analysis_date'] = args.analysis_date
     fout.attrs['input'] = sys.argv
     fout.attrs['cuts'] = args.template_cuts + args.trigger_cuts
-    fout.attrs['fit_function'] = args.fit_function
-    fout.attrs['fit_threshold'] = args.fit_threshold
     fout.attrs['sngl_ranking'] = args.sngl_ranking
```
titodalcanton, line 458: Should this line be kept?

examples/live/make_fit_coeffs.py
titodalcanton: Could you turn your note into a docstring for this file?

pycbc/events/single.py
```diff
@@ -141,13 +167,27 @@ def verify_args(args, parser, ifos):
 
     @classmethod
     def from_cli(cls, args, ifo):
+        # Allow None inputs
+        stat_files = args.statistic_files or []
+        stat_keywords = args.statistic_keywords or []
+
+        # flatten the list of lists of filenames to a single list
+        # (may be empty)
+        stat_files = sum(stat_files, [])
```
titodalcanton, line 176: Not sure I get this. How is `args.statistic_files` a list of lists?

pycbc/events/single.py
```diff
@@ -141,13 +167,27 @@ def verify_args(args, parser, ifos):
...
         return cls(
            ifo, newsnr_threshold=args.single_newsnr_threshold[ifo],
            reduced_chisq_threshold=args.single_reduced_chisq_threshold[ifo],
            duration_threshold=args.single_duration_threshold[ifo],
            fixed_ifar=args.single_fixed_ifar,
+           maximum_ifar=args.single_maximum_ifar[ifo],
            fit_file=args.single_fit_file,
-           sngl_ifar_est_dist=args.sngl_ifar_est_dist[ifo]
+           sngl_ifar_est_dist=args.sngl_ifar_est_dist[ifo],
+           statistic=args.ranking_statistic,
```
titodalcanton, line 187: I am not sure I understand why `--ranking-statistic` is relevant to the `LiveSingle` class. I probably missed or forgot some conversation…

pycbc/events/single.py
```diff
@@ -158,43 +198,62 @@ def check(self, trigs, data_reader):
...
+            trig_chisq = trigs['chisq']
+            trig_snr = trigs['snr']
+
         valid_idx = (trigs['template_duration'] >
                      self.thresholds['duration']) & \
-                    (trigs['chisq'] <
+                    (trig_chisq <
                      self.thresholds['reduced_chisq']) & \
-                    (trigs['snr'] >
+                    (trig_snr >
```
titodalcanton, line 214: This seems to effectively change the meaning of `--reduced-chisq-threshold` and `--newsnr-threshold` in case the PSD var statistic is used, i.e. "reduced chisq" and "newsnr" will no longer mean what people historically think they mean. Should we rename the two options then?

bin/live/pycbc_live_single_significance_fits
GarethCabournDavies: These are now in the IFO groups

pycbc/events/single.py
GarethCabournDavies: This is so that it can be passed into the stat_calculator property, which is used later (line 229)

pycbc/events/single.py
titodalcanton: Huh, I had no idea argparse behaved like that…

titodalcanton: Indeed I just verified this, apparently this comes from `action='append'`.

bin/live/pycbc_live_combine_single_significance_fits
```diff
@@ -152,6 +157,9 @@ fout['fits_dates'] = ad + start_time_n
 for ifo in args.ifos:
     logging.info(ifo)
     fout_ifo = fout.create_group(ifo)
+    for ifo in args.ifos:
+        fout_ifo.attrs['fit_threshold'] = fit_thresh[ifo]
+        fout_ifo.attrs['fit_function'] = fit_func[ifo]
```
titodalcanton, line 162: I think there is an extra loop now:
```suggestion
    fout_ifo.attrs['fit_threshold'] = fit_thresh[ifo]
    fout_ifo.attrs['fit_function'] = fit_func[ifo]
```

GarethCabournDavies: of course :picard:

titodalcanton (approved):
I think we can merge and start proper tests on this now, provided the CI passes.

PR #4690, opened 2024-04-09
generalize catalog retrieval

ahnitz:


ahnitz:
This is untested (so likely has some bugs still). I'll remove WIP when that is done.

ahnitz:
This PR  updates the catalog package so that catalogs from the LVK are retreived more programmatically and don't require updating when a  named "catalog" is added.

pycbc/catalog/__init__.py
```diff
@@ -40,31 +40,42 @@
...
                     name = mname
-                    self.data = data[name]
+                    data = data[name]
                     break
-            else:
-                raise ValueError('Did not find merger matching'
-                                 ' name: {}'.format(name))
+    else:
+        raise ValueError('Did not find merger matching'
+                         ' name: {}'.format(name))
```
titodalcanton, line 65:
```suggestion
        raise ValueError(f'Did not find merger matching name: {name}')
```

pycbc/catalog/__init__.py
```diff
@@ -40,31 +40,42 @@
 _aliases['z'] = _aliases['redshift'] = 'redshift'
 _aliases['distance'] = 'luminosity_distance'
 
+def find_event_in_catalog(source='any'):
```
titodalcanton, line 43: I suggest using None in place of the string 'any' (same in the constructor of Merger, below).

pycbc/catalog/catalog.py
```diff
@@ -37,16 +37,13 @@
...
-             'GWTC-2.1-marginal': 'LVC',
-             'GWTC-3-confident': 'LVC',
-             'GWTC-3-marginal': 'LVC'}
+
+def lvk_catalogs():
+    _catalog_source = "https://gwosc.org/eventapi/json/"
+    catalog_list = get_file(_catalog_source)
+    return catalog_list
+
+_catalogs = {cname: 'LVK' for cname in lvk_catalogs.keys()}
```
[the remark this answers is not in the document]
titodalcanton: I suggest moving the definition of `_catalogs` (and thus the call to `lvk_catalogs()`) from the global namespace into the functions that need it, below.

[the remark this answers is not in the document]
ahnitz: Yeah, connecting on every import is clearly bad. That I'll fix, and maybe add in caching if I can figure out how to do that with the caveats abovve.

pycbc/catalog/__init__.py
ahnitz: Ok

ahnitz:
Ok, I think this now works. The following for example now works as expected. 

```
import pycbc.catalog
from matplotlib import pyplot as plt
m = pycbc.catalog.Merger("GW230529_181500")
ts = m.strain('L1').whiten(4, 4).time_slice(m.time-8, m.time+2)

t, f, p = ts.qtransform(.01, logfsteps=200, qrange=(120, 120), frange=(30, 400))
plt.pcolormesh(t, f, p)
plt.axvline(m.time)
plt.yscale('log')
```
download

titodalcanton (approved):
Looks good.

PR #4691, opened 2024-04-10
Adding detector class for calculating LISA projections and response function

acorreia61201:


acorreia61201:
The `LISA_detector` class in my most recent commit should be sufficient for calculating LISA TDI channels from radiation frame PyCBC waveforms. I am able to replicate the results from the FastLISAResponse tutorial (shown [here](https://github.com/mikekatz04/lisa-on-gpu/blob/master/examples/fast_LISA_response_tutorial.ipynb)), with additional functionality for inputting reference times. By default, the TDI channels are calculated using `lisatools.detector.ESAOrbits` (a simulated realistic orbital file), but orbital information can be input using the `Orbits` base class in LISA Analysis Tools. All modifications use the most recent versions of FastLISAResponse (ver. 1.0.5) and LISA Analysis Tools.

One pending issue regards the usage of `Orbits.configure`, which will throw an error with the most recent version of LISA Analysis Tools. I have submitted a pull request to that repository that should solve the issue, but I am still awaiting a response from their development team. Additionally, I haven't been able to conduct timing tests on GPUs, as the most recent version of FastLISAResponse cannot be built with GPU support.

ahnitz:
@acorreia61201 You'll want to rebase this PR, get the unittest to pass with your PR, and then demonstrate how this compares to say calling BBHx directly. E.g. can you post an example that reproduces BBHx output but using this class and say starting from IMRPhenomD?

WuShichao:
@acorreia61201 Can you do this test: from the typical parameter space of SMBHB anD stellar-mass BBH to compute the match between BBHx plugin and this PR? Let's do 1000 draw and plot histogram of match.

pycbc/detector.py
```diff
@@ -655,134 +655,810 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
-    """For LISA detector
+class AbsSpaceDet(ABC):
     """
-    def __init__(self):
-        None
+    Parent class to space-based detector classes.
+
+    Parameters
+    ----------
+    reference_time : float (optional)
```
ahnitz, line 669: @acorreia61201 For a base class, you want to make sure it really is the minimal set of things that all implementations must have to work. I think there are few things here that don't necessarily qualify. It might be worth going through and pruning a bit here. 

A base class for example shouldn't have options just because a child class might have them. It's the child class's job to implement things that are specific to it.

pycbc/detector.py
```diff
@@ -655,134 +655,810 @@ def overhead_antenna_pattern(right_ascension, declination, polarization):
...
-        self.location = np.array([x, y, z])
+        hp_ssb = hp*cphi - hc*sphi
+        hc_ssb = hp*sphi + hc*cphi
 
-        return self.location
+        return hp_ssb, hc_ssb
 
-    def get_gcrs_pos(self, location):
-        """ Transforms ICRS frame to GCRS frame
+    def preprocess(self, hp, hc, polarization=0, reference_time=None):
```
ahnitz, line 751: What does this routine mean for a truly generic implementation?

ahnitz:
An organizational comment, this might be a good time to turn the module into a package itself as the file is getting quite long.

pycbc/detector.py
acorreia61201: Both the FLR and LDC versions had some common routines that needed to be applied to the waveforms before inputting them into the response generation. Mainly, this function is making sure that the signals fit into the orbital data being supplied. The original intent was to circumvent errors both backends throw when the signal extends past the orbital data, but I think this could be a generic way to ensure the waveform can be projected given the orbits.

There are some other functions in here (e.g. polarization, padding, start/ref time caching) that I added just to prevent them from being copied for each implementation, although in the interest of making the base class as generic as possible I could move these to the backend classes.

ahnitz:
@acorreia61201 It's fine to have common functionality / methods between the LDC / flr methods, but that's somewhat different than what should be in the abstract base class. You could have them share methods by using an auxiliary class that they both inherit from or a simple function that they both can call. So avoiding code duplication is definitely good. For classes though we also want to make sure there is a clear interface of what *must* be defined so that people can build against it.

acorreia61201:
@ahnitz When you say turn the module to a package, do you mean moving the space-based classes to a new file (say, `detector_space.py`)? If so, I could move those common methods into functions that can be called by the backends as needed.

ahnitz:
@acorreia61201 I mean this https://docs.python.org/3/tutorial/modules.html#packages

e.g. create a 'detector' folder rather than signle module we have now. Then split up the functions into separate files e.g. ground.py space.py etc. A package also has an __init__.py which can provide truly global / common functions and also preserve the expected namespace by importing things from the submodules so that nobody has to change their existing code and can still do say (pycbc.detector.Detector("H1")...

Does that make sense?

acorreia61201:
I think that makes sense; it seems simple enough looking at some of the other `__init__` files. I can set this up once I've restructured the base space class and child classes.

acorreia61201:
@ahnitz @WuShichao I've reworked the structure of the space classes and detector module as suggested. Are there any other checks or changes I should make once the unittests pass?

WuShichao:
@acorreia61201 Thanks, I will take a look this week.

pycbc/detector/space.py
```diff
@@ -0,0 +1,899 @@
...
+    """
+    cphi = cos(2*polarization)
+    sphi = sin(2*polarization)
+
+    hp_ssb = hp*cphi - hc*sphi
+    hc_ssb = hp*sphi + hc*cphi
+
+    return hp_ssb, hc_ssb
+
+def preprocess(hp, hc, orbit_start_time, orbit_end_time,
```
ahnitz, line 70: This might need a better name. Maybe it would be clearly by not mixing up applying polarization and checking the time span ?

pycbc/detector/space.py
```diff
@@ -0,0 +1,899 @@
...
+    @abstractmethod
+    def sky_coords(self):
+        """
+        List the sky coordinate names for the detector class.
+        Raises a NotImplementedError if not specified in child class.
+        """
+        return
+
+    @abstractmethod
+    def orbits_init(self):
```
ahnitz, line 228: I'm not sure if `get_links` or `orbits_init` should be in the base class or not. What does get_links return? For oribts_init, why wouldn't that just be a part of the init of the class itself. It doesn't seem to take any arguments. 

The standard arguments and expected return should be specified for the base class methods. Yes, it's not enforced, but it's in part so that people (and future developers) know what the API is supposed to be. Project wave for example should specify the minimum interface that can be expected (but not for example any optional argument or extensions).

ahnitz (approved):
@acorreia61201 I think this is good enough to move forward with as a starting point. There will likely need to be revision as we learn more. 

I'm approving this, but I'd like to at least see the minimum requirement arguments for the absbaseclass methods be added since those define the broadly compatible API. 

As we find what is truly common we can always move more things into the baseclass.

pycbc/detector/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .ground import *
```
ahnitz, line 1: If you do an import *, make sure the functions / classes that should be imported are included in an __all__ property at the bottom of the file. That way it only imports the things we actually want exposed at the higher level.

pycbc/detector/space.py
acorreia61201: `get_links` in the child classes returns the GW projected to the six laser links. This would be the intermediary step before calculating the TDI combinations. Initially, I thought it would be useful to have this be common to all classes in case, e.g., someone wanted to homebrew their own TDI combinations. I can see in retrospect, though, that this shouldn't be required. The same is true for `orbits_init`, although I've left them as methods for now just to keep things more organized.

cdcapano:
@WuShichao @ahnitz I'm just going to merge this now since it's passing all tests. If we want to make other changes they can be addressed in another PR.

WuShichao:
Yeah, I forgot to approve this, yes, it's time to work on injection PR for this. I will put more detailed comments on that PR.

PR #4692, opened 2024-04-11
Allow the user to specify the output directory for nessai

mj-will:
Update the nessai interface to allow the user to set the `output` option in an ini file. If not specified, the code falls back to the path to the checkpoint file rather than the CWD.

@HumphreyWang hopefully this addresses the issue you were having.

### DIFF
--- a/pycbc/inference/sampler/nessai.py
+++ b/pycbc/inference/sampler/nessai.py
@@ -85,10 +85,6 @@ class NessaiSampler(BaseSampler):
 
     def run(self, **kwargs):
         """Run the sampler"""
-        out_dir = os.path.join(
-            os.path.dirname(os.path.abspath(self.checkpoint_file)),
-            "outdir_nessai",
-        )
         default_kwds, default_run_kwds = self.get_default_kwds(
             importance_nested_sampler=self.extra_kwds.get(
                 "importance_nested_sampler", False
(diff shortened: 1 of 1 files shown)


HumphreyWang:
@mj-will I tried to run an example and it worked well, thanks again for the PR!

pycbc/inference/sampler/nessai.py
```diff
@@ -98,6 +94,15 @@ def run(self, **kwargs):
         extra_kwds = self.extra_kwds.copy()
         run_kwds = self.run_kwds.copy()
 
+        # Output in kwargs takes priority of extra kwds.
+        output = kwargs.pop("output", extra_kwds.pop("output", None))
+        # If neither have been specified, use the path from the checkpoint file
+        if output is None:
```
ahnitz, line 100: @mj-will So if you don't set an output dir, it will still create one? Is that intentional?

mj-will: @ahnitz yes, nessai takes an output directory as an input rather than a file, so I opted to have it create one by default rather than dumping everything into the current working directory. I don't have a strong preference though and happy to change it.

ahnitz: Ah, ok, so that's really no different than the current situation.

ahnitz (approved):

ahnitz:
@HumphreyWang has confirmed that this works so I think this good to go. The change itself is straightforward.

PR #4694, opened 2024-04-12
Add support for inessai sampler

mj-will:

This change changes: the `nessai` sampler to support the importance nested sampler version

This change: is tested via an example, follows style guidelines (See e.g. [PEP8](https://peps.python.org/pep-0008/)), has been proposed using the [contribution guidelines](https://github.com/gwastro/pycbc/blob/master/CONTRIBUTING.md)

This change will: does not break any existing code

## Motivation

With the release of nessai 0.12.0, a method for mapping from the prior to the unit hypercube is no longer required for the importance nested sampler (ins). This means we can support the ins (inessai) with a few minor changes to the existing implementation.

## Contents

Updates to the existing class so that it supports the `importance_nested_sampler` option. This is `False` by default.

I also added a second class `INessaiSampler` that has this default switch to `True` and has the name set to `inessai`.

This means there are two ways of calling the importance nested sampler.

## Testing performed

Run the example included in `examples/inference/samplers/`

## Additional notes

**To-do**

- [ ] Test checkpoint after this change, since it reworks the `samples` property.
- [ ] Test on BBH

### DIFF
--- a/companion.txt
+++ b/companion.txt
@@ -17,7 +17,7 @@ https://github.com/willvousden/ptemcee/archive/master.tar.gz
 # Force the cpu-only version of PyTorch
 --extra-index-url https://download.pytorch.org/whl/cpu
 torch
-nessai>=0.11.0
+nessai>=0.12.0
 snowline
 
 # useful to look at PyCBC Live with htop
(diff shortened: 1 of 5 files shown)


GarethCabournDavies:
@mj-will what is the status of this?

mj-will:
@GarethCabournDavies I think the code works but I haven't had a chance to test the remaining examples. It's relatively low-priority for me at the moment, so I probably won't get to it until the end of the summer at the earliest.

PR #4695, opened 2024-04-12
Add option to plot fits in bank corner plot

GarethCabournDavies:
Added a fits-file option to the `pycbc_plot_bank_corner` code so that we can use it to plot fit coefficient, number of templates per bin, etc. in a nice format

This is a new feature.
This change affects plotting, for possible use with the offline and live codes, and provides a minor bugfix for an edge case in a seldom-used module

## Motivation
As we are intending to use template fits which automatically update in pycbc live, we should make diagnostic plots whenever the file is updated.

## Contents
 - Added a fits-file option to pycbc_plot_bank_corner
 - Added the fits file into the default title
    - As this makes it quite long, I've added the --title option to the code to choose something else, and included line wrapping if using the default
 - As the fits can have default 'this hasn't worked' values of -1, I have updated some of the min/max calculations as appropriate

## Testing performed
I have plotted using the fit parameters as color, and as a parameter in the corner plot, in https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/bank_corner_fit_values/testing/outputs/

## Additional notes
A couple of minor fixes elsewhere:
 - the create_multidim_plot was broken when using 2 parameters but not the marginal plotting
 - when using duration directly from the bank, the bank_conversions get_bank_property was not dealing with this properly

### DIFF
--- a/bin/plotting/pycbc_plot_bank_corner
+++ b/bin/plotting/pycbc_plot_bank_corner
@@ -26,6 +26,8 @@ import h5py
 import numpy as np
 import argparse
 import logging
+from textwrap import wrap
+
 from pycbc import init_logging
 import pycbc.version
 from pycbc import __version__
@@ -38,6 +40,13 @@ from pycbc.results.scatter_histograms import create_multidim_plot
 from pycbc.results import metadata
 
 conversion_options = bconv.conversion_options
+_fit_parameters = [
+    'count_above_thresh',
+    'count_in_template',
+    'fit_coeff',
+    'median_sigma'
+]
+
 
 parser = argparse.ArgumentParser(usage='pycbc_plot_bank_corner [--options]',
                     description=__doc__)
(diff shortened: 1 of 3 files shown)


GarethCabournDavies:
Using the O3 broad bank with a random chunk from that, we can get: 
image

maxtrevor:
@GarethCabournDavies I've looked over the code. However I'm a bit confused about how to interpret the plot you've posted. When looking at a plot like that, what should I be looking for? What would indicate good / bad performance?

maxtrevor (approved):

GarethCabournDavies:
> @GarethCabournDavies I've looked over the code. However I'm a bit confused about how to interpret the plot you've posted. When looking at a plot like that, what should I be looking for? What would indicate good / bad performance?
> 
> edit to add that I'm approving, my questions are not a blocker.

In general it would be anything which pops out of the 'normal' range for the fit coefficients - so basically the scale of the colour bar. As well as spots showing up, e.g. if something like [this plot](https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/bank_corner_fit_values/testing/outputs/test_plot_fit_coeff_H1_FIT_BY_TEMPLATE.png) was showing up in fit_over_param plots, that would be very bad, as it would basically boost anything in that region, or blind us to it if it were very low.

I think for live running, looking at the code over multiple days and noticing particular differences between the plots would be the thing to look at.

PR #4697, opened 2024-04-15
Script to create hdf trigger merge files from PyCBC Live output trigger files

ArthurTolley:
This script is used to convert a large number of PyCBC Live trigger output files to a hdf trigger merge file.

This is a: new feature

This change affects: the live search

## Motivation

Using template fits in the PyCBC Live search (#4527) requires template fits to be made separate to using them. One of the requirements to make these files is a hdf trigger merge. This code produces that trigger merge file.

## Contents

The PyCBC Live search outputs a trigger file every stride (8s for Live, 1s for Early Warning) containing any triggers found within the previous stride. This code takes a number of options to be given these trigger files and will then collate them into a single hdf object: from a list of trigger files, from a directory containing subdirectories containing trigger files (typically one subdirectory for each day), from a start and end date, from a start date and a number of days, from a gps start and end time.

The template fit creation code expects the triggers to be in a certain format (the offline trigger file format) and therefore the Live triggers need to be converted to match that format, creating new datasets where needed and ensuring other datasets are correct: for example, PyCBC Live stores chisq as reduced chisq whereas the offline search doesn't, so we convert these PyCBC Live triggers to the offline format.

The triggers are also sorted by template_id and region references are created using the template_id boundaries to allow for rapid access in future codes.

- Template fits in Live: #4527  

## Testing performed

I have taken 2 days worth of PyCBC Live triggers with H1, L1 & V1 triggers and run the following scripts to test:

From a file containing a list of trigger files:
```
python ../pycbc_live_collate_triggers \
    --trigger-file-method file \
    --list-of-trigger-files test_file_list.txt \
    --ifos H1 L1 V1 \
    --output-trigger-file-list test_file_list_collated.txt \
    --output-dir ./ \
    --output-file-name test_collated_triggers.hdf
```

From a directory containing subdirectories containing trigger files:
```
python ../pycbc_live_collate_triggers \
     --trigger-file-method dir \
     --trigger-dir /home/arthur.tolley/PyCBC_changes/collate_triggers_script/pycbc/bin/live/testing \
     --ifos H1 L1 V1 \
     --output-trigger-file-list test_file_list_collated_dir.txt \
     --output-dir ./ \
     --output-file-name test_collated_triggers_dir.hdf
```

From a start and end date:
```
python ../pycbc_live_collate_triggers \
     --trigger-file-method start-end-date \
     --trigger-dir /home/arthur.tolley/PyCBC_changes/collate_triggers_script/pycbc/bin/live/testing \
     --start-date 2024-04-11 \
     --end-date 2024-04-12 \
     --ifos H1 L1 V1 \
     --output-trigger-file-list test_file_list_collated_start-end.txt \
     --output-dir ./ \
     --output-file-name test_collated_triggers_start-end.hdf
```

From a start date and a number of days:
```
python ../pycbc_live_collate_triggers \
     --trigger-file-method start-num-days \
     --trigger-dir /home/arthur.tolley/PyCBC_changes/collate_triggers_script/pycbc/bin/live/testing \
     --start-date 2024-04-11 \
     --num-days 2 \
     --ifos H1 L1 V1 \
     --output-trigger-file-list test_file_list_collated_start-num.txt \
     --output-dir ./ \
     --output-file-name test_collated_triggers_start-num.hdf
```

From a start and end gps time:
```
python ../pycbc_live_collate_triggers \
     --trigger-file-method gps-start-end-time \
     --trigger-dir /home/arthur.tolley/PyCBC_changes/collate_triggers_script/pycbc/bin/live/testing \
     --gps-start-time 1396828816 \
     --gps-end-time 1396844864 \
     --ifos H1 L1 V1 \
     --output-trigger-file-list test_file_list_collated_gps.txt \
     --output-dir ./ \
     --output-file-name test_collated_triggers_gps.hdf
```

### DIFF
--- a/bin/live/pycbc_live_collate_triggers
+++ b/bin/live/pycbc_live_collate_triggers
@@ -1,0 +1,302 @@
+import glob
+import numpy
+import argparse
+import h5py
+import os
+import logging
+import timeit
+from datetime import datetime, timedelta
+
+# Set up the command line argument parser
+parser = argparse.ArgumentParser(description='Find trigger files and combine '
+                                             'them into a single hdf trigger '
+                                             'merge file.')
+
+parser.add_argument('--trigger-file-method', type=str, required=True,
+                    help='The method to use to find the trigger files. '
+                         'Options are: file, dir, start-end-date, '
+                         ' start-num-days, gps-start-end-time.',
+                    choices=['file', 'dir', 'start-end-date',
+                             'start-num-days', 'gps-start-end-time'])
+
+parser.add_argument('--trigger-dir', type=str, required=False,
+                    help='The directory containing all of the PyCBC Live '
+                         'trigger files. The directory must contain sub-'
+                         'directories with naming schemes "YYYY-MM-DD".'
+                         'The trigger files must be of the format: '
+                         '"{ifos}-Live-GPSTIME-INCREMENT.hdf".')
+
+# Take trigger files from file
+parser.add_argument('--list-of-trigger-files', type=str,
+                    help='A file containing all the trigger files you '
+                    ' would like to collate.')
+
+# Take trigger files from start and end date
+parser.add_argument('--start-date', type=str, required=False,
+                    help='The first day of triggers you want to collate.'
+                         'If no end date is given, the start date is the '
+                         'only day.')
+parser.add_argument('--end-date', type=str, required=False,
+                    help='The final day of triggers to collect. If no end '
+                         'is provided then only the start day is used.')
+
+# Take trigger files from start and num days
+parser.add_argument('--num-days', type=int, required=False,
+                    help='The number of days after (and including) the start '
+                    'date to collate triggers from.')
+
+# Take trigger files from gps start and end time
+parser.add_argument('--gps-start-time', type=int, required=False,
+                    help='Start time of the collection of trigger files.')
+parser.add_argument('--gps-end-time', type=int, required=False,
+                    help='End time of the collection of trigger files.')
+
+parser.add_argument('--ifos', type=str, required=True, nargs='+',
+                    help='The detectors to extract triggers for in the '
+                         'trigger files.')
+
+# Option to output list of trigger files
+parser.add_argument('--output-trigger-file-list', type=str,
+                    help='Name of the output file to save the list of trigger files.')
+
+parser.add_argument('--output-dir', type=str, required=True,
+                    help='The directory to write the collated trigger file '
+                         'to. The file will be named with the following '
+                         'format: "{ifos}-Live-STARTGPS-ENDGPS.hdf" unless '
+                         'an output file name is provided.')
+parser.add_argument('--output-file-name', type=str, required=False,
+                    help='The name of the output file to save the collated '
+                         'trigger file to. If not provided, the file will be '
+                         'named with the format: "{ifos}-Live-STARTGPS-ENDGPS.hdf"')
+
+args = parser.parse_args()
+
+logging.basicConfig(level=logging.INFO)
+
+# GRAB THE TRIGGER FILES TO COLLATE
+if args.trigger_file_method == 'file':
+    assert args.list_of_trigger_files, 'Please provide a list of trigger files.'
+    trigger_files = numpy.loadtxt(args.list_of_trigger_files,
+                                  delimiter=',', dtype=str)
+
+if args.trigger_file_method == 'dir':
+    assert args.trigger_dir, 'Please provide a directory containing trigger files.'
+    trigger_files = [
+        l for l in glob.glob(args.trigger_dir + '/*/*-Live-*.hdf', recursive=True)
+    ]
+
+if args.trigger_file_method == 'start-end-date':
+    assert args.start_date, 'Please provide a start date.'
+    assert args.end_date, 'Please provide an end date.'
+    assert args.trigger_dir, 'Please provide a directory containing trigger files.'
+    # Convert dates to datetime to get the days array
+    start_date = datetime.strptime(args.start_date, '%Y-%m-%d').date()
+    end_date = datetime.strptime(args.end_date, '%Y-%m-%d').date()
+
+    # Create the days array
+    days = []
+    delta = timedelta(days=1)
+
+    # Loop through the days and add each date to the list
+    current_date = start_date
+    while current_date <= end_date:
+        days.append(current_date)
+        current_date += delta
+
+    # Convert to strings in a list
+    days = [str(date).replace('-', '_') for date in days]
+    num_days = str(len(days))
+
+    trigger_files = [
+        os.path.join(args.trigger_dir, day, trigger_file)
+        for day in days
+        for trigger_file in os.listdir(os.path.join(args.trigger_dir, day))
+    ]
+
+if args.trigger_file_method == 'start-num-days':
+    assert args.start_date, 'Please provide a start date.'
+    assert args.num_days, 'Please provide a number of days.'
+    assert args.trigger_dir, 'Please provide a directory containing trigger files.'
+    # Convert dates to datetime to get the days array
+    start_date = datetime.strptime(args.start_date, '%Y-%m-%d').date()
+    num_days = timedelta(days=args.num_days - 1)
+    end_date = start_date + num_days
+
+    # Create the days array
+    days = []
+    delta = timedelta(days=1)
+
+    # Loop through the days and add each date to the list
+    current_date = start_date
+    while current_date <= end_date:
+        days.append(current_date)
+        current_date += delta
+
+    # Convert to strings in a list
+    days = [str(date).replace('-', '_') for date in days]
+    num_days = str(len(days))
+
+    trigger_files = [
+        os.path.join(args.trigger_dir, day, trigger_file)
+        for day in days
+        for trigger_file in os.listdir(os.path.join(args.trigger_dir, day))
+    ]
+
+if args.trigger_file_method == 'gps-start-end-time':
+    assert args.gps_start_time, 'Please provide a GPS start time.'
+    assert args.gps_end_time, 'Please provide a GPS end time.'
+    assert args.trigger_dir, 'Please provide a directory containing trigger files.'
+
+    trigger_files = [
+        l for l in glob.glob(args.trigger_dir + '/*/*-Live-*.hdf', recursive=True)
+        if float(l.split('/')[-1][10:27]) > args.gps_start_time and float(l.split('/')[-1][10:27]) < args.gps_end_time
+    ]
+
+start = timeit.default_timer()
+logging.info(f" {len(trigger_files)} files found")
+
+if args.output_trigger_file_list:
+    logging.info(' Writing list of trigger files to: ' + args.output_trigger_file_list)
+    with open(args.output_trigger_file_list, 'w') as f:
+        for item in trigger_files:
+            f.write("%s\n" % item)
+
+# Record start and end gpstime for the output file
+#  This assumes the trigger files are sorted chronologically
+start_gpstime = float(trigger_files[0].split('/')[-1][10:27])
+end_gpstime = float(trigger_files[-1].split('/')[-1][10:27])
+
+###########################
+# COLLATE THE TRIGGER FILES
+###########################
+
+if args.output_file_name:
+    output_file = args.output_dir + args.output_file_name
+else:
+    ifo_string = "".join(args.ifos)
+    print(ifo_string)
+    output_file = args.output_dir + f'{ifo_string}-Live-' + start_gpstime + '-' + end_gpstime + '.hdf'
+
+logging.info(f" Creating a file at: {output_file}")
+with h5py.File(output_file, 'a') as destination:
+    # Create the ifo groups in the trigger file
+    for ifo in args.ifos:
+        if ifo not in destination:
+            destination.create_group(ifo)
+
+    file_count = 0
+    for source_file in trigger_files:
+        file_count += 1
+        trigger_file = os.path.basename(source_file)
+        if not (trigger_file.endswith('.hdf') and trigger_file.startswith('*-Live')):
+            continue
+
+        start_time = float(trigger_file.split('-')[2])
+        duration = float(trigger_file.split('-')[3][:-4])
+        end_time = start_time + duration
+
+        if file_count % 100 == 0:
+            logging.info(f" Files appended: {file_count}/{len(trigger_files)}")
+
+        with h5py.File(source_file, 'r') as source:
+            for ifo in args.ifos:
+                triggers = source[ifo]
+                if ('approximant' not in triggers) or (len(triggers['approximant']) == 0):
+                        continue
+
+                for name, dataset in triggers.items():
+
+                    if name in destination[ifo]:
+                        if triggers[name].shape[0] == 0:
+                            continue
+                        # Append new data to existing dataset in destination
+                        destination[ifo][name].resize((destination[ifo][name].shape[0] + triggers[name].shape[0]), axis=0)
+                        destination[ifo][name][-triggers[name].shape[0]:] = triggers[name]
+                    else:
+                        destination[ifo].create_dataset(name, data=dataset, chunks=True, maxshape=(None,))
+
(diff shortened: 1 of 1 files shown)


bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,306 @@
...
+
+if args.output_trigger_file_list:
+    logging.info(' Writing list of trigger files to: ' + args.output_trigger_file_list)
+    with open(args.output_trigger_file_list, 'w') as f:
+        for item in trigger_files:
+            f.write("%s\n" % item)
+
+# Record start and end gpstime for the output file
+#  This assumes the trigger files are sorted chronologically
+start_gpstime = float(trigger_files[0].split('/')[-1].split("-")[2])
```
GarethCabournDavies, line 166: Please add a fixme that the trigger files should have the gps start and end times of the file added as attributes, and that the start/end times should be obtained by getting the min/max of these over the file reading loop

GarethCabournDavies: we should not assume that the trigger files are sorted chronologically

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,306 @@
...
+    output_file = args.output_dir + f'{ifo_string}-Live-' + start_gpstime + '-' + end_gpstime + '.hdf'
+
+logging.info(f" Creating a file at: {output_file}")
+with h5py.File(output_file, 'a') as destination:
+    # Create the ifo groups in the trigger file
+    for ifo in args.ifos:
+        if ifo not in destination:
+            destination.create_group(ifo)
+
+    # The user must give the ifos in the same order as the trigger files
```
GarethCabournDavies, line 187: I don't understand this comment?

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,306 @@
...
+    #  For PyCBC Live it's always H1L1 or H1L1V1?
+    ifo_prefix = "".join(args.ifos)
+
+    file_count = 0
+    for source_file in trigger_files:
+        file_count += 1
+        trigger_file = os.path.basename(source_file)
+        if not (trigger_file.endswith('.hdf') and trigger_file.startswith(f'{ifo_prefix}-Live')):
+            continue
```
GarethCabournDavies, line 196: Can you do this part during the trigger_files calculation earlier? By doing it this way your progress logging is misleading

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,306 @@
...
+        tids = numpy.arange(numpy.max(template_ids) + 1, dtype=int)
+
+        sorted_indices = numpy.argsort(template_ids)
+        sorted_template_ids = template_ids[sorted_indices]
+        unique_template_ids, template_id_counts = numpy.unique(sorted_template_ids, return_counts=True)
+        index_boundaries = numpy.cumsum(template_id_counts)
+        template_boundaries = numpy.insert(index_boundaries, 0, 0)[:-1]
+
+        # Re-running purposes, comment out otherwise
+        #del triggers['template_boundaries']
```
GarethCabournDavies, line 264: So should this be included / removed?

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,306 @@
...
+        # Grab the ifo you want
+        triggers = output[ifo]
+        try:
+            template_ids = triggers['template_id'][:]
+        except:
+            logging.info(f"  No triggers for {ifo}, skipping")
+            continue
+
+        # 17.4 seconds to do this
+        tids = numpy.arange(numpy.max(template_ids) + 1, dtype=int)
```
GarethCabournDavies, line 255: This will add edge cases if templates dont have triggers - maybe include the bank file as an input so this can be obtained from there

In fact you would want the bank file anyway for template_hash sorting - trigger merge files are kept in hash order for offline, we should check whether that should be used here

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,306 @@
...
+        #   template_duration, template_hash, template_id
+        for key in triggers.keys():
+            if len(triggers[key]) == len(template_ids):
+                logging.info(f'  Sorting {key} by template id')
+                sorted_key = triggers[key][:][sorted_indices]
+                triggers[key][:] = sorted_key
+
+
+        # Chisq is saved as reduced chisq for live triggers but offline
+        #  code required original chisq. This converts it back.
```
GarethCabournDavies, line 280:
```suggestion
        # code requires original chisq. This converts it back.
```

GarethCabournDavies (changes requested):
There should be discussion of certain points I'm not sure of (i.e. do we need template_hash sorting in this), but this is broadly right.

I think the arguments regarding the file discovery should be reworked / simplified though. I think something like https://github.com/gwastro/pycbc/pull/4354 should work, so I'll put the work in for that to be pulled in here

bin/live/pycbc_live_collate_triggers
ArthurTolley: The tids variable isn't actually used anywhere (mysteriously) so I will just remove this

bin/live/pycbc_live_collate_triggers
ArthurTolley: There is probably a nice way to do this but the gist is:
--ifos needs to take 'H1 L1 V1' in the same order that the output files from pycbc have the prefix of: H1L1V1-Live-

ArthurTolley: You can't give L1H1V1 because then it looks for L1H1V1 files but these don't exist!

bin/live/pycbc_live_collate_triggers
ArthurTolley: I stopped being lazy and extracted the gpstimes from the filenames and took the min/max

bin/live/pycbc_live_collate_triggers
ArthurTolley: I guess you could do something like: `ifo_permutations = [''.join(p) for p in itertools.permutations(ifos)]` and then check for all of the premutations. Or, the live search outputs a known order of ifos every time it runs so we could just force it

bin/live/pycbc_live_collate_triggers
ArthurTolley: Good spot, I missed this when squishing the two scripts together. This will record every file (pngs, xmls, etc.) as a trigger file until checked later. I will move this check. Thanks!

ArthurTolley: Okay, I've moved it out and added the ifo permutation part

GarethCabournDavies:
We are using this for inputs to fit_by_template and bin_trigger_rates_dq, which both use stat-threshold as input and discard anything below that - does it then make sense to implement the same here? That could save:
 - storage
 - computational effort in the hash sorting stage

GarethCabournDavies:
During testing, this code is quite slow, so I thought about what the later uses are; `pycbc_fit_sngls_by_template` and `pycbc_bin_trigger_rates_dq` at the moment.

They both apply cuts straight away. As a result, I think adding cuts would be good here and we can drastically reduce
 1. how quickly this runs
 2. how much space the files take up

This should mean that we can reuse a lot of the file-reading code from `pycbc_live_single_significance_fits`, which can be put into a module

GarethCabournDavies:
I've just found a bug, but not sure how best to fix it - in the case that there are no triggers from that template, the boundaries are not being set properly, so there are not enough region references being made
```
/H1/chisq_dof_template   Dataset {729910}
/H1/chisq_template       Dataset {729910}
/H1/coa_phase_template   Dataset {729910}
/H1/end_time_template    Dataset {729910}
/H1/sg_chisq_template    Dataset {729910}
/H1/sigmasq_template     Dataset {729910}
/H1/snr_template         Dataset {729910}
/H1/template_duration_template Dataset {729910}
/L1/chisq_dof_template   Dataset {729913}
/L1/chisq_template       Dataset {729913}
/L1/coa_phase_template   Dataset {729913}
/L1/end_time_template    Dataset {729913}
/L1/sg_chisq_template    Dataset {729913}
/L1/sigmasq_template     Dataset {729913}
/L1/snr_template         Dataset {729913}
/L1/template_duration_template Dataset {729913}
```
I think I can find the bug / propose a fix fairly soon

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,333 @@
...
+        triggers = output[ifo]
+        try:
+            template_ids = triggers['template_id'][:]
+        except:
+            logging.info(f"  No triggers for {ifo}, skipping")
+            continue
+
+        sorted_indices = numpy.argsort(template_ids)
+        sorted_template_ids = template_ids[sorted_indices]
+        unique_template_ids, template_id_counts = numpy.unique(sorted_template_ids, return_counts=True)
```
spxiwh, line 291: This line here doesn't include the `0` template_id_counts. I think you'd want to add that in.

GarethCabournDavies: That's a separate issue (though it is fixed a couple of lines down) - I think the problem is that the region references are not being created further down, so when the fit_by_template script tries to call them, it fails

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,333 @@
...
+
+# Check if all files found are in the correct format and are trigger files
+ifo_permutations = [''.join(p) for p in itertools.permutations(args.ifos)]
+ifo_prefix_set = set(ifo_permutations)
+
+trigger_files = []
+for source_file in initial_trigger_files:
+    trigger_file = os.path.basename(source_file)
+    if any(trigger_file.startswith(prefix + '-Live') for prefix in ifo_prefix_set):
+        trigger_files.append(source_file)
```
titodalcanton, line 195: At some point here you should also check that they are HDF5 files (extension `.hdf` or `.h5`) though this is still not guaranteed to give you only the files you are looking for.

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,333 @@
...
+
+logging.info(f" {len(trigger_files)} confirmed trigger files")
+
+if args.output_trigger_file_list:
+    logging.info(' Writing list of trigger files to: ' + args.output_trigger_file_list)
```
titodalcanton, line 200:
```suggestion
    logging.info('Writing list of trigger files to %s', args.output_trigger_file_list)
```

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,333 @@
...
+ifo_permutations = [''.join(p) for p in itertools.permutations(args.ifos)]
+ifo_prefix_set = set(ifo_permutations)
+
+trigger_files = []
+for source_file in initial_trigger_files:
+    trigger_file = os.path.basename(source_file)
+    if any(trigger_file.startswith(prefix + '-Live') for prefix in ifo_prefix_set):
+        trigger_files.append(source_file)
+
+logging.info(f" {len(trigger_files)} confirmed trigger files")
```
titodalcanton, line 197:
```suggestion
logging.info("%s confirmed trigger files", len(trigger_files))
```

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,333 @@
...
+logging.info(f" Creating a file at: {output_file}")
+with h5py.File(output_file, 'a') as destination:
+    # Create the ifo groups in the trigger file
+    for ifo in args.ifos:
+        if ifo not in destination:
+            destination.create_group(ifo)
+
+    file_count = 0
+    for source_file in trigger_files:
+        file_count += 1
```
titodalcanton, line 230: Use `enumerate()` here instead of having to count "by hand".

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,333 @@
...
+        end_time = start_time + duration
+
+        if file_count % 100 == 0:
+            logging.info(f" Files appended: {file_count}/{len(trigger_files)}")
+
+        with h5py.File(source_file, 'r') as source:
+            for ifo in args.ifos:
+                triggers = source[ifo]
+                if ('approximant' not in triggers) or (len(triggers['approximant']) == 0):
+                        continue
```
titodalcanton, line 244: `continue` is over-indented by one level.

bin/live/pycbc_live_collate_triggers
titodalcanton: The reason here is that if you end up with a PNG file (for example), the code will throw an exception when you try to load it with `h5py`.

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,333 @@
...
+    output_file = args.output_dir + args.output_file_name
+else:
+    # Extract the gpstimes for every file and take min/max for start and end
+    trigger_file_times = [float(trigger_file.split('/')[-1].split("-")[2])
+                          for trigger_file in trigger_files]
+    start_gpstime = min(trigger_file_times)
+    end_gpstime = max(trigger_file_times)
+
+    ifo_string = "".join(args.ifos)
+    output_file = args.output_dir + f'{ifo_string}-Live-' + start_gpstime + '-' + end_gpstime + '.hdf'
```
titodalcanton, line 219:
```suggestion
    output_file = f'{args.output_dir}/{ifo_string}-Live-{start_gpstime}-{end_gpstime}.hdf'
```
…but note that the file naming convention calls for the last integer value to be the *duration* of the data (rounded up to the next integer second), not the end time.

ArthurTolley:
@titodalcanton Gareth has added all the trigger finding stuff (and resolved a lot of the comments you had before).
Sorry for taking a while to get to those.
I think this is in a stage now where it could be merged.
We just need to profile and fully test this version.

GarethCabournDavies:
Profile:
image

time -v output:
```
	User time (seconds): 1245.77
	System time (seconds): 12.20
	Percent of CPU this job got: 98%
	Elapsed (wall clock) time (h:mm:ss or m:ss): 21:19.87
	Maximum resident set size (kbytes): 1026320
	Minor (reclaiming a frame) page faults: 502261
	Voluntary context switches: 106347
	Involuntary context switches: 2862
	File system inputs: 6165104
	File system outputs: 7937024
```

So approx 20 minutes, 1Gb memory usage, and it looks like most time is spent in I/O

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,280 @@
...
+logging.info("%s files found", len(trigger_files))
+
+if args.output_trigger_file_list:
+    logging.info(
+        'Writing list of trigger files to %s ',
+        args.output_trigger_file_list
+    )
+    with open(args.output_trigger_file_list, 'w') as f:
+        for item in trigger_files:
+            f.write("%s\n" % item)
```
GarethCabournDavies, line 77: @ArthurTolley is this still wanted or is it for testing only?

bin/live/pycbc_live_collate_triggers
```diff
@@ -0,0 +1,280 @@
...
+            ifo,
+            abs(segs[ifo])
+        )
+        if n_triggers[ifo] != n_triggers_cut[ifo]:
+            logging.info(
+                "  %d triggers after cuts",
+                n_triggers_cut[ifo],
+            )
+
+with h5py.File(args.bank_file,'r') as bank_file:
```
GarethCabournDavies, line 211: Should this be moved to the start, then you can avoid closing/reopening the output file?

bin/live/pycbc_live_collate_triggers
ArthurTolley: It was good for testing because the file system does spend a lot of time looking for the files themselves depending on how many directories were being searched but I don't think we need it anymore.

bin/live/pycbc_live_collate_triggers
ArthurTolley: Looks like it should

GarethCabournDavies:
poke @titodalcanton on approval/final review for this

titodalcanton (approved):
Looks good.

PR #4701, opened 2024-04-17
Release v2.1.6

maxtrevor:
This patch prepares the build of pycbc that will be released as v2.1.6


maxtrevor:
4 different checks have been 'in progress' for several hours now, I'm not sure whats going on with them.

GarethCabournDavies (approved):
> 4 different checks have been 'in progress' for several hours now, I'm not sure whats going on with them.

Look like its the docs checks all timing out at the same point during inference. Everything else looks fine though, so I will approve as I don't think these changes actually touch that code

PR #4712, opened 2024-04-23
Use approximant parsing argument for make_bayestar

GarethCabournDavies:
When generating bayestar skymaps in the upload prep minifollowup, the approximant used was not the same as in the rest of the workflow, this caused issues.

Here we use the standard approximant interface used by other codes in order to work out the appropriate waveform.

This is a: bug fix
This change changes: result presentation / plotting, scientific output

This change will: require a new release on the v23 release branch so that it can be used for o4 LVK results

## Motivation
The correct approximant was not being used for the bayestar subprocess, this means the skymaps would be wrong

## Contents
I have used standard library code for obtaining the approximant from the command line interface given conditions, rather than only allowing one  input.

None

## Testing performed
Previously failing event now runs, with the right approximant. Boundary also moved to ensure that the SPAtmplt --> TaylorF2 check was implemented

### DIFF
--- a/bin/all_sky_search/pycbc_make_bayestar_skymap
+++ b/bin/all_sky_search/pycbc_make_bayestar_skymap
@@ -23,8 +23,14 @@ import glob
 import sys
 import random
 import tempfile
+
+from ligo.lw import lsctables, utils as ligolw_utils
+
 import pycbc.version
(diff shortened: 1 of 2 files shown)


spxiwh (approved):
Looks good, assuming it works!

PR #4714, opened 2024-04-24
Smarter coinc SNR in pygrb post-processing

pannarale:
A more straightforward way of calculating the coincident SNR in post-processing (when coherent and null SNR are both known). 


titodalcanton (approved):
Looks correct to me and it simplifies the code.

PR #4717, opened 2024-04-25
Removed unused imports and variables; improved comments and messages …

pannarale:
Here is more low-level work on `pycbc_multi_inspiral`.
- Comments improved
- Messages to user improved (help for options and logging)
- Unused imports and variables removed

### DIFF
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -237,7 +237,7 @@ args = parser.parse_args()
 init_logging(args.verbose)
 # Set GRB time variable for convenience
 t_gps = args.trigger_time
-# Arange IFOs alphabetically so they are always called in the same order
+# Arrange detectors alphabetically so they are always called in the same order
 args.instruments.sort()
 nifo = len(args.instruments[:])
 # Use class verification methods to check whether input CLI options provided
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -571,7 +571,7 @@ with ctx:
                 for ifo in args.instruments
             }
             sigma = {ifo: np.sqrt(sigmasq[ifo]) for ifo in args.instruments}
-            # Every time s_num is zero , run new_template to increment the
+            # Every time s_num is zero, run new_template to increment the
             # template index
             if s_num == 0:
                 event_mgr.new_template(tmplt=template.params, sigmasq=sigmasq)
@@ -855,10 +855,9 @@ with ctx:
                             ifo: snr_dict[ifo][coinc_idx_det_frame[ifo]]
                             for ifo in args.instruments
                         }
-                        # Calculate the powerchi2 and autochi2 values of
-                        # remaining triggers (this uses the SNR timeseries
-                        # before the time delay, so we need to undo it; the
-                        # same holds for normalisation)
+                        # Calculate the powerchi2 values of remaining triggers
+                        # (this uses the SNR timeseries before the time delay,
+                        # so we undo it; the same holds for normalisation)
                         chisq = {}
                         chisq_dof = {}
                         for ifo in args.instruments:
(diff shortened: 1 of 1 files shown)


bin/pycbc_multi_inspiral
```diff
@@ -68,25 +66,29 @@ parser.add_argument(
...
-    help="Cut triggers with NewSNR less than THRESHOLD",
+    help="Single IFO SNR threshold for trigger generation."
 )
+# Commenting out unused options: remove if they remain unused
+# parser.add_argument(
+#     "--newsnr-threshold",
+#     type=float,
+#     metavar='THRESHOLD',
+#     help="Cut triggers with NewSNR less than THRESHOLD",
+# )
```
pannarale, line 80: I will remove this in 2 PRs from now.  I have one coming up for further cleanup to this executable and then one on arguments that will affect these lines.

titodalcanton: Fair enough.

bin/pycbc_multi_inspiral
```diff
@@ -228,82 +237,72 @@ args = parser.parse_args()
 init_logging(args.verbose)
 # Set GRB time variable for convenience
 t_gps = args.trigger_time
-# The following line aranges the ifos in alphabetical order so they are
-# always called in the same order.
+# Arange IFOs alphabetically so they are always called in the same order
```
titodalcanton, line 240:
```suggestion
# Arrange detectors alphabetically so they are always called in the same order
```

bin/pycbc_multi_inspiral
```diff
@@ -228,82 +237,72 @@ args = parser.parse_args()
...
     args, args.instruments, inj_filter_rejector, dyn_range_fac=DYN_RANGE_FAC
 )
-# The following stores a dictionary of python slice objects, that indicate
-# where the segments begin and end,for each detector's timeseries.
+# Create a dictionary of Python slice objects that indicate where the segments
+# start and end for each detector timeseries.
 strain_segments_dict = strain.StrainSegments.from_cli_multi_ifos(
     args, strain_dict, args.instruments
 )
+# Python context
```
titodalcanton, line 270:
```suggestion
# Context manager to handle the various possible processing schemes
```

bin/pycbc_multi_inspiral
```diff
@@ -382,11 +376,21 @@ with ctx:
...
-    # Clustering happens at the end of the template loop.
+    logging.info("Setting up MatchedFilterControl at each IFO")
+    # Prototype container for the output of MatchedFilterControl and
+    # waveform.FilterBank (see below).
+    # Use tlen of the first IFO as it is the same across IFOs.
+    template_mem = zeros(tlen, dtype=complex64)
+
+    # We are currently using the same matched-filter parameters for all
+    # IFOs. Therefore only one MatchedFilterControl setup is needed.
+    # This can be change if needed.
```
titodalcanton, line 387: I actually see one MatchedFilterControl for each detector, so is this comment still appropriate?

titodalcanton (approved):
Approved modulo a comment on a comment.

As a general note, despite the good use of comments, the algorithm is quite long and might benefit from being broken down into several functions.

bin/pycbc_multi_inspiral
pannarale: Let me propose a different text.  What I meant is that (`segments[ifo]` aside) we call `MatchedFilterControl` the exact same way on all available detectors, whereas we would want a single IFO SNR threshold just for the two most sensitive ones.  This would require a non-zero `args.snr_threshold` for 2 detectors and a vanishing one for all other the others.

pannarale:
```
    # We are currently using the same matched-filter parameters for all
    # IFOs. Therefore we loop over one MatchedFilterControl setup.
    # This can be change if needed.
```

titodalcanton: Ah I see, then maybe rephrase "only one MatchedFilterControl setup is needed" as "all MatchedFilterControl instances are initialized in the same way"?

pannarale:
```
    # We are currently using the same matched-filter parameters for all
    # IFOs. Therefore only all MatchedFilterControl instances are initialized
    #  in the same way. This can be change if needed.
```

PR #4719, opened 2024-04-26
Remove PyGRB injection set name check

jakeb245:
These changes affect PyGRB.

This PR removes a check on the injection set name that can cause empty results from `pycbc_grb_inj_combiner`. It is up to the user to provide valid inputs to the executable, and placing a filename constraint is unnecessary.

### DIFF
--- a/bin/pygrb/pycbc_grb_inj_combiner
+++ b/bin/pygrb/pycbc_grb_inj_combiner
@@ -258,7 +258,7 @@ usefiles = filter(
     args.input_files,
 )
 
-if not usefiles:
+if not list(usefiles):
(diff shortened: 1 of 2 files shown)


pannarale (approved):

pannarale:
Agreed. The idea of filename dependent instructions/checks breaks easily.
The user has the responsibility of passing actual injection results files to this executable and the script will keep injections with inclination < what the user has asked.
If the user passes on a file that is not an injection results file, things will crash anyways and naming the file nicely does not prevent that.

pannarale:
@jakeb245, if you pull the latest changes from gwastro/pycbc/master, the tests failing will work and we can merge.

PR #4720, opened 2024-04-26
Add explicit error message for samplers

HumphreyWang:
This is a minor change for adding a more informative error message to enhance user experience.

People (especially beginners) may not realize the core issue when seeing, for example, this error message:
```
  File "/home/dice/anaconda3/lib/python3.11/site-packages/pycbc/inference/sampler/__init__.py", line 104, in load_from_config
    return samplers[name].from_config(cp, model, **kwargs)
           ~~~~~~~~^^^^^^
KeyError: 'emcee'
```
New error message gives explicit guidance.

### DIFF
--- a/pycbc/inference/sampler/__init__.py
+++ b/pycbc/inference/sampler/__init__.py
@@ -14,7 +14,7 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 """
-This modules provides a list of implemented samplers for parameter estimation.
+This module provides a list of implemented samplers for parameter estimation.
 """
 
 import logging
@@ -87,7 +87,7 @@ def load_from_config(cp, model, **kwargs):
         Config parser to read from.
     model : pycbc.inference.model
         Which model to pass to the sampler.
-    \**kwargs :
+    **kwargs :
         All other keyword arguments are passed directly to the sampler's
         ``from_config`` file.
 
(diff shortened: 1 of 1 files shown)


pycbc/inference/sampler/__init__.py
```diff
@@ -101,4 +101,10 @@ def load_from_config(cp, model, **kwargs):
...
 
     name = cp.get('sampler', 'name')
-    return samplers[name].from_config(cp, model, **kwargs)
+    try:
+        return samplers[name].from_config(cp, model, **kwargs)
+    except KeyError:
+        raise ImportError(f"No accessible sampler named {name}. Please check"
+                          f" if the name is correct or the required package"
+                          f" for this sampler is installed correctly."
+                          f" Accessible samplers: {list(samplers.keys())}")
```
GarethCabournDavies, line 110: Thanks for looking at this!

Super minor suggestions:
 - I've removed the fstrings which don't use formatting
 - moved the spaces within the strings to the end of the lines rather than the start
 - list formatting in fstrings isn't that nice, so I've changed it to comma-separated
 - Minor wording change from 'accessible' to 'available'

```suggestion
        raise ImportError(
            f"No available sampler named {name}. Please check "
            "if the name is correct or the required package "
            "for this sampler is installed correctly. "
            f"Available samplers: {', '.join(list(samplers.keys()))}"
        )
```

PR #4725, opened 2024-04-29
set TDI version in LDC example

WuShichao:

This is a: bug fix

This change affects: Set TDI version for the upcoming BBHx plugin update.

This change changes: documentation,

### DIFF
--- a/examples/inference/lisa_smbhb_ldc/lisa_smbhb_relbin.ini
+++ b/examples/inference/lisa_smbhb_ldc/lisa_smbhb_relbin.ini
@@ -30,6 +30,8 @@ q =
 tc =
 
 [static_params]
+; LDC-Sangria uses TDI-1.5
+tdi = 1.5
 ref_frame = LISA
 approximant = BBHX_PhenomD
 coa_phase = 4.275929308696054


ahnitz:
@WuShichao Why is this a bugfix? Shoudn't there be a default value?

ahnitz (approved):

WuShichao:
> @WuShichao Why is this a bugfix? Shoudn't there be a default value?

Set a default value is not a good idea, because initially I set it to be '1.5'. If I set TDI to be '2.0' but sometimes I forget to set it in the PE config file, results will be wrong, I guess users will encounter the same problem, so I just force users to set it.

ahnitz:
@WuShichao I'm not sure this should be a required argument. On the plugin side, it probably should have a default and one can optionally change the value.

WuShichao:
> @WuShichao I'm not sure this should be a required argument. On the plugin side, it probably should have a default and one can optionally change the value.

Just set it by default on one side, but not on other side, might be inconsistent. Currently we don't have a check in PyCBC Inference to make sure the TDI versions are consistent everywhere, this will cause problem, and users will likely miss the point.

PR #4731, opened 2024-05-02
Clean up of pycbc_pygrb_page_tables

pannarale:
This PR cleans up `pycbc_pygrb_page_tables`:
- empty lines, whitespaces, indentation, redundant backslashes, etc.;
- output files are no longer all required, but the user needs to specify enough output file paths when providing a found-missed injections file and enough output file paths when dealing with loudest onsource/offsource; there are checks in place for all this and these have made other checks further down in the code redundant;
- the `do-injections` flagged is removed: just check whether the user provided a `found_missed_file`;

### DIFF
--- a/bin/pygrb/pycbc_pygrb_page_tables
+++ b/bin/pygrb/pycbc_pygrb_page_tables
@@ -259,18 +259,20 @@ qf_h5_outfile = opts.quiet_found_injs_h5_output_file
 
 # Set output files and directories
 output_files = []
-if onsource_file is not None:
-    output_files = [lont_outfile, lont_h5_outfile, lofft_outfile,
-                    lofft_h5_outfile]
-    if len(output_files) != 4:
-        parser.error('Please provide both on-source output files ' +
-                     'and both off-source output file when using ' +
-                     '--onsource-file.')
-elif found_missed_file is not None:
+# The user may want to process injections (also against onsource)...
+if found_missed_file is not None:
     output_files = [qf_outfile, mf_outfile, qf_h5_outfile]
-    if len(output_files) != 3:
+    if None in output_files:
         parser.error('Please provide all 3 injections output files when ' +
                      'using --found-missed-file')
+# ...or just triggers in offsource and onsource
+elif onsource_file is not None:
+    output_files = [lont_outfile, lont_h5_outfile, lofft_outfile,
+                    lofft_h5_outfile]
+    if None in output_files:
+        parser.error('Please provide both on-source output files ' +
+                     'and both off-source output files when using ' +
+                     '--onsource-file.')
 logging.info("Setting output directory.")
 for output_file in output_files:
     if output_file:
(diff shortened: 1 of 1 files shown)


pannarale:
@sebastiangomezlopez, @titodalcanton, let's see whether Jacob can review, approve and merge.

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -156,54 +160,56 @@ def load_missed_found_injections(hdf_file, ifos, snr_threshold, bank_file,
...
 # =============================================================================
 parser = ppu.pygrb_initialize_plot_parser(description=__doc__,
                                           version=__version__)
 parser.add_argument("-F", "--offsource-file", action="store", required=True,
                     default=None, help="Location of off-source trigger file")
-# As opposed to offsource-file and trig-file, this only contains onsource
 parser.add_argument("--onsource-file", action="store", default=None,
                     help="Location of on-source trigger file.")
-ppu.pygrb_add_injmc_opts(parser)
+parser.add_argument("--found-missed-file", action="store", default=None,
```
titodalcanton, line 188: I think there is no need to explicitly use `default=None` here and below. By default, there is no default :)

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -268,15 +249,28 @@ lower_dist = opts.lower_inj_dist
...
-# Set output directories
+# Set output files and directories
+output_files = []
+if onsource_file is not None:
+    output_files = [lont_outfile, lont_h5_outfile, lofft_outfile,
+                    lofft_h5_outfile]
+    if len(output_files) != 4:
+        parser.error('Please provide both on-source output files ' +
+                     'and both off-source output file when using ' +
+                     '--onsource-file.')
```
jakeb245, line 268: Wouldn't this always be true even if one of the output files is `None`?

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -268,15 +249,28 @@ lower_dist = opts.lower_inj_dist
...
+elif found_missed_file is not None:
+    output_files = [qf_outfile, mf_outfile, qf_h5_outfile]
+    if len(output_files) != 3:
+        parser.error('Please provide all 3 injections output files when ' +
+                     'using --found-missed-file')
```
jakeb245, line 273: Same thing here. It looks like very basic Python lists count `None` entries in the `len()` function, but this could behave in a way I'm not expecting.

bin/pygrb/pycbc_pygrb_page_tables
pannarale: Good point.  I check if `None` is in `output_files` now.
This also made me realise that the initial checks here make some checks further down redundant.  E.g., I'm forcing the user to specify certain output files if `--found-missed-file` is provided, so it is pointless to test for those files again later.

jakeb245 (approved):

PR #4732, opened 2024-05-02
V23_RELEASE_BRANCH Implement minimum statistic storage in sngls_findtrigs

GarethCabournDavies:
sngls_statmap is taking a lot of memory, part of that is the lack of real cuts being applied to sngls_findtrigs

There is no scientific benefit to keeping anything below the stat-threshold being used in pycbc_sngls_statmap, so this way we can cut things out that we won't use later as early as possible

This is an efficiency update
This change changes:  result presentation / plotting, scientific output (a little bit)

## Motivation
sngls_statmap has very high memory use - this is mainly due to sngls_findtrigs not getting rid of anything

## Contents
Implement a --minimum-stat threshold cut in sngls-findtrigs

## Testing performed
Code run, statmap generated, and then plotted with pycbc_page_snrifar, I will link to plots once the runs are completed!

## Additional notes
I would prefer decimation of the triggers below this cut, so that the plots in particular can go lower, but I am struggling to implement that. This can go on the v23 release branch as a fix there

### DIFF
--- a/bin/all_sky_search/pycbc_sngls_findtrigs
+++ b/bin/all_sky_search/pycbc_sngls_findtrigs
@@ -43,6 +43,8 @@ parser.add_argument('--cluster-window', type=float,
                     help='Window (seconds) during which to keep the trigger '
                          'with the loudest statistic value. '
                          'Default=do not cluster')
+parser.add_argument("--minimum-stat", type=float,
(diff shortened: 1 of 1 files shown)


spxiwh (approved):
This looks good, and follows how the clustering would work. Assuming it's tested, it's good with me.

PR #4735, opened 2024-05-03
DRAFT: Supervise fit coeff dq files

GarethCabournDavies:
This is a draft so that people can see ongoping work with regard to the pycbc live collated triggers, FIT_BY_TEMPLATE, FIT_OVER_MULTIPARAM and DQ calculation codes


GarethCabournDavies:
@maxtrevor @titodalcanton and @ArthurTolley are interested parties

PR #4736, opened 2024-05-03
Cleanup sky stuff in multi_inspiral

titodalcanton:
Cleanup of the sky grid and antenna pattern calculations in `pycbc_multi_inspiral`.

This is a code cleanup and minor bugfix.

This change affects PyGRB.

No changes to results or scientific output are expected.

No functionality or API should be broken by this.

## Motivation

`pycbc_multi_inspiral` implements a very long algorithm right now, including many branches and multiple long loops. Looking through it, I noticed that the code handling the sky grid CLI and calculating the antenna pattern is more complicated than necessary, uses more variables than needed, and pollutes the namespace with several temporary variables. There is also a check for simultaneous usage of `--sky-grid`, `--ra` and `--dec` which is not entirely correct, for example it will miss the cases where `--sky-grid` and `--ra` are both given, or none are given at all.

## Contents

I created two small helper functions that handle the CLI arguments related to the sky grid (including a proper check of them) and the precalculation of the antenna pattern functions.

None.

## Testing performed

None, apart from the automatic CI tests.

## Additional notes

I think the present calculation/handling of the antenna pattern and time delays is not particularly efficient, as it uses dicts and lists, and is repeated identically by every `pycbc_multi_inspiral` job. I wonder how much total time this entails in a PyGRB workflow with a large sky grid. It may be useful to evaluate this, and switch to precalculating the antenna pattern and time delays just once, maybe in `pycbc_make_skygrid`, and using multidimensional Numpy arrays for carrying around the precalculated values.

### DIFF
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -80,7 +80,7 @@ def calculate_antenna_pattern(args, sky_pos_indices):
     """
     antenna_pattern = {}
     for ifo in args.instruments:
-        curr_det = Detector(ifo)
+        curr_det = detector.Detector(ifo)
         antenna_pattern[ifo] = [None] * len(sky_pos_indices)
         for index in sky_pos_indices:
             antenna_pattern[ifo][index] = curr_det.antenna_pattern(
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -58,7 +58,10 @@ def sky_grid_from_cli(parser, args):
         args.sky_grid is not None
         and (args.ra is not None or args.dec is not None)
     ):
-        parser.error('Give either a sky grid or a sky position, not both')
+        parser.error(
+            'Please provide either a sky grid via --sky-grid or a '
+            'single sky position via --ra and --dec, not both'
+        )
     if args.sky_grid is not None:
         with h5py.File(args.sky_grid, 'r') as sky_grid_file:
             ra = sky_grid_file['ra'][:]
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -51,6 +51,47 @@ from pycbc.types import zeros, float32, complex64
 from pycbc.vetoes import sgchisq
 
 
+def sky_grid_from_cli(parser, args):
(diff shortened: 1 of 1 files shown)


bin/pycbc_multi_inspiral
```diff
@@ -51,6 +51,47 @@ from pycbc.types import zeros, float32, complex64
...
+    return np.array([ra, dec])
+
+
+def calculate_antenna_pattern(args, sky_pos_indices):
+    """Precalculate the antenna pattern functions for all detectors and sky
+    positions.
+    """
+    antenna_pattern = {}
+    for ifo in args.instruments:
+        curr_det = Detector(ifo)
```
sebastiangomezlopez, line 83: Tests are complaining about this typo, we should switch  `Detector(ifo)` -> `detector(ifo)`

sebastiangomezlopez: I noticed the `detector.Detector` [class](https://github.com/gwastro/pycbc/blob/0d5f46c6bbbcaeb1dabab442c78640dcde0b6bba/pycbc/detector.py#L194), has this arg "reference_time" that defaults to a GW150914 time. Do we want this ?  otherwise we should use `detector.Detector(ifo, reference_time=None)`

titodalcanton: I am not sure that will have a big effect, but setting it to None or to the GRB time might be more correct in principle, and might avoid some numerical issues with the antenna pattern or time delay calculations.

I am not going to change this here, but if you want to do a comparison, go ahead.

bin/pycbc_multi_inspiral
```diff
@@ -51,6 +51,47 @@ from pycbc.types import zeros, float32, complex64
...
+    else:
+        parser.error(
+            'Please specify a sky grid via --sky-grid '
+            'or a single position via --ra and --dec'
+        )
+    return np.array([ra, dec])
+
+
+def calculate_antenna_pattern(args, sky_pos_indices):
+    """Precalculate the antenna pattern functions for all detectors and sky
```
pannarale, line 78:
```suggestion
    """Calculate the antenna pattern functions for all detectors and sky
```

pannarale (approved):

PR #4737, opened 2024-05-03
Options and cuts in pycbc_multi_inspiral

pannarale:
- Passing options from `pycbc_multi_inspiral` command line to the methods that calculate null SNR, cut in null SNR, and calculate reweighted SNR.
- Renaming `n_ifo` as `ifo_idx` to avoid confusion with the already existing `nifo`.
- Grouping together the definitions of auxiliary variables.

### DIFF
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -433,8 +433,7 @@ with ctx:
     # Directly use existing SingleDetPowerChisq, SingleDetBankVeto, and
     # SingleDetAutoChisq to calculate single detector chi-squares for
     # multiple IFOs
-    power_chisq = vetoes.SingleDetPowerChisq(args.chisq_bins,
-                                             args.chisq_snr_threshold)
+    power_chisq = vetoes.SingleDetPowerChisq(args.chisq_bins)
     bank_chisq = vetoes.SingleDetBankVeto(
         args.bank_veto_bank_file,
         flen,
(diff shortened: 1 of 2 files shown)


bin/pycbc_multi_inspiral
```diff
@@ -615,6 +640,12 @@ with ctx:
                 snr_dict[ifo] = (
                     snr_ts[matched_filter[ifo].segments[s_num].analyze] * norm
                 )
+                # Raise error if this segment has no data
+                if len(snr_dict[ifo]) == 0:
+                    raise RuntimeError(
+                        'The SNR triggers dictionary is empty. This '
+                        'should not be possible.'
+                    )
```
titodalcanton, line 648: We discussed this a bit on Slack, and I am still not quite understanding this check. `snr_dict[ifo]` seems to be a (portion of) SNR time series. So first of all, the error message is not correct, as we are not looking at triggers here. Second, since the error condition should never occur in *any* kind of normal operation, this check is better expressed using `assert` rather than an if statement and `raise`. So I suggest
```suggestion
                assert len(snr_dict[ifo]) > 0, f'SNR time series for {ifo} is empty'
```

bin/pycbc_multi_inspiral
```diff
@@ -901,15 +928,20 @@ with ctx:
                         # Calculate chisq reweighted SNR
                         if nifo > 2:
                             reweighted_snr = ranking.newsnr(
-                                rho_coh, network_chisq_dict
+                                rho_coh, network_chisq_dict,
+                                q=args.chisq_index, n=args.chisq_nhigh
```
titodalcanton, line 932: I suggest doing another pass through `black` to properly format these edited calls.

pannarale: The latest commit applied another round of `black`.

PR #4739, opened 2024-05-06
Page tables bug fix

MarcoCusinato:
Sorted ifos event_ids by the network one, allowing to display the correct result.

### DIFF
--- a/bin/pygrb/pycbc_pygrb_page_tables
+++ b/bin/pygrb/pycbc_pygrb_page_tables
@@ -130,8 +130,11 @@ def load_missed_found_injections(hdf_file, ifos, snr_threshold, bank_file,
     # parameters. This is used to marginalize over calibration errors.
     # also save the snr per ifo
     for ifo in ifos:
-        if not np.all(inj_data['network/event_id'][...] == 
+        if np.all(inj_data['network/event_id'][...] == 
                       inj_data['%s/event_id' % ifo][...]):
+            found_data['sigmasq_%s' % ifo] = inj_data['%s/sigmasq' % ifo][...]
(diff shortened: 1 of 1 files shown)


bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -136,8 +130,11 @@ def load_missed_found_injections(hdf_file, ifos, snr_threshold, bank_file,
     # parameters. This is used to marginalize over calibration errors.
     # also save the snr per ifo
     for ifo in ifos:
-        found_data['sigmasq_%s' % ifo] = inj_data['%s/sigmasq' % ifo][...]
-        found_data['snr_%s' % ifo] = inj_data['%s/snr' % ifo][...]
+        # Sort the ifo event_id with respect to the network event_id
+        ifo_sorted_indices = np.argsort(np.array(inj_data['network/event_id'])[np.argsort(
+            inj_data['network/event_id'])].searchsorted(inj_data['%s/event_id' % ifo][...]))
+        found_data['sigmasq_%s' % ifo] = np.array(inj_data['%s/sigmasq' % ifo])[ifo_sorted_indices]
+        found_data['snr_%s' % ifo] = np.array(inj_data['%s/snr' % ifo])[ifo_sorted_indices]
```
pannarale, line 137: This should not be necessary, in the sense that `(inj_data['network/event_id'][]==inj_data['%s/event_id' % ifo]).all()` is true for all IFOs.  I suggest keeping the 2 original lines and raising an error if the condition above does not hold.  Should we ever hit that error, we will need to fix it in the `inj_combiner` and `inj_finder`.

pannarale (changes requested):

pannarale (approved):

PR #4744, opened 2024-05-13
add str() to all tdi

WuShichao:
Force `tdi` to be `str` just like other places in the code. Otherwise, PyCBC Inference will read it as a number. that will cause error.

This is a: bug fix

## Motivation

Fix issues encountered in PE.

## Contents

Adding `strt()` to all `tdi` input.


HumphreyWang (approved):

WuShichao:
Thanks~

PR #4756, opened 2024-05-20
Add PyGRB Exclusion Distance Table Executable

jakeb245:

This PR adds an executable to create a static table for PyGRB exclusion distances and updates `pycbc_pygrb_efficiency` to output the necessary data to a JSON file.

This is a: new feature

This change affects: PyGRB

## Motivation

PyGRB result webpages currently write the exclusion distances in the plot captions. A table containing this information would look nicer.

## Contents

- Updates `pycbc_pygrb_efficiency` to write a JSON file with the exclusion distances
- Adds `pycbc_pygrb_exclusion_distance_table` which takes the JSON output files and outputs a result table.

## Testing performed

An example of the resulting table can be seen [here](https://ldas-jobs.ligo.caltech.edu/~jacob.buchanan/test_table.html). This is a quick example generated on the command line. For a final result, the table will be added to the results webpage within the workflow generator (to be done in a later PR).

## Additional notes


pannarale (changes requested):

bin/pygrb/pycbc_pygrb_exclusion_dist_table
```diff
@@ -0,0 +1,109 @@
...
+    with open(file_name, "r") as file:
+        file_contents.append(json.load(file))
+
+# Get list of trials (i.e. OFFTRIAL_i, ONSOURCE, etc.)
+trials = []
+for fc in file_contents:
+    trials.append(fc["trial_name"])
+# Check that at least one trial exists
+if len(trials) == 0:
+    raise ValueError("No trials found in input files.")
```
pannarale, line 57: Should we make `--trial-name` required in `pycbc_pygrb_efficiency` to prevent this?  This way we have a check in place here for the standalone script, and a way to dodge the error in the workflow.

pannarale: Now that I looked better, `pycbc_pygrb_efficiency` must check that `opts.trial_name` is not None when `exclusion_dist_output_file` is not None.

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -100,6 +101,8 @@ parser.add_argument("--background-output-file", default=None, required=True,
                     help="Detection efficiency output file.")
 parser.add_argument("--onsource-output-file", default=None, required=True,
                     help="Exclusion distance output file.")
+parser.add_argument("--exclusion-dist-output-file", default=None,
+                    help="JSON file containing exclusion distances.")
```
pannarale, line 105: All `default=None` can be removed.

bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -606,7 +612,17 @@ if do_injections:
             excl_dist = 0
             msg = "Efficiency below %d%% in first bin!" % (percentile)
             logging.warning(msg)
-        # TODO: include percentile, excl_dist on output pages
+
+    # Write 50% and 90% exclusion distances to JSON file
+    # Also include injection set name and trial name
+    if opts.exclusion_dist_output_file:
+        excl_dist_dict = {}
+        excl_dist_dict['inj_set'] = inj_set_name
```
pannarale, line 620: The user must provide `opts.injection_set_name` if `opts.exclusion_dist_output_file` and/or `opts.found_missed_file` are provided.
Looking at this made me wonder about something I had not caught in the past: does it even make sense to have `opts.found_missed_file` and `opts.found_missed_file` as optional?  I.e., if `do_injections` and/or `onsource_file` are False, does the code do something meaningful (and not crash)?  Otherwise we can promote those options to required.

pannarale (changes requested):

bin/pygrb/pycbc_pygrb_efficiency
jakeb245: The injections part of the code uses results from the onsource analysis, so I'd say it wouldn't hurt anything to make both files required, but there could be cases I'm not considering.

pannarale (approved):

PR #4762, opened 2024-05-29
fix cdfinv of external text file distribution when multiple are  used in a joint distribution

ahnitz:
@WuShichao There is a bug in the external from fun distribution in how it retrieves the parameter name. 

At the same time, the config file format was overly complicated. I've simplified this. so that one can do 

[prior-final_mass]
name = external_func_fromfile
file_path = spin.txt
column_index = 1

The change is backwards compatible. 

@WuShichao Can you please verify and review? Also add it to your todo list to put in a unittest for this class.

### DIFF
--- a/pycbc/distributions/external.py
+++ b/pycbc/distributions/external.py
@@ -119,7 +119,7 @@ class External(object):
 
 
 class DistributionFunctionFromFile(External):
-    """Evaluating PDF, logPDF, CDF and inverse CDF from the external
+    r"""Evaluating PDF, logPDF, CDF and inverse CDF from the external
         density function.
 
     To add to an inference configuration file:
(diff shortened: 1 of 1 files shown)


WuShichao:
@ahnitz Thanks for this fix. I have tried my previous 3G population mock data code, and it works fine:
image
Please rerun the LISA SMBHB LDC test in CI and fix the CC issue.

WuShichao (approved):
Looks good!

WuShichao:
@ahnitz After @koustavchandra tries it successfully, I will merge it.

WuShichao:
This PR fixes this issue https://github.com/gwastro/pycbc/issues/4312

koustavchandra:
Yes, it fixes it. Thanks!

PR #4766, opened 2024-05-30
Fix epoch in cached fft

mj-will:
Update `pycbc.strain.strain.execute_cached_fft` so that the epoch of the output frequency/time series is set to match that of the input rather than being zero.

This is a: bug fix

This change affects any code that makes use of `execute_cached_fft` or functions/classes that call it. This includes `pycbc.filter` and `pycbc.psd` given this, it could impact a large part of the codebase.

This change will: fix an existing bug

## Motivation

This fixes a bug in `pycbc.strain.strain.execute_cached_fft`

## Contents

* Specify the `_epoch` attribute of the output frequency/time series based on the input
* Add a basic unit test for `execute_cached_fft` and `execute_cached_ifft`

N/A

## Testing performed

Added unit tests to test the changes

## Additional notes

## Minimal example

```python
import pycbc
import pycbc.strain
ts = pycbc.types.TimeSeries([1, 2, 3], delta_t=0.2, epoch=1e9)
fs = pycbc.strain.strain.execute_cached_fft(ts)
print(ts._epoch, fs._epoch)
```
which prints:
```
1000000000 0
```

### DIFF
--- a/pycbc/strain/strain.py
+++ b/pycbc/strain/strain.py
@@ -1383,7 +1383,10 @@ def execute_cached_fft(invec_data, normalize_by_rate=True, ifft=False,
             outvec._data *= invec._delta_t
     if copy_output:
         outvec = outvec.copy()
-    outvec._epoch = invec_data._epoch
+    try:
+        outvec._epoch = invec_data._epoch
+    except AttributeError:
+        pass
     return outvec
 
 
--- a/pycbc/strain/strain.py
+++ b/pycbc/strain/strain.py
@@ -1383,6 +1383,7 @@ def execute_cached_fft(invec_data, normalize_by_rate=True, ifft=False,
             outvec._data *= invec._delta_t
     if copy_output:
         outvec = outvec.copy()
+    outvec._epoch = invec_data._epoch
     return outvec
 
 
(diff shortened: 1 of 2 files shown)


mj-will:
@spxiwh I opted to make the change in a `try/except` block so that the function still works with `Array` inputs. I'm happy to change this if you have other suggestions

titodalcanton:
Could you elaborate on what bug this fixes? Which scientific output does the bug affect (searches, PE…)?

mj-will:
> Could you elaborate on what bug this fixes? Which scientific output does the bug affect (searches, PE…)?

@titodalcanton scientific output might be a bit of a stretch but it was the best fit of the options listed.

The bug this fixes occurs if you call `executed_cached_fft` on an instance of `TimeSeries` (or on a `FrequencSeries` if using the inverse option). The output object does not have the `_epoch` attribute set, so it defaults to zero. I've updated the PR text with a minimal example that shows this.

This cropped up in some LISA-related short author work Ian, Gareth and I are working.  We're using this function for performing faster FFTs/iFFTs with large time series/frequency series for both search and inference. The bug described here was leading to inconsistent epochs between the data and generated waveforms, which in turn led to incorrect results in the inference.

spxiwh (approved):
Thanks Michael!

PR #4768, opened 2024-05-31
Hfile pycbc version information check

GarethCabournDavies:
Add a `pycbc_version` attribute when opening a HFile to write, and check that it exists/matches when opening for reading.

This is a new feature
This change (can) affect all codes which use h5py input/output, if wanted
This changes warnings seen when running the code, no scientific content or results plotting will be affected

I can add unit tests if desired

This change will warn when different versions of PyCBC are being used

## Motivation
When the code changes, there are (sometimes unwitting) differences in how the files from the code are output/input. This just adds a warning to the user to be wary of that.

## Contents
When a HFile object is opened, the file read/write permission will be used to determine whether or not to check / update the `pycbc_version` attribute.

 - If this version does not match the one being used, a warning is shown.
 - If using append permission, the version is changed, and a warning is shown

## Testing performed
Tested each case (same / different versions, and unversioned files), and the correct warnings are given each time. 
Tested that the permissions are carried through to the underlying h5py.File, and appropriate errors are raised.

### DIFF
--- a/bin/all_sky_search/pycbc_add_statmap
+++ b/bin/all_sky_search/pycbc_add_statmap
@@ -3,10 +3,11 @@
 with more than one ifo combination available. Cluster to keep foreground
 coincs with the highest stat value.
 """
-import h5py, numpy as np, argparse, logging, pycbc, pycbc.events, pycbc.io
+import numpy as np, argparse, logging, pycbc, pycbc.events, pycbc.io
 import pycbc.version
 import pycbc.conversions as conv
 from pycbc.events import coinc, significance
+from pycbc.io import HFile
 from ligo import segments
 import sys, copy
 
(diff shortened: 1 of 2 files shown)


GarethCabournDavies:
Note that the changes to `add_statmap` are so that this is checked in the CI - codes should change to use HFile rather than h5py.File gradually, but I won't change them all in one go

ahnitz:
Is there any more information on the motivation for this PR? I'm not sure how I feel about linking version of the *pycbc library* information into a file like this.

GarethCabournDavies:
@ahnitz sorry, I should have added you as a reviewer to get your opinion as well, as it is a proposal at the moment

Basically as different pycbc versions may not be able to work together, I thought it best to warn about this rather than anyone expecting the codes to work across different versions.

This would additionally help with code review as the reviewer could check that the pycbc version in the result files matches the one which _should_ have been run

GarethCabournDavies:
Note that the general option of pycbc always using HFile as a wrapper would go someway towards solving #1525 and #2058

PR #4774, opened 2024-06-06
pycbc_test_suite to be able to run from anywhere within the git repo

GarethCabournDavies:
Add a base directory location - the base directory of the repo - to the pycbc_test_suite bash script, so that it can be run from any directory within the repo.

## Motivation
Solve #2639


tools/pycbc_test_suite.sh
```diff
@@ -7,14 +7,18 @@ echo -e "\\n>> [`date`] Python Major Version:" $PYTHON_VERSION
 PYTHON_MINOR_VERSION=`python -c 'import sys; print(sys.version_info.minor)'`
 echo -e "\\n>> [`date`] Python Minor Version:" $PYTHON_MINOR_VERSION
 
+# This will work from anywhere within the pycbc directory
+PYCBC_BASE_DIR=`git rev-parse --show-toplevel`
```
ahnitz, line 11: I think there are problems if we assume git access in certain circumstances (e.g. it will make it hard to integrate into conda's testing if we want to improve that). Can we get this without that? 

I think one can do this because the script can know what directly it exists in, no?

ahnitz (changes requested):

tools/pycbc_test_suite.sh
GarethCabournDavies: I think run from any folder - the git problem raised makes sense, so I will look at that

GarethCabournDavies: I think this will work now

GarethCabournDavies:
poke on the re-review here

ahnitz (approved):

ahnitz:
@GarethCabournDavies You'll just have to rebase to pick up the new required tests.

Issue #2885, opened 2019-08-02
replace lal.YRJUL_SI with conv sec_to_year


tdent:
Some places in the code still import lal explicitly while others have been wrapped via pycbc.conversions. Want to clean up

PR #4775, opened 2024-06-06
Use pycbc conversions as single source of truth for converting seconds and years

GarethCabournDavies:
Use the pycbc conversions sec_to_year function throughout so that there is a single source of truth for the conversion of seconds and years, and we dont call into lal for YRJUL_SI so much

## Motivation
Solve #2885

## Contents
Whenever the previous constant was being used, use the conversion function instead

### DIFF
--- a/bin/plotting/pycbc_page_snrratehist
+++ b/bin/plotting/pycbc_page_snrratehist
@@ -145,31 +145,19 @@ histpeak = bins[exc_binvals[0].argmax()]
 
 # plot full background
 if not args.closed_box:
-    if h_inc_back_num == 0:
-        binweights = dec / conv.sec_to_year(f.attrs['background_time'])
-        pylab.hist(
-            bstat,
-            bins=bins,
-            histtype='step',
-            linewidth=2,
-            color='black',
-            log=True,
-            label='Full Background',
-            weights=binweights
-        )
-    else :
-        bg_time_sec = f.attrs['background_time_h%s' % h_inc_back_num]
-        binweights = dec / conv.sec_to_year(bg_time_sec)
-        pylab.hist(
-            bstat,
-            bins=bins,
-            histtype='step',
-            linewidth=2,
-            color='black',
-            log=True,
-            label='Full Background',
-            weights=binweights
-        )
+    bg_key = 'background_time' if h_inc_back_num == 0 \
+        else 'background_time_h%s' % h_inc_back_num
+    binweights = dec / conv.sec_to_year(f.attrs[bg_key])
+    pylab.hist(
+        bstat,
+        bins=bins,
+        histtype='step',
+        linewidth=2,
+        color='black',
+        log=True,
+        label='Full Background',
+        weights=binweights
+    )
 
 if fstat is not None and not args.closed_box:
     le, re = bins[:-1], bins[1:]
--- a/bin/plotting/pycbc_page_snrratehist
+++ b/bin/plotting/pycbc_page_snrratehist
@@ -3,10 +3,11 @@
     Also has the ability to plot inclusive backgrounds from different stages
     of hierarchical removal.
 """
-import argparse, h5py, numpy, logging, sys, lal
+import argparse, h5py, numpy, logging, sys
 import matplotlib
 matplotlib.use('Agg')
 import pylab, pycbc.results, pycbc.version
+from pycbc import conversions as conv
 from scipy.special import erf, erfinv
 
 def sigma_from_p(p):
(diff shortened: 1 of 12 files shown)


tdent:
Can you make this a draft until you think it's ready ?

GarethCabournDavies:
I think it is ready now, but have just had a couple of bugs

bin/plotting/pycbc_page_snrratehist
```diff
@@ -145,17 +146,27 @@ histpeak = bins[exc_binvals[0].argmax()]
...
-                   weights=dec / f.attrs['background_time'] * lal.YRJUL_SI)
+        pylab.hist(
+            bstat,
+            bins=bins,
+            histtype='step',
+            linewidth=2,
+            color='black',
+            log=True,
+            label='Full Background',
+            weights=dec / conv.sec_to_year(f.attrs['background_time'])
```
tdent, line 157: this looks correct, but it is a bit messy, as in fact it was before .. ideally one would define eg `w = dec / conv.sec_to_year(f.attrs['background_time'])` and then just `weights=w` in the hist call.

bin/plotting/pycbc_page_snrratehist
```diff
@@ -145,17 +146,27 @@ histpeak = bins[exc_binvals[0].argmax()]
...
-                   weights=dec / f.attrs['background_time_h%s' % h_inc_back_num] * lal.YRJUL_SI)
+        pylab.hist(
+            bstat,
+            bins=bins,
+            histtype='step',
+            linewidth=2,
+            color='black',
+            log=True,
+            label='Full Background',
+            weights=dec / conv.sec_to_year(f.attrs['background_time_h%s' % h_inc_back_num])
```
tdent, line 168: Same comment, preferable to calculate the weight array beforehand outside the call to `hist`

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -158,7 +159,7 @@ for event in coinc_table:
             coinc_inspiral_table_curr.append(coinc_insp)
 
     if args.min_ifar is not None and \
-            coinc_inspiral_table_curr[0].combined_far > 1./args.min_ifar/lal.YRJUL_SI:
+            coinc_inspiral_table_curr[0].combined_far > conv.sec_to_year(1. / args.min_ifar):
```
tdent, line 162: This is a bit tricky .. you are using the sec_to_year function on a _FAR_ (1/time) number, which makes it a factor pi*1e7 smaller, but it is rather difficult to parse/check as the function is written to convert times. 
Easier to deal with this in the same way as the previous ifar conversion :
```suggestion
            conv.sec_to_year(1. / coinc_inspiral_table_curr[0].combined_far) < args.min_ifar:  # event IFAR is smaller than the minimum
```

pycbc/io/live.py
```diff
@@ -202,7 +203,7 @@ def __init__(self, coinc_ifos, ifos, coinc_results, **kwargs):
         coinc_inspiral_row.end_time = sngl_populated.end_time
         coinc_inspiral_row.end_time_ns = sngl_populated.end_time_ns
         coinc_inspiral_row.snr = network_snrsq ** 0.5
-        far = 1.0 / (lal.YRJUL_SI * coinc_results['foreground/ifar'])
+        far = conv.sec_to_year(1.0 / coinc_results['foreground/ifar'])
```
tdent, line 206: we are usually writing '1.' for some reason ..
```suggestion
        far = conv.sec_to_year(1. / coinc_results['foreground/ifar'])
```

tdent (approved):
See comments, otherwise OK assuming all tests pass ...

bin/plotting/pycbc_page_snrratehist
```diff
@@ -165,7 +168,7 @@ if not args.closed_box:
             color='black',
             log=True,
             label='Full Background',
-            weights=dec / conv.sec_to_year(f.attrs['background_time_h%s' % h_inc_back_num])
+            weights=binweights
         )
```
tdent, line 172: Huh.  This if-else switch seems to be redundant as the same hist call is made in both cases. 
Can the background attribute be defined in one-liner - eg `bg_key = 'background_time' if h_inc_back_num == 0 else 'background_time_h%s' % h_inc_back_num`
then the switch can be completely removed ?

GarethCabournDavies: Added

tdent (approved):

Issue #2126, opened 2018-04-11
test_lalsim.py swap test only works for 2,2 mode


sfield17:
In the file test/test_lalsim.py, the test test_swapping_constituents should add $\pi$ to the variable coa_phase when swapping. This will allow the test to work for models that include harmonic modes different from $(2,\pm 2)$. 

```python
        hpswap, hcswap = get_waveform(self.p, mass1=self.p.mass2, mass2=self.p.mass1,
                spin1x=self.p.spin2x, spin1y=self.p.spin2y, spin1z=self.p.spin2z,
                spin2x=self.p.spin1x, spin2y=self.p.spin1y, spin2z=self.p.spin1z,
                lambda1=self.p.lambda2, lambda2=self.p.lambda1,
                coa_phase=self.p.coa_phase + lal.PI)  # fix in this line
```

PR #4777, opened 2024-06-07
Use π phase shift to allow swap test for more modes

GarethCabournDavies:
I'm not sure on whether this is actually tested for higher-order modes, but the sanity checks of `test/lalsim.py` should use this π phase shift in order to work for anything except 2,±2

This change affects the test suite

This change, follows style guidelines (See e.g. [PEP8](https://peps.python.org/pep-0008/)), has been proposed using the [contribution guidelines](https://github.com/gwastro/pycbc/blob/master/CONTRIBUTING.md)

## Motivation
#2126 

## Contents
Add a phase shift for the component swap test

### DIFF
--- a/test/lalsim.py
+++ b/test/lalsim.py
@@ -324,10 +324,20 @@ class TestLALSimulation(unittest.TestCase):
         #"""
 
         hp, hc = get_waveform(self.p)
-        hpswap, hcswap = get_waveform(self.p, mass1=self.p.mass2, mass2=self.p.mass1,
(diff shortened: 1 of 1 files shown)


titodalcanton:
Have you verified that more approximants pass this test after this change, especially ones with HoMs?

GarethCabournDavies:
> Have you verified that more approximants pass this test after this change, especially ones with HoMs?

I have not, but HoMs is not really my area, I just implemented the suggestion in the issue :wink:

GarethCabournDavies:
I ran the unit tests for lalsim and this makes no difference so I will close

Issue #2504, opened 2019-02-13
We should publish documentation for older release versions in addition to latest


cdcapano:
Right now you can only access documentation online for the current master version of pycbc. This can cause issues if changes occurred to master since the latest release. For example, as described in Issue #2502, the version of `pycbc_inference` on master gets the low frequency cutoff to use for the likelihood from the configuration file. The example configuration files (which can be downloaded from the documentation page) reflect this. However, in the current release version, the low frequency cutoff is set on the command line. This means that if a user installs pycbc using pip, then tries to run pycbc inference using the examples documented online, they'll get an error.

If docs pages existed for each version, then the user would know (or at least could figure out) that they should look at the version of the docs corresponding to the version of pycbc that they have. This would be similar to how numpy/scipy work.

PR #4778, opened 2024-06-07
Draft: Start pushing tags to permanently-kept documentation pages

GarethCabournDavies:
Here I am starting to put together an idea on how to provide documentation for every release. I.e. instead of only having https://pycbc.org/pycbc/latest/html/, we would have https://pycbc.org/pycbc/v2.4.X/html/ as well, which points to the documentation at that release

This change affects continuous integration and the documentation

This change will require a new release to ensure that it is correctly enacted

## Motivation
This is best practice anyway, but there is an issue at #2504

## Contents
 - Add a new job to the CI which is only run when a release is made.
 - This will also add the documentation to a new folder within the gh-pages repo
 - The gh-pages repo should then have the release kept whenever the documentation is built again, and will be pushed for each version

Plan for next steps
 - Get a dropdown of links for each version (plus latest) added to the bottom of the sidebar on the documentation pages
 - Add a link to the latest released version of the documentation to pypi

## Testing performed
I'm not sure how to test this without making a release!

### DIFF
--- a/.github/workflows/basic-tests.yml
+++ b/.github/workflows/basic-tests.yml
@@ -75,3 +75,23 @@ jobs:
         BRANCH: gh-pages
         FOLDER: _gh-pages
         SINGLE_COMMIT: true
+  deploy_documentation_release:
+    runs-on: ubuntu-latest
+    needs: build
+    if: github.ref == 'refs/head/master' && github.event_name == 'release'
+    steps:
+    - name: retrieve built documentation
+      uses: actions/download-artifact@v2
+      with:
+        name: documentation-page
+    - name: debug
+      run: |
+        mkdir _gh-pages
+        mv latest _gh-pages/${GITHUB_REF#refs/tags/}
+    - name: deploying tag ${GITHUB_REF#refs/tags/} to gh-pages
+      uses: JamesIves/github-pages-deploy-action@3.7.1
+      with:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        BRANCH: gh-pages
+        FOLDER: _gh-pages
+        SINGLE_COMMIT: true


ahnitz:
@GarethCabournDavies I think this is a long overdue good thing to do. Thank you for looking into this!

GarethCabournDavies:
I think the only way this can be tested is it it is just merged and see if it runs for the next release

.github/workflows/basic-tests.yml
```diff
@@ -75,3 +75,23 @@ jobs:
         BRANCH: gh-pages
         FOLDER: _gh-pages
         SINGLE_COMMIT: true
+  deploy_documentation_release:
```
ahnitz, line 78: @GarethCabournDavies Doesn't this have code duplication here?

ahnitz (changes requested):

.github/workflows/basic-tests.yml
ahnitz: To further explain, it looks like the documentation release is now duplicated rather than simply running on both master and the tag version.

GarethCabournDavies: yeah, I couldnt work out how to move to `latest _gh-pages/${GITHUB_REF#refs/tags/}` or `current` (or whatever) dynamically

ahnitz: @GarethCabournDavies I don't really follow. Wouldn't you just have an if statement in the yaml in the case of a release and then essentially do what you do here, except without all the duplicated bits?

GarethCabournDavies: Ah, I think I didn't realise you can effectively use shell if statements in the yaml, I will try deduplicating

GarethCabournDavies:
I forgot that I hadn't un-drafted and re-requested review for this!

.github/workflows/basic-tests.yml
```diff
@@ -58,20 +58,21 @@ jobs:
...
-    if: github.ref == 'refs/heads/master' && github.event_name == 'push'
+    if: github.ref == 'refs/heads/master' && (github.event_name == 'push' || github.event_name == 'release')
     steps:
     - name: retrieve built documentation
       uses: actions/download-artifact@v2
       with:
         name: documentation-page
     - name: debug
       run: |
+        if [ github.event_name == 'release']; then ENV_PAGE=${GITHUB_REF#refs/tags/}; else ENV_PAGE=current; fi
```
GarethCabournDavies, line 69:
```suggestion
        if [ github.event_name == 'release']; then ENV_PAGE=${GITHUB_REF#refs/tags/}; else ENV_PAGE=latest; fi
```

GarethCabournDavies:
Ping @ahnitz

duboism:
Hello,

Is there any plan to finish this soon ? As pointed in #2504, it is rather useful to have stable URL.

GarethCabournDavies:
Poke on this for review

GarethCabournDavies:
This is on the back-burner for now, waiting on review. Its good to poke reviewers though @ahnitz and @cdcapano

PR #4780, opened 2024-06-10
issue when using PSD from a file

aleynaakyuz:
I am working with the strain sensitivity curves for Cosmic Explorer, which are available [here](https://dcc.cosmicexplorer.org/CE-T2000017/public). When I read the data files using the pycbc.psd.read.from_txt function with df=0.1, low_freq_cutoff=5.1, and length=4000, I observed that the resulting curves appear flat, as shown in the attached plot.

After examining the issue, I found that this issue does not occur with different values for low_freq_cutoff. The problem arises due to floating point arithmetic. Specifically, the value of low_freq_cutoff is 5.1 and delta_f is 0.1, so kmin should ideally be 51. However, in Python, the calculation 5.1 / 0.1 yields 50.99999999999999, and since the code uses the int function, kmin is set to 50. This causes flow to be 5 and data_start to be -1 because the first value of freq_data is slightly above 50 for Cosmic Explorer.

I would suggest that you resolve this issue by replacing the int function with the round function in the relevant part of the code. This change will ensure that kmin is correctly calculated as 51 when low_freq_cutoff is 5.1 and delta_f is 0.1.

low_freq_cutoff5_1_40km


WuShichao:
Thanks, I know this `f_lower` issue, I just use `5.05 Hz` for CE. Any comments from @ahnitz?

ahnitz:
@WuShichao Can you verify your case as well, to see that you no longer have this restriction?

WuShichao:
@ahnitz I just realized that my issue was not caused by this, just because the initial frequency of CE asd file is `5.000000000000000888e+00` not `5`. I have checked Aleyna's case, this PR can fix her issue.

ahnitz:
@aleynaakyuz You'll need to rebase this PR and make sure the unittests pass along with the code climate report (found as a link towards the bottom of the checks).

PR #4787, opened 2024-06-13
Handle RA and dec in radians in pygrb

pannarale:
This PR simply uses radians for right ascension and declination throughout PyGRB.  Displaying them in degrees in the summary table on the results webpage is preserved.  This cleanup fixes the single IFO vs coherent IFO diagnostic plots.

This is a: bug fix

This change affects: PyGRB

## Motivation

The injections in plots such as
 
H1L1-PYGRB_PLOT_COH_IFOSNR_NSBH_GRB170728A_H1-1185258293-5648

with 2 IFOs or

H1L1V1-PYGRB_PLOT_COH_IFOSNR_ZOOMIN_NSBH_GRB170817A_H1-1187006058-5648

with 3 IFOs look wrong, but otherwise the behaviour of injections looks good.  The PR ensures that R.A. and declination are handled consistently throughout the workflow, so that the resulting plots are now correspondingly

H1L1-PYGRB_PLOT_COH_IFOSNR_NSBH_GRB170728A_H1-1185258293-5648

and

H1L1V1-PYGRB_PLOT_COH_IFOSNR_ZOOMIN_NSBH_GRB170817A_H1-1187006058-5648-1

## Testing performed
The updated plots were produced rerunning the entire workflow from injection production onwards (injection filtering, post-processing).

### DIFF
--- a/bin/pygrb/pycbc_pygrb_grb_info_table
+++ b/bin/pygrb/pycbc_pygrb_grb_info_table
@@ -47,10 +47,10 @@ parser.add_argument("--trigger-time", action="store",
                     help="GPS time of the GRB.")
 parser.add_argument("--ra", action="store",
                     default=None, required=True,
-                    help="Right ascention (degrees) of the GRB.")
+                    help="Right ascention (radians) of the GRB.")
 parser.add_argument("--dec", action="store",
                     default=None, required=True,
-                    help="Declination (degrees) of the GRB.")
+                    help="Declination (radians) of the GRB.")
 parser.add_argument("--sky-error", action="store",
                     default=0, required=False,
                     help="Sky-localisation error (degrees) of the GRB.")
@@ -73,10 +73,10 @@ utc_time = datetime(*lal.GPSToUTC(int(opts.trigger_time))[0:6]).strftime("%B %d 
 data[0].append(utc_time)
 headers.append('Coordinated Universal Time')
 
-data[0].append(str(opts.ra))
+data[0].append(str(numpy.rad2deg(float(opts.ra))))
 headers.append('R.A. (deg)')
 
-data[0].append(str(opts.dec))
+data[0].append(str(numpy.rad2deg(float(opts.dec))))
 headers.append('Dec (deg)')
 
 data[0].append(str(opts.sky_error))
@@ -88,8 +88,8 @@ headers.append('IFOs')
 for ifo in opts.ifos:
     antenna = Detector(ifo)
     factor = get_antenna_dist_factor(antenna,
-                                     numpy.deg2rad(float(opts.ra)),
-                                     numpy.deg2rad(float(opts.dec)),
+                                     float(opts.ra),
+                                     float(opts.dec),
                                      float(opts.trigger_time))
     data[0].append('%.3f' % factor)
     headers.append(ifo + ' Antenna Factor')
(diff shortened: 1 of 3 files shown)


bin/pygrb/pycbc_pygrb_grb_info_table
```diff
@@ -47,10 +47,10 @@ parser.add_argument("--trigger-time", action="store",
                     help="GPS time of the GRB.")
 parser.add_argument("--ra", action="store",
                     default=None, required=True,
-                    help="Right ascention (degrees) of the GRB.")
+                    help="Right ascention (radians) of the GRB.")
```
titodalcanton, line 50:
```suggestion
                    help="Right ascension (radians) of the GRB.")
```

bin/pygrb/pycbc_pygrb_grb_info_table
```diff
@@ -47,10 +47,10 @@ parser.add_argument("--trigger-time", action="store",
...
 parser.add_argument("--dec", action="store",
                     default=None, required=True,
-                    help="Declination (degrees) of the GRB.")
+                    help="Declination (radians) of the GRB.")
 parser.add_argument("--sky-error", action="store",
                     default=0, required=False,
                     help="Sky-localisation error (degrees) of the GRB.")
```
titodalcanton, line 56: Is the sky localization error also in radians, or is it actually degrees?

bin/pygrb/pycbc_pygrb_grb_info_table
```diff
@@ -88,8 +88,8 @@ headers.append('IFOs')
 for ifo in opts.ifos:
     antenna = Detector(ifo)
     factor = get_antenna_dist_factor(antenna,
-                                     numpy.deg2rad(float(opts.ra)),
-                                     numpy.deg2rad(float(opts.dec)),
+                                     float(opts.ra),
+                                     float(opts.dec),
```
titodalcanton, line 92: Unless I am missing some specific use case, the `--ra` and `--dec` args should be marked as `type=float`, and then you can drop the `float()` conversions here and above.

pycbc/workflow/matched_filter.py
```diff
@@ -243,12 +242,11 @@ def setup_matchedfltr_dax_generated_multi(workflow, science_segs, datafind_outs,
...
         exe_class = select_matchedfilter_class(match_fltr_exe)
-        # Right ascension + declination provided in degrees,
-        # so convert to radians
+        # Right ascension + declination must be provided in radians
         cp.set('inspiral', 'ra',
-               str(radians(float(cp.get('workflow', 'ra')))))
+               cp.get('workflow', 'ra'))
         cp.set('inspiral', 'dec',
-               str(radians(float(cp.get('workflow', 'dec')))))
+               cp.get('workflow', 'dec'))
```
titodalcanton, line 249: Does this revert back to having to specify the RA/dec in the .ini file in radians rather than degrees?

pannarale: It does. This way it is directly consistent with `pycbc_multi_inspiral` that [asks](https://github.com/gwastro/pycbc/blob/master/bin/pycbc_multi_inspiral#L246) for RA and dec or the sky-grid in radians.

Undoing this change would not be a problem, but it would imply something ugly, i.e., we want RA and dec in degrees in the ini file, but the sky-grid file content to be in radians.

titodalcanton: My understanding of the decision was that all user-facing angles would be in degrees, but all internal operations would be in radians. As long as the GRB information is translated into a PyGRB `.ini` file by reviewable code, however, I guess it is not a big deal to just have everything in radians.

titodalcanton: Discussed briefly offline, we will go with radians for now.

titodalcanton (approved):

PR #4789, opened 2024-06-19
require pre-numpy 2.0

ahnitz:
This pins to numpy < 2.0 while the transition is worked out. We'll need one final release with this requirement before switching to numpy 2.0 support.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,7 +34,7 @@ requires = []
 setup_requires = ['numpy>=1.16.0']
 install_requires = setup_requires + [
     'cython>=0.29',
-    'numpy>=1.16.0,!=1.19.0,numpy<2.0'
+    'numpy>=1.16.0,!=1.19.0,numpy<2.0',
     'scipy>=0.16.0',
     'astropy>=2.0.3,!=4.2.1,!=4.0.5',
     'matplotlib>=1.5.1',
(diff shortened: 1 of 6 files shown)


titodalcanton:
Do we need to edit `pyproject.toml` as well? I forget, but it was suggested to me.

setup.py
```diff
@@ -34,7 +34,7 @@
 setup_requires = ['numpy>=1.16.0']
 install_requires = setup_requires + [
     'cython>=0.29',
-    'numpy>=1.16.0,!=1.19.0',
+    'numpy>=1.16.0,!=1.19.0,numpy<2.0'
```
titodalcanton, line 37:
```suggestion
    'numpy>=1.16.0,!=1.19.0,numpy<2.0',
```

ahnitz:
> Do we need to edit `pyproject.toml` as well? I forget, but it was suggested to me.

Yes, if we want it to work for python3.9. Once we drop that, we can actually build against 2.0 and it will work in numpy versions supported back to those using python3.10. They changed the way this works in numpy 2.0 to support backwards compatibility, but only to numpy versions that support the currently supported python version.

ahnitz:
Not sure why build is failing on python 3.11. Running a test just to make sure that is the case and there aren't failures on the other versions. I can't reproduce on my laptop with python3.11 (no failures there) so unfortunately I'm resorting to trying here. Any ideas?

spxiwh (approved):

PR #4792, opened 2024-06-24
Use HFile instead of h5py.File in most places

GarethCabournDavies:
Use pycbc.io.hdf's HFile class rather than h5py.File directly so that we can make changes to the functionality.

For example checksumming, version information, and anything else we may want

This is a change to utilise potential future features
This change touches everywhere in the code, but affects nothing

## Contents
Not quite find-and-replace, but update everywhere which uses h5py.File to pycbc.io.HFile

#1525 #4768 

## Testing
Broadly, this should be checked by the CI running `--help` for all executables


GarethCabournDavies:
Note that I haven't touched `bin/plotting/pycbc_plot_bank_bins`, as that is already subclassing h5py.File, though I'm not sure whether the functionality which is there already exists elsewhere

GarethCabournDavies:
I have also avoided live (due to efficiency worries) and inference, so as not to interfere with the h5py.File wrapper used there

spxiwh (approved):
I'm okay with this, any issues @ahnitz ?

ahnitz (approved):
Yeah, I don't have a problem with this.

Issue #4761, opened 2024-05-28
Ensure that the changes made to the v23 branch for efficiency are also in main


GarethCabournDavies:
We made some changes required for the v23 branch to fix some points of huge memory usage in #4724 

This issue is to track that the changes made there are also in the main branch

The identified offending codes were:
| Executable | v23 fix | main fix |
| --- | --- | --- |
|`pycbc_plot_hist` | #4724 | #4545 |
|`pycbc_page_coinc_snrchi` | #4724  | #4764 |
|`pycbc_page_snrchi` | #4724 | #4729 |
|`page_foreground` | #4724 | #4764 |
|`pycbc_sngls_statmap` | #4732 | #4747 |
|`pycbc_page_coincinfo` | #4724 | #4726 |

I will comment below and add the PR to the table as appropriate

GarethCabournDavies:
After the above changes, all efficiency updates are in main

PR #4799, opened 2024-06-26
V23 RELEASE BRANCH: Efficiency savings in pycbc_injection_minifollowup

GarethCabournDavies:
pycbc_injection_minifollowup has already been fixed on master to cut according to end time and use a premask for SingleDetTriggers, but not on the v23 release branch - this implements those changes here

This isnt such a problem with small(er) injection files, but can be with larger ones, which is why we have only just seen the issue

This is an efficiency update

## Motivation
See efficiency testing below

## Contents
Use HFile.select() function to efficiently cut down the triggers in the injection minifollowup

Kind of the opposite of #4761 !

## Testing performed
Efficiency tested by using `time -v`:
v2.3.7:
```
	Maximum resident set size (kbytes): 2141384
```
current master (v2.4.2):
```
	Maximum resident set size (kbytes): 628760
```
After this change:
```
	Maximum resident set size (kbytes): 619912
```
Note that this used an injection set which wasn't particularly onerous - the savings will be even better for a larger hdf_trigger_merge file

### DIFF
--- a/bin/minifollowups/pycbc_injection_minifollowup
+++ b/bin/minifollowups/pycbc_injection_minifollowup
@@ -386,18 +386,36 @@ for num_event in range(num_events):
 
     for curr_ifo in args.single_detector_triggers:
         single_fname = args.single_detector_triggers[curr_ifo]
-        hd_sngl = SingleDetTriggers(single_fname, args.bank_file, None, None,
-                                    None, curr_ifo)
-        end_times = hd_sngl.end_time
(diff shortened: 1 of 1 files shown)


GarethCabournDavies:
I think this branch needs some cherrypicking of the numpy / scipy updates

spxiwh (approved):
Looks good to me. Yes, we will probably need to make sure the code does not try to use numpy >= 2 in this branch.

PR #4804, opened 2024-06-28
fix normalization of sky and time marginalization to reduce integral …

ahnitz:
…uncertainty

This reduces the uncertainty in the monte-carlo integral for time / sky marginalization. I was using a normalization that incorrectly introduced a large sample variance when in fact it was not needed. In testing this means that the likelihood (not log likelihood thankfully) std deviation is ~ 5-10% after 1000 samples and improves with root(N) as expected. For many cases, this is sufficient unless the level of accuracy required is very high. 

I've also added some comments for future improvements. The marginalizations implicitly assume a uniform time prior, which should be alleviated in a future pull request.

### DIFF
--- a/pycbc/inference/models/tools.py
+++ b/pycbc/inference/models/tools.py
@@ -364,14 +364,17 @@ class DistMarg():
         starts = []
         ends = []
 
-        tmin = tcmin + dt - snrs[iref].delta_t
-        tmax = tcmax + dt + snrs[iref].delta_t
+        delt = snrs[iref].delta_t
+        tmin = tcmin + dt - delt
+        tmax = tcmax + dt + delt
         if hasattr(self, 'tstart'):
             tmin = self.tstart[iref]
             tmax = self.tend[iref]
 
-        starts.append(max(tmin, snrs[iref].start_time))
-        ends.append(min(tmax, snrs[iref].end_time))
+        # Make sure we draw from times within prior and that have enough
+        # SNR calculated to do later interpolation
+        starts.append(max(tmin, snrs[iref].start_time + delt))
+        ends.append(min(tmax, snrs[iref].end_time - delt * 2))
 
         idels = {}
         for ifo in ifos[1:]:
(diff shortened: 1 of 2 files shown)


pycbc/inference/models/tools.py
```diff
@@ -45,16 +45,16 @@ def draw_sample(loglr, size=None):
     cdf = numpy.exp(loglr).cumsum()
     cdf /= cdf[-1]
     xl = numpy.searchsorted(cdf, x)
+    
+    loglr = loglr - logsumexp(loglr)
```
WuShichao, line 49: Thanks for the fix. What's the difference between normalization on `cdf` of `loglr` (previous code) and on `loglr` itself (this PR)? Can we remove the previous code? Here you generate `xl` twice.

pycbc/inference/models/tools.py
```diff
@@ -120,7 +120,7 @@ def pop_prior(param):
         self.reconstruct_phase = False
         self.reconstruct_distance = False
         self.reconstruct_vector = False
-        self.precalc_antennna_factors = False
+        self.precalc_antenna_factors = False
```
WuShichao, line 123: This typo has been fixed in another merged PR; please rebase it.

pycbc/inference/models/tools.py
ahnitz: I'll need to remove this. I accidentally left it in for debugging purposes. I wasn't quite sure where the problem was originally. It's not here, it was the fact that the prior normalization was tied to the samples rather than upfront. This caused a large uncertainty depending on how the samples ended up distributed.

pycbc/inference/models/tools.py
```diff
@@ -363,7 +368,7 @@ def draw_times(self, snrs, size=None):
         starts = []
         ends = []
 
-        tmin, tmax = tcmin - dt, tcmax + dt
+        tmin, tmax = tcmin + dt - snrs[iref].delta_t, tcmax + dt + snrs[iref].delta_t
```
WuShichao, line 371: Why did you change `tcmin - dt` to `tcmin + dt`? Is `snrs[iref].delta_t` the time resolution of the SNR time series for a detector?

WuShichao:
Some code climate issues need to be fixed.

pycbc/inference/models/tools.py
ahnitz: To make sure the boundary condition was satisfied.

ahnitz:
@WuShichao The remaining cc issues can't be resolved and were not introduced in this PR.

PR #4805, opened 2024-06-28
allow approximant to be read from config file

ahnitz:


yi-fan-wang:
@ahnitz This breaks the CI test

yi-fan-wang (approved):

yi-fan-wang:
@ahnitz Looks good to me. What is the context for "use-cross"? Is this for when you need to generate two polarizations?

PR #4806, opened 2024-06-28
add ability to cache detector response to marginalize time model and us…

ahnitz:
This implements two features into the marginalized time model. 

1) it can now take a 'sample_rate' argument. If given, a different sample rate can be used for the marginalization than the intrinsic one of the data. Note it really should be higher than the data sample rate to make sense.

2) The detector response factors are now retrieved from the precalculated ones which improves the marginalization speed by ~ 50 %.

### DIFF
--- a/examples/inference/margtime/margtime.ini
+++ b/examples/inference/margtime/margtime.ini
@@ -3,6 +3,9 @@
 name = marginalized_time
 low-frequency-cutoff = 30.0
 
+# This is the sample rate used for the model and marginalization
+sample_rate = 2048
+
 marginalize_vector_params = tc, ra, dec, polarization
 marginalize_vector_samples = 500
 
--- a/examples/inference/margtime/margtime.ini
+++ b/examples/inference/margtime/margtime.ini
@@ -4,10 +4,10 @@ name = marginalized_time
 low-frequency-cutoff = 30.0
 
 # This is the sample rate used for the model and marginalization
-sample_rate = 2048
+sample_rate = 4096
 
 marginalize_vector_params = tc, ra, dec, polarization
-marginalize_vector_samples = 500
+marginalize_vector_samples = 2000
 
 ; You shouldn't use phase marginalization if the approximant has
 ; higher-order modes
--- a/pycbc/inference/models/marginalized_gaussian_noise.py
+++ b/pycbc/inference/models/marginalized_gaussian_noise.py
@@ -19,6 +19,7 @@ distance.
 """
 
 import itertools
+import logging
 import numpy
 from scipy import special
 
@@ -207,6 +208,7 @@ class MarginalizedTime(DistMarg, BaseGaussianNoise):
     def __init__(self, variable_params,
                  data, low_frequency_cutoff, psds=None,
                  high_frequency_cutoff=None, normalize=False,
+                 sample_rate=None,
                  **kwargs):
 
         self.kwargs = kwargs
@@ -240,6 +242,13 @@ class MarginalizedTime(DistMarg, BaseGaussianNoise):
                     gates=self.gates, **kwargs['static_params'])
 
         self.dets = {}
+        
+        if sample_rate is not None:
+            logging.info("Using %s sample rate for marginalization", sample_rate)
+            for det in self._whitened_data:
+                tlen = int(round(float(sample_rate) * self.whitened_data[det].duration))
+                flen = tlen // 2 + 1
+                self._whitened_data[det].resize(flen)    
 
     def _nowaveform_loglr(self):
         """Convenience function to set loglr values if no waveform generated.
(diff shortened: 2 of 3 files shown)


ahnitz:
@cdcapano Can I merge this?

pycbc/inference/models/marginalized_gaussian_noise.py
```diff
@@ -241,6 +243,15 @@ def __init__(self, variable_params,
...
         self.dets = {}
 
+        if sample_rate is not None:
+            logging.info("Using %s sample rate for marginalization",
+                         sample_rate)
+            for det in self._whitened_data:
+                tlen = int(round(float(sample_rate) *
+                           self.whitened_data[det].duration))
+                flen = tlen // 2 + 1
+                self._whitened_data[det].resize(flen)
```
cdcapano, line 253: Is it worth raising an error here if the sample rate is smaller than the data?

Also, this might lead to a bug... the whitened data is set when the psd is set (see [here](https://github.com/gwastro/pycbc/blob/5a9d151dd9d013c1bb651770688f589eb84f8965/pycbc/inference/models/gaussian_noise.py#L263)). In principle you can change the psd after the class is initialized. I don't think it's done in an inference run anywhere currently, but could be done in a notebook and/or possibly implemented later. I think it might be better to do this by redefining `psd.setter` so that it calls the parent, then runs this block of code to update the whitened data. You'll need to save `sample_rate` as an attribute in that case. Maybe call it something else to distinguish it from the data sample rate (`likelihood_sample_rate`?).

pycbc/inference/models/marginalized_gaussian_noise.py
```diff
@@ -207,6 +208,7 @@ class MarginalizedTime(DistMarg, BaseGaussianNoise):
     def __init__(self, variable_params,
                  data, low_frequency_cutoff, psds=None,
                  high_frequency_cutoff=None, normalize=False,
+                 sample_rate=None,
```
cdcapano, line 211: What's the advantage of doing this instead of just increasing the sample rate of the data? Also, this might need some documentation, or maybe change the name of the argument to make it clearer what it is. People are going to think this is the data sample rate otherwise.

pycbc/inference/models/marginalized_gaussian_noise.py
```diff
@@ -325,15 +336,20 @@ def _loglr(self):
...
-            fp, fc = self.dets[det].antenna_pattern(
-                                    params['ra'],
-                                    params['dec'],
-                                    params['polarization'],
-                                    params['tc'])
-            dt = self.dets[det].time_delay_from_earth_center(params['ra'],
-                                                             params['dec'],
-                                                             params['tc'])
+
+            if self.precalc_antenna_factors:
```
cdcapano, line 340: How is `precalc_antenna_factors` set to True in a run? Is this an argument for the model? I couldn't find where it would be set.

cdcapano (commented):
@ahnitz Have some questions and concerns before approving. See below.

pycbc/inference/models/marginalized_gaussian_noise.py
ahnitz: Increasing the sample rate of the data means that you need to do data conditioning and parts of the generation at  a higher sample rate. That's only useful if you actually care about the frequency band at higher frequencies. If you just need better timing resolution that would be extremely slow and memory intensive.

I think the separation is clear as the data sample rate is in the data section. This is only about the model itself. BTW, we already have this distinction for the heterodyne likelihoods when doing sky marginalization. This was just advancing the same interface / feature to the more regular model

pycbc/inference/models/marginalized_gaussian_noise.py
ahnitz: It's already handled by the marginalization parent class. The features were implemented first for other cases, and this just extends the use to the marginalized time model.

pycbc/inference/models/marginalized_gaussian_noise.py
ahnitz: Ok, interesting. Really, I just need this done before we do the FFT in the likelihood to calculate the SNR time series. I'm not sure this make more sense in the psd setter, but maybe I should put it directly in the likelihood? Is the psd setter preferable over that?

Yeah, truncating should probably be an error.

cdcapano: If you're modifying the `_whitened_data` data than it would make more sense to do it in the PSD setter, to avoid the error I mentioned. But yeah, that is a round about way to just increase the sample rate of the likelihood time series. Don't you also need to zero pad the waveform in that case too? In which case, there's a bunch of 0*0 going on in the match function. What if you add this as an option to the `matched_filter_core` instead? That is, it takes in an optional sample rate argument, and if provided you zero pad `qtilde` just prior to doing the `ifft`? In the model you would store the sample rate as an attribute and pass it to `matched_filter_core` in the likelihood call.

pycbc/inference/models/marginalized_gaussian_noise.py
cdcapano: Ok, but I was asking how you set that in the config file and how that's passed to the model initialization. I couldn't figure it out from looking at the code, although I didn't dig too deep.

ahnitz: Ah, it's automatically populated if you do sky marginalization so it's the same set of settings really. These factors are calculated in any case when the sky grid is set up.

pycbc/inference/models/marginalized_gaussian_noise.py
ahnitz: @cdcapano Indeed. Actually the waveform is already zeropadding to whatever the whitened data is. I guess I could just do both there (probably not too slow to just copy out the data too, so it's not replacing the original values).

ahnitz: @cdcapano I'll make this change.

ahnitz: @cdcapano I think for now I'll make this directly in the likelihood. I thought about adding to matched filter core, but I'm not sure the best strategy at the moment.

ahnitz:
@cdcapano Is this now ok?

examples/inference/margtime/margtime.ini
```diff
@@ -3,8 +3,11 @@
 name = marginalized_time
 low-frequency-cutoff = 30.0
 
+# This is the sample rate used for the model and marginalization
+sample_rate = 4096
+
```
ahnitz, line 8: I updated the example to make explicit use of this feature.

cdcapano (approved):
Ok, this looks good. This seems like it would be a good thing to add to the matched filter core at some point, but this will do for now.

PR #4808, opened 2024-06-29
Edit make banksim to allow external injection file

divyajyoti09:

This is a: new feature

This change affects: `pycbc_make_banksim` function

This change changes: documentation, banksim generator

## Motivation
At the moment, only `lalapps_inspinj` function is allowed for creating injection file through the workflow generator `pycbc_make_banksim`. With this update, an externally created injection file can also be called.

## Contents
The code allows for option to use `inj.xml` provided by user if the user chooses `[external_injection]` section instead of `[inspinj]` in the config file.

It also indicates the same in the documentation text.

## Testing performed
I have tested it by generating the workflow using `inj.xml` generated by `pycbc_create_injections`. It works fine.


bin/pycbc_make_banksim
```diff
@@ -187,8 +187,19 @@ shutil.copy(banksim_prog, 'scripts/pycbc_banksim')
...
+    logging.info("Using lalapps_inspinj to create injections")
+    inj_str = "lalapps_inspinj " + get_ini_opts(confs, "inspinj") + "--output inj.xml"
+    os.system(inj_str)
+elif confs.has_section("external_injection"):
+    logging.info("Using external injection file. Please ensure the file is in sim_inspiral table (.xml) format.")
+    inj_file_path = confs.get("external_injection", "inj-file")
+    if inj_file_path == "inj.xml":
+        pass
+    else:
+        os.system("cp {} inj.xml".format(inj_file_path))
```
ahnitz, line 200: You should probably shutil instead of a system command, otherwise this seems fine.

ahnitz (approved):

PR #4809, opened 2024-07-02
PyGRB: Propagating slide_id information across postprocessing plotting scripts

sebastiangomezlopez:

This is a new feature that propagates slide_id information across some PyGRB's plotting scripts. In particular it implements a new way of displaying triggers from a specific slide_id.

Currently some PyGRB plots are overpopulated with the addition of timeslides in recent runs. For example, the following plot shows the background triggers obtained from zerlolag data + all timeslides (blue crosses), and just zerolag data, i.e. `slide_id=0` (black crosses) around GW170817

image

This change affects:  PyGRB

- pycbc/bin/pycbc_pygrb_plot_snr_timeseries
- pycbc/bin/pycbc_pygrb_plot_coh_ifo_snr
- pycbc/bin/pycbc_pygrb_plot_null_stats
- pycbc/bin/pycbc_pygrb_plot_chisq_veto

- pycbc/results/pygrb_plotting_utils.py
- pycbc/results/pygrb_postprocessing_utils.py


pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -172,6 +172,19 @@ def pygrb_add_bestnr_cut_opt(parser):
...
+# Wrapper to pick triggers with certain slide_ids 
+# =============================================================================
+def slide_filter(File, data, slide_id=None):
+    """   """
+    mask  = numpy.where(File['network/slide_id'][:]==slide_id)[0]
+
+    if slide_id is not None:
+        return data[mask]
+    else:
+        return data
```
titodalcanton, line 185: Some comments on this function:
* Instead of adding a comment before the function, put the same text as the function's docstring.
* When checking for None, invert the logic: if `slide_id` is None then return `data` right away, otherwise carry on (without an `else` block).
* `File` is a bad choice for a variable name in Python (variables should generally be lowercase, and `file` is a reserved word in Python). Maybe `trigger_file` or `trig_file` are better choices?

bin/pygrb/pycbc_pygrb_plot_snr_timeseries
```diff
@@ -166,8 +164,9 @@ trig_data_time = trig_data[x_key][:]
 inj_data_time = inj_data[x_key][:] if inj_file else None
 
 # Obtain SNRs
+
 trig_data_snr = trig_data[y_key][:]
-inj_data_snr = inj_data[y_key][:] if inj_file else None
+inj_data_snr = inj_data[y_key][:]  if inj_file else None
```
titodalcanton, line 169: Whitespace change that goes against PEP8. Please revert.

bin/pygrb/pycbc_pygrb_plot_snr_timeseries
```diff
@@ -203,7 +202,8 @@ xlims = [start, end]
 if opts.x_lims:
     xlims = opts.x_lims
     xlims = map(float, xlims.split(','))
+
 plu.pygrb_plotter([trig_data_time, trig_data_snr],
                   [inj_data_time, inj_data_snr],
                   "Time since %.3f (s)" % (central_time), y_label,
-                  opts, cmd=' '.join(sys.argv))
+                  opts, cmd=' '.join(sys.argv), fig=plt.figure())
```
titodalcanton, line 209: Is this change (and the corresponding one to `pygrb_plotter()` below) really needed? I do not see a functional change.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -194,6 +196,7 @@ init_logging(opts.verbose, format="%(asctime)s: %(levelname)s: %(message)s")
 
 # Check options
 trig_file = os.path.abspath(opts.trig_file)
+print(trig_file)
```
titodalcanton, line 199: This looks like a temporary debugging aid, please remove.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -315,3 +319,4 @@ plu.pygrb_plotter(trigs, injs, x_label, y_label, opts,
                   snr_vals=snr_vals, conts=conts, colors=colors,
                   shade_cont_value=cont_value, vert_spike=True,
                   cmd=' '.join(sys.argv))
+                                                                                                                                                                                                                                                                                                                                                                                                                                                         321,41        Bot
```
titodalcanton, line 322: This looks like an accidental paste, which is causing the tests to fail. Remove.

bin/pygrb/pycbc_pygrb_plot_snr_timeseries
sebastiangomezlopez: I added it because I thought it'd be nice to be able to pass an existing pyplot.figure object to multiple instances of  pygrb_plotter to plot different datasets when debugging.  For example, the [plot](https://private-user-images.githubusercontent.com/92443572/345035725-d7872fd6-7cf9-4245-88f3-f4e3fc2b35e1.png?jwt=eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3MjEzOTgwMzAsIm5iZiI6MTcyMTM5NzczMCwicGF0aCI6Ii85MjQ0MzU3Mi8zNDUwMzU3MjUtZDc4NzJmZDYtN2NmOS00MjQ1LTg4ZjMtZjRlM2ZjMmIzNWUxLnBuZz9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNDA3MTklMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjQwNzE5VDE0MDIxMFomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPWZhZTI3Mzc1MTQzNDBhZDFlODVlYTZmZTkxNzAwNzIxOWY1OGM5MjRlNzI5MjNlY2E0NTVhMjhkNjQ5NDhhOTAmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JmFjdG9yX2lkPTAma2V5X2lkPTAmcmVwb19pZD0wIn0.F2RS-9lSLWCvO9FzIoNMson6FobCPs1QewvscsdyR-0), was easily produced with such feature. 

In any case, I'm happy to remove it if you think such modification is not needed. Let me know.

pycbc/results/pygrb_plotting_utils.py
```diff
@@ -188,12 +188,14 @@ def axis_min_value(trig_values, inj_values, inj_file):
 # =============================================================================
 def pygrb_plotter(trigs, injs, xlabel, ylabel, opts,
                   snr_vals=None, conts=None, shade_cont_value=None,
-                  colors=None, vert_spike=False, cmd=None):
+                  colors=None, vert_spike=False, cmd=None, fig=None):
```
pannarale, line 191: Please revert.

pycbc/results/pygrb_plotting_utils.py
```diff
@@ -188,12 +188,14 @@ def axis_min_value(trig_values, inj_values, inj_file):
...
     """Master function to plot PyGRB results"""
     from matplotlib import pyplot as plt
 
     # Set up plot
-    fig = plt.figure()
+    if fig is None:
+        fig = plt.figure()
+
```
pannarale, line 198: Please revert.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
```diff
@@ -233,10 +235,10 @@ ifos, vetoes = ppu.extract_ifos_and_vetoes(
 )
 
 # Extract trigger data
-trig_data = load_data(trig_file, ifos, vetoes, opts)
+trig_data = load_data(trig_file, ifos, vetoes, opts, slide_id=0)
 
 # Extract (or initialize) injection data
-inj_data = load_data(found_file, ifos, vetoes, opts, injections=True)
+inj_data = load_data(found_file, ifos, vetoes, opts, injections=True, slide_id=0)
```
pannarale, line 241: I would add a `slide_id` option to plotting scripts like this one, with 0 as default, and explaining that `None` will force plotting all triggers from all timeslides; then I would pass this to `load_data` via the keyword `slide_id`

Injections are on unslid data so `inj_data = load_data(found_file, ifos, vetoes, opts, injections=True, slide_id=0)` can be left unchanged.

pannarale (changes requested):

pycbc/results/pygrb_postprocessing_utils.py
titodalcanton: The docstring is sufficient, anything else just adds lines to the file.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
sebastiangomezlopez: I just finished implementing this, let me know whether I misunderstood your request.

bin/pygrb/pycbc_pygrb_plot_snr_timeseries
```diff
@@ -97,7 +97,7 @@ def reset_times(data_time, trig_time):
 # =============================================================================
 parser = ppu.pygrb_initialize_plot_parser(description=__doc__)
 parser.add_argument("-t", "--trig-file", action="store",
-                    default=None, required=True,
+        default=None, required=True,
```
pannarale, line 100: This needs to be reverted.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -92,7 +92,9 @@ def pygrb_initialize_plot_parser(description=None):
     parser.add_argument('--plot-caption', default=None,
                         help="If provided, use the given string as the plot " +
                         "caption")
-
+    parser.add_argument("--slide-id", default=0, type=int,
+                        help="If None, plotting scripts will use triggers" +
+                        "from all short slides.")
```
pannarale, line 97: This option is not common to all PyGRB plotting scripts so it does not belong in `pygrb_initialize_plot_parser`.  Since 4 plotting scripts use it, you may want to create something like `pygrb_add_slide_opts(parser)` rather than copying and pasting this multiple times.
Also, the help string does not explain what the option is for; it dives straight into the consequences of setting it to `None`.

pannarale (changes requested):

bin/pygrb/pycbc_pygrb_plot_snr_timeseries
```diff
@@ -44,13 +44,12 @@ __version__ = pycbc.version.git_verbose_msg
 __date__ = pycbc.version.date
 __program__ = "pycbc_pygrb_plot_snr_timeseries"
 
-
```
pannarale, line 47: Do not remove the empty line.

bin/pygrb/pycbc_pygrb_plot_snr_timeseries
```diff
@@ -165,6 +168,7 @@ trig_data_time = trig_data[x_key][:]
 inj_data_time = inj_data[x_key][:] if inj_file else None
 
 # Obtain SNRs
+
```
pannarale, line 171: Remove empty line.

pannarale (changes requested):

pycbc/results/pygrb_postprocessing_utils.py
pannarale: I thought about this more and the first line will not allow the user to set this option to `None`.  I don't _think_ you can specify multiple type options (`None` or `int`); a workaround would be 
```
parser.add_argument("--slide-id", default='0',
                    help="...")
...
if slide_id.isdigit():
    slide_id = int(count)
elif slide_id=='None':
    slide_id = None
else:
    Throw an error to say that --slide-id has to be an int or None
```
@titodalcanton, any other suggestion?  We are trying to keep the zero-lag (`slide_id = 0`) as the default for safety, and to enable plotting data from all timeslides at once (`slide_id=None`)

pannarale (changes requested):

pycbc/results/pygrb_postprocessing_utils.py
titodalcanton: I do not see a better alternative, though I would rather use `--slide-id all` instead of `--slide-id None`, which sounds very nonsensical from a UX point of view.

pannarale: True: given there is no way of exploiting `None`, `'all'` makes more sense.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -92,10 +92,29 @@ def pygrb_initialize_plot_parser(description=None):
...
+    """
+       This function overwrites the types of input slide_id information
+       when loading data in postprocessing scripts.
+    """
+    if args.slide_id.isdigit():
+        args.slide_id = int(args.slide_id)
+    elif args.slide_id.lower() == "all":
+        args.slide_id = None
+    else:
+        raise ValueError("--slide-id must be all or int")
```
pannarale, line 115:
```suggestion
        raise ValueError("--slide-id must be the string all or an int")
```

pannarale (dismissed):

titodalcanton (approved):
Looks good to me. Thanks!

PR #4810, opened 2024-07-03
Update pycbc_brute_bank

ahnitz:
f_ref should not be manually set. If someone would like to set this, they should do so from a config file.


yi-fan-wang (approved):

WuShichao (approved):
Yeah, we need to make use of config from now on :)

WuShichao:
One of the tests takes a long time.

PR #4811, opened 2024-07-05
Add a standard --version option to common pycbc options

GarethCabournDavies:
Add an option to give --version information, allowing the user to specify how verbose the information is

This is a code management improvement
This change touches all areas of the codebase
This change affects no output except the versioning page on reults pages

## Motivation
This, along with the logging standardization changes previously made, should make it easier to contribute to PyCBC, particularly if making a new executable

## Contents
Removed all cases of '--version' arguments being added to a parser when `add_common_pycbc_options` is used

Modify (and rename) the pycbc.version.Version class to be able to take in a modifier defining how verbose the library version information is.

## Testing performed
Various executables run with `--help` and `--version`, `--version 0`, `--version 1`, `--version 2`, `--version 3` as appropriate

The CI will confirm if I have removed an `import pycbc.version` when it is needed for an implicit pycbc import - which I probably have

Next steps for testing are to run the `pycbc_page_versioning` script with every executable as input, and modify it to have some of the more verbose version information at the top (probably the first executable will be given the modifier)

## Additional notes
I have avoided pycbc_live, for a similar reason to why I have not touched it in the logging changes ; I do not want to damage any of the codes which work on the output of that

### DIFF
--- a/bin/bank/pycbc_aligned_stoch_bank
+++ b/bin/bank/pycbc_aligned_stoch_bank
@@ -25,7 +25,6 @@ import numpy
 import logging
 
 import pycbc
-import pycbc.version
 from pycbc import tmpltbank
 # Old ligolw output functions no longer imported at package level
 import pycbc.tmpltbank.bank_output_utils as bank_output
@@ -34,10 +33,6 @@ import pycbc.psd
 import pycbc.strain
 from pycbc.pnutils import named_frequency_cutoffs
 
-
-__author__  = "Ian Harry <ian.harry@astro.cf.ac.uk>"
-__version__ = pycbc.version.git_verbose_msg
-__program__ = "pycbc_aligned_stoch_bank"
 
 # Read command line option
 _desc = __doc__[1:]
--- a/bin/bank/pycbc_aligned_stoch_bank
+++ b/bin/bank/pycbc_aligned_stoch_bank
@@ -45,7 +45,6 @@ parser = argparse.ArgumentParser(description=_desc,
            formatter_class=tmpltbank.IndentedHelpFormatterWithNL)
 
 # Begin with code specific options
-parser.add_argument("--version", action="version", version=__version__)
 pycbc.add_common_pycbc_options(parser)
 parser.add_argument("-V", "--vary-fupper", action="store_true", default=False,
                     help="Use a variable upper frequency cutoff in laying "
--- a/bin/pygrb/pycbc_make_offline_grb_workflow
+++ b/bin/pygrb/pycbc_make_offline_grb_workflow
@@ -20,13 +20,6 @@
 Make workflow for the archival, targeted, coherent inspiral pipeline.
 """
 
-import pycbc.version
-
-__author__ = "Andrew Williamson <andrew.williamson@ligo.org>"
-__version__ = pycbc.version.git_verbose_msg
-__date__ = pycbc.version.date
-__program__ = "pycbc_make_offline_grb_workflow"
-
 import sys
 import os
 import argparse
--- a/bin/pygrb/pycbc_make_offline_grb_workflow
+++ b/bin/pygrb/pycbc_make_offline_grb_workflow
@@ -46,7 +46,6 @@ workflow_name = "pygrb_offline"
 # Parse command line options and instantiate pycbc workflow object
 parser = argparse.ArgumentParser()
 add_common_pycbc_options(parser)
-parser.add_argument("--version", action="version", version=__version__)
 _workflow.add_workflow_command_line_group(parser)
 _workflow.add_workflow_settings_cli(parser)
 args = parser.parse_args()
--- a/bin/workflows/pycbc_make_offline_search_workflow
+++ b/bin/workflows/pycbc_make_offline_search_workflow
@@ -153,7 +153,6 @@ def check_stop(job_name, container, workflow, finalize_workflow):
 
 parser = argparse.ArgumentParser(description=__doc__[1:])
 pycbc.add_common_pycbc_options(parser)
-parser.add_argument('--version', action='version', version=__version__)
 wf.add_workflow_command_line_group(parser)
 wf.add_workflow_settings_cli(parser)
 args = parser.parse_args()
--- a/bin/workflows/pycbc_make_offline_search_workflow
+++ b/bin/workflows/pycbc_make_offline_search_workflow
@@ -22,10 +22,6 @@ finding and ranking then generate post-processing
 and plots.
 """
 import pycbc
-import pycbc.version
-__version__ = pycbc.version.git_verbose_msg
-__date__    = pycbc.version.date
-__program__ = "pycbc_offline"
 
 import sys
 import socket
--- a/bin/workflows/pycbc_make_sbank_workflow
+++ b/bin/workflows/pycbc_make_sbank_workflow
@@ -171,7 +171,6 @@ class CombineHDFBanksExecutable(wf.Executable):
 _desc = __doc__[1:]
 parser = argparse.ArgumentParser(description=_desc)
 pycbc.add_common_pycbc_options(parser)
-parser.add_argument('--version', action='version', version=__version__)
 parser.add_argument("--output-file", type=str, default=None,
                     help="Specify the output file name. Either a name can be "
                          "provided or a full path to file. Is this is not "
--- a/bin/workflows/pycbc_make_sbank_workflow
+++ b/bin/workflows/pycbc_make_sbank_workflow
@@ -28,15 +28,8 @@ import os
 import argparse
 
 import pycbc
-import pycbc.version
 import pycbc.workflow as wf
 import pycbc.workflow.pegasus_workflow as pwf
-
-# Boiler-plate stuff
-__author__  = "Ian Harry <ian.harry@ligo.org>"
-__version__ = pycbc.version.git_verbose_msg
-__date__    = pycbc.version.date
-__program__ = "pycbc_make_sbank_workflow"
 
 # We define classes for all executables used in the workflow
 
--- a/pycbc/_version.py
+++ b/pycbc/_version.py
@@ -83,7 +83,7 @@ class PyCBCVersionAction(argparse._StoreAction):
     """
     default_help = (
         'Display PyCBC version information and exit. '
-        'Can be supplied a modifier integer to control the '
+        'Can optionally supply a modifier integer to control the '
         'verbosity of the version information. 0 and 1 are the '
         'same as --version; 2 provides more detailed PyCBC library '
         'information; 3 provides information about PyCBC, '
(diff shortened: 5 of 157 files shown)


bin/workflows/pycbc_make_sbank_workflow
```diff
@@ -171,7 +171,6 @@ class CombineHDFBanksExecutable(wf.Executable):
 _desc = __doc__[1:]
 parser = argparse.ArgumentParser(description=_desc)
 pycbc.add_common_pycbc_options(parser)
-parser.add_argument('--version', action='version', version=__version__)
```
titodalcanton, line 174: Looks like you can also drop the definition of `__version__` above, and maybe one import?

bin/workflows/pycbc_make_offline_search_workflow
```diff
@@ -153,7 +153,6 @@ def check_stop(job_name, container, workflow, finalize_workflow):
 
 parser = argparse.ArgumentParser(description=__doc__[1:])
 pycbc.add_common_pycbc_options(parser)
-parser.add_argument('--version', action='version', version=__version__)
```
titodalcanton, line 156: Drop the definition of `__version__` too?

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -46,7 +46,6 @@ workflow_name = "pygrb_offline"
 # Parse command line options and instantiate pycbc workflow object
 parser = argparse.ArgumentParser()
 add_common_pycbc_options(parser)
-parser.add_argument("--version", action="version", version=__version__)
```
titodalcanton, line 49: Drop the definition of `__version__`?

bin/bank/pycbc_aligned_stoch_bank
```diff
@@ -45,7 +45,6 @@ parser = argparse.ArgumentParser(description=_desc,
            formatter_class=tmpltbank.IndentedHelpFormatterWithNL)
 
 # Begin with code specific options
-parser.add_argument("--version", action="version", version=__version__)
```
titodalcanton, line 48: Drop the definition of `__version__` as well?

pycbc/_version.py
```diff
@@ -75,40 +76,82 @@ def get_lal_info(module, lib_glob):
...
-    def __call__(self, parser, namespace, values, option_string=None):
-        import pycbc
+    default_help = (
+        'Display PyCBC version information and exit. '
+        'Can optionally supply a modifier integer to control the '
+        'verbosity of the version information. 0 and 1 are the '
+        'same as --version; 2 provides more detailed PyCBC library '
+        'information; 3 provides information about PyCBC, '
+        'LAL and LALSimulation packages (if installed)'
+    )
```
titodalcanton, line 91: What is the idea behind 0 and 1 both being equivalent to no option?

titodalcanton: Ah let me guess, `--version --version` is equivalent to `--version 2`, so 0 and 1 are both equivalent to a single `--version`?

pycbc/_version.py
```diff
@@ -28,6 +28,7 @@
 import logging
 
 logger = logging.getLogger('pycbc._version')
+SUPPRESS = "==SUPPRESS=="
```
titodalcanton, line 31: What is the point of this variable?

tools/pycbc_test_suite.sh
```diff
@@ -44,10 +44,28 @@ if [ "$PYCBC_TEST_TYPE" = "help" ] || [ -z ${PYCBC_TEST_TYPE+x} ]; then
         else
             echo -e "    Pass."
         fi
+	if [[ `echo $prog | egrep '(pycbc_copy_output_map|pycbc_submit_dax|pycbc_stageout_failed_workflow)'` ]] ; then
+	    continue
+	fi
+	echo -e ">> [`date`] running $prog --version"
+	$prog --version &> $LOG_FILE
+	if test $? -ne 0 ; then
```
titodalcanton, line 52: These lines need an extra indentation.

tools/pycbc_test_suite.sh
```diff
@@ -44,10 +44,28 @@ if [ "$PYCBC_TEST_TYPE" = "help" ] || [ -z ${PYCBC_TEST_TYPE+x} ]; then
...
     done
-    # also check that --version works for one executable
-    echo -e ">> [`date`] running pycbc_inspiral --version"
+    # also check that --version with increased modifiers works for one executable
+    echo -e ">> [`date`] running pycbc_inspiral --version with modifiers"
     pycbc_inspiral --version &> $LOG_FILE
+    pycbc_inspiral --version 0 &> $LOG_FILE
+    pycbc_inspiral --version 1 &> $LOG_FILE
+    pycbc_inspiral --version 2 &> $LOG_FILE
+    pycbc_inspiral --version 3 &> $LOG_FILE
```
titodalcanton, line 68: This will miss failure modes that are triggered by all invocations except for the last one, since the last one will succeed and set `$?` to zero regardless of the previous ones. I suggest doing a for loop with a similar structure to the loop just above.

titodalcanton: I.e.
```bash
for modifier in "" 0 1 2 3`
do
    pycbc_inspiral --version ${modifier} &> $LOG_FILE
    if test $? -ne 0
        …
    fi
done
```

titodalcanton:
I made it through 160 files! This looks mostly ok but I left some comments above; some of them can be grouped into a single "there are more things you can remove" comment.

bin/workflows/pycbc_make_sbank_workflow
GarethCabournDavies: Yeah I saw that - it was the fact that all the `__author__` information was around it and I wasn't sure if that should be kept for some reason

titodalcanton: I doubt it, it might be a legacy from an old Python writing style.

pycbc/_version.py
GarethCabournDavies: Thats what I _wanted_ to happen, (similar to the verbose action), but argparse's version and count asctions did not play nicely with one another when I tried to combine them.

Basically telling it to count, but then also to exit during the argument parsing stage wasn't possible

pycbc/_version.py
GarethCabournDavies: Ah, thats a leftover from trying to combine argparse actions. I have deleted it now

tools/pycbc_test_suite.sh
GarethCabournDavies: Thats a really good point. 

I've also move all of the `if test $? -ne 0 ; then` statement into its own function, as it is reused _a lot_ in this script

tools/pycbc_test_suite.sh
GarethCabournDavies: Fixed by moving to function

bin/workflows/pycbc_make_sbank_workflow
GarethCabournDavies: Okay, removed (and with the below examples as well)

pycbc/_version.py
titodalcanton: Ok, so in some sense the specific choise of values is a slight inconvenient due to limitations in argparse?

tools/pycbc_test_suite.sh
titodalcanton: Does moving to a function really work though? Does bash let you change a global variable like that from inside a function?

pycbc/_version.py
GarethCabournDavies: yeah, though I can put it back so that 0,1,2 give the different options if preferred

tools/pycbc_test_suite.sh
GarethCabournDavies: yup, I checked as I had the same worry - so long as the `local` keyword isnt set within the function, it should work

titodalcanton: Nope, that does not work:
```
(base) tito@newman:~ $ cat foo.sh 
ANTANI=0

echo $ANTANI

function set_antani {
    ANTANI=1
}

echo $ANTANI
(base) tito@newman:~ $ bash foo.sh 
0
0
```

pycbc/_version.py
titodalcanton: I guess the only odd thing is zero being a valid option (since, to me, `--version 0` sounds like "give me version info, but none of it"). But we can leave it as is.

tools/pycbc_test_suite.sh
```diff
@@ -11,77 +11,63 @@ LOG_FILE=$(mktemp -t pycbc-test-log.XXXXXXXXXX)
...
-        else
-            echo -e "    Pass."
-        fi
+	test_result
+	if [[ `echo $prog | egrep '(pycbc_copy_output_map|pycbc_submit_dax|pycbc_stageout_failed_workflow)'` ]] ; then
+	    continue
+	fi
+	echo -e ">> [`date`] running $prog --version"
+	$prog --version &> $LOG_FILE
+	test_result
```
titodalcanton, line 48: The indentation here is still wrong. Looks like you added tabs while we should keep using spaces.

tools/pycbc_test_suite.sh
```diff
@@ -11,77 +11,63 @@ LOG_FILE=$(mktemp -t pycbc-test-log.XXXXXXXXXX)
...
     pushd examples/inspiral
     bash -e run.sh
-    if test $? -ne 0 ; then
-        RESULT=1
-        echo -e "    FAILED!"
-        echo -e "---------------------------------------------------------"
-    else
-        echo -e "    Pass."
-    fi
+    test_result
```
titodalcanton, line 64: `test_result` will try to cat `$LOG_FILE` which does not exist for this test. Maybe you could add a check in `test_result` that only cats it if it exists? That may still cat it in unwanted situations though, so maybe an argument is better?

🐈

tools/pycbc_test_suite.sh
```diff
@@ -11,77 +11,63 @@ LOG_FILE=$(mktemp -t pycbc-test-log.XXXXXXXXXX)
...
     pushd examples/tmpltbank
     bash -e testNonspin2.sh
-    if test $? -ne 0 ; then
-        RESULT=1
-        echo -e "    FAILED!"
-        echo -e "---------------------------------------------------------"
-    else
-        echo -e "    Pass."
-    fi
+    test_result
```
titodalcanton, line 70: 🐈 problem also here.

tools/pycbc_test_suite.sh
GarethCabournDavies: I added a :black_cat:`_output` global variable which gets checked whether to give log file output, and this is set to false when we aren't checking the output any more.

If we ever want to have a mix though, we need to remember to set/unset it as appropriate

titodalcanton (approved):
Looks good.

PR #4813, opened 2024-07-12
Add a script to supervise template fitting in pycbc live

GarethCabournDavies:
Adding a script, similar to the one for significance fits, to collate triggers and run fit_by_template and fit_over_multiparam on them

This is a new feature

## Motivation
This is a code so that the ranking statistics which use trigger fit files can be used in the pycbc live search. The triggers will be collected, and fit_over_multiparam files generated using a rolling window of daily fit_by_template

## Contents
(Didn't realise I'd missed this originally - added in an edit)
 - Move a lot of the shared content of this and the significance fits into a module
 - Make a new file which runs the collation and fitting: bin/live/pycbc_live_supervise_collated_triggers

A general overview of the supervision script is as follows:
 - Define which options should be read from the input config file for each supervised code
 - Make some functions which run the individual codes
 - the fit_over_multiparam function also checks for previous days' fit_by_template files if needed
A lot of this is similar to the significance fits code, so I could go through and remove duplicated effort, but it works for now

#4527, #4670, #4689, #4695, #4697

## Testing performed
Run daily (manually) so that plots are produced - see https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/pycbclive/supervise_fit_coeff_dq_files/testing/output_plots/2024/07/ for latest plots


bin/live/pycbc_live_supervise_collated_triggers
```diff
@@ -0,0 +1,432 @@
...
+        elif len(value.split()) > 1:
+            # Option is a list
+            value = value.split()
+        else:
+            # Option is a single value
+            value = value.strip()
+
+        option_values[option] = value
+
+    return option_values
```
titodalcanton, line 161: Can we rely on configparser for this functionality?

bin/live/pycbc_live_supervise_single_significance_fits
```diff
@@ -128,29 +67,8 @@ def check_trigger_files(filenames, test_options, controls):
...
-                          command_output.stderr.decode()]
-        mail_volunteers_error(
-            controls,
-            error_contents,
-            f"PyCBC live could not run {command_arguments[0]}"
-        )
-        err_msg = f"Could not run {command_arguments[0]}"
-        raise subprocess.SubprocessError(err_msg)
+        sv.mail_volunteers_error(controls, mail_body_lines,
+            'PyCBC Live single trigger fits extreme value(s)')
```
titodalcanton, line 71: Please keep the formatting as `black` would.

pycbc/live/supervision.py
```diff
@@ -0,0 +1,134 @@
...
+#
+#                                   Preamble
+#
+# =============================================================================
+#
+""" This module contains functions for optimizing the signal-to-noise ratio
+of triggers produced by PyCBC Live. Also contained within this module are the
+command line arguments required and options group for the SNR optimization.
+This module is primarily used in the pycbc_optimize_snr program.
+"""
```
titodalcanton, line 28: Looks like the docstring needs to be rewritten.

bin/live/pycbc_live_supervise_collated_triggers
titodalcanton: If not, can we at least reduce the duplication of the same code in `pycbc_live_supervise_single_significance_fits`?

bin/plotting/pycbc_plot_bank_corner
```diff
@@ -78,10 +78,6 @@ parser.add_argument("--parameters",
          "property of that parameter will be used. If not "
          "provided, will plot all of the parameters in the "
          "bank.")
-parser.add_argument("--log-parameters",
-    nargs="+",
-    default=[],
-    help="Plot these parameters on a log scale")
```
maxtrevor, line 84: Is there a reason the `log-parameters` argument was removed? I don't see how this is related to the rest of the changes in this MR

bin/live/pycbc_live_supervise_single_significance_fits
```diff
@@ -270,9 +188,9 @@ def do_fitting(args, day_dt, day_str):
 
         daily_options['gps-start-time'] = f'{gpstime.utc_to_gps(day_dt).gpsSeconds:d}'
         daily_options['gps-end-time'] = f'{gpstime.utc_to_gps(day_dt + timedelta(days=1)).gpsSeconds:d}'
-        daily_args += dict_to_args(daily_options)
+        daily_args += sv.dict_to_args(daily_options)
 
-        run_and_error(daily_args, controls)
+        sv.run_and_error(daily_args)
```
maxtrevor, line 193: Doesn't this line still need to pass `controls` ? 
`sv.run_and_error` takes two arguments

bin/live/pycbc_live_supervise_single_significance_fits
```diff
@@ -286,15 +204,15 @@ def do_fitting(args, day_dt, day_str):
             daily_plot_output,
             '--log-colormap'
         ]
-        run_and_error(daily_plot_arguments, controls)
+        sv.run_and_error(daily_plot_arguments)
```
maxtrevor, line 207: Same question about needing to pass controls

bin/live/pycbc_live_supervise_single_significance_fits
```diff
@@ -388,9 +306,9 @@ def do_fitting(args, day_dt, day_str):
     combined_options['trfits-files'] = ' '.join(trfits_files)
 
     combined_args = ['pycbc_live_combine_single_significance_fits']
-    combined_args += dict_to_args(combined_options)
+    combined_args += sv.dict_to_args(combined_options)
 
-    run_and_error(combined_args, controls)
+    sv.run_and_error(combined_args)
```
maxtrevor, line 311: same question about needing to pass `controls`

bin/live/pycbc_live_supervise_single_significance_fits
```diff
@@ -425,7 +343,7 @@ def do_fitting(args, day_dt, day_str):
         '--log-colormap'
     ]
 
-    run_and_error(combined_plot_arguments, controls)
+    sv.run_and_error(combined_plot_arguments)
```
maxtrevor, line 346: Same question about needing to pass `controls`

bin/live/pycbc_live_supervise_collated_triggers
```diff
@@ -0,0 +1,432 @@
...
+    logging.info(fit_over_args)
+    sv.run_and_error(fit_over_args, option_values)
+    if 'variable-fit-over-param' in option_values:
+        variable_fits = option_values['variable-fit-over-param'].format(ifo=ifo)
+        sv.symlink(fit_over_full, variable_fits)
+
+    return fit_over_full, file_id_str
+
+
+def do_collation_fits_dq(args, option_values, day_dt, day_str):
```
maxtrevor, line 367: why does the name of this function include `_dq` ? Is the intention that I add the dq fits into this method after this is merged?

bin/plotting/pycbc_plot_bank_corner
titodalcanton: I had missed this. We definitely want to keep this option, since a hyperbank-like bank cannot be meaningfully displayed in linear component masses or total mass.

pycbc/live/supervision.py
titodalcanton: I also suggest removing the "preamble" box above. It makes the file longer and I never understood its purpose throughout other files in PyCBC and elsewhere.

bin/plotting/pycbc_plot_bank_corner
GarethCabournDavies: This is a somewhat-related bug, as the supervision script calls this script. The option isn't actually used, if wanting a log parameter, we would have to pass "log10(mchirp)" (or similar) to `--parameters`

bin/live/pycbc_live_supervise_single_significance_fits
GarethCabournDavies: I've realised that the supervision scripts need combining into one anyway, so this will be fixed when that is done

bin/live/pycbc_live_supervise_collated_triggers
GarethCabournDavies: That was the plan. The name was intended to be: Do the collation, do the fits, and do the dq

GarethCabournDavies:
Big changes in the last batch of commits: moving to use one supervision script is the main one. 

Fixed a couple of bugs now that we are dealing with fit files which may not contain triggers from the past, which may not have information for templates which triggered today.

I think there isn't too much code duplication and have tried to log / comment / make things obvious at each step

pycbc/live/supervision.py
```diff
@@ -0,0 +1,157 @@
...
+
+def ensure_directories(control_values, day_str):
+    """
+    Ensure that the required directories exist
+    """
+    output_dir = os.path.join(
+        control_values['output-directory'],
+        day_str
+    )
+    run_and_error(['mkdir', '-p', output_dir], control_values)
```
titodalcanton, line 145: No need to spawn an external program here, we have `pycbc.makedir()`.

bin/live/pycbc_live_supervise_collated_trigger_fits
```diff
@@ -0,0 +1,626 @@
...
+        first_date = date_out
+        # reset the "missed files" counter, and add to the "found files"
+        missed_files = 0
+        found_files += 1
+        daily_files.append(daily_full)
+
+    if found_files == 0:
+        raise RuntimeError("No files found")
+
+    if missed_files == 10:
```
titodalcanton, line 193: If not too difficult, could we rather take this value from the config file?

bin/live/pycbc_live_supervise_collated_trigger_fits
```diff
@@ -0,0 +1,626 @@
...
+else:
+    # run just once
+    if args.date:
+        day_str = args.date
+        day_dt = datetime.strptime(args.date, '%Y_%m_%d')
+    else:
+        # Get the date string for yesterday's triggers
+        day_dt = datetime.utcnow() - timedelta(days=1)
+        day_dt = datetime.combine(day_dt, datetime.min.time())
+        day_str = day_dt.strftime('%Y_%m_%d')
```
titodalcanton, line 625: These four lines are repeated identically in the other branch. I suggest moving them to a function `get_yesterday_str()` and using the comment for its docstring.

bin/live/pycbc_live_supervise_collated_trigger_fits
GarethCabournDavies: I have done, but I've made it so that if it isnt given, then 10 is used as a default, as I think that is fairly sensible

PR #4814, opened 2024-07-13
fix a memory leakage bug in pycbc_brute_bank

yi-fan-wang:
https://github.com/gwastro/pycbc/pull/4803 has a memory leakage bug because, as far as I understand, `atexit` makes a reference to the multiprocessing pool so it doesn't completely terminate after each call. This PR add a method to close the pool and remove the reference. I did a test with what failed before, the memory leakage doesn't occur again.


yi-fan-wang:
@raffienficiaud Much appreciate if you can take a look!

ahnitz (approved):

PR #4820, opened 2024-07-18
PyGRB trigger clustering with short timeslides

pannarale:
This PR enables taking into account short timeslides when clustering triggers in PyGRB.  Prior to this, the clustering does not fail but it ignores what slide triggers belong to.

This is a: bug fix (in the sense that the code could run doing things incorrectly, but we were aware this enhancement had to be done and it was not a surprise).

This change affects: PyGRB

## Contents
A loop over timeslides was added, ensuring that triggers from different slides are kept separate.

The plotting scripts also need to plot information from a single slide (typically the zero-lag, i.e., unslid data): this is handled by https://github.com/gwastro/pycbc/pull/4809.

The number of slides performed, ensuring independent background events, was set in https://github.com/gwastro/pycbc/pull/4754

### DIFF
--- a/bin/pygrb/pycbc_grb_trig_cluster
+++ b/bin/pygrb/pycbc_grb_trig_cluster
@@ -173,10 +173,10 @@ all_clusters = []
 
 # load necessary information from all triggers
 
-with h5py.File(args.trig_file, "r") as h5f:
-    all_times = h5f["network"]["end_time_gc"][()]
-    all_snrs = h5f["network"][args.rank_column][()]
-    slide_ids = h5f["network"]["slide_id"][()]
+with HFile(args.trig_file, "r") as h5f:
+    all_times = h5f["network/end_time_gc"][()]
+    all_snrs = h5f[f"network/{args.rank_column}"][()]
+    slide_ids = h5f["network/slide_id"][()]
 
 # empty file (no triggers), so just copy the file
 if not all_times.size:
@@ -301,6 +301,8 @@ for slide_id in unique_slide_ids:
     logging.info(msg)
 
 logging.info('Total clustered triggers: '+str(len(all_clusters)))
+
+# -- write output --------------------------------
 
 slice_hdf5(
     args.trig_file,
--- a/bin/pygrb/pycbc_grb_trig_cluster
+++ b/bin/pygrb/pycbc_grb_trig_cluster
@@ -168,105 +168,143 @@ outfile = os.path.join(
     ),
 )
 
-# -- generate clustering bins -------------------
+# this list contains the indexing of clusters from all slides
+all_clusters = []
 
-nbins = int((end - start) // win + 1)
-bins = [[] for i in range(nbins)]
-loudsnr = numpy.zeros(nbins)
-loudtime = numpy.zeros(nbins)
-clusters = []
+# load necessary information from all triggers
 
-# -- cluster ------------------------------------
-
-with HFile(args.trig_file, "r") as h5f:
-    time = h5f["network"]["end_time_gc"][()]
-    snr = h5f["network"][args.rank_column][()]
+with h5py.File(args.trig_file, "r") as h5f:
+    all_times = h5f["network"]["end_time_gc"][()]
+    all_snrs = h5f["network"][args.rank_column][()]
+    slide_ids = h5f["network"]["slide_id"][()]
 
 # empty file (no triggers), so just copy the file
-if not time.size:
+if not all_times.size:
     shutil.copyfile(args.trig_file, outfile)
     msg = "trigger file is empty\n"
     msg += "copied input file to {}".format(outfile)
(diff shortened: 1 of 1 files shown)


bin/pygrb/pycbc_grb_trig_cluster
```diff
@@ -168,105 +168,143 @@ outfile = os.path.join(
...
+all_clusters = []
 
-nbins = int((end - start) // win + 1)
-bins = [[] for i in range(nbins)]
-loudsnr = numpy.zeros(nbins)
-loudtime = numpy.zeros(nbins)
-clusters = []
+# load necessary information from all triggers
 
-# -- cluster ------------------------------------
```
sebastiangomezlopez, line 179: Do we still need this comment ?

bin/pygrb/pycbc_grb_trig_cluster
```diff
@@ -168,105 +168,143 @@ outfile = os.path.join(
...
-
```
sebastiangomezlopez, line 180: We would not need this empty line if we remove the comment above

bin/pygrb/pycbc_grb_trig_cluster
```diff
@@ -168,105 +168,143 @@ outfile = os.path.join(
...
-with HFile(args.trig_file, "r") as h5f:
-    time = h5f["network"]["end_time_gc"][()]
-    snr = h5f["network"][args.rank_column][()]
+with h5py.File(args.trig_file, "r") as h5f:
+    all_times = h5f["network"]["end_time_gc"][()]
```
sebastiangomezlopez, line 177: This is more a matter of taste. Would you consider rewriting it as h5f["network/end_time_gc"][()]? Same for the other lines used to load hdf data.

bin/pygrb/pycbc_grb_trig_cluster
```diff
@@ -168,105 +168,143 @@ outfile = os.path.join(
...
+    all_snrs = h5f["network"][args.rank_column][()]
```
sebastiangomezlopez, line 178: In case you agree, here an f-string would do the job, h5f[f"network/{args.rank_column"}][()]

bin/pygrb/pycbc_grb_trig_cluster
```diff
@@ -168,105 +168,143 @@ outfile = os.path.join(
...
-                postfix=dict(nclusters=0), **TQDM_KW)
-for i, bin_ in enumerate(bar):
-    if not bin_:  # empty
-        continue
-
-    for idx in bin_:
-        t, s = time[idx], snr[idx]
-
-        if s < loudsnr[i]:  # not loudest in own bin
+# -- cluster ------------------------------------
```
sebastiangomezlopez, line 189: This whole block of code looks good

bin/pygrb/pycbc_grb_trig_cluster
pannarale: Yes, because this where we cluster.  I will add a comment to flag where the output is written.

bin/pygrb/pycbc_grb_trig_cluster
pannarale: `h5f[f"network/{args.rank_column}"][()]`

bin/pygrb/pycbc_grb_trig_cluster
pannarale: I did the same with the slide_id

bin/pygrb/pycbc_grb_trig_cluster
sebastiangomezlopez: Sorry, I commented on lines you removed. Now I see the comment in line 189 of your new code

bin/pygrb/pycbc_grb_trig_cluster
```diff
@@ -168,105 +168,145 @@ outfile = os.path.join(
...
+                bar.set_postfix(nclusters=nclusters)
+
+        # update things for next time
+        first = False
+        last = i == nbins - 1
+        prev += 1
+        nxt_ += 1
+
+        bar.update()
+
```
sebastiangomezlopez, line 292: Mostly rearranging lines, looks good.

bin/pygrb/pycbc_grb_trig_cluster
```diff
@@ -168,105 +168,145 @@ outfile = os.path.join(
...
-    bar.update()
+# -- write output --------------------------------
 
 slice_hdf5(
     args.trig_file,
     outfile,
-    numpy.asarray(clusters),
+    numpy.asarray(all_clusters),
     verbose=args.verbose,
 )
```
sebastiangomezlopez, line 312: Interesting approach overall. I'm ok with merging once tests are passed.

sebastiangomezlopez (approved):

PR #4825, opened 2024-07-26
changed_apply_fd_time_shift_to_apply_fseries_time_shift_in_FDomainDetFrameTwoPolNoRespGenerator

labani-01:
Correction to the "generate" function of FDomainDetFrameTwoPolNoRespGenerator - possible correction is to replace "apply_fd_time_shift" by "apply_fseries_time_shift" in the generate function of FDomainDetFrameTwoPolNoRespGenerator. 

Marginalized_time model uses FDomainDetFrameTwoPolNoRespGenerator to generate the waveform. The "generate" function of FDomainDetFrameTwoPolNoRespGenerator uses "apply_fd_time_shift" for the time domain approximant to shift it by an amount tshift(1/df - abs(hp.epoch)). The goal of this step is to shift the merger at the boundary. But if we follow the source code of "apply_fd_time_shift" then it will be clear that it will shift the waveform by (shifttime - hp.epoch) amount.

Here shiftime = tshift and hp.epoch is negative, so (shifttime - hp.epoch) = (tshift - hp.epoch) = (1/df - abs(hp.epoch) - hp.epoch) = 1/df = length of the signal. So, the fucntion "apply_fd_time_shift" is not doing any shift to the signal and giving us the same time domain waveform.

To solve this problem we can use "apply_fseries_time_shift" function (which gives a relative shift). This function is giving the correct shift to the time domain waveform by an amount tshift = 1/df - abs(hp.epoch) and the merger is at the boundary.

Necessity of the the time-shift - pycbc inference models convert waveforms in frequency domain before calculating the likelihood. marginalized_time is one of the pycbc_inference models. Time marginalization is applied after generating the waveform (We integrate over tc). Each tc is just a time-shift of the waveform. The way this model incorporates different time-shifts is by doing inverse fourier transform of the inner product of signal and data at every discrete elements of every possible time shifts. We can only talk about the fourier transform at different time-shifts if there is some convention on the signal. A good likelihood value implies high value for the inner product between data and the signal. We know the standard reference in the signal is the location of the merger. The convention is to put the merger at the beginning of the vector then every time shift will give how much the signal needs to be shifted. Frequency domain waveform gives this setup by default. But for the time domain approximant we need to apply a time shift so that merger will be at the boundary or at the start of the signal vector.

### DIFF
--- a/pycbc/waveform/.ipynb_checkpoints/Untitled-checkpoint.ipynb
+++ b/pycbc/waveform/.ipynb_checkpoints/Untitled-checkpoint.ipynb
@@ -1,6 +1,0 @@
-{
- "cells": [],
- "metadata": {},
- "nbformat": 4,
- "nbformat_minor": 5
-}
--- a/pycbc/waveform/.ipynb_checkpoints/Untitled-checkpoint.ipynb
+++ b/pycbc/waveform/.ipynb_checkpoints/Untitled-checkpoint.ipynb
@@ -1,0 +1,6 @@
+{
+ "cells": [],
+ "metadata": {},
+ "nbformat": 4,
+ "nbformat_minor": 5
+}
--- a/pycbc/waveform/generator.py
+++ b/pycbc/waveform/generator.py
@@ -38,7 +38,7 @@ from pycbc import transforms
 from pycbc.types import TimeSeries
 from pycbc.waveform import parameters
 from pycbc.waveform.utils import apply_fseries_time_shift, taper_timeseries, \
-                                 ceilpow2, apply_fd_time_shift  
+                                 ceilpow2, apply_fd_time_shift
 from pycbc.detector import Detector
 from pycbc.pool import use_mpi
 import lal as _lal
--- a/pycbc/waveform/generator.py
+++ b/pycbc/waveform/generator.py
@@ -37,7 +37,7 @@ from . import waveform_modes
 from pycbc import transforms
 from pycbc.types import TimeSeries
 from pycbc.waveform import parameters
-from pycbc.waveform.utils import apply_fd_time_shift, taper_timeseries, \
+from pycbc.waveform.utils import apply_fseries_time_shift, taper_timeseries, \
                                  ceilpow2
 from pycbc.detector import Detector
 from pycbc.pool import use_mpi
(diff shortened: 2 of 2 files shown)


cdcapano:
If the tshift isn't being calculated correctly (I leave @ahnitz to verify). I think the correct fix would be to still call `apply_fd_time_shift` with a modified `tshift` rather than call `apply_fseries_time_shift`. `apply_fseries` assumes that the input array is equally spaced in frequency. That might always be the case here since this is used by the time marginalization model, but not sure. `apply_fd_time_shift` can handle both, albeit needing to take times rather than time shifts (something I regret doing, but would be too difficult to change now).

ahnitz:
> If the tshift isn't being calculated correctly (I leave @ahnitz to verify). I think the correct fix would be to still call `apply_fd_time_shift` with a modified `tshift` rather than call `apply_fseries_time_shift`. `apply_fseries` assumes that the input array is equally spaced in frequency. That might always be the case here since this is used by the time marginalization model, but not sure. `apply_fd_time_shift` can handle both, albeit needing to take times rather than time shifts (something I regret doing, but would be too difficult to change now).

It will be equally spaced actually, so I don't think that's a difference here.

ahnitz:
@cdcapano I'll be out of contact, so can @cdcapano you review this after @labani-01 has updated it to address the issues.

pycbc/waveform/.ipynb_checkpoints/Untitled-checkpoint.ipynb
```diff
@@ -0,0 +1,6 @@
+{
+ "cells": [],
+ "metadata": {},
+ "nbformat": 4,
```
ahnitz, line 4: Be careful not to commit files you don't intend to.

labani-01: If we don't change the generate.py script (keeping apply_fd_time_shift in the FDomainDetFrameTwoPolNoRespGenerator) and want to get the time domain waveform using a time domain approximant 'SEOBNRv4', then it will give the plot below. 
```
params={'mass1' : 36.09431175153813,
'mass2' : 33.61286267925508,
'distance' :  500.60060060060056} 
from pycbc.waveform.generator import FDomainDetFrameTwoPolNoRespGenerator
from pycbc.waveform.generator import TDomainCBCGenerator
generator = FDomainDetFrameTwoPolNoRespGenerator(TDomainCBCGenerator, 0, variable_args=['mass1', 'mass2', 'distance'],  delta_f=1./68, delta_t=1/2048, f_lower=20., approximant='SEOBNRv4')
no_det = generator.generate(**params) #frequency domain waveform
print(no_det['RF'])
plt.plot(no_det['RF'][0])
plt.show()
``` 
add_fd_frequency

```
plt.plot(no_det['RF'][0].to_timeseries())
plt.show()
```
add_fd_timedomain
If we notice the time domain plot then it is clear that merger is not shifted at the beginning of the waveform or the merger is not at the boundary.

But if we do the required changes (changing apply_fd_time_shift to  apply_fseries_time_shift in the FDomainDetFrameTwoPolNoRespGenerator) then it is giving the expected nature of the time domain plot. 

add_fseries_frequency - this is the frequency domain waveform after the change.

add_fseries_timedomain
- this is the expected time domain plot which we are getting after doing the above mentioned change.

ahnitz (approved):
@labani-01 Thanks. I think this looks good now. Assuming the tests pass, this may now be merged.

PR #4826, opened 2024-07-28
qtransform interpolation should be linear to avoid negative values

ahnitz:
In a previous update scipy removed interp2d. Per the guide, we moved to rectbivariatespline, however, this meant that the default changed from linear to cubic interpolation. This isn't ideal for qtransform plots due to the fact that it results in negative values. These show up as "snow" when plotting the amplitude in results pages for example. e.g. 
data

This was repoted by @yi-fan-wang and others.


GarethCabournDavies (approved):
Makes sense, solves the problem. Approved

PR #4827, opened 2024-07-29
Update pycbc_live to allow debug logging

GarethCabournDavies:
Allow debug logging in pycbc live. To get debug logging, we want args.verbose to use 'count', as in the 'common_options' interface.


titodalcanton (approved):
Minimal change, looks good.

Could you just rebase to get the latest CI?

PR #4833, opened 2024-07-31
Set back to development

GarethCabournDavies:


titodalcanton (approved):

GarethCabournDavies:
Failures look genuine. I think it could be that #4834 is needed, so I will include that in this PR

titodalcanton:
If BBHx requires Python 3.12, we cannot use it for the tests running on older Pythons.

GarethCabournDavies:
> If BBHx requires Python 3.12, we cannot use it for the tests running on older Pythons.

ugh, of course. I will revert

PR #4836, opened 2024-08-01
PyCBC Live: fix errors from stat refresh code when not being used

titodalcanton:
#4816 leads to an error if `--statistic-refresh-rate` is not used. The error causes the refresh threads to crash, but this is not detected in the CI tests, which only check the final results.

This is mostly a bug fix but includes a few minor improvements and simplifications.

This change affects the live search.

## Motivation

The core of the error is this message (possibly mangled due to multithreading, sorry):
```
  File "…/lib/python3.9/site-packages/pycbc/events/coinc.py", line 1427, in refresh_statistic
    if since_stat_refresh > self.statistic_refresh_rate:
TypeError: '>' not supported between instances of 'float' and 'NoneType'
```

## Contents

Adds a check for None, cleans up a few logging messages, improves an argparse help string, uses simpler timing functions.

Improves #4816.

## Testing performed

Error discovered by staring at the Live example while it runs. Testing performed in the same way.

### DIFF
--- a/bin/pycbc_live
+++ b/bin/pycbc_live
@@ -1049,7 +1049,8 @@ parser.add_argument('--psd-variation', action='store_true',
                          "the search. Required when using a single detector "
                          "ranking statistic that includes psd variation.")
 parser.add_argument("--statistic-refresh-rate", type=float,
-                    help="How often to refresh the statistic object")
+                    help="How often to refresh the statistic object, "
+                         "in seconds. If omitted, no refreshing is done.")
 
 scheme.insert_processing_option_group(parser)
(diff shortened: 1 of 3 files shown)


titodalcanton:
As discussed on Slack, I do not see why using threads was really necessary in #4816. I think the same behavior and very similar performance would be obtained by doing the refresh inside the `get_coinc()` function in the main `pycbc_live` executable, just before calling the `add_singles()` method (or perhaps even directly inside that method). Maybe worth some testing…

GarethCabournDavies (approved):

PR #4846, opened 2024-08-06
Fix small typo in pycbc_live_supervise_collated_trigger_fits

GarethCabournDavies:


bin/live/pycbc_live_supervise_collated_trigger_fits
```diff
@@ -569,7 +569,7 @@ def get_yesterday_date():
     day_dt = datetime.utcnow() - timedelta(days=1)
     day_dt = datetime.combine(day_dt, datetime.min.time())
     day_str = day_dt.strftime('%Y_%m_%d')
-    return date_dt, date_str
+    return day_dt, date_str
```
maxtrevor, line 572: Also need to change `date_str` to `day_str`

GarethCabournDavies:
```suggestion
    return day_dt, day_str
```

maxtrevor (approved):

Issue #4267, opened 2023-02-16
Add a sky location distribution that draws from a HEALPix skymap


titodalcanton:
I would like to take care of this with some students here.

titodalcanton:
Addressed by #4848.

PR #4848, opened 2024-08-08
add a distribution to draw sky locations from a HEALPix map

JulienGreuter:
Add a class HealpixSky to `pycbc/pycbc/distributions/sky_location.py` that allow one to sample the distribution given by a HEALPix sky map by using the rejection sampling method.

This change follows style guidelines (See e.g. [PEP8](https://peps.python.org/pep-0008/)) and includes running `black` to reformat `sky_location.py`.

This change adds a new feature to the distributions module, mainly for use in PyGRB.

## Motivation

Versions of PyGRB used for past analyses approximated the uncertainty in the GRB sky location with circular patches, or using "IPN boxes". In many cases, the uncertainties actually are not circular at all. This is typically the case for IPN localizations, which can resemble thin rings over large patches of the sky. Nowadays, most GRBs have their sky location uncertainties represented as complete probability densities over the sky, encoded via the HEALPix scheme, and stored in FITS files. This change will allow PyGRB to directly read those FITS files, without making approximations.

## Contents

The class HealpixSky samples the distribution given by a HEALPix map by using the rejection sampling method. 
To have an practical acceptance rate even with precise localizations, a bounding rectangle in (alpha,delta) is first determined as a preprocessing step, which encloses a given amount of probability (specified by the `coverage` parameter). In order to do this in a simple way, the map is temporarily rasterized to a resolution specified by `rasterization_nside`. The choice of this parameter is not very critical, it mainly affects the run time. It is probably possible to modify the algorithm so that this step is not necessary, and the `coverage` and `rasterization_nside` can be dropped. This is something to be looked at in a future pull request. Anyway, the present method seems to work, providing 10000 samples in order of tens of seconds max for most IPN maps.

In order to use this distribution with `pycbc_create_injections`, the corresponding `.ini` file must contain a section that looks like this:
```
[prior-ra+dec]
name = healpix_sky
; Path to the skymap file in FITS format. Can be gzipped (.fits.gz).
healpix_file =  /path/to/skymap.fits
; Must be a power of 2, 128 seems to works best in terms of run time.
; Optional parameter, internal default set to 64.
rasterization_nside =  128
; Must be between 0 and 1, 1 meaning that you consider all the space where the data is non zero.
; Optional parameter, internal default set to 0.9999.
coverage = 0.999
```

This addresses #4267.

## Testing performed

Test on the following GRBs are to be done to check accuracy and speed:
```
GRB211124A_IPN_map_hpx.fits.gz
GRB220203A_IPN_map_hpx.fits.gz
GRB220209A_IPN_map_hpx.fits.gz
GRB220211A_IPN_map_hpx.fits.gz
GRB220711C_IPN_map_hpx.fits.gz
GRB220905A_IPN_map_hpx.fits.gz
GRB220927A_IPN_map_hpx.fits.gz
GRB221022C_IPN_map_hpx.fits.gz
GRB221124A_IPN_map_hpx.fits.gz
GRB221226A_IPN_map_hpx.fits.gz
GRB230511D_IPN_map_hpx.fits.gz
GRB230518A_IPN_map_hpx.fits.gz
GRB230625B_IPN_map_hpx.fits.gz
GRB230712A_IPN_map_hpx.fits.gz
GRB230712C_IPN_map_hpx.fits.gz
GRB230715D_IPN_map_hpx.fits.gz
GRB231018A_IPN_map_hpx_moc.fits.gz
GRB231224A_IPN_map_hpx_moc.fits.gz
GRB240119A_IPN_map_hpx_moc.fits.gz
GRB240124B_IPN_map_hpx.fits.gz
GRB240125A_IPN_map_hpx_moc.fits.gz
GRB240125B_IPN_map_hpx_moc.fits.gz
GRB240205A_IPN_map_hpx_moc.fits.gz
GRB240209A_IPN_map_hpx_moc.fits.gz
GRB240410A_IPN_map_hpx.fits.gz
GRB240504A_IPN_map_hpx_moc.fits.gz
GRB240514B_IPN_map_hpx_moc.fits.gz
GRB240529A_IPN_map_hpx_moc.fits.gz
GRB240531A_IPN_map_hpx_moc.fits.gz
GRB240603A_IPN_map_hpx_moc.fits.gz
GRB240607A_IPN_map_hpx_moc.fits.gz
GRB240615A_IPN_map_hpx_moc.fits.gz
GRB240628B_IPN_map_hpx_moc.fits.gz
GRB240715A_IPN_map_hpx_moc.fits.gz
```

## Additional notes

This code requires mhealpy 0.3.4 or later. Previous versions have a bug that prevents a function from working.

### DIFF
--- a/pycbc/distributions/sky_location.py
+++ b/pycbc/distributions/sky_location.py
@@ -191,13 +191,13 @@ class HealpixSky:
     healpix_file : str
         path to a fits file containing probability distribution encoded in a 
         HEALPix scheme
-    coverage : {float
+    coverage : {float, 0.9999}
         percentage of the map covered by the method
-        default value : coverage = 99.99
-    rasterisation_nside : int
+        
+    rasterization_nside : int
         nside of the rasterized map used to determine the 
         boundaries of the input map.
-        default value : rasterisation_nside = 64
+        default value : rasterization_nside = 64
     """
     name = 'healpix_sky'
     _params = ['ra', 'dec']
@@ -207,8 +207,8 @@ class HealpixSky:
         import mhealpy
         
         #give the boundaries of a map m in radian, and following the 
-        #ra-dec convention delta in [-pi/2, pi/2] alpha in [0,2pi]
-        def boundaries(file_name,nside,X): 
+        #radec convention delta in [-π/2, π/2] alpha in [0,2π]
+        def boundaries(file_name,nside,coverage): 
             m = mhealpy.HealpixMap.read_map(file_name)
             
             delta_max= -numpy.pi/2
@@ -216,10 +216,10 @@ class HealpixSky:
             alpha_max= 0
             alpha_min = 2*numpy.pi
             
-            rasterize_map = m.rasterize(scheme = 'NESTED',nside = min(nside,m.nside)) #min si jamais nside < m.nside pour eviter pb # scheme pas important ? RING par default
-            data = rasterize_map.data
+            rasterized_map = m.rasterize(scheme = 'NESTED',nside = min(nside,m.nside)) #min si jamais nside < m.nside pour eviter pb # scheme pas important ? RING par default
+            data = rasterized_map.data
             non_zero_data = data[data != 0]
-            A = non_zero_data.sum() # cst de renormalisation
+            A = non_zero_data.sum() # renormalization_constant
             
             normalized_data = data/A
             sort_normalized_data = - numpy.sort(-non_zero_data/A) # TRI DECROISSANT DE non_zero_data
@@ -232,11 +232,11 @@ class HealpixSky:
             #print(f'{N} pixels non nuls')
             for i in range(N):
                 j = numpy.argmax(normalized_data == sort_normalized_data[i]) 
-                pix[i] = rasterize_map.uniq[j]-4*nside*nside
+                pix[i] = rasterized_map.uniq[j]-4*nside*nside
                 delta[i],alpha[i] =  healpy.pix2ang(
                     nside = nside,
                     ipix = pix[i],
-                    nest = rasterize_map.is_nested,
+                    nest = rasterized_map.is_nested,
                     lonlat = False
                 )
                 delta[i] = numpy.pi/2 - delta[i] # conversion d'angle colatitude -> latitude 
@@ -245,7 +245,7 @@ class HealpixSky:
                 delta_max,delta_min = max(delta_max,delta[i]),min(delta_min,delta[i])
                 alpha_max,alpha_min = max(alpha_max,alpha[i]),min(alpha_min,alpha[i])
                     
-                if S > X/100 : 
+                if S > coverage : 
                     break
             #ajout de la securité : taille_pix ~ np.pi/4nside-1
             secu = 2*numpy.pi/(4*nside -1)
@@ -262,22 +262,22 @@ class HealpixSky:
         if 'coverage' in params:
             coverage = params['coverage']
         else :
-            coverage = 99.99    
-        if coverage > 100 or coverage < 0 : #mettre entre 0 et 1 ? => modif de boundaries()
+            coverage = 0.9999    
+        if coverage > 1 or coverage < 0 : 
             raise ValueError(
-                'Coverage must be between 0% and 100%'
+                'Coverage must be between 0 and 1'
                 )
-        if 'rasterisation_nside' in params:
-            rasterisation_nside = params['rasterisation_nside']
+        if 'rasterization_nside' in params:
+            rasterization_nside = params['rasterisation_nside']
         else :
-            rasterisation_nside = 64 # or 128
-        if rasterisation_nside not in [int(2**n) for n in range(20)]: #math.log2(nside).is_integer()
+            rasterization_nside = 64 # or 128
+        if rasterization_nside not in [int(2**n) for n in range(20)]: #math.log2(nside).is_integer()
             raise ValueError( #bin(nside).count('1') == 1
                 'Nside must be in [1,2,4,8,16,32,64,128,256,512,1024, ...],'
                 '(preferably between 16 and 256 for better efficiency)'
                 )
         self.healpix_map = mhealpy.HealpixMap.read_map(file_name)
-        self.boundaries = boundaries(file_name,rasterisation_nside,coverage) #nside = 64 ou 128 utiliser self.m à la place de file_name ?
+        self.boundaries = boundaries(file_name,rasterization_nside,coverage) #nside = 64 ou 128 utiliser self.m à la place de file_name ?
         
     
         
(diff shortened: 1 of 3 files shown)


pycbc/distributions/sky_location.py
```diff
@@ -161,18 +159,246 @@ def rvs(self, size):
...
+            data = rasterized_map.data
+            renormalization_constant = data.sum()
+
+            normalized_data = data / renormalization_constant
+            sorter = numpy.argsort(-normalized_data)
+
+            map_coverage = 0
+
+            for i in range(len(data)):
+                # j = numpy.argmax(normalized_data == sort_normalized_data[i])
```
titodalcanton, line 248: Let's drop this comment.

pycbc/distributions/sky_location.py
```diff
@@ -161,18 +159,246 @@ def rvs(self, size):
...
+                j = sorter[i]
+                pix = rasterized_map.uniq[j] - 4 * nside * nside
+                delta, alpha = rasterized_map.pix2ang(pix, lonlat=False)
+                # converts colatitude to latitude
+                delta = numpy.pi / 2 - delta
+
+                # map_coverage += sort_normalized_data[i]
```
titodalcanton, line 255: Drop this comment too.

titodalcanton:
Reporting a summary of the tests that @JulienGreuter did on a variety of GRB IPN skymaps.

Drawing 100000 samples takes between 5 and 300 s depending on the properties of the input map (uniform resolution vs MOC, maximum resolution, morphology). The slowest cases appear to be dominated by the code that determines the bounding rectangle at initialization. We can probably optimize this later if needed.

To test the correctness of the method, we plotted the number of samples that fall in each pixel as a function of the probability contained in that pixel. This should show a linear trend with slope equal to the total number of samples, and a scatter given by the Poisson distribution. Some examples follow.

image

image

image

image

In a couple cases, the result is not that good:

image

image

This happens because the HealpixSky class uses linear interpolation to get the probability density between pixels, while the test plotting code assumes that the probability density is constant across each pixel. It turns out that a few IPN maps are sampled with a relatively coarse resolution, comparable to the angular scale of the variation of the probability density, and this makes the distinction between the two codes important. In most cases, however, the pixels are small enough that this distinction is irrelevant.

ahnitz:
@JulienGreuter @titodalcanton 

Out of curiosity, is there a reason that rejection sampling is used here? Given that healpix format is a discretization of the sky space, one should be able to do a simple inverse cdf draw by treating it as a 1-d space over the pixel id. That would be both faster, and would allow one do say use a healpix map as a prior.

titodalcanton:
@ahnitz good question. We started with rejection sampling for simplicity, and we found (perhaps surprisingly) that it works sufficiently well for sampling. I am not sure how available Julien will be for further contribution, but I do have an alternative implementation that does as you suggest. It is somewhat simpler (in the sense that it does not take additional parameters apart from the file path) and, *for some skymaps*, much faster (much slower for others, however). I would like to merge this one first to acknowledge Julien's work, and then I will probably open a new PR with the new algorithm.

pycbc/distributions/sky_location.py
```diff
@@ -161,18 +159,244 @@ def rvs(self, size):
...
+
+    name = 'healpix_sky'
+    _params = ['ra', 'dec']
+
+    def __init__(self, **params):
+        import mhealpy
+
+        def boundaries(healpix_map, nside, coverage):
+            """Boundaries of the part of the celestial sphere which we are
+            looking to do the distribution.
```
pannarale, line 202:
```suggestion
            """Boundaries of the portion of the celestial sphere where we
             set up the distribution.
```

pycbc/distributions/sky_location.py
```diff
@@ -161,18 +159,244 @@ def rvs(self, size):
...
+            return delta_min, delta_max, alpha_min, alpha_max
+
+        file_name = params['healpix_file']
+
+        coverage = params['coverage']
+
+        if coverage > 1 or coverage < 0:
+            raise ValueError(
+                f'Coverage must be between 0 and 1, {coverage} is not correct'
+            )
```
pannarale, line 285: Consider checking this right away, at the start of the function.

pycbc/distributions/sky_location.py
```diff
@@ -161,18 +159,244 @@ def rvs(self, size):
...
+
+        rasterization_nside = params['rasterization_nside']
+
+        if bin(rasterization_nside).count('1') != 1:
+            raise ValueError(
+                f'Rasterization_nside must be a power of 2,'
+                f'{rasterization_nside} is not correct'
+            )
```
pannarale, line 293: And this check, too, may be anticipated.

pycbc/distributions/sky_location.py
```diff
@@ -161,18 +159,244 @@ def rvs(self, size):
...
+    def from_config(cls, cp, section, variable_args):
+        tag = variable_args
+        variable_args = variable_args.split(VARARGS_DELIM)
+        if set(variable_args) != set(cls._params):
+            raise ValueError(
+                "Not all parameters used by this distribution "
+                "included in tag portion of section name"
+            )
+        healpix_file = str(cp.get_opt_tag(section, 'healpix_file', tag))
+        coverage = 0.9999
```
pannarale, line 314: This value and the one for `rasterization_nside` should be reported in the class documentation.

pycbc/distributions/sky_location.py
```diff
@@ -161,18 +159,244 @@ def rvs(self, size):
...
+            healpix_file=healpix_file,
+            coverage=coverage,
+            rasterization_nside=rasterization_nside,
+        )
+
+    def rvs(self, size):
+        def simple_rejection_sampling(healpix_map, size, boundaries):
+            """Start from a uniform distribution of points, and accepts those
+             whose values on the map are greater than a random value
+             following a uniform law
```
pannarale, line 333:
```suggestion
            """Starts from a uniform distribution of points, and accepts those
            whose values on the map are greater than a random value
            following a uniform law.
```

pycbc/distributions/sky_location.py
```diff
@@ -161,18 +159,244 @@ def rvs(self, size):
...
+             healpix_map  : HealpixMap instance
+             size : int
+                 number of points tested by the method
+             boundaries : list -> tuple of 4 floats
+                 delta_min,delta_max,alpha_min,alpha_max = boundaries
+                 delta is the declination in radians [-pi/2, pi/2]
+                 alpha is the right ascention in radians [0,2pi]
+
+             Returns
+             -------
```
pannarale, line 346: Something is funny with the indentation here. Perhaps this is what codeclimate raised, but I haven't checked.

pannarale (approved):
This looks good to me.  I do not know how easy it is to apply the suggeted changes to this fork of pycbc, and I know that @titodalcanton has other updates to submit on this topic, so I'm happy if this gets merged and improved in an upcoming PR.

titodalcanton:
Thanks @pannarale. Some of your comments above will become moot with my next pull request on this. Those that remain valid will be addressed there.

PR #4849, opened 2024-08-09
pycbc_plot_bank_corner improvements for live

GarethCabournDavies:
Improvements are needed for the fits plotting.
This allows log ranges to be used for all parameters including the colorbar parameter, and also adds the IFO to the title of the plot when it is called by the live supervisor

This is a new feature
This change affects: plotting from the live search

## Motivation
Before these changes, we would have to use log(param) in the --parameters argument. This leads to unwieldy labels and non-intuitive axis tick values

## Contents
 - Add two options to pycbc_plot_bank_corner for log parameters and log-colormap
 - check that the log-parameters are sensible
 - Modify the logic for extending the axes so that if it is on a log scale, this will not break things by becomeing a negative value
 - pass the colorbar argument into `create_multidim_plot`, and use manual normalisation objects rather than the default linear norm
 - set semilogy and semilogx on the axis dictionary values when appropriate
 - I've also added the IFO into the plot title when it is called by the supervisor script.

## Testing performed
See https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/fit_plots_improvements/L1-FIT_OVER_PARAM-2024_08_07.png for a plot which uses log scale in x axis and linear elsewhere.
https://ldas-jobs.ligo.caltech.edu/~gareth.cabourndavies/testoutput/fit_plots_improvements/L1-FIT_BY_TEMPLATE-2024_08_07.png uses log scale on x and y, and also on the color scale
Other plots in that directory show:
 - where 2 parameters are used, with a histogram (special case where the bottom right histogram plot is rotated 90 degrees)
 - where 3 parameters are used, with a histogram
 - where no log-scaled parameters are used

### DIFF
--- a/bin/live/pycbc_live_supervise_collated_trigger_fits
+++ b/bin/live/pycbc_live_supervise_collated_trigger_fits
@@ -270,7 +270,10 @@ def plot_fits(
     ]
     fits_plot_arguments += sv.dict_to_args(plot_fit_options)
 
-    title = "Fit parameters for pycbc-live, triggers from " + day_title_str
+    title = "Fit parameters for pycbc-live, triggers from {}, {}".format(
+        ifo,
+        day_title_str
+    )
     if smoothed == True:
         title += ', smoothed'
     fits_plot_arguments += ['--title', title]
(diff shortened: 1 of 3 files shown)


maxtrevor (approved):
Looks good to me

PR #4850, opened 2024-08-09
Idq live

maxtrevor:

This update allows the DQExpFitFgBgNormStatistic ranking statistic to be used in the low-latency PyCBC Live Search

This change primarily affects code used in the live search. No difference is expected in the Live search unless the DQExpFitFgBgNormStatistic is chosen. The same statistic code is also used in the offline search, but this update should not change the results of using the DQ stat in offline.

This change changes scientific output of the low-latency search if DQExpFitFgBgNormStatistic is selected as the ranking statistic.

## Contents
While this change allows the DQExpFitFgBgNormStatistic to be used in low-latency, it does not include code changes needed to produce a new statistic file for that stat to use.

I plan to make a separate merge request in the next few days building on https://github.com/gwastro/pycbc/pull/4813 that will allow a new statistic file to be produced and read in daily.

## Testing performed
Ran on the O3 replay MDC for a day


maxtrevor:
Noting that this is the completed version of a previous draft MR: https://github.com/gwastro/pycbc/pull/4818

bin/pycbc_live
```diff
@@ -1310,18 +1312,27 @@ with ctx:
                     for key in results[ifo]:
                         if len(results[ifo][key]):
                             results[ifo][key] = results[ifo][key][idx]
-                if data_reader[ifo].idq is not None:
-                    logging.info("Checking %s's iDQ information", ifo)
+                if data_reader[ifo].idq:
```
GarethCabournDavies, line 1315: Why remove the 'is not None'?

pycbc/frame/frame.py
```diff
@@ -906,16 +906,14 @@ def indices_of_flag(self, start_time, duration, times, padding=0):
         duration: int
             Number of seconds to check.
         padding: float
-            Number of seconds to add around flag inactive times to be considered
-        inactive as well.
+            Amount of time in seconds to flag around glitchy times
```
GarethCabournDavies, line 909: glitchy doesnt feel right - maybe `Amount of time in seconds to extend around idq-flagged times` or something similar?

pycbc/frame/frame.py
```diff
@@ -929,8 +927,10 @@ def indices_of_flag(self, start_time, duration, times, padding=0):
         glitch_times = stamps[glitch_idx]
         starts = glitch_times - padding
         ends = starts + 1.0 / sr + padding * 2.0
-        idx = indices_outside_times(times, starts, ends)
-        return idx
+        idx = indices_within_times(times, starts, ends)
+        out = numpy.zeros(len(times), dtype=bool)
```
GarethCabournDavies, line 931: `out` isn't super informative, maybe 'flagged_bool'

pycbc/events/stat.py
```diff
@@ -2437,24 +2461,35 @@ def lognoiserate(self, trigs):
...
         dq_rate = numpy.maximum(dq_rate, 1)
 
         logr_n = ExpFitFgBgNormStatistic.lognoiserate(
                     self, trigs)
         logr_n += numpy.log(dq_rate)
         return logr_n
 
+    def single(self, trigs):
+        # make sure every trig has a dq state
+        try:
```
GarethCabournDavies, line 2474: I think an `if hasattr(self, 'ifo')` would be nicer here than the try/except. However I know that is against EAFP, and this is how it is done elsewhere

pycbc/frame/frame.py
GarethCabournDavies, line 909: `times` doesn't have an entry in the docstring

pycbc/frame/frame.py
```diff
@@ -906,16 +906,14 @@ def indices_of_flag(self, start_time, duration, times, padding=0):
...
-        indices: numpy.ndarray
-            Array of indices marking the location of triggers within valid
-        time.
+        flag_state: numpy.ndarray
+            Boolean array of whether flag was on at given times
         """
-        from pycbc.events.veto import indices_outside_times
+        from pycbc.events.veto import indices_within_times
         sr = self.idq.raw_buffer.sample_rate
         s = int((start_time - self.idq.raw_buffer.start_time - padding) * sr) - 1
```
GarethCabournDavies, line 918: If you could add comments here as to what each block of a few lines is doing, that would help to understand. e.g. 
```
# Convert start time into sample number for the raw idq buffer
...
# Get the appropriate idq state for the start time and duration given
...
# Get the times when idq false alarm probability is below the threshold
...
# Apply padding around the glitch times
...
# Work out the indices of trigger times which fall into the padded flags
```

pycbc/frame/frame.py
GarethCabournDavies: (This preference may be due to having watched too much [Go-Jetters](https://gojetters.fandom.com/wiki/Grandmaster_Glitch))

pycbc/events/stat.py
```diff
@@ -2368,35 +2395,32 @@ def update_file(self, key):
...
-                ''.join(self.ifos),
+                ifo,
                 key
             )
-            self.assign_dq_rates(key)
-            self.assign_template_bins(key)
-            self.setup_segments(key)
+            self.dq_rates_by_state[ifo] = self.assign_dq_rates(key)
+            self.dq_bin_by_tid[ifo] = self.assign_template_bins(key)
             return True
```
GarethCabournDavies, line 2406: Thanks for sorting this!

pycbc/events/stat.py
```diff
@@ -2368,35 +2395,32 @@ def update_file(self, key):
...
+            ifo = trigs.get('ifo', None)
+            if ifo is None:
+                ifo = self.ifos[0]
+            assert ifo in self.ifos
 
-        dq_val = numpy.zeros(len(dq_state))
+        dq_state = trigs['dq_state']
+        dq_val = numpy.ones(len(dq_state))
 
+        tnum = self.curr_tnum
```
GarethCabournDavies, line 2423: I just looked and think its right that this is _always_ run after `curr_tnum` is initialised, but it would be good to confirm

pycbc/events/stat.py
```diff
@@ -2368,35 +2395,32 @@ def update_file(self, key):
...
-    def find_dq_noise_rate(self, trigs, dq_state):
+    def find_dq_noise_rate(self, trigs):
         """Get dq values for a specific ifo and dq states"""
 
-        try:
-            tnum = trigs.template_num
-        except AttributeError:
-            tnum = trigs['template_id']
-
         try:
```
GarethCabournDavies, line 2412: This try/except is used a couple of times now. It would be good to be in a function. Codeclimate will probably complain

pycbc/events/stat.py
```diff
@@ -2271,14 +2271,43 @@ def __init__(self, sngl_ranking, files=None, ifos=None,
...
+        Returns
+        -------
+        None
+        """
+        ifo = key.split('-')[0]
+        with h5py.File(self.files[key], 'r') as dq_file:
+            ifo_grp = dq_file[ifo]
+            if 'dq_segments' in ifo_grp.keys():
+                # if segs are in stat file, we are not in LL
+                assert not self.low_latency, 'Should not have segments in LL'
```
GarethCabournDavies, line 2305: Style grumble: I don't really like the asserts here, they are more for debugging, and would rather have a `raise ValueError('Low-latency DQ statistic should not have DQ state segments')` within more if statements. But that is stylistic and not too important

GarethCabournDavies (changes requested):
Minor changes, mostly to do with style/conventions. Functionally this looks like it does what I would expect

pycbc/events/stat.py
maxtrevor: This was why I moved where `self.curr_tnum` was set to earlier within `ExpFitFgBgNormStatistic.single` method.
Otherwise I was getting an error that `self.curr_tnum` was none the first time this line was reached

pycbc/events/stat.py
maxtrevor: No problem!

bin/pycbc_live
maxtrevor: I think this was accidental but didn't actually break anything

bin/pycbc_live
```diff
@@ -1311,17 +1313,26 @@ with ctx:
...
                             results[ifo][key] = results[ifo][key][idx]
                 if data_reader[ifo].idq is not None:
-                    logging.info("Checking %s's iDQ information", ifo)
+                    logging.info("Reading %s's iDQ information", ifo)
                     start = data_reader[ifo].start_time
                     times = results[ifo]['end_time']
-                    idx = data_reader[ifo].idq.indices_of_flag(
+                    flag_active = data_reader[ifo].idq.flag_at_times(
                             start, valid_pad, times,
                             padding=data_reader[ifo].dq_padding)
```
GarethCabournDavies, line 1321:
```suggestion
                            padding=data_reader[ifo].dq_padding
                        )
```

bin/pycbc_live
```diff
@@ -1311,17 +1313,26 @@ with ctx:
...
-                    logging.info('Keeping %d/%d %s triggers after iDQ',
-                                 len(idx), len(times), ifo)
-                    for key in results[ifo]:
-                        if len(results[ifo][key]):
-                            results[ifo][key] = results[ifo][key][idx]
+
+                    if args.idq_reweighting:
+                        logging.info('iDQ flagged %d/%d %s triggers',
+                                     numpy.sum(flag_active), len(times), ifo)
```
GarethCabournDavies, line 1325:
```suggestion
                        logging.info(
                            'iDQ flagged %d/%d %s triggers',
                            numpy.sum(flag_active),
                            len(times),
                            ifo
                        )
```

bin/pycbc_live
```diff
@@ -1311,17 +1313,26 @@ with ctx:
...
+                        results[ifo]['dq_state'] = flag_active.astype(int)
+                    else:
+                        # use idq as a veto
+                        keep = numpy.logical_not(flag_active)
+                        logging.info('Keeping %d/%d %s triggers after iDQ',
+                                     numpy.sum(keep), len(times), ifo)
```
GarethCabournDavies, line 1331:
```suggestion
                        logging.info(
                            'Keeping %d/%d %s triggers after iDQ',
                            numpy.sum(keep),
                            len(times),
                            ifo
                        )
```

pycbc/events/stat.py
```diff
@@ -2271,14 +2271,44 @@ def __init__(self, sngl_ranking, files=None, ifos=None,
...
+        None
+        """
+        ifo = key.split('-')[0]
+        with h5py.File(self.files[key], 'r') as dq_file:
+            ifo_grp = dq_file[ifo]
+            if 'dq_segments' not in ifo_grp.keys():
+                # if segs are not in file, we must be in LL
+                if self.dq_state_segments is not None:
+                    raise ValueError(
+                        'Either all dq stat files must have segments or none')
```
GarethCabournDavies, line 2307:
```suggestion
                    raise ValueError(
                        'Either all dq stat files must have segments or none'
                    )
```

pycbc/events/stat.py
```diff
@@ -2271,14 +2271,44 @@ def __init__(self, sngl_ranking, files=None, ifos=None,
...
+                self.low_latency = True
+            elif self.low_latency:
+                raise ValueError(
+                    'Either all dq stat files must have segments or none')
```
GarethCabournDavies, line 2311:
```suggestion
                raise ValueError(
                    'Either all dq stat files must have segments or none'
                )
```

GarethCabournDavies (approved):
Minor comments re: formatting, but I think this is good otherwise

titodalcanton:
Please rebase on master to pick up #4855 and get the tests to pass.

PR #4863, opened 2024-09-03
Try forcing installing list pypmc branch

ahnitz:


ahnitz:
TODO

[] - fix the virtualenv build as this will now need to build ligo.skymap from source (and hence needs gsl)
[] - update the documentation once PRs are opened to point to the active PRs so we know when to revert pointing to forked libraries.

spxiwh (approved):

PR #4866, opened 2024-09-05
Add dq fitting to Live supervised fitting script

maxtrevor:
This MR completes the statistic upgrade for PyCBC Live by allowing the daily supervised fitting script to produce DQ trigger rate files for use with the DQ reweighting statistic

This is a new feature.

This change primarily affects the live search. 

Two files are modified which are used in the offline search; they are bin/all_sky_search/pycbc_bin_templates and bin/plotting/pycbc_plot_dq_flag_likelihood . In both cases, the change should not cause any functional difference.

This changes the scientific output of the live search when the DQ reweighted stat is being used, otherwise it does not make a difference.

This also makes a number of changes to bin/live/pycbc_live_supervise_collated_trigger_fits beyond just adding the dq fitting process. Changes include moving the code used to calculate the original date of replay data into its own function, code style improvements, removing unused imports, changing variable names to be less ambiguous, etc.

Two new scripts are added to the bin/live/ folder, these are used for performing the dq fitting. The first gathers dq statistical information for a day of data, the second combines the information from multiple days an produces a statistic file to be ingested by the search.

After this is merged, we should be ready to create a new release branch for the live analysis.


bin/live/pycbc_live_supervise_collated_trigger_fits
```diff
@@ -390,26 +389,27 @@ def combine_significance_fits(
...
 
-    if 'variable-significance-fits' in combined_controls:
+    if 'variable-significance-fits' in combined_fit_controls:
         logging.info("Linking to variable significance fits file")
         sv.symlink(
-            combined_options['output'],
-            combined_controls['variable-significance-fits']
+            combined_fit_options['output'],
+            combined_fit_controls['variable-significance-fits']
         )
```
GarethCabournDavies, line 409: We had a conversation on one of the PyCBC calls about symlinking to local vs copying - I think this was one where copying was wanted, if that could be altered

bin/live/pycbc_live_supervise_collated_trigger_fits
```diff
@@ -460,6 +460,140 @@ def plot_combined_significance_fits(
...
+
+    combine_dq_args = ['pycbc_live_combine_dq_trigger_rates']
+    combine_dq_args += sv.dict_to_args(combined_dq_options)
+
+    sv.run_and_error(combine_dq_args, controls)
+
+    if 'variable-dq-trigger-rates' in combined_dq_controls:
+        logging.info("Linking to variable significance fits file")
+
+        sv.symlink(
```
GarethCabournDavies, line 545: Change to copy rather than symlink

bin/live/pycbc_live_supervise_collated_trigger_fits
```diff
@@ -598,6 +780,10 @@ parser.add_argument(
     help='Date to analyse, if not given, will analyse yesterday (UTC). '
          'Format YYYY_MM_DD. Do not use if using --run-daily-at.'
 )
+parser.add_argument(
+    '--collated-trigger-file',
+    help='Collated trigger file to use, if not given, will be created.'
```
GarethCabournDavies, line 785:
```suggestion
    help='Collated trigger file to use, if not given, will be created. '
         'Useful for testing.'
```

bin/live/pycbc_live_supervise_collated_trigger_fits
```diff
@@ -499,18 +637,22 @@ def supervise_collation_fits_dq(args, day_dt, day_str):
...
 
-    merged_triggers = trigger_collation(
-        day_dt,
-        day_str,
-        collation_control_options,
-        collation_options,
-        output_dir,
-        controls
-    )
+    if args.collated_trigger_file is not None:
```
GarethCabournDavies, line 640:
```suggestion
    if args.collated_trigger_file is not None:
        logging.info(
            "Using trigger file %s" from command line input,
            args.collated_trigger_file
        )
```

bin/live/pycbc_live_combine_dq_trigger_rates
```diff
@@ -0,0 +1,103 @@
...
+
+pycbc.init_logging(args.verbose)
+
+daily_files = args.daily_dq_files
+daily_files.sort()
+
+# need all files to use compatible settings
+# get settings hash from the last file
+# we will only use files that have the same hash
+with h5.File(daily_files[-1], 'r') as last_file:
```
GarethCabournDavies, line 41: Use pycbc.io HFile (and lines 59, 81)

bin/live/pycbc_live_combine_dq_trigger_rates
```diff
@@ -0,0 +1,103 @@
...
+total_livetime = 0
+flagged_livetime = 0
+total_triggers = np.zeros(num_bins)
+flagged_triggers = np.zeros(num_bins)
+for fpath in daily_files:
+    with h5.File(fpath, 'r') as f:
+        if f.attrs['settings_hash'] != settings_hash:
+            warning_str = f'File {fpath} has incompatible settings, skipping'
+            logging.warning(warning_str)
+            continue
```
GarethCabournDavies, line 63: I'm tempted to suggest to raise an error here, but I am sure there is a reason not to

GarethCabournDavies (changes requested):
This looks good - mainly stylistic/standardisation comments from me with a couple of minor questions

bin/live/pycbc_live_daily_dq_trigger_rates
```diff
@@ -0,0 +1,149 @@
...
+# option) any later version.
+#
+# This program is distributed in the hope that it will be useful, but
+# WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General
+# Public License for more details.
+
+"""
+Calculate the noise trigger rate adjustment as a function of data-quality state
+for a day of PyCBC Live triggers.
```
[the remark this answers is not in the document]
maxtrevor: It is somewhat different.
The main difference between the online and offline versions of the overall dq workflow has to do with where you get dq flagged segments from.
In offline the segments are precomputed and fed into pycbc as a segment file.
In online we obviously don't know the segments of bad dq problems ahead of time. This script has to ingest a day's worth of frame files and check how much of the time in that day was flagged, which is not something the offline code has to deal with.
Longer term I will think bout how these can possibly be united into a single script but I don't want to delay getting this update out.

bin/live/pycbc_live_daily_dq_trigger_rates
```diff
@@ -0,0 +1,149 @@
...
+        bin_num = int(bin_name[3:])
+        bgrp = bin_group.create_group(bin_name)
+        bgrp.create_dataset('tids', data=template_bins[bin_name])
+        bgrp.create_dataset('total_triggers', data=bin_total_triggers[bin_num])
+        bgrp.create_dataset('dq_triggers', data=bin_dq_triggers[bin_num])
+
+        bg_triggers = bin_total_triggers[bin_num] - bin_dq_triggers[bin_num]
+        num_trigs = np.array([bg_triggers, bin_dq_triggers[bin_num]])
+        trig_rates = num_trigs / state_time
+        mean_rate = bin_total_triggers[bin_num] / daily_livetime
```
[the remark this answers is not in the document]
maxtrevor: Using the total trigger rate as the normalization is intentional, it is the same as what is done in offline.
In some cases we have seen that the trigger rate for long duration templates is actually higher during clean times than flagged times. Using the total trigger rate as normalization allows us to downweight those situations as well as the usual case when trigger rates are higher during flagged times.

bin/live/pycbc_live_combine_dq_trigger_rates
maxtrevor: Yes, we want to have the flexibility to change the settings mid-run.
If we adjust the settings mid-run, this way will just ignore the files with old settings. Having an error would prevent this job from being able to finish after a settings change.

bin/live/pycbc_live_daily_dq_trigger_rates
```diff
@@ -0,0 +1,149 @@
...
+# for each bin, get total number of triggers and number of dq flagged triggers
+bin_total_triggers = np.zeros(num_bins)
+bin_dq_triggers = np.zeros(num_bins)
+
+with h5.File(args.trigger_file, 'r') as trigf:
+    for bin_name in template_bins.keys():
+        # bins are named as 'bin{bin_num}'
+        bin_num = int(bin_name[3:])
+        template_ids = template_bins[bin_name]
+        for template_id in template_ids:
```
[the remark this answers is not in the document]
maxtrevor: It hasn't been noticeably slow. In testing, the trigger collation process takes the majority of the time out of the collate and fit routine.

bin/live/pycbc_live_collated_dq_trigger_rates
```diff
@@ -0,0 +1,157 @@
...
+            # update trigger counts
+            bin_total_triggers[bin_num] += len(dq_states)
+            bin_dq_triggers[bin_num] += np.sum(dq_states)
+
+# write outputs to file
+logging.info(f'Writing results to {args.output}')
+with h5.File(args.output, 'w') as f:
+    ifo_group = f.create_group(args.ifo)
+    ifo_group.create_dataset('observing_livetime', data=livetime)
+    ifo_group.create_dataset('dq_flag_livetime', data=flagged_time)
```
GarethCabournDavies, line 127: I've found the issue with using HFile, [this line](https://github.com/gwastro/pycbc/blob/93f2856ae1b2c88d5f19a0d87ac12db5a234eb66/pycbc/io/hdf.py#L52) uses hasattr('dtytpe') to check it is is a numpy array, which is needed to allow the fletcher32 keyword to be set.

However numpy number-types have dtype attributes as well. The options are either:
 - change that line to use `isinstance(data, np.ndarray)` instead [should work]
 - change this line to give dq_flag_livetime as an attribute [will work]
 - Change this line to cast dq_flag_livetime as a numpy array before saving it [I think this will work, and allow this dataset to be checksummed]

bin/plotting/pycbc_plot_dq_flag_likelihood
```diff
@@ -76,7 +90,10 @@ ax2.set_ylabel('DQ Log Likelihood Penalty')
 ax2.set_ylim(numpy.log(ymin), numpy.log(ymax))
 
 # add meta data and save figure
-plot_title = f'{ifo}:{args.dq_label} DQ Trigger Rates'
+if args.title:
```
GarethCabournDavies, line 93: Use `if args.title is not None` - If someone wanted to use an empty string as title, this would not do it right

GarethCabournDavies (approved):
All my remaining comments are pedantry/optional, so approving

bin/live/pycbc_live_supervise_collated_trigger_fits
```diff
@@ -34,14 +32,50 @@ def read_options(args):
     del config_opts['environment']
     return config_opts
 
+
+def get_true_date(replay_day_dt, controls):
```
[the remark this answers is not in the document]
maxtrevor: I moved the if statement into the function

bin/live/pycbc_live_collated_dq_trigger_rates
maxtrevor: I think changing that line to use `isinstance(data, np.ndarray)` is the cleanest fix. It will also prevent this issue from arising again in the future for some other unsuspecting student.

PR #4868, opened 2024-09-09
allow using constraint in the config which sets vector margin params

WuShichao:

This is a: bug fix

This change will: nothing

## Motivation

If there are `marginalize_vector_params` in the config file and `constraint` sections to some variables, PyCBC will
apply those constraints to those marginalized parameters array, it will show `array-shape mismatch in array` error.

## Contents

Not pass those marginalized parameters to `self.prior_distribution`.

## Additional notes

### DIFF
--- a/pycbc/inference/models/base.py
+++ b/pycbc/inference/models/base.py
@@ -563,7 +563,13 @@ class BaseModel(metaclass=ABCMeta):
     def _logprior(self):
         """Calculates the log prior at the current parameters."""
         logj = self.logjacobian
-        logp = self.prior_distribution(**self.current_params) + logj
+        # allow using constraint in the config which sets vector marginalized parameters
+        if hasattr(self, 'marginalize_vector_params'):
+            self.current_params_no_vec_margin = {
+                k: v for k, v in self.current_params.items()
+                if k not in self.marginalize_vector_params
+            }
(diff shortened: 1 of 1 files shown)


WuShichao:
@ahnitz, Can this CC issue be ignored? https://codeclimate.com/github/gwastro/pycbc/pull/4868

pycbc/inference/models/base.py
```diff
@@ -563,7 +563,16 @@ def logprior(self):
     def _logprior(self):
         """Calculates the log prior at the current parameters."""
         logj = self.logjacobian
-        logp = self.prior_distribution(**self.current_params) + logj
+        # allow using constraint in the config which sets
```
ahnitz, line 566: This seems to leak the interface of the marginalization parameters so that it is coded in multiple places. Would it be better to have the DistMarg class perhaps just override this function instead?

WuShichao: Can you explain more?

WuShichao: @ahnitz It turns out Keisi encountered the same issue; this PR works for her.

ahnitz: @WuShichao I think you still need to address this comment. The point is that knowledge about marginalization should not exist strewn about the codebase, especially in the base model.

PR #4876, opened 2024-09-11
Update __init__.py

WuShichao:

This is a: bug fix

This change affects: inference, workflow

This change changes: plotting

This change will: fix the plotting injected values from mutliband demargin hdf file's injections

## Motivation

the default operation is to look for `injections` on the top-level in hdf file, this fix lets it to look it under the primary model

## Contents

 lets it to look `injections` under the primary model

## Additional notes

### DIFF
--- a/pycbc/inference/io/__init__.py
+++ b/pycbc/inference/io/__init__.py
@@ -765,7 +765,12 @@ def injections_from_cli(opts):
     # loop over all input files getting the injection files
     for input_file in input_files:
         fp = loadfile(input_file, 'r')
-        these_injs = fp.read_injections()
+        cp = fp.read_config_file()
+        if cp.has_option('model', 'primary_model'):
+            primary_model_label = cp.get('model', 'primary_model')
+            these_injs = fp.read_injections(group='/%s' % primary_model_label)
+        else:
+            these_injs = fp.read_injections()
         # apply mapping if it was provided
         if param_map:
             mapvals = {sp: these_injs[ip] for ip, sp in param_map}


ahnitz:
This has more to do with how we want to handle the heirarchical analysis. @cdcapano Should weigh in

WuShichao:
@ahnitz @cdcapano This PR fixes my current issue but can be more general for other hierarchical-based models.

cdcapano (changes requested):
Thanks for doing this; it's a good thing to fix. However, the patch as-is will only fix the case for the JointPrimaryMarginalized model, and will only load the injection from the primary model at that. It would be good to fix this for more generic hierarchical models. Currently, plotting injection parameters is broken for all hierarchical runs, since the injections are specific to each sub-model.

I think the way to go about this is to check if there's a `submodels` argument in the `[model]` section of the config file, and if so, get their names. Then cycle over all of the submodels and load the injection from each one, but remap the parameter names of the injections to `<submodel_label>__<param>`. You'd want to do that anyway since the template parameters will have the submodel name pre-pended to them. I think that should work for all models that inherit from the HierarchicalModel, including the JointPrimaryMarginalized, but you and @ahnitz should verify since you wrote the JointPrimary.

WuShichao:
> Thanks for doing this; it's a good thing to fix. However, the patch as-is will only fix the case for the JointPrimaryMarginalized model, and will only load the injection from the primary model at that. It would be good to fix this for more generic hierarchical models. Currently, plotting injection parameters is broken for all hierarchical runs, since the injections are specific to each sub-model.
> 
> I think the way to go about this is to check if there's a `submodels` argument in the `[model]` section of the config file, and if so, get their names. Then cycle over all of the submodels and load the injection from each one, but remap the parameter names of the injections to `<submodel_label>__<param>`. You'd want to do that anyway since the template parameters will have the submodel name pre-pended to them. I think that should work for all models that inherit from the HierarchicalModel, including the JointPrimaryMarginalized, but you and @ahnitz should verify since you wrote the JointPrimary.

Thanks for the comments, my `JointPrimaryMarginalized` is different, in the config file it doesn't have `submodels` like the original `HierarchicalModel`:
```
[model]
name = joint_primary_marginalized
primary_model = 3g
other_models = lisa
static_margin_params_in_other_models = True
```
So `fp.read_config_file()` will not see `submodels`, but after initializing the model by `from_config` it will have `self.submodels`, while it will not be initialized during plotting.

cdcapano:
Ok, then add a static method to the `HierarchicalModel`, something like `get_model_labels(cp)`, that loads the model names given a config file. In the `HierarchicalModel` that will look for the `submodels` argument. Override that method in the `JointPrimaryMarginalizedModel` to load the `primary_model` and `other_models` instead. In either case, the `get_model_labels` should return the model labels. Then in `injections_from_cli` you can just call `get_model_labels` without needing to worry about looking for specific arguments. That will future proof it for any other hierarchical model derivative that might use different names.

PR #4879, opened 2024-09-13
Fix numpy scalar edge case bug

GarethCabournDavies:
Creating a dataset in HFile using a numpy scalar would pass this check as it has a dtype attribute, which isn't what we need here

This is a bug fix
This change changes nothing

## Motivation
Bug fix

This was causing problems in #4866

## Testing performed
None


maxtrevor (commented):
Does what we need

maxtrevor (approved):

PR #4881, opened 2024-09-13
Bugfix for live fits

maxtrevor:
We discovered a bug where if an ifo had no data for an entire day, it would cause an error in the daily supervised fitting. This fixes that bug.

This affects only the PyCBC Live search, not offline.

### DIFF
--- a/bin/live/pycbc_live_supervise_collated_trigger_fits
+++ b/bin/live/pycbc_live_supervise_collated_trigger_fits
@@ -668,6 +668,8 @@ def supervise_collation_fits_dq(args, day_dt, day_str):
     # Store the locations of files needed for the statistic
     stat_files = []
 
+    # daily fits should only be done for IFOs that had data
+    # and were included in the config file
     all_ifos = controls['ifos'].split()
     active_ifos = get_active_ifos(merged_triggers)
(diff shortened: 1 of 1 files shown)


bin/live/pycbc_live_supervise_collated_trigger_fits
```diff
@@ -656,7 +667,11 @@ def supervise_collation_fits_dq(args, day_dt, day_str):
         )
     # Store the locations of files needed for the statistic
     stat_files = []
-    for ifo in config_opts['control']['ifos'].split():
+
+    all_ifos = controls['ifos'].split()
+    active_ifos = get_active_ifos(merged_triggers)
+    active_ifos = [ifo for ifo in active_ifos if ifo in all_ifos]
```
GarethCabournDavies, line 673: I think here we are looking at removing any IFOs which are in the trigger file but not in the config file? That may be worth commenting, as I got confused for a minute

GarethCabournDavies (approved):
Minor comment re a comment, but this LGTM

bin/live/pycbc_live_supervise_collated_trigger_fits
maxtrevor: Added a comment

PR #4882, opened 2024-09-17
Update offline source probability for change to mchirp area

tdent:
Fix a from_cli bug and remove mass gap related options and code, as mass gap classification does not need to be supported for offline calculation

This is a: bug fix

This change changes: ability to produce a scientific result vs no result

This change: follows style guidelines (See e.g. [PEP8](https://peps.python.org/pep-0008/))

Bug fix

## Motivation
Offline source probability code fails to run as it was not updated for a module change

## Contents
Try to make it run

## Testing performed
To be tested by local install and attempting to run offline source probability

The author of this pull request confirms they will adhere to the [code of conduct](https://github.com/gwastro/pycbc/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/bin/pycbc_source_probability_offline
+++ b/bin/pycbc_source_probability_offline
@@ -27,8 +27,6 @@ parser.add_argument('--search-tag', required=True,
 parser.add_argument('--ifar-threshold', type=float, default=None,
                     help='Select only candidate events with IFAR '
                          'above threshold.')
-parser.add_argument('--include-mass-gap', action='store_true',
-                    help='Option to include the Mass Gap region.')
 mchirp_area.insert_args(parser)
 args = parser.parse_args()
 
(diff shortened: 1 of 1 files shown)


AnaLorenzoMedina (approved):
yep, it works now. Everything ok

AnaLorenzoMedina (dismissed):
The first change was necessary but the code still didn't work, we needed one more fix for the missing gap entry. Moreover, even if you run the 'run_probabilities_offline.sh' script with the '--include-mass-gap' option, the output file does not have a 'Mass Gap' entry.

AnaLorenzoMedina (commented):
The first change was necessary but the code still didn't work, we needed one more fix for the missing gap entry. Moreover, even if you run the 'run_probabilities_offline.sh' script with the '--include-mass-gap' option, the output file does not have a 'Mass Gap' entry.

tdent:
OK, given the changes in the interface to mchirp_area.py to make no mass gap the default (https://github.com/gwastro/pycbc/commit/45c9560c72a83b9364b23af746b4eeea5942c35b) it seems we would need more work on this script to be able to actually produce classifications with mass gap. 

However it doesn't seem that there is any need to support this option for offline results, so I will just remove it completely to simplify the script

tdent:
Hopefully it works without any issue after removing all mass gap option related code

AnaLorenzoMedina (approved):
I checked locally and it should run fine now.

PR #4890, opened 2024-09-24
Don't force global multiprocessing start method

ahnitz:

This is a: bug fix

This change affects library useage of pycbc within a setting where the multiprocessing start context is already set. We currently *force* a setting of "fork" which doesn't work in some cases and may not be what someone wants for all uses. 

Any executable and library code which relies on multiprocessing. In these cases, the correction recommended by @xkzl is used. 

This change will ideally have no effect on current usage. 

Error reported in https://github.com/gwastro/pycbc/issues/4853

## Additional notes

### DIFF
--- a/pycbc/inference/models/brute_marg.py
+++ b/pycbc/inference/models/brute_marg.py
@@ -20,7 +20,7 @@ import math
 import logging
 import numpy
 
-from multiprocessing import Pool
+from pycbc.pool import BroadcastPool as Pool
 from scipy.special import logsumexp
 
 from .gaussian_noise import BaseGaussianNoise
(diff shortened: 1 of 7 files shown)


pycbc/inference/models/brute_marg.py
```diff
@@ -20,7 +20,7 @@
 import logging
 import numpy
 
-from multiprocessing import Pool
```
GarethCabournDavies, line 23: Possibly stupid question - some places have multiprocessing.Pool, other have multiprocessing.pool.Pool. This is how it was before these changes as well, so it must be okay, but wanted to flag it

GarethCabournDavies (approved):
Looks good to me; one minor question but it isn't a showstopper

pycbc/inference/models/brute_marg.py
xkzl: If using 'multiprocessing.Pool' I believe this will use default context as defined by set_start_context. In this case one can use 'get_context("fork/spawn").Pool()'

PR #4892, opened 2024-09-24
Do not force pygrb_efficiency to produce all output: let the user decide

pannarale:
This PR allows the user to choose which output files (if any) `pygrb_efficiency` must produce.

This is adds flexibility to an existing post-processing script.

This change affects: PyGRB

This change changes: the ability to specifiy what plots and tables `pygrb_efficiency` must create.

## Motivation
There is no need for post-processing scripts to produce all the output they can possibly produce.  The user may or may not need all outputs when calling a script.

## Contents
`pygrb_efficiency` can produce three outputs so the lines to produce them are now encapsulated in three `if`s.

Additionally, the code makes sure that all input files necessary to produce the requested output files are passed.

## Testing performed
All plots in sections 5 and 6.02 of [this webpage](https://ldas-jobs.ligo.caltech.edu/~francesco.pannarale/LVC/pygrb_sep2024_2/) were made with this updated version of `pygrb_efficiency`.

### DIFF
--- a/bin/pygrb/pycbc_pygrb_efficiency
+++ b/bin/pygrb/pycbc_pygrb_efficiency
@@ -98,11 +98,11 @@ parser.add_argument("-F", "--trig-file", action="store", required=True,
 parser.add_argument("--onsource-file", action="store",
                     help="Location of on-source trigger file (or a " +
                     "background trigger file to be treated as such).")
-parser.add_argument("--background-output-file", required=True,
+parser.add_argument("--background-output-file",
                     help="Detection efficiency output file.")
-parser.add_argument("--onsource-output-file", required=True,
+parser.add_argument("--onsource-output-file",
                     help="Exclusion distance output file.")
-parser.add_argument("--exclusion-dist-output-file", required=True,
+parser.add_argument("--exclusion-dist-output-file",
                     help="JSON file containing exclusion distances.")
 parser.add_argument("-g", "--glitch-check-factor", action="store",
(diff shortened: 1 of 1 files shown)


bin/pygrb/pycbc_pygrb_efficiency
```diff
@@ -164,10 +171,15 @@ num_mc_injs = opts.num_mc_injections
...
-    os.makedirs(outdir)
+# Set output directories
+outdir = None
+for output_file in [opts.exclusion_dist_output_file,
+    opts.background_output_file, opts.onsource_output_file]:
+    if output_file is not None:
+        logging.info("Setting output directory.")
+        outdir = os.path.split(os.path.abspath(output_file))[0]
+        if not os.path.isdir(outdir):
+            os.makedirs(outdir)
```
MarcoCusinato, line 182: I would add a `break` at the end to avoid repeating logging messages.

MarcoCusinato (changes requested):

bin/pygrb/pycbc_pygrb_efficiency
pannarale: Would it make more sense to move the message inside the last if and specify the directory being created?

MarcoCusinato: I agree. Maybe we can also add the directroy name to the log, for a more complete output.

MarcoCusinato (approved):

PR #4896, opened 2024-09-29
add zero-noise PE example for LISA

WuShichao:

This is a: doc update

This change affects: inference, doc

This change changes: documentation

## Motivation

In the current PyCBC Doc, it seems there are no `zeroNoise` examples for PE. Also, someone asked how to perform LISA PE with zero noise injection while still passing extra arguments to PSD functions in the likelihood calculation.

## Contents

Add code to show how to use `psd-extra-args` with `psd-model`, when `fake-strain = zeroNoise`.

## Testing performed

`zero noise`
lisa_smbhb_mass_tc
`noise`
lisa_smbhb_mass_tc_noise

## Additional notes

### DIFF
--- a/examples/inference/lisa_smbhb_inj/lisa_smbhb_relbin.ini
+++ b/examples/inference/lisa_smbhb_inj/lisa_smbhb_relbin.ini
@@ -52,8 +52,7 @@ tc =
 
 [static_params]
 ; Change it to "ref_frame = SSB", if you use SSB frame in injection file.
-; ref_frame = LISA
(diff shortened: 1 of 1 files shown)


WuShichao:
@ahnitz Can we merge it?

ahnitz:
@WuShichao Shouldn't it be a separate example rather than overwriting the existing one? I know you left stuff in the comments, but I think a separate file would be better. Otherwise, if this works after that then you are good to go.

PR #4897, opened 2024-10-02
Update setup.py

ahnitz:
Prepare for release to pick up numpy 2.0 support + other features


spxiwh (approved):
I forced a CodeClimate rerun, which succeeded, but doesn't show up here. Approved.

PR #4900, opened 2024-10-06
always save the latest lognl in the demargin hdf

WuShichao:

This is a: bug fix,

This change changes: result presentation / plotting, scientific output

This change will: nothing

## Motivation

There is a precision issue in lognl calculation, during reconstruct process, model will get slightly different lognl and loglr values (so loglikelihood). This becomes a problem when loglr is very small compared to lognl, which numerical precision is domainted by lognl. If still use the original lognl before the reconstruct process, `pycbc_inference_plot_posterior` will get wrong loglr/snr for plotting.

## Contents

Save the latest `lognl` into the demargin hdf file.

### DIFF
--- a/bin/inference/pycbc_inference_model_stats
+++ b/bin/inference/pycbc_inference_model_stats
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/work/shichao.wu/anaconda3/envs/bbhx_env/bin/python3.9
 
 # Copyright (C) 2020 Collin Capano
 #
@@ -17,8 +17,9 @@
 # 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
 
 """Recalculates log likelihood and prior for points in a given inference or
-posterior file and writes them to a new file. Also records auxillary model
-stats that may have been ignored by the sampler.
+posterior file and writes them to a new file, overwrites all lognl in the
+samples group attrs. Also records auxillary model stats that may have been
+ignored by the sampler.
 """
 
 import os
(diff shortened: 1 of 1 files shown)


bin/inference/pycbc_inference_model_stats
```diff
@@ -92,20 +93,34 @@ model = models.read_from_config(cp)
...
 
-# create function for calling the model to get the stats
+# create function for calling the model to get the stats, lognl, and rec
 def callmodel(arg):
     iteration, paramvals = arg
     # calculate the logposterior to get all stats populated
     model.update(**{p: paramvals[p] for p in model.variable_params})
     _ = model.logposterior
     stats = model.get_current_stats()
+    if hasattr(model, 'submodels'):
```
ahnitz, line 103: @WuShichao Wouldn't it make more sense for heirarchical models to make it so that `get_current_stats` includes any submodels rather than hardcoding knowledge about submodels into this script? It's best to keep information compartmentalized and use clearer interfaces to avoid technical debt in the future and prevent flexibility. 

Similarly, why do we need to hardnode anything about lognl here? Maybe it should just be included in the stats when it is a fixed number like this, no?

ahnitz (commented):
@WuShichao See comments. I thinn tk probably the right place for changes here is actually on the model side. 

@cdcapano Thoughts?

bin/inference/pycbc_inference_model_stats
WuShichao: The `lognl` issue is not the hierarchical-model-only problem; whenever users use this script and `lognl` is much larger than `loglr`, the same problem will occur. So instead of modifying all likelihood models, I only change in one place. The reason why I hardcode `lognl` is I want to match the keys in the original PE file.

ahnitz: @WuShichao (1) You are explicitly checking through submodels here. That really should be handled by the hierarchical model itself. 

(2) lognl should probably just be a model stat if it is a constant. No modifications are then needed here.

PR #4908, opened 2024-10-11
Make coinc_statmap work in the case that no foreground is present

GarethCabournDavies:
Prasia found a bug in pycbc_coinc_statmap if no foreground events have been found,
this fixes that

This is a: bug fix
This change changes: nothing
This change: has been proposed using the [contribution guidelines](https://github.com/gwastro/pycbc/blob/master/CONTRIBUTING.md)

## Motivation
Fix bug for a legitimate use-case in the offline search

## Contents

A variable was being used later in the code which was only set in the case that there was some foreground

This sets the variable and ensures that it is usable later

https://github.com/gwastro/pycbc/issues/4907

## Testing performed
To be tested with downstream codes

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_statmap
+++ b/bin/all_sky_search/pycbc_coinc_statmap
@@ -291,9 +291,9 @@ if fore_locs.sum() > 0:
     f['foreground/ifar_exc'] = conv.sec_to_year(ifar_exc)
     f['foreground/fap_exc'] = fap_exc
 else:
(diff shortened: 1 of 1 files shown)


Prasia-Pankunni (approved):

Prasia-Pankunni (approved):
The statmap job works fine

GarethCabournDavies:
I forgot that reviewers need write access -  @spxiwh ?

spxiwh (approved):

PR #4913, opened 2024-10-21
Adding warning for no injections

aleynaakyuz:
This raises a warning if no injections are applied during an inference job.

### DIFF
--- a/pycbc/inject/inject.py
+++ b/pycbc/inject/inject.py
@@ -568,6 +568,7 @@ class CBCHDFInjectionSet(_HDFInjectionSet):
             # safety buffer
             start_time = inj.tc - 2 * (inj_length + 1)
             if end_time < t0 or start_time > t1:
+                logger.warning("No injections applied.")
                 continue
             signal = self.make_strain_from_inj_object(inj, delta_t,
                      detector_name, f_lower=f_l,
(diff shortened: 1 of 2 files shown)


ahnitz:
@aleynaakyuz I don't think this will do what you think. I think you'd only see a message if the injection-file option wasn't given. I think what you need to do is to add a logging message into the InjectionSet class itself where the injections are applied.

ahnitz:
Think about potential changes here. https://github.com/gwastro/pycbc/blob/master/pycbc/inject/inject.py#L561

pycbc/inject/inject.py
```diff
@@ -568,6 +568,7 @@ def apply(self, strain, detector_name, f_lower=None, distance_scale=1,
             # safety buffer
             start_time = inj.tc - 2 * (inj_length + 1)
             if end_time < t0 or start_time > t1:
+                logger.warning("No injections applied.")
```
ahnitz, line 571: Shouldn't this be after the for loop?

ahnitz: See discussion here https://book.pythontips.com/en/latest/for_-_else.html

aleynaakyuz: After which loop? If there are no injections, `if end_time < t0 or start_time > t1:` will be True and the rest of the interaction will be skipped, right? That's why I put the warning in the loop.

ahnitz: That's only true for a *specific* injection. What if the file has multiple things which could be injected? The injection file can have many signals in it (perhaps for different times or even overlapping).

PR #4914, opened 2024-10-22
Fix problem on pycbc live due to numpy 2 

### LINKED ISSUE #4912: Numpy 2 leads to numerical underflows in PyCBC Live
Numpy 2 introduced [changes to data type promotion](https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion), in particular this one:
> np.float32(3) + 3. now returns a float32 when it previously returned a float64.

This introduces a subtle bug in PyCBC Live. PSD estimates are internally represented as single-precision float arrays, which is possible because the GW data is scaled by the "dynamic range factor" (see FindChirp paper for details) before doing math on it. In a couple places, however, we convert the PSD estimate from the internal representation back to "physical" values by doing this:
```
actual_psd = single_precision_psd / DYN_RANGE_FAC ** 2
```
where `single_precision_psd` is typically around 1e-4, and `DYN_RANGE_FAC` is [~6e20](https://github.com/gwastro/pycbc/blob/master/pycbc/__init__.py#L175). This line used to work before Numpy 2 because `single_precision_psd` would be promoted to double precision before being divided. In Numpy 2, the promotion no longer happens, so the division underflows and we get an array of (single-precision) zeroes.

In practice this causes the checks in the example script to fail (well done, checks!) and it causes the PSD estimates in the LIGOLW XML files to be all zeroes (which would break whatever reads them later, like BAYESTAR).

I think this went unnoticed because Numpy 2 was never pulled in any of the CI tests due to other dependencies. Thanks to @Thomas-JACQUOT for spotting it while trying the PyCBC Live example on his laptop :)

Thomas-JACQUOT:
This fixes #4912.

I explictly convert the PSD to double precision before the division, as soon numpy 2.x is not doing it automatically anymore. 

I also add another check of the XML files to make sure that the PSDs in there are valid.

### DIFF
--- a/examples/live/check_results.py
+++ b/examples/live/check_results.py
@@ -11,7 +11,7 @@ from pycbc.io.ligolw import LIGOLWContentHandler
 from ligo.lw.utils import load_filename as load_xml_doc
 from ligo.lw import lsctables
 from pycbc import conversions as conv
-import LigoLWEventSource
+from ligo.skymap.io import LigoLWEventSource
 
 def close(a, b, c):
     return abs(a - b) <= c
--- a/examples/live/check_results.py
+++ b/examples/live/check_results.py
@@ -39,7 +39,7 @@ def check_single_results(args):
                     continue
 
                 # check that PSD is sane
-                psd = group['psd'][:] / pycbc.DYN_RANGE_FAC ** 2
+                psd = group['psd'][:].astype(np.float64) / pycbc.DYN_RANGE_FAC ** 2
                 psd_df = group['psd'].attrs['delta_f']
                 psd_f = np.arange(len(psd)) * psd_df
                 psd_epoch = group['psd'].attrs['epoch']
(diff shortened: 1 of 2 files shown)


titodalcanton:
One potential concern if we move *everything* to double would be memory usage. PyCBC Live already uses quite a bit of RAM, which would presumably double if we did this. Using double just for the PSD would not be a problem of course, but then again, what do we gain?

maxtrevor:
Given that this is currently blocking a new release for PyCBC Live, I would strongly prefer that we use the fastest fix available. We can revisit larger changes like moving everything to float64 on a longer timescale.

titodalcanton:
@maxtrevor this only blocks PyCBC Live releases if you decide to run it under Numpy 2.

maxtrevor:
> @maxtrevor this only blocks PyCBC Live releases if you decide to run it under Numpy 2.

Ah right. At this point we have to actively choose not to use numpy 2 when making an environment, so I'd still prefer to get this in but you;re right its not necessarily a blocker.

ahnitz (approved):
Feel free to address the codeclimate message, but otherwise this can be merged whenever you are happy. The change is straightforward and make sense.

ahnitz:
@spxiwh The psd calculation itself is actually not problematic. It can be done in either float32 or float64 already just by choice of input (e.g. whatever the precision of the data is). The issue is just that float32 only can have exponents within 10^(+- ~38).

PR #4916, opened 2024-10-28
Adding the option to choose preferred domain

kkacanja:
Incorporated the option to set a preferred domain for waveform generation. This is useful for waveform models that are available in both frequency domain and time domain. Previously it automatically chose the frequency domain.

Example Usage:

[static-params]
preferred-domain = td

The options are td for time domain,  fd for frequency domain, and none for default


pycbc/inference/models/gaussian_noise.py
```diff
@@ -1201,7 +1201,8 @@ def create_waveform_generator(
     except KeyError:
         raise ValueError("no approximant provided in the static args")
 
-    generator_function = generator_class.select_rframe_generator(approximant)
+    domain = static_params.get('preferred_domain', 'none')
```
cdcapano, line 1204: Any reason why you're using the string 'none' instead of the type `None`? The latter will be a bit better to work with, to avoid spelling errors.

pycbc/waveform/generator.py
```diff
@@ -1198,57 +1199,57 @@ def generate(self, **kwargs):
...
-    >>> from pycbc.waveform import ringdown
-    >>> ringdown.ringdown_fd_approximants.keys()
-
-    Get generator object:
-    >>> from pycbc.waveform.generator import select_waveform_generator
-    >>> select_waveform_generator(waveform.fd_approximants()[0])
-    """
-    # check if frequency-domain CBC waveform
+def select_waveform_generator(approximant, domain):
+    """Returns the single-IFO generator for the approximant."""
```
cdcapano, line 1203: Add some documentation here about what domain is, and what the accepted options are. Also, I'd set the default domain to None (`domain=None`), to keep the calling signature the same as it was in the past.

pycbc/waveform/generator.py
```diff
@@ -1198,57 +1199,57 @@ def generate(self, **kwargs):
...
+                'TdQNMfromFinalMassSpin': TDomainMassSpinRingdownGenerator,
+                'TdQNMfromFreqTau': TDomainFreqTauRingdownGenerator,
+            },
+            'supernovae': {
+                'CoreCollapseBounce': TDomainSupernovaeGenerator
+            }
+        }
+    }
+
+    gen = None
```
cdcapano, line 1225: Add a check here that domain is set to one of the recognized values (`None`, `'td'`, or `'fd'`). If not, raise a ValueError.

pycbc/waveform/generator.py
```diff
@@ -1198,57 +1199,57 @@ def generate(self, **kwargs):
...
+
+    # Check if the approximant exists in any domain
     if approximant in waveform.fd_approximants():
-        return FDomainCBCGenerator
-    # check if time-domain CBC waveform
+        if domain in ['fd', 'none']:
```
cdcapano, line 1229: Rather than doing `if domain in ...` do `if domain == 'fd' or domain is None:`

pycbc/waveform/generator.py
```diff
@@ -1198,57 +1199,57 @@ def generate(self, **kwargs):
...
+            gen = generator_map['fd']['cbc']
+        elif domain == 'td' and approximant in waveform.td_approximants():
+            gen = generator_map['td']['cbc']
+        elif domain == 'td':
+            raise ValueError(f"{approximant} is a frequency-domain model.")
```
cdcapano, line 1234: Add to the error message "Requested domain is 'td', but...", to give the user a better idea of what went wrong if you get here.

pycbc/waveform/generator.py
```diff
@@ -1198,57 +1199,57 @@ def generate(self, **kwargs):
...
     elif approximant in waveform.td_approximants():
-        return TDomainCBCGenerator
-    # check if frequency-domain ringdown waveform
+        if domain in ['td', 'none']:
```
cdcapano, line 1236: Same comment as above: do an or check rather than looking for values in a list.

cdcapano (changes requested):
Thanks for doing this. I have just a few code-related requests.

pycbc/waveform/generator.py
ahnitz: @kkacanja If there are good reason, one can sometimes bend / ignore the CC recommendations. One shouldn't do that by default, but we shouldn't hold so strictly to them that we make it impossible to write sensible code.

pycbc/waveform/generator.py
```diff
@@ -1225,30 +1229,57 @@ def select_waveform_generator(approximant):
...
-            return TDomainSupernovaeGenerator
-    # otherwise waveform approximant is not supported
-    else:
-        raise ValueError("%s is not a valid approximant." % approximant)
+        gen = generator_map['td']['supernovae'].get(approximant, None)
+
+    if gen is None:
+        raise ValueError(f"{approximant} is not a valid approximant.")
+
+    return gen
```
cdcapano, line 1282: I kind of agree with code climate here... the logic is challenging to follow. More importantly, it's not good that domain is silently ignored if you are using a ringdown or supernova approximants. I think this could be simplified and made more uniform if you make the initial decision based on domain rather than approximant. Something like:
```
if domain == 'td':
    gen = get_td_generator(approximant)
elif domain == 'fd':
    gen = get_fd_generator(approximant)
elif domain is None:
    try:
        gen = get_fd_generator(approximant)
    except ValueError:
         gen = get_td_generator(approximant)
```
Then write separate functions `get_td_generator` and `get_fd_generator` that will pull out the appropriate cbc, ringdown, or supernova generator based on the approximant, raising a ValueError if it can't find one. That will also be more straight forward to explain: if the domain is none, it will first try the FD approximants, then try the TD approximants, then fail if it can't find either, and would apply equally to all types of signals.

Doing this means you'll have to move the generator_map out of the function definition, since get_td_generator and get_fd_generator will need to access it. Better yet, since you'll have separate functions for this, you can drop the dictionary and just have each function apply the appropriate logic to select the generator.

pycbc/waveform/generator.py
```diff
@@ -1225,30 +1229,57 @@ def select_waveform_generator(approximant):
...
 
 
 def select_waveform_modes_generator(approximant):
```
cdcapano, line 1285: Could you add the domain argument to this function too, for consistency? Use the same logic you use above. You can use the `get_td_generator` and `get_fd_generator` functions I suggested writing above, just add a `modes=False` keyword argument to them so that they get the modes generators for CBC waveforms if `modes` is set to True. Then in this function you would do the same logic, just with `modes=True`.

cdcapano (changes requested):
Sorry, more change requests. Also, please address the 79-character limits that code climate flagged.

kkacanja:
> Sorry, more change requests. Also, please address the 79-character limits that code climate flagged.
You initially made the comment for
            raise ValueError(f"{approximant} is a frequency-domain model.")
to have a more informative error message. The message above is 75 characters long. Do you want me to revert back to this to limit it to 79 characters or keep the more informative error message?

pycbc/waveform/generator.py
kkacanja: I can do that as well

pycbc/waveform/generator.py
kkacanja: Thanks for the feedback, I agree. I initially had the code logic set up in this way. If I revert to it codeclimate might set up a flag for having 20 branches. If it is alright, I can revert these changes

pycbc/waveform/generator.py
```diff
@@ -1198,14 +1199,38 @@ def generate(self, **kwargs):
...
 
-def select_waveform_generator(approximant):
+# Updated code to make the logic clearer and simplify decision-making based on the domain
+
+def get_td_generator(approximant, modes=False):
+    """Returns the time-domain generator for the given approximant."""
+    if approximant in waveform.td_approximants():
+        return TDomainCBCModesGenerator if modes else TDomainCBCGenerator
+    if approximant in ringdown.ringdown_td_approximants:
+        return TDomainMassSpinRingdownGenerator if approximant == 'TdQNMfromFinalMassSpin' else TDomainFreqTauRingdownGenerator
```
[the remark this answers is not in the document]
kkacanja: Ok sounds good, I'll apply these changes

pycbc/waveform/generator.py
```diff
@@ -1225,51 +1250,47 @@ def select_waveform_generator(approximant):
...
-    if approximant in waveform.fd_approximants():
-        return FDomainCBCModesGenerator
-    # check if time-domain CBC waveform
-    elif approximant in waveform.td_approximants():
-        return TDomainCBCModesGenerator
-    # otherwise waveform approximant is not supported
-    raise ValueError("%s is not a valid approximant." % approximant)
+
+    if domain not in {None, 'td', 'fd'}:
+        raise ValueError(f"Invalid domain '{domain}'. Must be one of: None, 'td', or 'fd'.")
```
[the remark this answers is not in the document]
kkacanja: Oh cool, I didn't know you could do that. Thank you for all your suggestions

PR #4922, opened 2024-11-04
try macos latest version

ahnitz:


ahnitz:
@titodalcanton This patch solves the immediate problem of macos 12 going away by bumping up our versions. I think for the builds we should stick with a fixed version so this advances us to 14, but for the tests I think it makes sense to work with the latest. 

The old PR tried to do some additional things to merge the macos test and linux ones. That turned out to be more complicated than I'd like. I might revisit it at some point, but this separates out the needed part.

titodalcanton (approved):
Looks good, I agree with sticking to an explicit version for builds.

PR #4924, opened 2024-11-04
Fix bug in fetching injection group param

tdent:
The injection param fetching has an obvious bug in a branch which I am guessing has never been used .. see the change

This is a: bug fix

This change affects: plotting utility code currently not used in offline search

This change will: not have any effect other than improving ability to retrieve injection params

## Testing performed
Works to retrieve the `injections/tc` dataset.

### DIFF
--- a/.github/workflows/distribution.yml
+++ b/.github/workflows/distribution.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


ahnitz (approved):
Yeah, that does seem like an obvious bug.

ahnitz:
@tdent You'll need to rebase from the PR I just merged to address the macos build issues.

PR #4929, opened 2024-11-07
Preparing for veto file handling in PyGRB

pannarale:
This is the first PR of a series that will enable PyGRB to handle an xml veto file when producing final results (plots, p-values, tables, etc.)

This is a: a new feature in PyGRB.  As an aside, some utilities and scripts will be streamlined.

This change affects: PyGRB

This change will probably break current functionalities at the moment.  If the standard automated test running `--help` on all scripts fails on PyGRB plotting scripts, I will add some workarounds to avoid this.  If needed, these will likely be empty functions: the plotting scripts will be progressively renovated in PRs following this one.

## Motivation
PyGRB in its current version is not using data vetoes.

## Contents
- PyGRB used to handle veto definer files "collating" CAT1, CAT2, etc. xml veto definer files.  Now this can go through the PyCBC +/-flag syntax and a single file.  This is reflected in the post-processing utilities file, where a single veto vile is handled and a bunch of functions to combine multiple files were removed.
- While upgrading the post-processing utilities file, I cleaned up other minor things, included more logging messages, inserted a few additional checks, and tried increasing robustness in general.
- At workflow level the switch to a single veto file is reflected in `grb_utils.py` and the veto file is handled in parallel with the segments files throughout the functions (as noted in https://github.com/gwastro/pycbc/issues/4887 the segments files will need to become a single json file eventually).
- I did not work on the json file yet, but I did remove some hackiness in file handling from `jobsetup.py`.

## Testing performed

### DIFF
--- a/bin/pygrb/pycbc_make_offline_grb_workflow
+++ b/bin/pygrb/pycbc_make_offline_grb_workflow
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 
-# Copyright (C) 2015 Andrew R. Williamson
+# Copyright (C) 2015 Andrew R. Williamson, Francesco Pannarale
 #
 # This program is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by the
@@ -79,7 +79,6 @@ sciSegsFile = _workflow.get_segments_file(wflow,
                                           'science',
                                           'segments-science',
                                           seg_dir)
-
 sciSegs = {}
 for ifo in wflow.ifos:
     sciSegs[ifo] = sciSegsFile.segment_dict[ifo+':science']
@@ -130,14 +129,15 @@ else:
 
 sciSegs = segmentlistdict(sciSegs)
 if onSrc is None:
-    plot_met = make_grb_segments_plot(wflow, sciSegs, triggertime, triggername,
-                                      seg_dir, fail_criterion=offSrc)
+    plot_met = make_grb_segments_plot(
+        wflow, segmentlistdict(), triggertime, triggername,
+        seg_dir, fail_criterion=offSrc)
     logging.info("Making segment plot and exiting.")
     sys.exit()
 else:
     plot_met = make_grb_segments_plot(
-            wflow, sciSegs, triggertime, triggername, seg_dir,
-            coherent_seg=offSrc[tuple(offSrc.keys())[0]][0])
+        wflow, sciSegs, triggertime, triggername, seg_dir,
+        coherent_seg=offSrc[tuple(offSrc.keys())[0]][0])
     segs_plot = _workflow.File(plot_met[0], plot_met[1], plot_met[2],
                                file_url=plot_met[3])
     segs_plot.add_pfn(segs_plot.cache_entry.path, site="local")
@@ -193,9 +193,11 @@ if wflow.cp.has_option("workflow-condition_strain", "do-gating"):
                                            int(sciSegs[ifo][0][1]))
 
 ifos = sorted(sciSegs.keys())
-ifo = ifos[0]
 wflow.ifos = ifos
 datafind_veto_files = _workflow.FileList([])
+
+# Convenience variable to operate on sciSegs within this executable
+ifo = ifos[0]
 
 # GATING
 if wflow.cp.has_option("workflow-condition_strain", "do-gating"):
@@ -234,12 +236,17 @@ if wflow.cp.has_option("workflow-condition_strain", "do-gating"):
         gated_frames.convert_to_lal_cache().tofile(
                             open(gated_cache.storage_path, "w"))
         datafind_files.append(gated_cache)
-
 datafind_veto_files.extend(datafind_files)
-ifo_list = sorted(sciSegs.keys())
-ifo = ifo_list[0]
-ifos = ''.join(ifo_list)
-wflow.ifos = ifos
+
+# Retrieve vetoes
+veto_file = None
+if wflow.cp.has_option("workflow-segments", "segments-vetoes"):
+    veto_file = _workflow.get_segments_file(wflow,
+                                            'vetoes', 
+                                            'segments-vetoes',
+                                            seg_dir,
+                                            tags=['veto'])
+    datafind_veto_files.append(veto_file)
 
 # Config file consistency check for IPN GRBs
 if wflow.cp.has_option("workflow-inspiral", "ipn-search-points") \
@@ -262,6 +269,7 @@ else:
 if wflow.cp.has_option('workflow-inspiral', 'bank-veto-bank-file'):
     bank_veto_file = configparser_value_to_file(wflow.cp, 'workflow-inspiral',
                                                 'bank-veto-bank-file')
+    bank_veto_file.description += '_BANK_VETO_BANK'
     bank_veto_file = _workflow.FileList([bank_veto_file])
     datafind_veto_files.extend(bank_veto_file)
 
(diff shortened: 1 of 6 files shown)


pannarale:
@spxiwh, @titodalcanton I added both of you as reviewers since there will be a combination of workflow and "pure PyGRB" content in this chain of PRs.
I will now watch the tests breaking and add the fixes to avoid them, but I will not push the full set of changes I made all at once because it is too much for one PR.

pycbc/workflow/jobsetup.py
```diff
@@ -275,21 +275,20 @@ def multi_ifo_coherent_job_setup(workflow, out_files, curr_exe_job,
...
+    for f in datafind_outs:
+        if 'IPN' in f.description:
+            ipn_sky_points = f
+            frame_files.remove(f)
+        elif 'vetoes' in f.description:
+            veto_file = f
+            frame_files.remove(f)
+        elif 'INPUT_BANK_VETO_BANK' in f.description:
+            bank_veto = f
+            frame_files.remove(f)
```
titodalcanton, line 291: A couple comments on this block:
1. It looks like `datafind_outs` and `frame_files` may contain items that do not correspond to frame files. The variable names should then be more general, e.g. `file_list`.
2. The string "IPN" is just three characters long. How high is the risk of clashing with random occurrences of the same string in other file names? Does it make sense to use "_IPN_"?
3. In the spirit of potentially helping debug issues arising from point 2, please add logging statements at each decision, e.g. "Using %s for the IPN sky points", "Using %s for the veto file".

pycbc/workflow/grb_utils.py
```diff
@@ -643,7 +626,7 @@ def make_pygrb_injs_tables(workflow, out_dir, bank_file, off_file, seg_files,
     # Handle input/output for onsource/offsource
     else:
         src_type = 'offsource-trigs'
-        if on_file is not None:
+        if on_file:
             src_type = 'onsource-trig'
             # Onsource input file (passed as File instance)
             onsource_file = on_file
```
titodalcanton, line 632: What is the point of defining this extra variable? You could just pass `on_file` to `add_input_opt()` no?

pycbc/workflow/jobsetup.py
pannarale: Thanks, @titodalcanton.  I renamed `frame_files` `input_files`.  The search is now for `IPN_SKY_POINTS` which is more robust and consistent with the naming that is set up [here](https://github.com/gwastro/pycbc/blob/master/bin/pygrb/pycbc_make_offline_grb_workflow#L272).
Do you think the logging is still necessary?  I'm asking because with `IPN_SKY_POINTS` it more robust and jobsetup.py has not `logging` usage so this may be an intended choice (@spxiwh?).

pycbc/workflow/grb_utils.py
pannarale: Ah, you mean `onsource_file`.  Yes, that is pointless.

pycbc/workflow/jobsetup.py
titodalcanton: I definitely would not expect clashes if you use IPN_SKY_POINTS, so indeed you can skip point 3.

pycbc/workflow/grb_utils.py
titodalcanton: Yes sorry I meant `onsource_file`. And I think I spotted a similar case a few lines above this one.

pannarale: `fm_file = inj_file`... there it is

pannarale: I found two of those.  I grep-ed `if ` and piped it to `grep _file` to look for more, but this is it.

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -130,14 +129,15 @@ else:
 
 sciSegs = segmentlistdict(sciSegs)
 if onSrc is None:
-    plot_met = make_grb_segments_plot(wflow, sciSegs, triggertime, triggername,
-                                      seg_dir, fail_criterion=offSrc)
+    plot_met = make_grb_segments_plot(
+        wflow, segmentlistdict(), triggertime, triggername,
+        seg_dir, fail_criterion=offSrc)
```
spxiwh, line 134: How can it make the plot without the segments??

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -40,11 +40,7 @@
 
 # All/most of these final imports will become obsolete with hdf5 switch
 try:
-    from ligo.lw import utils
-    from ligo.lw.table import Table
     from ligo.segments.utils import fromsegwizard
```
spxiwh, line 43: Can we get rid of `fromsegwizard` as well ... That one is even older than XML files for segments!!

spxiwh (approved):
This only touches PyGRB code, so if it does what you want, then no issues from me. I leave a couple of optional comments. Should also wait to hear approval from Tito.

titodalcanton (approved):
Looks good apart from several minor comments and questions.

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -234,12 +236,17 @@ if wflow.cp.has_option("workflow-condition_strain", "do-gating"):
         gated_frames.convert_to_lal_cache().tofile(
                             open(gated_cache.storage_path, "w"))
         datafind_files.append(gated_cache)
-
 datafind_veto_files.extend(datafind_files)
-ifo_list = sorted(sciSegs.keys())
-ifo = ifo_list[0]
-ifos = ''.join(ifo_list)
-wflow.ifos = ifos
```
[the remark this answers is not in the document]
titodalcanton: I saw the code that uses this after posting the comment. This looks correct because `wflow` has a `ifos_string` attribute which turns the list into a string.

bin/pygrb/pycbc_make_offline_grb_workflow
pannarale: Ah, good catch.  This was intended to be `segmentlistdict(sciSegs)` to make it uniform with lines 117 and 112.
To answer @spxiwh's question, this `segmentlistdict()` is actually needed on line 112 when no data is available.  I'm attaching an example (ignore the GRB name).  The plot will show you the onsource interval and the trigger-time, overlayed to emptiness. 
GRB190605974_segments-1

pycbc/results/pygrb_postprocessing_utils.py
pannarale: I'll remove the comment.  I was planning to wipe out `fromsegwizard` when I get to the way the segment files are handled in general (i.e., 3 txts vs 1 json file).

pannarale: I also removed the `try/except` to match `pycbc_grb_inj_finder` and `pycbc_grb_trig_combiner`.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -710,7 +605,10 @@ def construct_trials(seg_files, seg_dict, ifos, slide_dict, vetoes):
 
                 iter_int += 1
 
-    return trial_dict
+    total_trials = sum([len(trial_dict[slide_id]) for slide_id in slide_dict])
+    logging.info(f"{total_trials} trials generated.")
```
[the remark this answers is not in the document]
pannarale: Argh, I was getting this from pylint a while back, if I recall correctly.  I may have sprinkled it in a few places.  I'll `git grep` and undo.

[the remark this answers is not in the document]
pannarale: It turns out I'm in luck, but there are scripts in `all_sky_search` and pycbc-live ones that would need cleaning.

PR #4930, opened 2024-11-07
Save overlapping PSDs with inspiral and use precomputed psds for inspiral

bhooshan-gadre:
TOP-LEVEL SUMMARY: Save overlapping PSDs with inspiral and use precomputed psds for inspiral

This is a new feature and efficiency update

This change *may* affect but not necessarily if the option is not given: the offline search,  PyGRB

## Contents
1. Overlapping PSDs can be saved while running inspiral in calculate_psd format. 
2. Inspiral can also be run with a precomputed PSD file as an input. 
3. These PSD files can then later be merged. Thus, PSD computing workflow is not needed during offline search.
4. The saved PSD can be used to filter injections.
5.  Another use case is to use null stream PSD for filtering strain from other IFOs of ET

## Testing performed
1. Test that the code works. It saves PSDs and inspiral can ingest precomputed PSD as expected. 
2. Still both outputs of inspiral (with and w/o precomputed PSDs) do not match exactly.

### DIFF
--- a/pycbc/events/ranking.py
+++ b/pycbc/events/ranking.py
@@ -26,7 +26,7 @@ def newsnr(snr, reduced_x2, q=6., n=2.):
     reduced chi-squared values. See http://arxiv.org/abs/1208.3491 for
     definition. Previous implementation in glue/ligolw/lsctables.py
     """
-    nsnr = numpy.array(snr, ndmin=1, dtype=numpy.float64)
+    nsnr = numpy.array(numpy.abs(snr), ndmin=1, dtype=numpy.float64)
     reduced_x2 = numpy.array(reduced_x2, ndmin=1, dtype=numpy.float64)
 
     # newsnr is only different from snr if reduced chisq > 1
--- a/pycbc/psd/__init__.py
+++ b/pycbc/psd/__init__.py
@@ -241,8 +241,14 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
     psd_options.add_argument("--psdvar-high-freq", type=float, metavar="HERTZ",
                              help="Maximum frequency to consider in strain "
                              "bandpass.")
+    psd_options.add_argument("--precomputed-psd-file", metavar='FILE', type=str,
+                            help='Given HDF5 MERGE PSD file as the option, '
+                            'appropriate PSD for each strain segment will be '
+                            'chosen and use which is precomputed. PSD from the '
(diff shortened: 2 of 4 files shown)


pycbc/psd/read.py
```diff
@@ -187,3 +192,90 @@ def from_xml(filename, length, delta_f, low_freq_cutoff, ifo_string=None,
 
     return from_numpy_arrays(freq_data, noise_data, length, delta_f,
                              low_freq_cutoff)
+
+
+class PrecomputedTimeVaryingPSD(object):
+    def __init__(self, opt, length, delta_f, sample_rate):
```
ahnitz, line 198: @bhooshan-gadre I think this is a nice way to do this, but I think it might also be useful to be able to invoke this class by hand. Would it not make sense to have a from_cli class method that uses a option/arg  input and then a regular init method that could be invoked by hand ?

ahnitz (changes requested):

pycbc/psd/__init__.py
```diff
@@ -534,26 +575,45 @@ def associate_psds_to_segments(opt, fd_segments, gwstrain, flen, delta_f, flow,
...
-            if psd_seg.intersects(inp_seg):
-                curr_overlap = abs(inp_seg & psd_seg)
-                if curr_overlap > psd_overlap:
-                    psd_overlap = curr_overlap
-                    best_psd = psd
-        if best_psd is None:
-            err_msg = "No PSDs found intersecting segment!"
-            raise ValueError(err_msg)
-        fd_segment.psd = best_psd
+    if opt.precomputed_psd_file:
```
ahnitz, line 578: Wouldn't it make sense to have the precomputed psd be gotten from the pycbc.psd.from_cli method instead? That way it can be invoked more broadly.

pycbc/events/ranking.py
```diff
@@ -26,7 +26,7 @@ def newsnr(snr, reduced_x2, q=6., n=2.):
     reduced chi-squared values. See http://arxiv.org/abs/1208.3491 for
     definition. Previous implementation in glue/ligolw/lsctables.py
     """
-    nsnr = numpy.array(snr, ndmin=1, dtype=numpy.float64)
+    nsnr = numpy.array(numpy.abs(snr), ndmin=1, dtype=numpy.float64)
```
tdent, line 29: Is this change needed in the PR?

PR #4933, opened 2024-11-11
Passing vetoes to scripts in pygrb results sub workflow

pannarale:
This PR is the second in the series started in PR #4929.

This change affects: PyGRB

If this change breaks the standard automated test running `--help` for PyGRB plotting scripts, I will add some workarounds to avoid this.  If needed, these will likely be empty functions: the plotting scripts will be progressively renovated in the whole series of PRs.

## Motivation
Prior to these PRs, PyGRB was not using data vetoes.

## Contents
- The results workflow generator for PyGRB is now passing segments files and the vdf to all results scripts
- The utility functions involved in loading trigger/injeciton data in pygrb were generalised
- `pycbc_pygrb_plot_skygrid` was updated accordingly as a first example.

## Testing performed

### DIFF
--- a/bin/pygrb/pycbc_pygrb_results_workflow
+++ b/bin/pygrb/pycbc_pygrb_results_workflow
@@ -78,7 +78,7 @@ def labels_in_files_metadata(labels, rundir, file_paths):
 # Function to retrieve the segments plot (produced in the preprocessing stage)
 # that ensures its copy is saved in the output directory before returning to
 # the original working directory. Appropriate meta data is added to the plot.
-def display_seg_plot(output_dir, segments_dir):
+def display_seg_plot(output_dir, segment_dir):
     """Return the File of the segments plot (which was already produced during
     the pre-processing) after adding the appropriate metadata to it.
     """
@@ -90,7 +90,7 @@ def display_seg_plot(output_dir, segments_dir):
         'GRB' + wflow.cp.get('workflow', 'trigger-name') + '_segments.png'
     )
     segments_plot = _workflow.resolve_url_to_file(
-        os.path.join(args.segment_dir, segments_plot_name)
+        os.path.join(segment_dir, segments_plot_name)
     )
     segments_plot_path = segments_plot.storage_path
     im = Image.open(segments_plot_path)
@@ -134,7 +134,6 @@ parser.add_argument(
     "-i",
     "--inj-files",
     action="store",
-    default=None,
     nargs="+",
     help="Location(s) of input injection results file(s)",
 )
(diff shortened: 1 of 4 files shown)


pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -287,13 +287,21 @@ def _dataset_iterator(g, prefix=''):
 # =============================================================================
 # Functions to load triggers
 # =============================================================================
-def load_triggers(input_file, ifos, vetoes, rw_snr_threshold=None,
-                  slide_id=None):
+def _load_triggers(input_file, ifos, data_tag=None, rw_snr_threshold=None,
+                   slide_id=None):
     """Loads triggers from PyGRB output file, returning a dictionary"""
 
+    logging.info("Loading triggers.")
```
titodalcanton, line 294: Kind of a silly point, but looking a few lines below, I wonder if this should say "Loading injections" if `data_tag` is "injs"?

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -336,12 +354,29 @@ def load_triggers(input_file, ifos, vetoes, rw_snr_threshold=None,
...
             if trigs_dict[path].size == trigs['network/slide_id'][:].size:
-                trigs_dict[path] = slide_filter(trigs, trigs_dict[path],
-                                                slide_id=slide_id)
+                trigs_dict[path] = _slide_filter(trigs, trigs_dict[path],
+                                                 slide_id=slide_id)
 
     return trigs_dict
 
 
+# Wrapper function to load trigger/injection data
```
titodalcanton, line 363: Is it really necessary to introduce such a function? It seems you could just keep using `_load_triggers()`, and just add
```python
if not input_file:
    return None
```
at the very beginning.

bin/pygrb/pycbc_pygrb_results_workflow
```diff
@@ -218,6 +220,12 @@ if not set(inj_sets).issubset(eff_secs):
 inj_sets = [i.upper() for i in inj_sets]
 inj_files = labels_in_files_metadata(inj_sets, start_rundir, args.inj_files)
 
+# File instance of the veto file
+veto_file = args.veto_file
+if veto_file: 
+    veto_file = os.path.join(start_rundir, args.veto_file)
```
titodalcanton, line 226: Why is this necessary? Should we not assume that the workflow generator will pass a complete file path to `--veto-file`?

titodalcanton: Also not sure you really need the `resolve_url_to_file()` call below…

titodalcanton: Ah wait, this is the results workflow generator. Then it makes more sense.

pannarale: Correct, this is workflow generation stuff, so the files are getting converted to `File` instances.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: I blended the two functions together.
The idea of returning None was to avoid that check in all plotting scripts, which may or may not receive an `--inejection-file`.  I don't mind putting it there, i.e.,
```python
if args.injection_file:
    inj_data = load_data(...)
```

titodalcanton: I see, then I think it makes sense.

titodalcanton (approved):

PR #4937, opened 2024-11-13
Fix logic for determining frame type in `pycbc_make_skymap`

titodalcanton:
Fixes a crash in `pycbc_make_skymap` triggered by using simulated strain and a detector that is not H1, L1 or V1.

This is a bug fix which affects only `pycbc_make_skymap`.

## Motivation

A student reported that adding KAGRA to our example produces this error:
```
  File "/home/tito/virtualenvs/pycbc/bin/pycbc_make_skymap", line 574, in <module>
    main(opt.trig_time, opt.mass1, opt.mass2,
  File "/home/tito/virtualenvs/pycbc/bin/pycbc_make_skymap", line 111, in main
    frame_types[ifo] = default_frame_type(mean_trig_time, ifo)
  File "/home/tito/virtualenvs/pycbc/bin/pycbc_make_skymap", line 56, in default_frame_type
    raise ValueError('Detector {} not supported at time {}'.format(ifo, time))
ValueError: Detector K1 not supported at time 1272790260.0
```

## Contents

The logic that determines the frame type to pass to `pycbc_single_template` was wrong in a trivial way. It was trying to guess the frame type based on the detector and trigger time even when the fake strain is being used, i.e. when no guessing is necessary. I reworked the logic to only guess the frame type when necessary.

N/A

## Testing performed

I ran the example before and after the fix, first with the default HLV network and then by adding KAGRA as well. Here are the results, just for fun:

image

image

## Additional notes

Hopefully we will see KAGRA at 80 Mpc soon 🤞

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -106,14 +106,6 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
         frame_types = {}
     if channel_names is None:
         channel_names = {}
-    for ifo in ifos:
-        if ifo not in frame_types:
-            frame_types[ifo] = default_frame_type(mean_trig_time, ifo)
-        if ifo not in channel_names:
-            if fake_strain[ifo] is not None:
-                channel_names[ifo] = ifo + ':FAKE_DATA'
-            else:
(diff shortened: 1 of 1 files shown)


spxiwh (approved):
Thanks!

PR #4941, opened 2024-11-14
Handling vetoes and timeslides in pygrb mini followups

pannarale:
This PR is the third in the series started in PR #4929.

This change affects: PyGRB

If this change breaks the standard automated test running `--help` for PyGRB plotting scripts, I will add some workarounds to avoid this.  If needed, these will likely be empty functions: the plotting scripts will be progressively renovated in the whole series of PRs.

## Motivation
Now that the workflow generator passes around the veto definer file, the mini followups need to handle it use it.  In checking `pycbc_pygrb_minifollowups` I noticed that it is better to enforce the correct timesmlide is being plotted.

## Contents
- Enable passing veto and segments files to `pycbc_pygrb_minifollowups` and the jobs it prepares.
- Figure out the correct timeslide to be plotted (e.g., zero-lag for the loudest onsource event) and ensure it is the one used in time series plots.
- Somewhat unrelated, but in preparation of other PRs, the utility `construct_trials` was improved and the utility `extract_basic_trig_properties ` was generalized to `extract_trig_properties` in which the user specifies what datasets need to be extracted from the trigger file.

## Testing performed

### DIFF
--- a/bin/pygrb/pycbc_pygrb_minifollowups
+++ b/bin/pygrb/pycbc_pygrb_minifollowups
@@ -56,22 +56,22 @@ def add_wiki_row(outfile, cols):
 
 
 def make_timeseries_plot(workflow, trig_file, snr_type, central_time,
-                         shift_time, out_dir, ifo=None, seg_files=None,
-                         veto_file=None, slide_id=0, tags=None):
+                         out_dir, ifo=None, seg_files=None,
+                         veto_file=None, tags=None):
     """Adds a node for a timeseries of PyGRB results to the workflow"""
 
     tags = [] if tags is None else tags
 
-    # If necessary, overwrite the slide-id configuration option to ensure only
-    # the correct timeslide is used in these follow-up plots
+    # Ensure that zero-lag data is used in these follow-up plots and store
+    # the slide-id option originally provided
     orig_slide_id = None
     if workflow.cp.has_option('pygrb_plot_snr_timeseries','slide-id'):
         orig_slide_id = workflow.cp.get('pygrb_plot_snr_timeseries','slide-id')
-        workflow.cp.add_options_to_section(
-            'pygrb_plot_snr_timeseries',
-            [('slide-id', str(slide_id))],
-            True
-        )
+    workflow.cp.add_options_to_section(
+        'pygrb_plot_snr_timeseries',
+        [('slide-id', '0')],
+        True
+    )
 
     # Initialize job node with its tags
     grb_name = workflow.cp.get('workflow', 'trigger-name')
@@ -97,10 +97,7 @@ def make_timeseries_plot(workflow, trig_file, snr_type, central_time,
     node.add_opt('--y-variable', snr_type)
     if ifo is not None:
         node.add_opt('--ifo', ifo)
-    reset_central_time = shift_time - central_time
-    # Horizontal axis range the = prevents errors with negative times
-    x_lims = str(-5.+reset_central_time)+','+str(reset_central_time+5.)
-    node.add_opt('--x-lims='+x_lims)
+    node.add_opt('--x-lims=-5.,5.')
     # Plot title
     if ifo is not None:
         title_str = f"'{ifo} SNR at {central_time:.3f} (s)'"
@@ -113,13 +110,16 @@ def make_timeseries_plot(workflow, trig_file, snr_type, central_time,
     # Add job node to workflow
     workflow += node
 
-    # Revert back the original slide-id option if necessary
+    # Revert the config parser back to how it was given
     if orig_slide_id:
         workflow.cp.add_options_to_section(
             'pygrb_plot_snr_timeseries',
             [('slide-id', orig_slide_id)],
             True
         )
+    else:
+        workflow.cp.remove_option('pygrb_plot_snr_timeseries','slide-id')
+
 
     return node.output_files
 
@@ -200,9 +200,10 @@ except:
 # Loop over triggers/injections to be followed up
 for num_event in range(num_events):
     files = FileList([])
-    logging.info('Processing event: %s', num_event)
+    logging.info('Processing event: %s', num_event+1)
     gps_time = fp['GPS time'][num_event]
     gps_time = gps_time.astype(float)
+    tags = args.tags + [str(num_event+1)]
     if wiki_file:
         row = []
         for key in fp.keys():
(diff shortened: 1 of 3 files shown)


bin/pygrb/pycbc_pygrb_minifollowups
```diff
@@ -163,29 +208,41 @@ for num_event in range(num_events):
...
+                                          snr_type, gps_time, gps_time,
+                                          args.output_dir, ifo=None,
+                                          seg_files=seg_files,
+                                          veto_file=veto_file,
+                                          slide_id=0,
+                                          tags=args.tags + [str(num_event)])
+        for ifo in ifos:
+            files += mini.make_qscan_plot(workflow, ifo, gps_time,
+                                          args.output_dir,
+                                          tags=args.tags + [str(num_event)])
```
pannarale, line 224: This mostly swaps the `if` and `else`,

bin/pygrb/pycbc_pygrb_minifollowups
```diff
@@ -163,29 +208,41 @@ for num_event in range(num_events):
...
     # Handle off/on-source loudest triggers follow-up (which are on slid data)
-    if not is_injection_followup:
+    else:
+        # If the file has information about slides, this is the follow up of an
+        # offsource event, otherwise it is the loudest onsource event, and
+        # zero-lag data must be used
+        slide_id = 0
+        if 'Slide Num' in fp.keys():
+            slide_id = fp['Slide Num'][num_event]
```
pannarale, line 232: I think there is a better way to do this.  I will give it a try.

bin/pygrb/pycbc_pygrb_minifollowups
```diff
@@ -142,50 +173,68 @@ trig_file = resolve_url_to_file(os.path.abspath(args.trig_file))
...
 
-
 # Loop over triggers/injections to be followed up
 for num_event in range(num_events):
     files = FileList([])
-    logging.info('Processing event: %s', num_event)
+    logging.info('Processing event: %s', num_event+1)
     gps_time = fp['GPS time'][num_event]
     gps_time = gps_time.astype(float)
+    tags = args.tags + [str(num_event+1)]
```
pannarale, line 204: Getting these to match the row numbering we see in the loudest offsource event / quite injections tables.

bin/pygrb/pycbc_pygrb_minifollowups
```diff
@@ -142,50 +173,68 @@ trig_file = resolve_url_to_file(os.path.abspath(args.trig_file))
...
             files += mini.make_qscan_plot(workflow, ifo, gps_time,
                                           args.output_dir,
-                                          tags=args.tags + [str(num_event)])
+                                          tags=tags)
+    # Handle off/on-source loudest triggers follow-up (which may be on slid
+    # data in the case of the off-source)
+    else:
+        for i_ifo, ifo in enumerate(ifos):
+            time_shift = fp[ifo+' time shift (s)'][num_event]
+            ifo_time = gps_time + time_shift
```
pannarale, line 228: The change from - to + is intentional here.  This was sliding the wrong way!

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -460,38 +457,35 @@ def sort_trigs(trial_dict, trigs, slide_dict, seg_dict):
...
+                found_trigs[key][slide_id] = get_coinc_snr(trigs)[indices] \
+                    if key == 'network/coincident_snr' else trigs[key][indices]
+            else:
+                found_trigs[key][slide_id] = numpy.asarray([])
 
-    logger.info("Time, SNR, and BestNR of triggers extracted.")
+    logging.info("Triggers information extracted.")
 
-    return trig_time, trig_snr, trig_bestnr
+    return found_trigs
```
MarcoCusinato, line 488: `pycbc_pygrb_efficiency`, `pycbc_pygrb_page_tables`, and `pycbc_pygrb_plot_stats_distribution` are using this function and need three outputs not to break.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -460,38 +457,35 @@ def sort_trigs(trial_dict, trigs, slide_dict, seg_dict):
...
-    """Extract and store as dictionaries time, SNR, and BestNR of
-    time-slid triggers"""
+def extract_trig_properties(trial_dict, trigs, slide_dict, seg_dict, keys):
+    """Extract and store as dictionaries specific keys of time-slid
+    triggers (trigs) compatibly with the trials dictionary (trial_dict)"""
 
     # Sort the triggers into each slide
     sorted_trigs = sort_trigs(trial_dict, trigs, slide_dict, seg_dict)
-    logger.info("Triggers sorted.")
+    n_surviving_trigs = sum([len(i) for i in sorted_trigs.values()])
```
MarcoCusinato, line 466: I would leave the log info here

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -460,38 +457,35 @@ def sort_trigs(trial_dict, trigs, slide_dict, seg_dict):
 # =============================================================================
 # Extract trigger properties and store them as dictionaries
 # =============================================================================
-def extract_basic_trig_properties(trial_dict, trigs, slide_dict, seg_dict,
-                                  opts):
-    """Extract and store as dictionaries time, SNR, and BestNR of
-    time-slid triggers"""
+def extract_trig_properties(trial_dict, trigs, slide_dict, seg_dict, keys):
+    """Extract and store as dictionaries specific keys of time-slid
```
MarcoCusinato, line 461: I agree with the change of name of the function. But since this and its precursor do two different things, shouldn't we keep them both? (see also comment on the output)

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -460,38 +457,35 @@ def sort_trigs(trial_dict, trigs, slide_dict, seg_dict):
...
-            trig_time[slide_id] = trigs['network/end_time_gc'][
-                indices]
-            trig_snr[slide_id] = trigs['network/coherent_snr'][
-                indices]
-        else:
-            trig_time[slide_id] = numpy.asarray([])
-            trig_snr[slide_id] = numpy.asarray([])
-        trig_bestnr[slide_id] = reweightedsnr_cut(
-            trigs['network/reweighted_snr'][indices],
-            opts.newsnr_threshold)
```
MarcoCusinato, line 490: Aren't we using this anymore? If just one of the functions is kept, maybe one could return the `found_trigs`as well as the `bestNR`?

MarcoCusinato (changes requested):
The changes on the `pygrb_postprocessing_utils.py` as right now would break  `pycbc_pygrb_efficiency`, `pycbc_pygrb_page_tables`, and `pycbc_pygrb_plot_stats_distribution`. So these three files need some adjustments

pannarale:
> The changes on the `pygrb_postprocessing_utils.py` as right now would break `pycbc_pygrb_efficiency`, `pycbc_pygrb_page_tables`, and `pycbc_pygrb_plot_stats_distribution`. So these three files need some adjustments

Yes, correct, and the changes will be part of the next PRs in this series.  If I were to produce a single PR, it would be too big to parse by the reviewer(s).

bin/pygrb/pycbc_pygrb_minifollowups
```diff
@@ -71,28 +83,42 @@ def make_timeseries_plot(workflow, trig_file, snr_type, central_time,
...
                              '--output-file', tags=extra_tags)
     # Quantity to be displayed on the y-axis of the plot
     node.add_opt('--y-variable', snr_type)
     if ifo is not None:
         node.add_opt('--ifo', ifo)
-    reset_central_time = shift_time - central_time
-    # Horizontal axis range the = prevents errors with negative times
-    x_lims = str(-5.+reset_central_time)+','+str(reset_central_time+5.)
-    node.add_opt('--x-lims='+x_lims)
+    node.add_opt('--x-lims=-5.,5.')
```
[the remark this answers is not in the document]
pannarale: This is centred at the (possibly shifted) gps time of the trigger.

E.g., if there is loud offsource trigger in slide 1 at time t0, and we have an onsource window of 6s and H1, L1, V1, the 3 follow up plots will show:
1.  the H1 timeseries with a horizontal axis with times relative to t0 (so 5s before and after)
2. the H1 timeseries with a horizontal axis with times relative to t0 + 6s, and
3. the V1 timeseries with a horizontal axis with times relative to t0 + 12s.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: Correct, in the development version these scripts will call `extract_trig_properties` to get these quantities.  But, I am breaking down the many changes in multiple PRs so that the diffs may be parsed with reasonable effort: one big PR would be complicated to handle for a reviewer.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: See the answer above.  The call to the old function will be removed throughout PyGRB.

pycbc/results/pygrb_postprocessing_utils.py
MarcoCusinato: I agree with breaking it down, but approving this PR before any of the others would find difficult to run workflows without issues.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: I removed this because the logging info below is more informative, while is still telling the user the triggers were sorted.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: The user will get the `'network/reweighted_snr'` by calling `extract_trig_properties` with `'network/reweighted_snr'` among the `keys`.   So this function is not returning the `found_trigs` dictionary with the `keys` requested (and `'network/reweighted_snr'` may be one of these).

pycbc/results/pygrb_postprocessing_utils.py
pannarale: This was true for quite a while until PR #4909, when PyCBC allowed to generate a (working) PyGRB workflow, but without vetoes.  As for previous "PR series" that took us to that point, the idea is that we are enabling a new big feature and, by breaking changes down, accepting that the intermediate states of PyCBC between PR #4909 and the completion of this PR review work will break PyGRB again.  At that point we will have a PyCBC where  PyGRB workflows are fully working and with vetoes.  It will then be the right moment to have a new CI/CD test with a small PyGRB search and hence stop this modus operandi.

PR #4942, opened 2024-11-14
Fix bug when plotting maximum likelihood point in `create_multidim_plot`

acorreia61201:
This is a bugfix to the `create_multidim_plot` function in `scatter_histograms.py`. Specifying `--plot-maxl` does not plot the expected maximum likelihood point when used in conjunction with `--plot-scatter` in multidimensional plots. Currently, `plot-maxl` selects the index of the max likelihood sample before `plot-scatter` sorts the samples by their z argument. When both options are specified, the selected maxL index will generally not correspond to the maxL sample in the newly sorted array.

This PR moves the indexing selection for `plot-maxl` after the samples are sorted by `plot-scatter`. This should have no effect if using either option separately with codes using `create_multidim_plot` (e.g. `pycbc_inference_plot_posterior`), but should now correctly plot the maxL sample if both are specified.


cdcapano (approved):
Thanks for fixing

PR #4948, opened 2024-11-19
Add a feature to generate frequency domain waveforms at reduced frequencies

Kanchan-05:

This is a: new feature 

This change affects: pycbc_brute_bank script

## Motivation

Template bank generation using pycbc_brute_bank is computationally intensive for very long-duration signals, primarily due to the match calculations. The computation time can be reduced by generating frequency-domain signals at lower frequencies specifically for the match calculations.

N/A

## Testing performed

N/A

## Additional notes

### DIFF
--- a/bin/bank/pycbc_brute_bank
+++ b/bin/bank/pycbc_brute_bank
@@ -53,7 +53,7 @@ parser.add_argument('--approximant',  required=False,
 parser.add_argument('--minimal-match', default=0.97, type=float)
 parser.add_argument('--buffer-length', default=2, type=float,
     help='size of waveform buffer in seconds')
-parser.add_argument('--buffer-high-pass-length', default=None, type=float,
+parser.add_argument('--full-resolution-buffer-length', default=None, type=float,
     help='size of waveform buffer in seconds')
 parser.add_argument('--max-signal-length', type= float,
                     help="When specified, it cuts the maximum length of the waveform model to the lengh provided")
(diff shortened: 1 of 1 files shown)


ahnitz:
The main purpose of this is for *time domain signals* since they must be generated at full resolution first before conversion to the frequency-domain. This adds the ability to then decimate their frequency-domain representation and use that for match calculations. 

We do this already for frequency-domain waveform by just setting the buffer length to something shorter than the actual duration. However, that will now work for time domain signals.

bin/bank/pycbc_brute_bank
```diff
@@ -53,6 +53,8 @@ parser.add_argument('--approximant',  required=False,
 parser.add_argument('--minimal-match', default=0.97, type=float)
 parser.add_argument('--buffer-length', default=2, type=float,
     help='size of waveform buffer in seconds')
+parser.add_argument('--buffer-high-pass-length', default=None, type=float,
```
ahnitz, line 56: The name / help might be clearer. Maybe --full-resolution-buffer-length or maybe you have a better idea? Suggestions?

Kanchan-05: --full-resolution-buffer-length sounds good! If it feels too lengthy, perhaps --full-res-buffer-length could be a more concise alternative.

Kanchan-05:
@spxiwh I looked at the cProfile output (see attachment) for low-mass template bank generation, and it turns out the FFT operations for match calculations are more time-consuming than generating waveforms in the time domain.  

If we go with the reduced frequency approach, we can reduce the cost of match calculations quite a bit. After that, waveform generation becomes the main bottleneck, but I think we can handle that by parallelizing it across multiple cores.

cprofile

ahnitz (commented):
Can you double check the code style? Codeclimate doesn't appear to be automatically running. Make sure that your code is following PEP8, etc.

ahnitz (approved):

spxiwh:
Okay, so 3600 waveform generations and 160000 FFTs … Makes sense then!

I do note that you are using the numpy backend for FFTs, which is slow, and not using the faster Class based FFTs (there is a “do cached FFTs” function that I wrote to leverage the class based stuff in an easy way, which would probably be useful here) … This code would be an interesting test case for the CUPY backend I am playing with as well (especially if using a waveform that can be generated on a GPU).

Kanchan-05:
@spxiwh That's interesting. Could you please point me to the class that you mentioned?

spxiwh:
https://github.com/gwastro/pycbc/blob/b7cb8bccef48deebfbbd677df7f11902addfaace/pycbc/strain/strain.py#L1330 is the function I mean. This won't do anything with the numpy FFT backend (which is always slow), but if you install+use MKL or FFTW it will help quite a bit.

PR #4950, opened 2024-11-20
Final version of `pycbc_pygrb_plot_chisq_veto` and `pycbc_pygrb_plot_coh_ifosnr`

pannarale:
This PR is the fifth in the series started in PR #4929.  Much like PR #4947 did with `pycbc_pygrb_plot_null_stats`, this PR provides a version of `pycbc_pygrb_plot_chisq_veto` and `pycbc_pygrb_plot_coh_ifosnr` that uses veto and segments files, as well as the utilities introduced recently to streamline the PyGRB results scripts.

This change affects: PyGRB

Should this change break the standard automated test running `--help` for PyGRB plotting scripts, I will add some workarounds to avoid this.  If needed, these will likely be empty functions: the plotting scripts will be progressively renovated in the whole series of PRs.

## Motivation
Now that the workflow generator passes the veto definer file to the jobs where needed, its usage in the PyGRB results scripts is possible.

## Testing performed

### DIFF
--- a/bin/pygrb/pycbc_pygrb_plot_chisq_veto
+++ b/bin/pygrb/pycbc_pygrb_plot_chisq_veto
@@ -150,7 +150,7 @@ if opts.plot_title is None:
     if veto_type != 'network':
         opts.plot_title = ifo + opts.plot_title
     if snr_type == 'single':
-        opts.plot_title += " vs " + ifo + " SNR"
+        opts.plot_title += f" vs {ifo} SNR"
     else:
         opts.plot_title += " vs " + snr_type.capitalize() + " SNR"
 if opts.plot_caption is None:
--- a/bin/pygrb/pycbc_pygrb_plot_chisq_veto
+++ b/bin/pygrb/pycbc_pygrb_plot_chisq_veto
@@ -262,7 +262,7 @@ logging.info("Plotting...")
 # Determine x-axis values of triggers and injections
 # Default is coherent SNR
 x_label = ifo if snr_type == 'single' else snr_type.capitalize()
-x_label = x_label + " SNR"
+x_label += " SNR"
 
 # Determine the minumum and maximum SNR value we are dealing with
 x_min = 0.9*plu.axis_min_value(found_trigs[x_key], found_injs[x_key],
--- a/bin/pygrb/pycbc_pygrb_plot_chisq_veto
+++ b/bin/pygrb/pycbc_pygrb_plot_chisq_veto
@@ -276,7 +276,7 @@ y_min = 0.9*plu.axis_min_value(found_trigs[y_key], found_injs[y_key],
 y_max = 1.1*plu.axis_max_value(found_trigs[y_key], found_injs[y_key],
                                found_missed_file)
 
-# Determine y-axis minimum value and label
+# Determine y-axis label
 y_label = "Network power chi-square" if veto_type == 'network' \
     else ifo + " Single " + veto_labels[veto_type].lower() + " chi-square"
 
--- a/bin/pygrb/pycbc_pygrb_plot_chisq_veto
+++ b/bin/pygrb/pycbc_pygrb_plot_chisq_veto
@@ -278,7 +278,7 @@ y_max = 1.1*plu.axis_max_value(found_trigs[y_key], found_injs[y_key],
 
 # Determine y-axis label
 y_label = "Network power chi-square" if veto_type == 'network' \
-    else ifo + " Single " + veto_labels[veto_type].lower() + " chi-square"
+    else f"{ifo} Single {veto_labels[veto_type].lower()} chi-square"
 
 # Determine contours for plots
 conts = None
--- a/bin/pygrb/pycbc_pygrb_plot_chisq_veto
+++ b/bin/pygrb/pycbc_pygrb_plot_chisq_veto
@@ -253,7 +253,8 @@ if len(keys) == 3:
     found_trigs[keys[1]] /= found_trigs[keys[2]]
     found_injs[keys[1]] /= found_injs[keys[2]]
 
-# Floor single IFO chi-square at 0.005
+# Single detector chi-squares are initialized to 0: we floor possible
+# remaining 0s to 0.005 to avoid asking for logarithms of 0 in the plot
 numpy.putmask(found_trigs[y_key], found_trigs[y_key] == 0, 0.005)
 
 # Generate plots
--- a/bin/pygrb/pycbc_pygrb_plot_chisq_veto
+++ b/bin/pygrb/pycbc_pygrb_plot_chisq_veto
@@ -213,10 +213,10 @@ else:
     y_key = 'network/my_network_chisq'
 
 keys = [x_key, y_key]
-# The network chi-square is already normalized so dof_key is not needed
+# The network chi-square is already normalized so it does not require a key
+# for the number of degrees of freedom
 if veto_type != 'network':
-    dof_key = y_key + '_dof'
-    keys += [dof_key]
+    keys += [y_key + '_dof']
 
 # Extract needed trigger properties and store them as dictionaries
 # Based on trial_dict: if vetoes were applied, trig_* are the veto survivors
--- a/bin/pygrb/pycbc_pygrb_plot_chisq_veto
+++ b/bin/pygrb/pycbc_pygrb_plot_chisq_veto
@@ -152,7 +152,7 @@ if opts.plot_title is None:
     if snr_type == 'single':
         opts.plot_title += f" vs {ifo} SNR"
     else:
-        opts.plot_title += " vs " + snr_type.capitalize() + " SNR"
+        opts.plot_title += f" vs {snr_type.capitalize()} SNR"
 if opts.plot_caption is None:
     opts.plot_caption = ("Blue crosses: background triggers. ")
     if found_missed_file:
--- a/bin/pygrb/pycbc_pygrb_plot_chisq_veto
+++ b/bin/pygrb/pycbc_pygrb_plot_chisq_veto
@@ -48,79 +48,6 @@ __program__ = "pycbc_pygrb_plot_chisq_veto"
 # =============================================================================
 # Functions
 # =============================================================================
-# Function to load trigger data: includes applying cut in reweighted SNR
-def load_data(input_file, ifos, vetoes, opts, injections=False, slide_id=None):
-    """Load data from a trigger/injection file"""
-
-    snr_type = opts.snr_type
-    veto_type = opts.y_variable
-
-    # Initialize the dictionary
-    data = {}
-    data[snr_type] = None
-    data[veto_type] = None
-    data['dof'] = None
-
-    # Ensure that newtwork power chi-square plots show all the data to see
-    # the impact of the reweighted SNR cut
-    rw_snr_threshold = 0. if veto_type=='network' else opts.newsnr_threshold
-
-    if input_file:
-        if injections:
-            logging.info("Loading injections...")
-            # This will eventually become load_injections
-            trigs_or_injs = \
-                ppu.load_triggers(input_file, ifos, vetoes,
-                                  rw_snr_threshold=rw_snr_threshold,
-                                  slide_id=slide_id)
-        else:
-            logging.info("Loading triggers...")
-            trigs_or_injs = \
-                ppu.load_triggers(input_file, ifos, vetoes,
-                                  rw_snr_threshold=rw_snr_threshold,
-                                  slide_id=slide_id)
-
-        # Count surviving points
-        num_trigs_or_injs = len(trigs_or_injs['network/reweighted_snr'])
-
-        if snr_type in ['coherent', 'null', 'reweighted']:
-            data[snr_type] = trigs_or_injs['network/%s_snr' % snr_type][:]
-        elif snr_type == 'single':
-            key = opts.ifo + '/snr'
-            data[snr_type] = trigs_or_injs[key][:]
-
-        # Calculate coincident SNR
-        elif snr_type == 'coincident':
-            data[snr_type] = ppu.get_coinc_snr(trigs_or_injs)
-
-        # Tags to find vetoes in HDF files
-        veto_tags = {'power': 'chisq',
-                     'bank': 'bank_chisq',
-                     'auto': 'auto_chisq',
-                     'network': 'my_network_chisq'}
-
-        # This chi-square is already normalized
-        if veto_type == 'network':
-            chisq_key = 'network/my_network_chisq'
-            data['dof'] = 1.
-        else:
-            chisq_key = opts.ifo + '/' + veto_tags[veto_type]
-            dof_key = '%s/%s_dof' % (opts.ifo, veto_tags[veto_type])
-            data['dof'] = trigs_or_injs[dof_key][:]
-
-        # Normalize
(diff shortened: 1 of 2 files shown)


bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -243,54 +170,115 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
+    )
+
+# Gather injections found surviving vetoes
+found_injs, *_ = ppu.apply_vetoes_to_found_injs(
+    opts.found_missed_file,
+    inj_data,
+    ifos,
+    veto_file=opts.veto_file,
+    keys=keys
+)
```
pannarale, line 243: This is equivalent to what is now in pycbc_pygrb_plot_null_stats.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
```diff
@@ -230,65 +141,158 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
+# Generate segments dictionary
+segment_dict = ppu.load_segment_dict(trig_file)
+
+# Construct trials removing vetoed times
+trial_dict, total_trials = ppu.construct_trials(
+    opts.seg_files,
+    segment_dict,
+    ifos,
+    slide_dict,
+    opts.veto_file
```
pannarale, line 159: This is equivalent to what is now in `pycbc_pygrb_plot_null_stats`.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
```diff
@@ -230,65 +141,158 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
 
-# Extract trigger data
-trig_data = load_data(trig_file, ifos, vetoes, opts, slide_id=opts.slide_id)
+# Load triggers/injections (apply reweighted SNR cut, not vetoes)
+trig_data = ppu.load_data(trig_file, ifos, data_tag='trigs',
+                          rw_snr_threshold=opts.newsnr_threshold,
+                          slide_id=opts.slide_id)
+inj_data = ppu.load_data(found_missed_file, ifos, data_tag='injs',
+                         rw_snr_threshold=opts.newsnr_threshold,
+                         slide_id=0)
```
pannarale, line 168: This is equivalent to what is now in `pycbc_pygrb_plot_null_stats`.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
```diff
@@ -314,7 +318,7 @@ if loudness_index < 2:
     polyy.extend([limy, limy])
     ax.fill(polyx, polyy, color='#dddddd')
 opts.plot_title += (
-    " (%s average sensitivity)" % loudness_labels[loudness_index]
+    " (" + loudness_labels[loudness_index] + " average sensitivity)"
```
titodalcanton, line 321:
```suggestion
    f" ({loudness_labels[loudness_index]} average sensitivity)"
```

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -219,13 +146,13 @@ veto_labels = {'network': "Network Power",
...
-    opts.plot_title = " %s Chi Square" % veto_labels[veto_type]
+    opts.plot_title = veto_labels[veto_type] + " Chi Square"
     if veto_type != 'network':
         opts.plot_title = ifo + opts.plot_title
     if snr_type == 'single':
-        opts.plot_title += " vs %s SNR" % (ifo)
+        opts.plot_title += " vs " + ifo + " SNR"
     else:
-        opts.plot_title += " vs %s SNR" % snr_type.capitalize()
+        opts.plot_title += " vs " + snr_type.capitalize() + " SNR"
```
titodalcanton, line 155:
```suggestion
        opts.plot_title += f" vs {snr_type.capitalize()} SNR"
```

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -243,54 +170,115 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
+    x_key = ifo + '/snr'
+else:
+    x_key = 'network/' + snr_type + '_snr'
+# Dataset name for the vertical direction and for normalization
+if veto_type == 'power':
+    y_key = opts.ifo + '/chisq'
+elif veto_type in ['bank', 'auto']:
+    y_key = opts.ifo + '/' + veto_type +'_chisq'
+else:
+    y_key = 'network/my_network_chisq'
```
titodalcanton, line 213: `my_network_chisq` sounds like a dummy placeholder value. Does it really exist?

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -243,54 +170,115 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
+
+keys = [x_key, y_key]
+# The network chi-square is already normalized so dof_key is not needed
+if veto_type != 'network':
+    dof_key = y_key + '_dof'
+    keys += [dof_key]
```
titodalcanton, line 219: Looks like `dof_key` is not used anywhere else, so I would avoid defining `dof_key` and turn these lines into a single line.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
MarcoCusinato: Yes it does, it is the network chi square

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -243,54 +170,115 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
+y_min = 0.9*plu.axis_min_value(found_trigs[y_key], found_injs[y_key],
                                found_missed_file)
-y_max = 1.1*plu.axis_max_value(trig_data[veto_type], inj_data[veto_type],
+y_max = 1.1*plu.axis_max_value(found_trigs[y_key], found_injs[y_key],
                                found_missed_file)
 
 # Determine y-axis minimum value and label
 y_label = "Network power chi-square" if veto_type == 'network' \
-    else "%s Single %s chi-square" % (ifo, veto_labels[veto_type].lower())
+    else ifo + " Single " + veto_labels[veto_type].lower() + " chi-square"
```
titodalcanton, line 281:
```suggestion
    else f"{ifo} Single {veto_labels[veto_type].lower()} chi-square"
```
Though I think here it is more readable to do
```python
if veto_type == 'network':
    y_label = "Network power chi-square"
else:
    y_label = f"{ifo} Single {veto_labels[veto_type].lower()} chi-square"
```
Also, the comment just above is incorrect in saying that the minimum value is being determined here.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
```diff
@@ -230,65 +141,158 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
+    )
 
-# Extract (or initialize) injection data
-inj_data = load_data(found_file, ifos, vetoes, opts, injections=True, slide_id=0)
+# Complete the dictionary found_trigs
+# 1) Do not assume individual IFO and network event_ids are sorted the same way
+for ifo in ifos:
+    sorted_ifo_ids = numpy.array(
+        [
+            numpy.where(found_trigs[ifo+'/event_id'] == idx)[0][0]
```
titodalcanton, line 202: `where()` is overcomplicated for this use case, as it just calls `nonzero()`.
```suggestion
            numpy.nonzero(found_trigs[ifo + '/event_id'] == idx)[0][0]
```

MarcoCusinato (commented):
I agree with @titodalcanton  on the chi `square == 0`. Also minor comment, would it be possible to rename the variable `senstvty` and `ifo_senstvty`  to `sensitivity` in `pycbc_pygrb_plot_coh_ifosnr`?

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
```diff
@@ -230,65 +141,158 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
+
+# 3) Calculate the mean, max, and min sigmas
+for ifo in ifos:
+    try:
+        sigma_norm = found_trigs[ifo+'/sigmasq'] / found_trigs['sigma_tot']
+        found_trigs[ifo+'/sigma_mean'] = sigma_norm.mean()
+        if ifo == opts.ifo:
+            found_trigs['sigma_max'] = sigma_norm.max()
+            found_trigs['sigma_min'] = sigma_norm.min()
+    except ValueError:
```
titodalcanton, line 231: I do not see what could cause a ValueError in the lines above.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
titodalcanton: In that case, I suggest an explicit `elif veto_type == whatever_thats_called:` and then catch everything else (presumably an invalid value) with an exception.

pannarale: @titodalcanton, do we really need to do that if we have
```
parser.add_argument("-y", "--y-variable", required=True,
                    choices=['network', 'bank', 'auto', 'power'],
                    help="Quantity to plot on the vertical axis.")
```
[...]
```
veto_type = opts.y_variable
```
?

bin/pygrb/pycbc_pygrb_plot_chisq_veto
```diff
@@ -243,54 +170,115 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
 
-# Sanity checks
-if trig_data[snr_type] is None and inj_data[snr_type] is None:
-    err_msg = "No data to be plotted on the x-axis was found"
-    raise RuntimeError(err_msg)
-if trig_data[veto_type] is None and inj_data[veto_type] is None:
-    err_msg = "No data to be plotted on the y-axis was found"
-    raise RuntimeError(err_msg)
+# Floor single IFO chi-square at 0.005
+numpy.putmask(found_trigs[y_key], found_trigs[y_key] == 0, 0.005)
```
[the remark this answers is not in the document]
pannarale: This a "trick" inherited from the old code (`pylal_cbc_cohptf_sbv_plotter `) to avoid taking logarithms of 0.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
```diff
@@ -230,65 +141,158 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
+    antenna = Detector(ifo)
+    ifo_f_resp = ppu.get_antenna_responses(
+        antenna,
+        found_trigs['network/ra'],
+        found_trigs['network/dec'],
+        found_trigs['network/end_time_gc']
+    )
+    # Get the average for f_resp_mean and calculate sigma_tot
+    found_trigs[ifo+'/f_resp_mean'] = ifo_f_resp.mean()
+    found_trigs['sigma_tot'] += found_trigs[ifo+'/sigmasq'] * ifo_f_resp
```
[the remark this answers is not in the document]
pannarale: `ifo_f_resp` is a 1D numpy array containing f+^2 + fx^2 values.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
pannarale: This essentially never happens, but you could be loading 0 triggers, and then
```
found_trigs['sigma_tot'] = numpy.zeros(len(found_trigs[x_key]))
```
is empty so that min and max give you a ValueError.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
```diff
@@ -230,65 +141,158 @@ outdir = os.path.split(os.path.abspath(opts.output_file))[0]
...
+        sigma_norm = found_trigs[ifo+'/sigmasq'] / found_trigs['sigma_tot']
+        found_trigs[ifo+'/sigma_mean'] = sigma_norm.mean()
+        if ifo == opts.ifo:
+            found_trigs['sigma_max'] = sigma_norm.max()
+            found_trigs['sigma_min'] = sigma_norm.min()
+    except ValueError:
+        found_trigs[ifo+'/sigma_mean'] = 0
+        if ifo == opts.ifo:
+            found_trigs['sigma_max'] = 0
+            found_trigs['sigma_min'] = 0
```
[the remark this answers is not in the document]
pannarale: If you prefer, I'm happy to initialise these quantities to 0 and change them only if found_trigs is not empty.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
titodalcanton: Good point, nevermind then.

bin/pygrb/pycbc_pygrb_plot_chisq_veto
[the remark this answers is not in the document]
titodalcanton: I see, and I imagine we want to be able to recognize on the plot which points were originally zero, so we have to be aware that 0.005 is a magic value. This is a bit hacky and dirty, I would rather catch the zeroes explicitly and plot them differently (maybe with vertical bars or arrows pointing downward). But that means more code and time, so I am willing to pass on this right now :)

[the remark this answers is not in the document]
titodalcanton: Could you perhaps just add a comment above to indicate why some points may be zero, and why we want to convert them to 0.005 instead?

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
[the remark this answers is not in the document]
titodalcanton: Yes I prefer that, I think it makes the intent clearer.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
titodalcanton: Ok, see response to your other comment above.

bin/pygrb/pycbc_pygrb_plot_coh_ifosnr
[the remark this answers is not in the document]
titodalcanton: Actually, I just tested a couple things, and:
1. The first line (definition of `sigma_norm`) will handle the case of empty `found_trigs` just fine (it will produce another empty array) so you can move it outside the check block.
2. You can pass the argument `initial` to `min()` and `max()` to handle the case of empty inputs, though you need to be careful to not mess up the result when the input is *not* empty (i.e. it may not actually work).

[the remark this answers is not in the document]
pannarale: I opted for avoiding `initial` because it does this:
```
>>> numpy.max([-1], initial=0)
0
>>> numpy.min([1], initial=0)
0
>>> 
```
which is not the behaviour we want.

titodalcanton (approved):
No further comments from me.

PR #4951, opened 2024-11-22
Save statmap significance fits information

GarethCabournDavies:
The significance fits information gets discarded and there is not way to get it back - this change saves the fit coefficient, its approximate error and the rate of triggers above threshold as attributes in the statmap output files

This is an information loss bug fix
This change changes: output which can be used for later review / plotting

## Motivation
We want to be able to get the FAR calculation extrapolation values for review etc.
We can reverse-engineer fit coefficient and count above threshold using e.g. the FAR / stat values above the threshold, but not the error on fit coefficient

## Contents
Add a significance info dictionary as output from the get_far function
Report the values in this dictionary into the attributes of the various statmap job outputs

## Testing performed
tested for sngls_statmap outputs, ensuring that these match reverse-engineered outputs

### DIFF
--- a/pycbc/events/significance.py
+++ b/pycbc/events/significance.py
@@ -84,7 +84,7 @@ def count_n_louder(bstat, fstat, dec,
 
     unsort = sort.argsort()
     back_cum_num = n_louder[unsort]
-    return back_cum_num, fore_n_louder
+    return back_cum_num, fore_n_louder, {}
 
 
 def n_louder_from_fit(back_stat, fore_stat, dec_facs,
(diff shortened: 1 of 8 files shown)


pycbc/events/significance.py
```diff
@@ -84,7 +84,7 @@ def count_n_louder(bstat, fstat, dec,
 
     unsort = sort.argsort()
     back_cum_num = n_louder[unsort]
-    return back_cum_num, fore_n_louder
+    return back_cum_num, fore_n_louder, {}
```
tdent, line 87: Hmm, I am not getting what is the logic behind returning an empty dictionary here.  I don't see it getting used downstream so maybe this was a placeholder change that in the end isn't needed?

tdent: Ah, it looks like this is to keep using the same signature between the time slide n-louder method vs the exp fit one. 
Then add a comment before this line to point out that this is in parallel to n_louder_from_fit

tdent (changes requested):
Seems OK, just requesting one additional comment (or actually I think the docstring for Returns also has to be changed) for the time slide n_louder function.

pycbc/events/significance.py
tdent: As noted in the review, the 'Returns' part of the docstring should mention that an empty dict is also returned

PR #4953, opened 2024-11-22
Fix a couple of output/upload value issues in pycbc live

GarethCabournDavies:
Update the chisquared uploaded to gracedb to be reduced, and save the ranking statistic value

This is a bug fix

## Motivation / contents

We noticed that the chisquared for singles was not the reduced chisquared - I tracked it down to where I'd missed that I was using the same trigger set for statistic calculation as I was for the upload, which is not what is done for coincs, this is now set back to the un-modified set of triggers.

We also realised that the ranking statistic isnt being saved - I have added this to the gracedb upload/xml by utilising the "likelihood" value in the coinc row of the table.

## Testing performed
Need to test this, but thought it best to get the PR in first

### DIFF
--- a/pycbc/events/single.py
+++ b/pycbc/events/single.py
@@ -303,7 +303,8 @@ class LiveSingle(object):
 
         # fill in a new candidate event
         candidate = {
-            f'foreground/{self.ifo}/{k}': cutall_trigs[k][i] for k in trigs
+            f'foreground/{self.ifo}/{k}': cut_trigs[k][sngl_idx][i]
+            for k in trigs
         }
         candidate['foreground/stat'] = rank[i]
         candidate['foreground/ifar'] = ifar
--- a/pycbc/io/gracedb.py
+++ b/pycbc/io/gracedb.py
@@ -136,7 +136,7 @@ class CandidateForGraceDB(object):
         coinc_event_row.time_slide_id = lsctables.TimeSlideID(0)
         coinc_event_row.process_id = proc_id
         coinc_event_row.coinc_event_id = coinc_id
-        coinc_event_row.likelihood = 0.
+        coinc_event_row.likelihood = coinc_results['foreground/stat']
         coinc_event_table.append(coinc_event_row)
         outdoc.childNodes[0].appendChild(coinc_event_table)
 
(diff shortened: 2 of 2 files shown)


pycbc/io/gracedb.py
```diff
@@ -136,7 +136,7 @@ def __init__(self, coinc_ifos, ifos, coinc_results, **kwargs):
         coinc_event_row.time_slide_id = lsctables.TimeSlideID(0)
         coinc_event_row.process_id = proc_id
         coinc_event_row.coinc_event_id = coinc_id
-        coinc_event_row.likelihood = 0.
+        coinc_event_row.likelihood = coinc_results['foreground/stat']
```
titodalcanton, line 139: Will this also work for offline uploads and for `pycbc_make_skymap`, or do we need to check if `coinc_results` has that key?

GarethCabournDavies: For offline uploads, it looks like they use the `to_coinc_xml_object` function from `pycbc/io/hdf.py`, which already does this

`pycbc_make_skymap` doesn't provide this, so it should be checked, Ive added that now

maxtrevor:
@GarethCabournDavies would it be possible for you to get single detector candidate info into the foreground section of the 8 second trigger files as well? or too much for this one MR?

maxtrevor:
> > @GarethCabournDavies would it be possible for you to get single detector candidate info into the foreground section of the 8 second trigger files as well? or too much for this one MR?
> 
> Probably too much for this MR (and avoiding scope-creep of this one).
> 
> I'll add an issue to say that a lot of this (saving/upload of candidates) seems to be repeated imperfectly between coincs and singles, so trying to get them to use the same code path would be better

Ok. Getting the likelihood into the xml is already an improvement, thanks!

pycbc/events/single.py
```diff
@@ -303,7 +303,8 @@ def check(self, trigs, data_reader):
 
         # fill in a new candidate event
         candidate = {
-            f'foreground/{self.ifo}/{k}': cutall_trigs[k][i] for k in trigs
+            f'foreground/{self.ifo}/{k}': cut_trigs[k][sngl_idx][i]
```
maxtrevor, line 306: I'm a bit confused by the change to this line? Why do we need sngl_idx here now when we didn't before?

GarethCabournDavies: `cutall_trigs` already had the `sngl_idx` included (line 285 above), but `cut_trigs` doesnt have that

maxtrevor: Ah I missed the change from `cutall_trigs` to `cut_trigs`

maxtrevor (approved):
This looks good to me, approving

PR #4959, opened 2024-11-26
Adding --fake-strain-from-file and channel names/frame types for O4 data

DD-GitH:
Current version of pycbc_make_skymap doesn't support usage of O4 strain. Generation of fake strain is also only available through `--fake-strain` using simulated ASDs stored within PyCBC and not external/custom ASD. 

- This change enables usage of local ASD files as input for fake strain generation
- This change allows reading O4 strain at CIT if one wants to generate injections on real-time data

## Motivation
I wanted to use `pycbc_make_skymap` to investigate impact of transient noise on GW sky localisation during O4b for Virgo, but this required the proposed changes to be made.

## Contents
- Argument `--fake-strain-from-file` to compute fake strain based on custom ASD file as input, which is already available in `pycbc_single_template`
- Input file for ASD should be a two column file, line per line: FREQUENCY ASD_VALUE. Example in [DCC](https://dcc.ligo.org/LIGO-T2400104)
- Added strain channel names and frame types for H1/L1/V1 during O4, according to definitions [here](https://wiki.ligo.org/LSC/JRPComm/ObsRun4).
- GPStime start of O4a set to 1368975618, but we can also include ER15 times if needed

## Testing performed
Code tested both for usage of O4 strain and `--fake-strain-from-file` argument.

### DIFF
--- a/bin/pycbc_make_skymap
+++ b/bin/pycbc_make_skymap
@@ -58,7 +58,7 @@ def default_frame_type(time, ifo):
         if ifo == 'V1':
             return 'HoftOnline'
         elif ifo in ['H1', 'L1']:
-            return ifo + '_HOFT_C00'
+            return ifo + '_HOFT_C00_AR'
     raise ValueError('Detector {} not supported at time {}'.format(ifo, time))
 
 def default_channel_name(time, ifo):
@@ -85,7 +85,7 @@ def default_channel_name(time, ifo):
         if ifo == 'V1':
             return ifo + ':Hrec_hoft_16384Hz'
         elif ifo in ['H1', 'L1']:
-            return ifo + ':GDS-CALIB_STRAIN_CLEAN'
+            return ifo + ':GDS-CALIB_STRAIN_CLEAN_AR'
     raise ValueError('Detector {} not supported at time {}'.format(ifo, time))
 
 def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
@@ -169,8 +169,6 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
     st_psd_paths = {}
     st_out_paths = {}
     for ifo in ifos:
-        if ifo not in frame_types:
-            frame_types[ifo] = default_frame_type(mean_trig_time, ifo)
         if ifo not in channel_names:
             if (fake_strain[ifo] is not None or fake_strain_from_file[ifo] is not None):
                 channel_names[ifo] = ifo + ':FAKE_DATA'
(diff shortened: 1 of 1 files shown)


bin/pycbc_make_skymap
```diff
@@ -157,8 +169,10 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
     st_psd_paths = {}
     st_out_paths = {}
     for ifo in ifos:
+        if ifo not in frame_types:
+            frame_types[ifo] = default_frame_type(mean_trig_time, ifo)
```
titodalcanton, line 173: These two lines should not be needed here, please remove.

bin/pycbc_make_skymap
```diff
@@ -223,6 +237,13 @@ def main(trig_time, mass1, mass2, spin1z, spin2z, f_low, f_upper, sample_rate,
             if fake_strain_seed[ifo] is not None:
                 command.append("--fake-strain-seed")
                 command.append(str(fake_strain_seed[ifo]))
+        elif fake_strain_from_file[ifo] is not None:
+            # use fake strain for this ifo
```
titodalcanton, line 241:
```suggestion
            # use fake strain for this ifo with ASDs given from files
```

bin/pycbc_make_skymap
```diff
@@ -68,12 +74,18 @@ def default_channel_name(time, ifo):
...
         if ifo == 'V1':
             return ifo + ':Hrec_hoft_16384Hz'
         elif ifo in ['H1', 'L1']:
             return ifo + ':DCS-CALIB_STRAIN_CLEAN_SUB60HZ_C01'
+    elif time > 1368975618:
+        # O4
+        if ifo == 'V1':
+            return ifo + ':Hrec_hoft_16384Hz'
+        elif ifo in ['H1', 'L1']:
+            return ifo + ':GDS-CALIB_STRAIN_CLEAN'
```
titodalcanton, line 88: We should probably default to the analysis ready channel for LIGO:
```suggestion
            return ifo + ':GDS-CALIB_STRAIN_CLEAN_AR'
```

bin/pycbc_make_skymap
```diff
@@ -47,12 +47,18 @@ def default_frame_type(time, ifo):
...
         if ifo == 'V1':
             return 'V1Online'
         elif ifo in ['H1', 'L1']:
             return ifo + '_HOFT_CLEAN_SUB60HZ_C01'
+    elif time >= 1368975618:
+        # O4
+        if ifo == 'V1':
+            return 'HoftOnline'
+        elif ifo in ['H1', 'L1']:
+            return ifo + '_HOFT_C00'
```
titodalcanton, line 61: We should probably default to the analysis ready type for LIGO:
```suggestion
            return ifo + '_HOFT_C00_AR'
```

titodalcanton (changes requested):
Thank you very much @DD-GitH. This looks good but please see the specific comments above.

titodalcanton (approved):

PR #4971, opened 2024-12-04
Fix bug in reading single detector event fit_threshold

maxtrevor:
The `fit_threshold` was being read from the wrong location in an hdf file when computing the significance of single detector events in PyCBC Live.

The location it was reading from used to be correct, but was not updated to reflect changes made to pycbc_live_combine_single_significance_fits executable in https://github.com/gwastro/pycbc/commit/d486ec29d4c52963a3cbff691ce361200475eed9


titodalcanton (approved):

titodalcanton:
I do not think it is stored anywhere, as far as I remember it is generated on the spot with this script:
https://github.com/gwastro/pycbc/blob/master/examples/live/make_singles_significance_fits.py
So I think it is just a matter of fixing that script.

PR #4982, opened 2024-12-11
Modify injection module to support space-based detector class

acorreia61201:
This PR adds support for generating injections using a space-based detector response. 

This change affects: injection generation.

This change changes: primarily the `inject` module, with some minor additions to the `detector` and `distributions` modules to be compatible with the `inject` changes. The latter changes should be purely additive, though; the behaviors of ground-based detectors should remain unchanged from current usage.

## Motivation

A method for generating and reading frame files for space-based signals is required for eventual space-based PE. While #4691 adds the ability to generate space-based detector responses by-hand, the changes here allow for this to be written to a frame file via the `inject` module. Similar functionality is available via BBHx for MBHBs. However, this PR seeks to add support for general space-based signals natively in PyCBC, ideally matching the current output of BBHx for the same injection parameters.

## Contents

- Modify the `inject` module to apply simulated signals to a strain using a space-based detector response.
  - Add the ability to read in generalized sky coordinates from the detector class
  - Allow for `project_wave` method in detector classes to accept kwargs from injection file
  - Generalize light travel time buffers based on coordinate system (i.e. SSB, geocentric) and detector (i.e. arm length)
  - Generalize estimation of injection length based on the scale of the system
- Add properties to `detector` classes that print off accepted sky location parameters (e.g. ra/dec, ecliptic longitude/latitude)
- Add support for reading empty strings in injection configuration files as True in `distributions`

The space-based response is applied using the code changes from PR #4691. Any changes to that PR will be applied here once that PR is merged to master.


pycbc/detector/__init__.py
```diff
@@ -1,2 +1,4 @@
 from .ground import *
+from .ground import _ground_detectors
```
ahnitz, line 2: If we want to import these, they shouldn't have a _ prefix. The prefix is to indicate that the initial intention was only use within the module and through functions in that module. I would check if you really need to access this or if you can use a method of the Detector class directly.

pycbc/inject/inject.py
```diff
@@ -541,6 +564,14 @@ def apply(self, strain, detector_name, f_lower=None, distance_scale=1,
         if self.table[0]['approximant'] in fd_det:
             t0 = float(strain.start_time)
             t1 = float(strain.end_time)
+        elif detector_name in get_available_space_detectors():
+            au_travel_time = lal.AU_SI / lal.C_SI
+            try:
+                arm_travel_time = _space_detectors[detector_name]['armlength'] / lal.C_SI
```
ahnitz, line 570: This is where you'll  need to think about what the time coordinate means in the injection file. Is it in the SSB time or the detector frame? If the SSB frame, this isn't the right time to consider. If the detector frame, it is close, but do we want to rely on the detector geometry having a single fixed arm length argument (this is a candidate for the detector class providing this information if it is really needed).

pycbc/inject/inject.py
```diff
@@ -757,13 +796,15 @@ def make_strain_from_inj_object(self, inj, delta_t, detector_name,
...
     def end_times(self):
         """Return the approximate end times of all injections.
-
-        Currently, this just assumes all ringdowns are 2 seconds long.
         """
         # the start times are the tcs
         tcs = self.table.tc
-        # FIXME: this could be figured out using the ringdown module
-        return tcs + 2
+        mfs = self.table.final_mass
```
ahnitz, line 802: Does this belong in this PR? Or should this be a small separate one that can be reviewed quickly?

pycbc/distributions/__init__.py
```diff
@@ -170,9 +170,13 @@ def read_params_from_config(cp, prior_section='prior',
             # float (as we would expect for string arguments)
             static_args[key] = float(val)
         except ValueError:
-            # try converting to a list of strings; this function will just
-            # return val if it does not begin (end) with [ (])
-            static_args[key] = _convert_liststring_to_list(val)
+            if val is None or val == '':
```
ahnitz, line 173: Does this belong in this PR or a separate one?

pycbc/detector/space.py
```diff
@@ -67,12 +67,13 @@ def apply_polarization(hp, hc, polarization):
 
     return hp_ssb, hc_ssb
 
-def check_signal_times(hp, hc, orbit_start_time, orbit_end_time,
-                       offset=TIME_OFFSET_20_DEGREES, pad_data=False, t0=1e4):
+def preprocess(hp, hc, orbit_start_time, orbit_end_time,
```
ahnitz, line 70: I think you may need to rebase your PR from master.

pycbc/inject/inject.py
```diff
@@ -638,8 +672,13 @@ def make_strain_from_inj_object(self, inj, delta_t, detector_name,
             # compute the waveform time series
             hp, hc = get_td_waveform(inj, delta_t=delta_t, f_lower=f_l,
                                      **self.extra_args)
-            strain = projector(detector_name,
-                               inj, hp, hc, distance_scale=distance_scale)
+            if detector_name in get_available_space_detectors():
```
ahnitz, line 675: Can we make the output uniform here so that we don't need the if statement?

pycbc/inject/inject.py
ahnitz: I guess the choice here is safe though, assuming the orbit is at an AU (probably ok?).

PR #4984, opened 2024-12-12
Make HealpixSky easier to use and more efficient

titodalcanton:
Followup of #4848 that implements a more efficient approach for HealpixSky.

This is an efficiency update.

This change affects the distributions module.

This change adds a dependency on mhealpy. I am open to discussion if others prefer not to make it an explicit dependency. Edit: the mhealpy dependency has been made optional as suggested below.

## Motivation

The initial implementation of HealpixSky worked well, but it had a couple drawbacks:
* It used rejection sampling, which is not always efficient (this depends on the actual morphology of the specific skymap).
* It required a couple extra parameters to make rejection sampling efficient, making its use a bit less easy than ideal.

## Contents

The new algorithm works as follows. We first choose a random pixel of the skymap based on its contained probability. Then we draw a point on the sky with a uniform probability inside that pixel. The latter step still requires rejection sampling because HEALPix pixels are generally rotated with respect to RA and dec, but that is implemented such that the acceptance rate is always around 50%, so just a few iterations are sufficient to obtain the sample.

Followup of #4848.

## Testing performed

See below.

## Additional notes

### DIFF
--- a/pycbc/distributions/sky_location.py
+++ b/pycbc/distributions/sky_location.py
@@ -18,6 +18,7 @@ right ascension and declination.
 """
 
 import logging
+import warnings
 import numpy
 
 
@@ -194,7 +195,7 @@ class HealpixSky:
             # scale of the density function is much larger than the pixel area,
             # so that we can just scale the density by the pixel area.
             pix_areas = self.healpix_map.pixarea(
-                np.arange(self.healpix_map.npix)
+                numpy.arange(self.healpix_map.npix)
             )
             self.pix_probs = (self.healpix_map * pix_areas).data
         else:
@@ -203,9 +204,9 @@ class HealpixSky:
 
         # Sanity-check the probabilities, and ensure they are normalized
         # correctly (sum to one).
-        assert type(self.pix_probs) == np.ndarray
+        assert type(self.pix_probs) == numpy.ndarray
         sum_pix_probs = sum(self.pix_probs)
-        if not np.isclose(sum_pix_probs, 1):
+        if not numpy.isclose(sum_pix_probs, 1):
             warnings.warn(
                 f'Sum of probs in HEALPix map is {sum(self.pix_probs)}, '
                 'far from 1. Something might be wrong with that map'
@@ -236,7 +237,7 @@ class HealpixSky:
         # FIXME boundaries() works for an input array and is fast, but only for
         # single-resolution maps. It raises an exception with MOC maps. Another
         # bugfix for mhealpy?
-        return np.array(
+        return numpy.array(
             [self.healpix_map.boundaries(pi, step=1) for pi in indices]
         )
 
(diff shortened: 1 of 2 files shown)


pycbc/distributions/sky_location.py
```diff
@@ -22,6 +22,7 @@
 
 
 from scipy.spatial.transform import Rotation
+from mhealpy import HealpixMap
```
ahnitz, line 25: I'd make it an optional dependency, otherwise I think it's fine. There is a small tool that can help with that if you use the import_optional function from libutils.

titodalcanton:
[Tests done](http://leela.fisica.indivia.net/~tito/pycbc/test_pr4984/test_results.html) using a variety of GRB skymaps from the IPN, Fermi/GBM and a few handmade test cases. For the latter ones, I think I have normalized the probabilities incorrectly, causing an apparent bias in the validation plots. The most important ones are the real GRB ones.

Happy for this to be reviewed.

pycbc/distributions/sky_location.py
```diff
@@ -306,91 +224,115 @@ def from_config(cls, cp, section, variable_args):
...
+        return numpy.array(
+            [self.healpix_map.boundaries(pi, step=1) for pi in indices]
         )
 
+    def normalize_azimuth(self, phi):
+        """Helper function to ensure the azimuthal coordinate stays within
+        [0,2π).
+        """
+        phi[phi < 0] += 2 * numpy.pi
+        phi[phi >= 2 * numpy.pi] -= 2 * numpy.pi
```
pannarale, line 243: It may be safer to place these in a loop, i.e., keep adding (subtracting) 2 pi until phi is greater than 0 (less than 2 pi).

pannarale: I'm not sure whether more than one adjustment is ever needed, but it wouldn't hurt.

pycbc/distributions/sky_location.py
```diff
@@ -306,91 +224,115 @@ def from_config(cls, cp, section, variable_args):
...
+        boundaries_z_max = boundaries_vec[:,2,:].max(axis=1)
+
+        # Find ranges of azimuthal angle. `arctan2()` produces angles in the
+        # range [-π,π] which is not the right convention, so fix that right
+        # away.
+        boundaries_phi = numpy.arctan2(
+            boundaries_vec[:,1,:], boundaries_vec[:,0,:]
+        )
+        del boundaries_vec
+        boundaries_phi[boundaries_phi < 0] += 2 * numpy.pi
```
pannarale, line 278: Same question as above.

pycbc/distributions/sky_location.py
```diff
@@ -306,91 +224,115 @@ def from_config(cls, cp, section, variable_args):
...
+        boundaries_phi_min = boundaries_phi.min(axis=1)
+        boundaries_phi_max = boundaries_phi.max(axis=1)
+        # Pixels that straddle phi = 0 need a special treatment :( Shift them
+        # by 90 deg so that they are not straddlers for the time being.
+        # We will shift the corresponding samples back after drawing them.
+        straddler_mask = (boundaries_phi_max - boundaries_phi_min) > numpy.pi / 2
+        boundaries_phi[straddler_mask,:] -= numpy.pi / 2
```
pannarale, line 285: And again.

pycbc/distributions/sky_location.py
titodalcanton: The values here come directly from `numpy.arctan2()`, which returns values in [-π,π], so this is safe (only one "loop" is needed in any case).

I just realized that there is an incorrect edge case here when both arguments of `arctan2()` are negative zeroes, resulting in -π (which should ideally become 0 or -0, but becomes π with the current code). However your proposed solution would not fix this problem either, this would really need a special case.

pycbc/distributions/sky_location.py
titodalcanton: This line is not changing the angular convention, it is meant to really rotate the points.

pannarale (approved):

PR #4987, opened 2024-12-12
Allow relbin's fiducial parameters to not be floats

cdcapano:
Currently the Relative model tries to cast all of the reference parameters specified in the model section to floats. This is a problem if you want to pass in, say, a string (like a different approximant name than what's in the static params) or another parameter type to be  passed to the waveform generartor. This patch fixes the issue by first checking if it's a boolean parameter, then catching any attempt to cast to float that fails.

### DIFF
--- a/pycbc/inference/models/relbin.py
+++ b/pycbc/inference/models/relbin.py
@@ -655,11 +655,23 @@ class Relative(DistMarg, BaseGaussianNoise):
         )
 
         # get fiducial params from config
-        fid_params = {
-            p.replace("_ref", ""): float(cp.get("model", p))
-            for p in cp.options("model")
-            if p.endswith("_ref")
-        }
+        fid_params = {}
+        for p in cp.options("model"):
+            if p.endswith("_ref"):
+                val = cp.get("model", p)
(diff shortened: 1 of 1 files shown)


pycbc/inference/models/relbin.py
```diff
@@ -655,11 +655,23 @@ def extra_args_from_config(cp, section, skip_args=None, dtypes=None):
...
-        fid_params = {
-            p.replace("_ref", ""): float(cp.get("model", p))
-            for p in cp.options("model")
-            if p.endswith("_ref")
-        }
+        fid_params = {}
+        for p in cp.options("model"):
+            if p.endswith("_ref"):
+                val = cp.get("model", p)
+                if val == '':
```
ahnitz, line 662: There's a couple places now where we try to do this str -> something python conversion. Maybe we should centralize this now?

ahnitz: Well, if we add it to a parser, it should be the base InterpolatingConfigParser, but yeah, that approach makes sense, e.g. extend get_opt

PR #4992, opened 2024-12-16
Allowing the FisherSky class to have angle unit suffix rad deg

Thomas-JACQUOT:
This PR aims to allow `FisherSky` class to have angle unit suffix rad deg

This is a new feature, which make the  `FisherSky` class able to deal with angle unit suffixes

This change will: change nothing in other parts of the code, it adds only a new feature for the `FisherSky` class.
## Motivation

@titodalcanton, @pannarale and I wanted to change the `[prior-ra+dec]` section of `GRBXXXX.ini` files in order to be more clear (see discussion [here](https://git.ligo.org/ligo-cbc/pycbc-config/-/merge_requests/414#note_1224603)). To do so, it needs to change the `FisherSky` class to make it able to deal with angle unit suffixes (just as @titodalcanton did [here](https://github.com/gwastro/pycbc/pull/4965))
## Contents

### DIFF
--- a/pycbc/distributions/sky_location.py
+++ b/pycbc/distributions/sky_location.py
@@ -26,6 +26,7 @@ from scipy.spatial.transform import Rotation
 from pycbc.distributions import angular
 from pycbc import VARARGS_DELIM
 from pycbc.io import FieldArray
+from pycbc.types import angle_as_radians
 
 logger = logging.getLogger('pycbc.distributions.sky_location')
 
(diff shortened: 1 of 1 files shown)


pycbc/distributions/sky_location.py
```diff
@@ -85,9 +85,25 @@ class FisherSky:
...
+                    ({rematch_ra.group(2)}) dec ({rematch_dec.group(2)}) units 
+                    must be the same"
+                )
+            mean_ra = float(rematch_ra.group(1))
+            mean_dec = float(rematch_dec.group(1))
+            sigma = float(rematch_sigma.group(1))
+        except:
+            mean_ra = params['mean_ra']
+            mean_dec = params['mean_dec']
+            sigma = params['sigma']
```
titodalcanton, line 106: I suggest using `angle_as_radians` from `pycbc.types` instead, so you do not have to do all the parsing yourself.

titodalcanton:
After this change, `angle_units` will no longer be necessary, so I suggest removing it. I do not think any analysis actually used FisherSky yet, so it should be ok to break backwards compatibility.

Also, you will have to rebase this on master after #4991 is merged for the tests to pass.

pycbc/distributions/sky_location.py
Thomas-JACQUOT: I wanted to use the whole code to check if the angle unit in the rematch is the same as `params['angle_unit']`, but I understand your point.

pycbc/distributions/sky_location.py
```diff
@@ -75,23 +76,15 @@ class FisherSky:
...
-        mean_ra = params['mean_ra']
-        mean_dec = params['mean_dec']
-        sigma = params['sigma']
-        if params['angle_unit'] == 'deg':
-            mean_ra = numpy.deg2rad(mean_ra)
-            mean_dec = numpy.deg2rad(mean_dec)
-            sigma = numpy.deg2rad(sigma)
+        mean_ra = angle_as_radians(params['mean_ra'])
+        mean_dec = angle_as_radians(params['mean_dec'])
+        sigma = angle_as_radians(params['sigma'])
```
titodalcanton, line 87: If you do the conversions here in the constructor, we need a couple more things:
1. Remove the float conversions in the `from_config()` method below, because they will not work if we have the suffixes, and the constructor now takes the strings directly.
2. Edit the docstring above to indicate that the parameters can now be float or string, and add the explanations of the suffixes like done [here](https://github.com/gwastro/pycbc/blob/master/bin/pycbc_multi_inspiral#L218-L219).

titodalcanton (changes requested):

pycbc/distributions/sky_location.py
Thomas-JACQUOT: I don't think we need to remove these conversions because [here](https://github.com/gwastro/pycbc/pull/4965/files#diff-7d65dce5b43e3ccad1a03ced076b28c76d415f7acbbc519cdb712dd823938897R566), the function is returning only the value and not the full string variable.

titodalcanton: `float('24deg')` will fail.

titodalcanton (approved):
Looks perfect, thanks.

PR #4994, opened 2024-12-18
Making the check on sigma for FisherSky class more relax (allowing 0 value)

Thomas-JACQUOT:
This PR is allowing 0 values for `sigma` in the `FisherSky` class. The `FisherSky` class follows a `numpy.random.rayleigh` distribution which allows 0 values for `sigma` (see this [comment](https://git.ligo.org/ligo-cbc/pycbc-config/-/merge_requests/414#note_1228331))

This change affects: PyGRB 

This change will: allow to use `FisherSky` distribution for GRBs that has a `sky_error` equals to 0.
## Motivation
I want to do this PR in order to use PyGRB with GRBs that has a `sky_error` equals to 0 and uses `FisherSky` class as a prior for the right ascension and the declination.

### DIFF
--- a/pycbc/distributions/sky_location.py
+++ b/pycbc/distributions/sky_location.py
@@ -98,7 +98,7 @@ class FisherSky:
                 'The mean declination must be between '
                 f'-π/2 and π/2, {mean_dec} rad given'
             )
-        if sigma <= 0 or sigma > 2 * numpy.pi:
+        if sigma < 0 or sigma > 2 * numpy.pi:
             raise ValueError(
                 'Sigma must be positive and smaller than 2π '
                 '(preferably much smaller)'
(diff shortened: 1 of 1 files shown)


pycbc/distributions/sky_location.py
```diff
@@ -98,7 +98,7 @@ def __init__(self, **params):
                 'The mean declination must be between '
                 f'-π/2 and π/2, {mean_dec} rad given'
             )
-        if sigma <= 0 or sigma > 2 * numpy.pi:
+        if sigma < 0 or sigma > 2 * numpy.pi:
             raise ValueError(
                 'Sigma must be positive and smaller than 2π '
```
titodalcanton, line 103: Change "positive" to "non-negative" in the error message here.

titodalcanton (approved):

PR #4996, opened 2024-12-18
Add waveform error handling to Relbin (and other models)

cdcapano:
The `Relative` model (and models that inherit from it) do not catch failed waveform errors like the GaussianNoise models do. This adds that functionality to those models.

To avoid repeating code, and to standardize catching these errors, I created a decorator `catch_waveform_error` that can be added to any method that will call a waveform generator. The decorator will catch any `NoWaveformError`, `FailedWaveformError`, and `RuntimeError`. It will then call the class's `_nowaveform_handler` method (for the latter two errors this will only happen if the class's `ignore_failed_waveforms` attribute is set to True). In order for this to work, I had to rename some of the class's `_nowaveform_loglr` and `_nowaveform_logl` to `_nowaveform_handler`.

All models that previously caught these errors have been updated to use the decorator, along with the new addition to the Relative model. I added unit tests to test that the models appropriately catch each error.

One slight issue: in order to catch waveform errors in the Relative model, I also had to catch RuntimeErrors. This is being done at the level of the loglr function. This does mean that if a RuntimeError is raised elsewhere in the loglr function (not by the waverform generator), then the `_nowaveform_handler` will be triggered. I figured this would be ok since it's usually the waveform generator that raise RuntimeErrors.

### DIFF
--- a/pycbc/inference/models/relbin.py
+++ b/pycbc/inference/models/relbin.py
@@ -37,6 +37,7 @@ from pycbc.detector import Detector
 from pycbc.types import Array, TimeSeries
 
 from .gaussian_noise import (BaseGaussianNoise, catch_waveform_error)
+from pycbc.waveform import FailedWaveformError
 from .relbin_cpu import (likelihood_parts, likelihood_parts_v,
                          likelihood_parts_multi, likelihood_parts_multi_v,
                          likelihood_parts_det, likelihood_parts_det_multi,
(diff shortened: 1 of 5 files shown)


pycbc/inference/models/relbin.py
```diff
@@ -881,3 +898,13 @@ def _loglr(self):
         else:
             results = loglr
         return results
+
+    def _nowaveform_handler(self):
+        """Sets loglr values if no waveform generated.
+        """
+        loglr = sh_total = hh_total = -numpy.inf
+        if self.return_sh_hh:
+            results = (sh_total, hh_total)
```
ahnitz, line 907: I'm not sure it makes sense to catch the case where the return products are different. These should never be called in such a case and you can't be certain you are matching the return product format (e.g. is it a vector or a scalar?). This path should never be used for normal calling of the model (e.g. only reconstruction or interaction with other model which should itself probably decide what to do with its own handler rather than work out how to handle unexpected input.

cdcapano: I was just trying to match the return signature. What do you suggest doing in this case? Just return -numpy.inf?

test/test_infmodel.py
```diff
@@ -166,8 +169,181 @@ def test_brute_pol_phase_marg(self):
         model.update(**self.q1)
         self.assertAlmostEqual(self.a2, model.loglr, delta=0.002)
 
+
+class TestWaveformErrors(unittest.TestCase):
```
ahnitz, line 173: Thank you for adding a unittest! 

I think this mostly looks great. There was one minor comment. Otherwise, once you are happy with the PR and assuming all the tests now pass, I think this is good to go.

pycbc/inference/models/relbin.py
ahnitz: Yeah, I think that case should just be an uncaught error e.g. it probably *should* blow up on the user since I think they are doing something wrong to end up in that situation.

PR #5016, opened 2025-01-22
Allow PyGRB single IFO cut to be applied to a number of IFOs specified by the user

pannarale:
The current version of PyGRB requires the single IFO SNR to exceed a threshold (provided by the user) in at least 2 IFOs, whenever the analysis uses more than 1 IFO.  I promoted the 2 to a parameter specified by the user, with 2 as default to preserve the current behaviour.  This allows for this cut to be tuned in a more flexible way.

I also removed an unused utility function I found in this context.  Finally I took the liberty to pass `pycbc_multi_inspiral` through `flake8`.

This change affects: PyGRB.


pannarale:
Looking at [this](https://lscsoft.docs.ligo.org/lalsuite/lalapps/coh___p_t_f__option_8c_source.html#l00939) line, the help message is inline with your second item, but I distinctly remember going through lines of code with Andrew where the whole "2 most sensitive" thing was coded up (your item 1).  Let me dig through some code and get back to you.

pannarale:
I edited this PR completely and rescoped it.

bin/pycbc_multi_inspiral
```diff
@@ -236,14 +236,22 @@ parser.add_argument(
     help="Triggers with coincident/coherent SNR below this "
     "value will be discarded.",
 )
+parser.add_argument(
+    "--nifo-sngl-snr-threshold",
+    action="store",
+    type=int,
+    default=2,
+    help="Number of detectors requiring a minimum SNR value (specified "
+    "with --sngl-snr-threshold) value for trigger generation (default: 2).",
```
titodalcanton, line 245: I suggest "How many detectors must pass the --sngl-snr-threshold cut in order to generate a trigger (default 2)."

bin/pycbc_multi_inspiral
```diff
@@ -330,6 +338,16 @@ init_logging(args.verbose)
 num_slides = slide_limiter(args)
 # Arrange detectors alphabetically so they are always called in the same order
 args.instruments.sort()
+# Check that the number of IFOs requested to have single SNR above threshold
+# is compativle with the number of detectors in the analysis
+if len(args.instruments) > 1:
```
titodalcanton, line 343: I think the logic is still correct even without this check. If you only have one detector and you ask `--nifo-sngl-snr-threshold 2`, raising the error is still the correct behavior, no?

pycbc/events/coherent.py
```diff
@@ -59,9 +62,12 @@ def get_coinc_indexes(idx_dict, time_delay_idx):
...
             )
     # Search through coinc_idx for repeated indexes. These must have been loud
-    # in at least 2 detectors.
+    # in at least min_nifos detectors if the analysis uses more than 2 detectors.
     counts = np.unique(coinc_list, return_counts=True)
-    coinc_idx = counts[0][counts[1] > 1]
+    min_nifos_low = min_nifos - 1
+    coinc_idx = (
+        counts[0][counts[1] > min_ifos_low] if len(idx_dict) > 2 else counts[0]
+    )
```
titodalcanton, line 70: I am confused by the "if the analysis uses more than 2 detectors" part here. Did you mean more than 1 detector (consistently with the edits above)? If 2 is really what you meant, maybe it should be explained in the help string of `--nifo-sngl-snr-threshold`.

bin/pycbc_multi_inspiral
pannarale: I added this `if` because later on if `nifo` (which is equal to `len(args.instruments)`) is 1 the single IFO SNR cut is not used, so  `--nifo-sngl-snr-threshold` can be anything in that case and the code will still work (correctly).

pycbc/events/coherent.py
pannarale: Which implies: I am happy to remove it if you think it's just a wasteful slow down of things.

bin/pycbc_multi_inspiral
titodalcanton: I see the point, though I think this makes the CLI confusing. Perhaps this is something else to explain in the help for `--nifo-sngl-snr-threshold`.

One more thing, there is no need to nest the two if statements, and there is an extra call to `parser.error()`.

pycbc/events/coherent.py
titodalcanton: Seems ok to keep it, though I would rather do
```
if len(idx_dict) == 1:
    return counts[0]
```
right after the call to `np.unique()`.

The first sentence of the docstring of this function also needs updating.

bin/pycbc_multi_inspiral
```diff
@@ -236,14 +237,23 @@ parser.add_argument(
     help="Triggers with coincident/coherent SNR below this "
     "value will be discarded.",
 )
+parser.add_argument(
+    "--nifo-sngl-snr-threshold",
+    action="store",
+    type=int,
+    default=2,
+    help="How many detectors must pass the --sngl-snr-threshold cut in order "
+    "to generate a trigger; the cut may used only when more than 1 detector "
```
titodalcanton, line 246:
```suggestion
    "to generate a trigger; the cut is only applied when more than 1 detector "
```

bin/pycbc_multi_inspiral
```diff
@@ -330,6 +340,15 @@ init_logging(args.verbose)
...
 args.instruments.sort()
+# Check that the number of IFOs requested to have single SNR above threshold
+# is compatible with the number of detectors in the analysis
+if len(args.instruments) > 1 and \
+   len(args.instruments) < args.nifo_sngl_snr_threshold:
+    parser.error("--nifo-sngl-snr-threshold "
+                 f"({args.nifo_sngl_snr_threshold}) cannot exceed the "
+                 "number of detectors in the analysis "
+                 f"({len(args.instruments)}) because this would "
+                 "always yield no triggers.")
```
titodalcanton, line 351:
```suggestion
                 "never yield any trigger.")
```

PR #5019, opened 2025-01-22
Move to igwn-ligolw

titodalcanton:
Move from python-ligo-lw to the recently released and more modern igwn-ligolw.

This is a depencency change.
This change should have no impact on any functionality or result.
This change will drop the dependency on python-ligo-lw and introduce a depencency on igwn-ligolw.

## Motivation

igwn-ligolw is more actively maintained and supports modern Python versions and more architectures than python-ligo-lw.

## Contents

Mainly replacing imports, but the transition from python-ligo-lw 1.8.3 to igwn-ligolw 2.0.0 has the side effect of introducing a few breaking changes which never made it to a python-ligo-lw release yet, as well as a couple deprecation warnings, so those are handled as well. We can also further simplify `tox.ini`, consistently with #4857.

This is essentially a counterpart to #4999. Improves #4857.

## Testing performed

Ran `tox` and a few of the tests manually on a local machine, but I am relying on the CI for now.

## Additional notes

I am not sure I want to push this to production codes yet, at least not before more significant testing is done with workflows and GraceDB interactions. In particular, I think we should create a release branch for PyGRB's O4a analysis before merging this (@pannarale). So I am applying a "on hold" label.

### DIFF
--- a/bin/all_sky_search/pycbc_coinc_hdfinjfind
+++ b/bin/all_sky_search/pycbc_coinc_hdfinjfind
@@ -4,7 +4,7 @@ files.
 """
 
 import argparse, logging, types, numpy, os.path
-from ligo.lw import lsctables, utils as ligolw_utils
+from igwn_ligolw import lsctables, utils as ligolw_utils
 import igwn_segments as segments
 import pycbc
 from pycbc import events, init_logging
--- a/bin/all_sky_search/pycbc_make_bayestar_skymap
+++ b/bin/all_sky_search/pycbc_make_bayestar_skymap
@@ -22,7 +22,7 @@ import os
 import glob
 import tempfile
 
-from ligo.lw import lsctables, utils as ligolw_utils
+from igwn_ligolw import lsctables, utils as ligolw_utils
 
 import pycbc
 from pycbc.waveform import bank as wavebank
--- a/bin/all_sky_search/pycbc_prepare_xml_for_gracedb
+++ b/bin/all_sky_search/pycbc_prepare_xml_for_gracedb
@@ -29,8 +29,8 @@ matplotlib.use('agg')
 
 import lal
 import lal.series
-from ligo.lw import lsctables
-from ligo.lw import utils as ligolw_utils
+from igwn_ligolw import lsctables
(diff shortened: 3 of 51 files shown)


titodalcanton:
Started to do some more thorough tests. Noting down a few more deprecations to address:
```
UserWarning: igwn_ligolw.utils.process.register_to_xmldoc() is deprecated.  use igwn_ligolw.ligolw.Document.register_p
rocess() instead.
UserWarning: igwn_ligolw.lsctables.New() is deprecated.  use igwn_ligolw.Table.new() instead.
UserWarning: igwn_ligolw.lsctables.use_in() is deprecated.  simply importing the lsctables module automatically enables
the custom table support features it provides.
```

titodalcanton:
I made a test run of the PyCBC Live example uploading to GraceDB test and I do not see particular issues, so I am gaining confidence about this.
image

titodalcanton:
I now repeated the PyCBC Live example test after the latest commits, and I have uploaded two sets of candidates from before and after this PR. I have compared the resulting candidates from two of the injections side-by-side on GraceDB. I have checked both the original and SNR-optimized events. The things that should be identical are identical. So I think this is safe enough to read, especially if we do not plan to cherry-pick this into the offline analyses yet.

titodalcanton:
I also think this is due to lalsuite, not igwn-ligolw.

spxiwh:
Hmm, this may mean that this patch is on hold until a new lalsuite release that does not use ligo-segments?

titodalcanton:
I found a workaround that lets the Mac tests pass even with the current lalsuite version. Some tests are failing now because git.ligo.org is down.

I propose we consider merging this.

spxiwh (approved):
Thanks @titodalcanton ! We should keep an eye on the lalsuite issue so that we can remove the old ligolw from tox.ini as soon as it's possible to.

ahnitz:
@spxiwh @spxiwh Do you know if there is a timescale for that? Weeks, months, unknown?

titodalcanton:
I would estimate about a month, this issue should be a good way to track it: https://git.ligo.org/lscsoft/lalsuite/-/merge_requests/2340.

PR #5022, opened 2025-01-23
Enabling onsource vs dq vetoes check in PyGRB

pannarale:
In the PyGRB coherent search, the onsource and offsource trials must have the same lengths.  This PR ensures that no vetoed time is included in the onsource segment, to avoid having onsource and offsource trials differing in length.  If the onsource contains vetoed times, the workflow generator exits with an informative message.

This is a: new feature.

See issue https://github.com/gwastro/pycbc/issues/4981

## Testing performed

I took the GRB 170817A configuration file, and the workflow generation is completed.  I then artificially moved the trigger time to a time with dq issues in one of the IFOs, and the workflow generation process exits in an informative way.

## Additional notes


bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -247,6 +248,15 @@ if wflow.cp.has_option("workflow-segments", "segments-vetoes"):
...
     datafind_veto_files.append(veto_file)
+# Check that the onsource is free from vetoes
+for ifo in ifos:
+    avail_segs = select_segments_by_definer(veto_file.storage_path, ifo=ifo)
+    avail_segs.coalesce()
+    vetoed_segs = offSrc[ifo] - avail_segs
+    if onSrc[ifo].intersects(vetoed_segs):
+        logging.warning("The onsource %s overlaps with vetoed time in %s.",
+                        onSrc[ifo], ifo)
+        sys.exit()
```
titodalcanton, line 259: `logging.warning()` is not the best choice here, because the condition results in an unrecoverable termination. Either use `logging.error()`, or raise a ValueError instead.

titodalcanton: I would also print the vetoed segment to provide all the information, something like this:
```
The onsource [123, 129) contains vetoed time [125, 126) in H1.
```

titodalcanton:
Just to confirm, this behavior reproduces what cohPTF used to do in the past, right? I am happy to proceed like this for the time being, though see my comments above.

bin/pygrb/pycbc_make_offline_grb_workflow
pannarale: I will add the additional information.

Regarding the use of `logging.warning()`, this follows what the rest of the script does when "the segments are not enough".  In all these cases, it is not an error, because the generator is doing the right thing and exiting with an informative message.  Thinking more about it, now that you bring this up, it is not even a warning, rigorously speaking: should I change them all to `logging.info()` maybe?

titodalcanton: \begin{pedantic rant}
The intended meaning of info/warning/error is documented here: https://docs.python.org/3/library/logging.html#logging-levels. According to that table, if you use `info()` then you are saying "things are working as expected", and the workflow should just drop that detector and carry on. If instead you decide to stop, I would say that "the software has not been able to perform some function" (the user wanted to use the detector, and the workflow has decided that is not possible) and therefore you should use `error()`.
\end{pedantic rant}

pannarale:
An example output message
```2025-01-28T22:49:54.853-08:00 ERROR : The onsource (1187009321, 1187009327) contains the time (1187009326.0, 1187009327) that is vetoed in H1.```

titodalcanton (approved):

Issue #4783, opened 2024-06-12
Be efficient with imports


GarethCabournDavies:
Sometimes the whole package isnt needed, so we can use e.g. `from pycbc import init_logging`.

We should also check what is being imported when, I cant remember exactly where, but the other day I got an error to do with the neutron star equation of state when trying to import an argument parser group

GarethCabournDavies:
helped by #5002

titodalcanton:
I think we still have several cases of `from xyz import *`, especially in the tests, which is probably not a good idea.

PR #5024, opened 2025-01-24
Efficient imports

GarethCabournDavies:
In an attempt to remove "from xxx import *"-style imports from more places, I've used `git grep import | grep "*" | grep -v css| grep -v __init__` to find any places where this is used.

I then commented out the import, ran the tests to find what needs importing and getting that explicitly (not the most sophisticated, but seems to have worked).

After this change there is now 1 place which uses "from xxx import *", which is the versioning module, and I think that is valid.

The changes are mainly in the test suite - I've tried as much as possible to get tests to now pass, using cpu and cuda schemes, as these were what was catered for before!

I'd like to have other schemes / backends working, and some cuda tests fail, but that is for another PR

This is a code quality change

## Motivation
#4783

## Testing performed
Ran tests manually - they worked. The CI will also check this

### DIFF
--- a/bin/pycbc_make_banksim
+++ b/bin/pycbc_make_banksim
@@ -273,7 +273,7 @@ f = open("scripts/pycbc_banksim_match_combine", "w")
 f.write("""#!/usr/bin/env python
 from os.path import isfile
 from optparse import OptionParser
-from numpy import *
+from numpy import array
 import glob
 parser = OptionParser()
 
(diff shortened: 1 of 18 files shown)


titodalcanton (approved):
This touches only tests and banksim/faithsim engines, so I think it is ok to proceed. Just noting that it is not immediately obvious to me what is going to happen when changing the schemes like that in some of the tests, though I suspect that has no effect.

PR #5025, opened 2025-01-30
Ensure pygrb webpage is generated after all results files are produced

pannarale:
PyGRB webpages often had gaps that could be fixed by running the `pycbc_make_html_page` job by hand, after a full workflow was completed.  This was a sign of incorrect dependencies in the workflow (i.e., `pycbc_make_html_page` was being called too early).

To fix this, this PR simply moves the call to `make_results_web_page` further down in the `pycbc_pygrb_results_workflow` workflow generator, so that the job dependencies are correct: **all** nodes that will give jobs that produce plots, tables, etc. are now in the collection of dependencies passed to `make_results_web_page`, so `pycbc_make_html_page` runs after those jobs.

This is a: bug fix.

This change affects: PyGRB

I produced [this](https://ldas-jobs.ligo.caltech.edu/~francesco.pannarale/LVC/pygrb_tutorial/) results webpage from a short workflow with no need to "sync" it by hand after the workflow had finished running.

### DIFF
--- a/bin/pygrb/pycbc_pygrb_results_workflow
+++ b/bin/pygrb/pycbc_pygrb_results_workflow
@@ -735,16 +735,6 @@ _workflow.makedir(rdir['workflow/input_map'])
 _workflow.makedir(rdir['workflow/output_map'])
 _workflow.makedir(rdir['workflow/planning'])
 
-logging.info(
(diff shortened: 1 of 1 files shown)


sebastiangomezlopez (approved):
These changes look good to me.

PR #5026, opened 2025-01-30
Handling empty table in PyGRB

pannarale:
This PR enables the production of tables, within PyGRB, when the table needs to be empty.

This is a: new feature

This change affects: PyGRB

## Motivation
The current version of `pycbc_pygrb_page_tables` (designed to handle missed/vetoed injections and loudest events) will not work when there is no injection/trigger to be displayed.

## Contents
This PR fixes this issue.  The script no longer assumes there are injections to be displayed; further, it checks whether there is an onsource trigger to be reported: if there is not, the h5 file will be empty, where as the html file will say "There are no events" in the first column and report "-" in the remaining columns, to leave a human readable message.

## Testing performed
The functioning of this can be checked in sections 3.01, 3.03, and 6 of [this](https://ldas-jobs.ligo.caltech.edu/~francesco.pannarale/LVC/pygrb_tutorial/) webpage.

## Additional notes
The code to handle the case in which the offsource yields no triggers is along the lines of the one implemented for the injections, but admittedly it is not tested in this run.

### DIFF
--- a/bin/pygrb/pycbc_pygrb_page_tables
+++ b/bin/pygrb/pycbc_pygrb_page_tables
@@ -497,10 +497,11 @@ if onsource_file:
         on_trigs_bestnrs = on_trigs['network/reweighted_snr'][...]
 
         # Gather bestNR index
-        bestNR_event = np.argmax(on_trigs_bestnrs)
-        loud_on_bestnr_trigs, loud_on_bestnr = \
-            (on_trigs['network/event_id'][bestNR_event],
-             on_trigs_bestnrs[bestNR_event])
+        if on_trigs_bestnrs.size > 0:
+            bestNR_event = np.argmax(on_trigs_bestnrs)
+            loud_on_bestnr_trigs, loud_on_bestnr = \
+                (on_trigs['network/event_id'][bestNR_event],
+                 on_trigs_bestnrs[bestNR_event])
     # If the loudest event has bestnr = 0, there is no event at all!
     if loud_on_bestnr == 0:
         loud_on_bestnr_trigs = None
@@ -536,9 +537,6 @@ if onsource_file:
             [on_trigs[ifo+'/snr'][ifo_trig_index[ifo]] for ifo in ifos] + \
             [loud_on_bestnr]
         td.append(d)
-    else:
-        td.append(["There are no events"] + [0 for number in range(11)] +
-                  [0 for ifo in ifos] + [0])
 
     # Table header
     th = ['p-value', 'GPS time', 'Rec. m1', 'Rec. m2', 'Rec. Mc',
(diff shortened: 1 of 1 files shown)


bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -497,10 +497,11 @@ if onsource_file:
...
         # Gather bestNR index
-        bestNR_event = np.argmax(on_trigs_bestnrs)
-        loud_on_bestnr_trigs, loud_on_bestnr = \
-            (on_trigs['network/event_id'][bestNR_event],
-             on_trigs_bestnrs[bestNR_event])
+        if on_trigs_bestnrs.size > 0:
+            bestNR_event = np.argmax(on_trigs_bestnrs)
+            loud_on_bestnr_trigs, loud_on_bestnr = \
+                (on_trigs['network/event_id'][bestNR_event],
+                 on_trigs_bestnrs[bestNR_event])
```
titodalcanton, line 509: I think this assignment is more readable in two lines.

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -547,6 +545,10 @@ if onsource_file:
 
     td = list(zip(*td))
 
+    # Handle the case in which there is no data to be placed in the table
+    if not td:
+        td = [[] for i in np.arange(len(th))]
```
titodalcanton, line 555: This is more concise but perhaps more obscure:
```suggestion
        td = [[]] * len(th)
```

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -646,6 +657,10 @@ if found_missed_file is not None:
                  len(td))
     td = list(zip(*td))
 
+    # Handle the case in which there is no data to be placed in the table
+    if not td:
+        td = [[] for i in np.arange(len(th))]
```
titodalcanton, line 662: See comment on similar block above.

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -430,6 +430,11 @@ if lofft_outfile:
     th.extend([ifo+' time shift (s)' for ifo in ifos])
     th.append('BestNR')
 
+    # When len(offsource_trigs) == 0, the loop above leaves td = [] unchanged
+    # and this case needs to be handled adequately prior to moving on
+    if not td:
+        td = [[] for i in np.arange(len(th))]
```
titodalcanton, line 436: See comment on similar block below.

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -562,7 +569,11 @@ if onsource_file:
     format_strings.extend(['##.##' for ifo in ifos])
     format_strings.extend(['##.##'])
 
-    # Table data
+    # Table data: assemble human readable message when no trigger is recovered
+    if not loud_on_bestnr_trigs:
+        td = [["There are no events"] + ["-" for number in range(11)] +
+              ["-" for ifo in ifos] + ["-"]]
+        td = list(zip(*td))
```
titodalcanton, line 576: I think this is simpler but maybe not so clear:
```suggestion
        td = [list("-" * len(format_strings))]
        td[0][0] = "There are no events"
```

titodalcanton (approved):
Looks good, but I have some suggestions that may make this code better.

PR #5028, opened 2025-01-31
fix a bug to save model current stats for maxL polarization

yi-fan-wang:
I think an index `idx` is missing to save the polarization that gives the maximum likelihood in the polarization marginalization model


cdcapano (approved):
Thanks for catching this. Looks good to me.

PR #5033, opened 2025-02-06
Fix strange usage of pycbc.version.release

titodalcanton:
This fixes a couple minor weird things I noticed about how we report PyCBC's version in a couple places.

1. In `setup.py`, we always set `vinfo.release` to a string like `'False'` or `'True'` which I always found odd. Why not just a bool? Here I turn it into a bool.
2. While debugging #5019 I noticed that the PyCBC version string reported on GraceDB is "Using PyCBC version 0.0a9071 (release)" which is wrong. This happens because of point 1, as the code that sets up that string assumes `release` is bool.
3. In a recent test run of PyGRB, PyCBC's version information page says "ID : 0.0a9055" and "Version : False", both of which make no sense to me; for context, for the other libraries "ID" is the git hash. This is also fixed to show the git hash in "ID" and to give the actual version, plus the optional string "(release)" if appropriate, in "Version". This makes the "Version" line consistent with what is uploaded on GraceDB.


pycbc/results/versioning.py
```diff
@@ -92,9 +92,11 @@ def add_info_new_version(info_dct, curr_module, extra_str):
 
     pycbcinfo = {}
     pycbcinfo['Name'] = 'PyCBC'
-    pycbcinfo['ID'] = pycbc.version.version
+    pycbcinfo['ID'] = pycbc.version.git_hash
     pycbcinfo['Status'] = pycbc.version.git_status
-    pycbcinfo['Version'] = pycbc.version.release or ''
+    pycbcinfo['Version'] = pycbc.version.version
+    if pycbc.version.release:
+        pycbcinfo['Version'] += ' (release)'
```
ahnitz, line 99: Why is this string being appended? Doesnt that make it harder to actually check the version string?

ahnitz, line 99: Do we want the text appended to the version string? Won't that make it an invalid format for release?

titodalcanton:
No I have not made any manual tests, but I should.

titodalcanton:
I tested this by building two wheels via `python setup.py bdist_wheel`, one from this unmodified branch (non-release) and the other after editing `setup.py` to make a mock 2.8.0 release. Both wheels build and install fine, and the version shows up correctly in the output of `pip list`. I then ran `pycbc_page_versioning` before and after this change, from release and non-release installs, and here is what I get in the "PyCBC Version Information" section.

Before change, non-release:
```
 Name : PyCBC
ID : 0.0a9082
Status : CLEAN: All modifications committed
Version : False
Tag : None
Author : Gareth S Cabourn Davies
Builder : Tito Dal Canton
Branch : master
Committer : GitHub
Date : 2025-03-10 20:03:56 +0000 
```

Before change, release 2.7.3:
```
 Name : PyCBC
ID : 2.7.3
Status : CLEAN: All modifications committed
Version : True
Tag : v2.7.3
Author : maxtrevor <65971534+maxtrevor@users.noreply.github.com>
Builder : Unknown User <>
Branch : None
Committer : GitHub
Date : 2025-02-25 19:36:47 +0000 
```

After change, non-release:
```
 Name : PyCBC
ID : cca7c047cdef873751674f26073da81a8c857c85
Status : CLEAN: All modifications committed
Version : 0.0a9064
Tag : None
Author : Tito Dal Canton
Builder : Tito Dal Canton
Branch : fix-release-mess
Committer : Tito Dal Canton
Date : 2025-03-10 18:15:51 +0000 
```

After change, mock 2.8.0 release:
```
 Name : PyCBC
ID : ccc31cbe3c422e10b24e53b1eafe70b5979bf66d
Status : CLEAN: All modifications committed
Version : 2.8.0 (release)
Tag : v2.8.0
Author : Tito Dal Canton
Builder : Tito Dal Canton
Branch : fix-release-mess
Committer : Tito Dal Canton
Date : 2025-03-10 19:34:49 +0000 
```

Everything looks ok to me, so I am merging.

PR #5038, opened 2025-02-11
PyGRB: minor fixes to workflow generator 

sebastiangomezlopez:

This is a: bug fix.

This change affects: PyGRB

This change changes: PyGRB workflow generation.

## Motivation

This PR contains 2 separate bugfixes uncovered while testing the code on single detector GRB runs using O4a data.

## Contents

1. In the context of handling Veto definer files, `pycbc_make_offline_grb_workflow` checks for veto segments present in the onsource timewindow, i.e. around the GRB trigger, and throws an error if that's the case to avoid submitting jobs and  end up wasting CPU hours.

Currently lines [252-264](https://github.com/gwastro/pycbc/blob/a83736d49ffdf21b4fe3c6630e1067ff207981d6/bin/pygrb/pycbc_make_offline_grb_workflow#L252) lack an if statement to avoid operating with an empty list, when `avail_segs` has no elements. 

2. Single detector runs were not tested in the past. However by recently doing so, we uncovered a minor bug in `pycbc/workflow/jobsetup.py`. Specifically lines [690&691](https://github.com/gwastro/pycbc/blob/a83736d49ffdf21b4fe3c6630e1067ff207981d6/pycbc/workflow/jobsetup.py#L690) were mistakenly using the method `get_opts` from the wrong object.

### DIFF
--- a/bin/pygrb/pycbc_make_offline_grb_workflow
+++ b/bin/pygrb/pycbc_make_offline_grb_workflow
@@ -252,6 +252,7 @@ if wflow.cp.has_option("workflow-segments", "segments-vetoes"):
 for ifo in ifos:
     avail_segs = select_segments_by_definer(veto_file.storage_path, ifo=ifo)
     avail_segs.coalesce()
+    # In case no relevant vetoes are available, avail_segs becomes an empty list
     if len(avail_segs) > 0:
         vetoed_segs = offSrc[ifo] - avail_segs
         if onSrc[ifo].intersects(vetoed_segs):
@@ -263,13 +264,6 @@ for ifo in ifos:
                 ifo
             )
             sys.exit()
-    else:
-        vetoed_segs = offSrc[ifo]
-        logging.info(
-                  "%s has no vetoed segments in the onsource timewindow %s.",
-                  ifo,
-                  onSrc[ifo][0][:]
-                  )
 
 # Generate sky grid if needed
 skygrid_file = None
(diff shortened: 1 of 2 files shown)


bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -252,16 +252,24 @@ if wflow.cp.has_option("workflow-segments", "segments-vetoes"):
...
+            intersection = onSrc[ifo] & vetoed_segs
+            logging.error(
+                "The onsource %s contains the time %s that is vetoed in %s.",
+                onSrc[ifo][0][:],
+                intersection,
+                ifo
+            )
+            sys.exit()
+    else:
+        vetoed_segs = offSrc[ifo]
```
pannarale, line 267: This is not necessary.

bin/pygrb/pycbc_make_offline_grb_workflow
```diff
@@ -252,16 +252,24 @@ if wflow.cp.has_option("workflow-segments", "segments-vetoes"):
 for ifo in ifos:
     avail_segs = select_segments_by_definer(veto_file.storage_path, ifo=ifo)
     avail_segs.coalesce()
```
pannarale, line 254: Please add a comment here to explain that avail_segs is empty if no relevant vetoes are found.

pannarale (changes requested):

PR #5039, opened 2025-02-13
Add a colour map for consistent coincidence colors

GarethCabournDavies:
As we are now making plots with difference coincidences on one chart, I thought we should make a consistent chart to refer to each detector combination.

This is a new feature

This change changes result presentation / plotting

## Contents
Add a dictionary of colour choices and a function to wrap it.

How this is made is probably the more interesting thing:

This was made from combining the `_ifo_color_map` colors and generating a maximally-distinct colour based on a few things.
I set this up by starting with an IFO-precedence list: H1 L1 V1 K1 I1 G1
I then created coincidences containing the most-preferred IFOs and extended the color palette using [glasbey](https://glasbey.readthedocs.io/en/latest/) to maximise the distinctness of the new color(s).

In words:
 - From H (red) and L (blue) colors, I got the maximally distinct color (lime green) for HL
 - I then added in V, and got the maximally-distinct set of colors which would extend H, L, V and HL's already-defined colorset, and used it to get the HV, LV and HLV colours
 - This is repeated in turn for K, I and then G
 - I then asked github copilot to name all the colors (though I don't really know what 'olive drab' and 'rosy brown' are)

The code I ran is here:
```
def generate_precedence_color_map(n_ifos=None):
    # Create a color map which adds in the coincidences containing ifos in the order of ifo_precedence
    # This means that the coincs containing the most preferred IFOs will be colored first, and the less preferred IFOs will be colored last
    # As a result the most-used combinations will be maximally distinguishable
    if n_ifos == 1:
        # Single-IFO "coincs" just use the color of the ifo from the original _ifo_color_map
        return {ifo_precedence[0]: _ifo_color_map[ifo_precedence[0]]}
    
    n_ifos = n_ifos if n_ifos is not None else len(ifo_precedence)
    # Generate the color map for the IFOs up to the IFO before this one
    color_map = generate_precedence_color_map(n_ifos=n_ifos-1)
    # Now we add the coincidences formed with the 
    new_coincs = []
    for k in list(color_map.keys()) + ['']:
        # Add the new IFO into the key in alphabetical order
        # split the string into each IFO, these will be 2 characters long
        ifos = [k[i:i+2] for i in range(0, len(k), 2)]
        # Add the new IFO into the list, put it into alpabetical order, and rejoin the string
        new_coinc = ''.join(sorted(ifos + [ifo_precedence[n_ifos-1]]))
        if new_coinc in _ifo_color_map:
            color_map[new_coinc] = _ifo_color_map[new_coinc]
        else:
            new_coincs.append(new_coinc)

    new_palette = glasbey.extend_palette(list(color_map.values()), palette_size=len(new_coincs) + len(color_map))
    color_map.update({s: new_palette[i + len(color_map)] for i, s in enumerate(new_coincs)})=
    return color_map

precedence_cmap = generate_precedence_color_map()
# Remove the single-ifo combinations
precedence_cmap = {k: v for k, v in precedence_cmap.items() if k not in _ifo_color_map}
```

I didn't want to add the code directly because
1. It can take a while to run the same thing every time
2. I didn't want to add a new dependency

Right now, nothing links to this function/dictionary, so I expect all tests to pass, but we can use this for different coincs on the same plot

### DIFF
--- a/pycbc/results/color.py
+++ b/pycbc/results/color.py
@@ -21,84 +21,85 @@ _source_color_map = {
 }
 
 _coinc_color_map = {
-    'H1L1': '#caff00',
-    'H1V1': '#088e49',
-    'L1V1': '#00ffe3',
-    'H1L1V1': '#ca9210',
-    'H1K1': '#0cd75d',
-    'K1L1': '#df9aff',
-    'H1K1L1': '#ff866d',
-    'K1V1': '#6d61ff',
-    'H1K1V1': '#b65910',
-    'K1L1V1': '#00c2ce',
-    'H1K1L1V1': '#82a200',
-    'H1I1': '#20869e',
-    'I1L1': '#b2baff',
-    'H1I1L1': '#4db68e',
-    'I1V1': '#8a7520',
-    'H1I1V1': '#92e7ff',
-    'I1L1V1': '#928ad2',
-    'H1I1L1V1': '#ffe77d',
-    'I1K1': '#c2c600',
-    'H1I1K1': '#10ff00',
-    'I1K1L1': '#087ddb',
-    'H1I1K1L1': '#5dffa6',
-    'I1K1V1': '#f36100',
-    'H1I1K1V1': '#ce5dff',
-    'I1K1L1V1': '#00927d',
-    'H1I1K1L1V1': '#00ae31',
-    'G1H1': '#0000aa',
-    'G1L1': '#650400',
-    'G1H1L1': '#550075',
-    'G1V1': '#004500',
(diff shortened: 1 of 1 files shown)


GarethCabournDavies:
Adding a few plots so that we can actually see teh changes:
Full map (including singles): 
image

HLV network:
image

All 2-ifo coincs:
image

Coincs and Singles: 
image

GarethCabournDavies:
Requires #5041

GarethCabournDavies:
Updated given #5041 closure, whitespace removed and docstring clarified. I think this is ready for review

PR #5041, opened 2025-02-14
Rename I1 to A1 everywhere

GarethCabournDavies:
LIGO India's short code is now changed to A1, so we change I1 to A1 wherever it is in the codebase.

It looks like I1 still works as legacy, but A1 will be better going forward.

While editing this, I listened to the '90s boyband A1 [their best song](https://www.youtube.com/watch?v=zPKvxYrgBV4)

## Testing performed
Manually ran the tests which I expect to change from this, they all pass

### DIFF
--- a/bin/plotting/pycbc_page_segplot
+++ b/bin/plotting/pycbc_page_segplot
@@ -26,6 +26,7 @@ from itertools import cycle
 from pycbc import add_common_pycbc_options, init_logging
 from pycbc.events.veto import get_segment_definer_comments
 from pycbc.results.mpld3_utils import LineTooltip
+from pycbc.results.color import ifo_color
 from pycbc.workflow import SegFile
 
 # parse command line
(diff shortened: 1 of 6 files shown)


GarethCabournDavies:
Poke for review

tdent:
Not urgent as far as I know & I am prioritizing critical offline search tasks (and giving feedback on some thesis chapters)

tdent:
Thesis is submitted, so coming back to this - can it be rebased as github claims there are conflicts ?

tdent (commented):
I don't think I will check every place explicitly, but apart from the CI checks the main question here is whether 'git grep I1' gives you anything, hopefully not.

tdent:
Ah, so you now need to do some patching when getting detectors from LAL .. Is there an open issue in lalsuite to fix this as well ?

GarethCabournDavies:
It looks like there wasn't - I've added one https://git.ligo.org/lscsoft/lalsuite/-/issues/801

tdent:
https://git.ligo.org/groups/operations/-/epics/3

PR #5046, opened 2025-02-17
Bug fix: Adding min and max tau values for inverse td_model in p_tau function

divyajyoti09:
This change introduces a minimum and maximum tau for the `inverse` model in `pycbc.population.population_models.p_tau()` and fixes this issue: https://github.com/gwastro/pycbc/issues/5043

This is a: bug fix

This change affects: PyCBC population models module

This change has been tested at the level of creating `merger_rate_density` plots.

## Motivation

This change is made because the current state of the `inverse` model in the `p_tau` function is practically non-functional as it doesn't include any time delay. This has been demonstrated in the issue here: https://github.com/gwastro/pycbc/issues/5043

https://github.com/gwastro/pycbc/issues/5043

## Testing performed
This code has been tested to generate plots for `merger_rate_density` function


WuShichao:
Thank you for this PR, I will review it this week.

WuShichao:
Can you show the new curve for `inverse`?

WuShichao:
Can you show the new curve for the `inverse` time delay model?

divyajyoti09:
@WuShichao  The new plot looks like this
Screenshot from 2025-02-18 12-46-52

I agree that the units are different for SFR and merger rate but I normalized them to see whether the time delay was introducing a shift in the peak (which it wasn't because all the probabilities were peaking at 0 time delay). After introducing the `td_min` limit, the curve is shifting as expected.

Also, although I agree that `power_law` is very close to the `inverse` model, I still think this bug in the inverse model should be fixed since it is being used as default in various functions.

ahnitz:
@divyajyoti09 @WuShichao Is this ready to merge? What is the blocking issue here?

WuShichao:
> @divyajyoti09 @WuShichao Is this ready to merge? What is the blocking issue here?

Sorry for the delay, I will be back to it this week. I need to test it with other simulations.

ahnitz:
@divyajyoti09 Can you rebase this PR?

WuShichao (commented):
Look good.

Issue #4995, opened 2024-12-18
Add the ability to use a FITS skymap as input to `pycbc_make_sky_grid`


titodalcanton:

titodalcanton:
Done in #5085.

PR #5048, opened 2025-02-18
Enabling single ifo runs in PyGRB workflows

sebastiangomezlopez:

This is a: bug fix

This change affects:  PyGRB

Disclaimer: **This PR  has not been fully tested yet. DO NOT merge**

The proposed changes in this Pull request include:
- An expansion of the single IFO case at the `pycbc_make_sky_grid` level.
- minor fixes to `pycbc_multi_inspiral` that were preventing the creation of "trivial" `time_delay` dictionaries when `slide_num=0`, and `slide_id` related lists and dictionaries only contain the 0 element.

## Additional notes


bin/pycbc_multi_inspiral
```diff
@@ -412,8 +412,10 @@ with ctx:
     )
 
     logging.info("Determining time slide shifts and time delays")
+    # create an array of slide_ids using the syntax: start, stop, step
+    # where values are generated within the half-open interval [start, stop)
+    slide_ids = np.arange(0, num_slides+1, 1)
     # Create a dictionary of time slide shifts; IFO 0 is unshifted
-    slide_ids = np.arange(num_slides)
```
pannarale, line 416: I don't understand how the combination
```
 num_slides = 0
slide_ids = np.arange(0, num_slides+1, 1)
```
is different from the original
```
num_slides = 1
slide_ids = np.arange(num_slides)
```

pannarale:
I would keep the changes to the 2 files in 2 distinct PRs.  The changes to `pycbc_make_skygrid` are solid and this PR could be renamed appropriately.  Changes to `pycbc_multi_inspiral` require more thought, in my opinion.

titodalcanton:
The changes to make_sky_grid relate to some design considerations I made in #4995.

I am not a big fan of structuring code like this:
```
if some_condition:
    # do something short and simple
else:
    # do something very long and complicated
```
I would prefer something like this instead:
```
def make_single_det_grid(args):
    # do something short and simple

def make_multi_det_grid(args):
    # do something very long and complicated

if one_det_given:
    make_single_det_grid(args)
else:
    make_multi_det_grid()
```

There is also the consideration that @Thomas-JACQUOT is working on a deep rewrite of make_skymap anyway to support the HEALPix input… Maybe we should spend some time at the next call to find a good design for the various use cases.

PR #5049, opened 2025-02-19
Avoid nan values in network volume term

GarethCabournDavies:
When testing live with the new ranking statistic, we were getting a lot of weird FARs - this is due to getting stat values which were nan, and these were considered louder than everything else in the count_n_louder.

This is a bug fix
This change affects the live search, and touches code of the offline search, but shouldn't affect anything

## Contents
We combat this problem in three ways, which should prevent it occurring as much as possible:
1. If median sigma from fits files for any benchmark-network detector is NaN, then we ignore it. As a result, e.g. NaNs in L1 should not affect H1 singles, which was the case before. Hopefully this will cut most cases, as it goes from being an "or" to an "and" for H1 and L1 
2. If the benchmark is NaN - i.e. both detectors have NaNs in the fits files - then we do not apply the log sensitivity term.
3. As a final fail-safe, in the significance calculation, move any background with statistic NaN to have statistic of `-np.inf`. This shouldn't happen due to the protections above

Downsides of this approach:
 - If there is a much-less-sensitive detector which has NaN values, then the down-ranking from this factor will not be as it applied
 - If there is a foreground event which has statistic NaN, then it would be ranked higher than all background, I don't think this should be possible given the protections 1 and 2 in place, but this is a concern about protection 3

#5050 

## Testing performed
None yet

### DIFF
--- a/pycbc/events/significance.py
+++ b/pycbc/events/significance.py
@@ -197,7 +197,7 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
             "Removing %d NaN background statistic values",
             np.count_nonzero(nanmask),
         )
-    bstat = copy.deepcopy(back_stat)
+    back_stat = copy.deepcopy(back_stat)
     back_stat[nanmask] = -np.inf
 
     return _significance_meth_dict[method](
--- a/pycbc/events/significance.py
+++ b/pycbc/events/significance.py
@@ -194,8 +194,8 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
     nanmask = np.isnan(back_stat)
     if any(nanmask):
         logging.warning(
-            "Removing %d NaN background statistic values",
-            np.count_nonzero(nanmask),
+            "Setting %d NaN background statistic values to -inf",
+            nanmask.sum(),
         )
         back_stat = copy.deepcopy(back_stat)
         back_stat[nanmask] = -np.inf
--- a/pycbc/events/significance.py
+++ b/pycbc/events/significance.py
@@ -197,8 +197,8 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
             "Removing %d NaN background statistic values",
             np.count_nonzero(nanmask),
         )
-    back_stat = copy.deepcopy(back_stat)
-    back_stat[nanmask] = -np.inf
+        back_stat = copy.deepcopy(back_stat)
+        back_stat[nanmask] = -np.inf
 
     return _significance_meth_dict[method](
         back_stat,
--- a/pycbc/events/significance.py
+++ b/pycbc/events/significance.py
@@ -187,6 +187,14 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
     Wrapper to find the correct n_louder calculation method using standard
     inputs
     """
+    # This nan to inf conversion is a safety shield; there shouldn't be any
+    # nan statistic values, but if there are, they should be considered as
+    # quiet as possible. nans are considered greater than any floats in an
+    # argsort
+    nanmask = np.isnan(back_stat)
+    bstat = copy.deepcopy(back_stat)
+    back_stat[nanmask] = -np.inf
+
     return _significance_meth_dict[method](
         back_stat,
         fore_stat,
--- a/pycbc/events/significance.py
+++ b/pycbc/events/significance.py
@@ -192,16 +192,16 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
     # quiet as possible. nans are considered greater than any floats in an
     # argsort
     nanmask = np.isnan(back_stat)
+
     if any(nanmask):
         logging.warning(
             "Setting %d NaN background statistic values to -inf",
             nanmask.sum(),
         )
-        back_stat = copy.deepcopy(back_stat)
-        back_stat[nanmask] = -np.inf
+    back_stat_nonan = np.where(nanmask, -np.inf, back_stat)
 
     return _significance_meth_dict[method](
-        back_stat,
+        back_stat_nonan,
         fore_stat,
         dec_facs,
         **kwargs)
--- a/pycbc/events/significance.py
+++ b/pycbc/events/significance.py
@@ -192,6 +192,11 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
     # quiet as possible. nans are considered greater than any floats in an
     # argsort
     nanmask = np.isnan(back_stat)
+    if any(nanmask):
+        logging.warning(
+            "Removing %d NaN background statistic values",
+            np.count_nonzero(nanmask),
+        )
     bstat = copy.deepcopy(back_stat)
     back_stat[nanmask] = -np.inf
 
--- a/pycbc/events/stat.py
+++ b/pycbc/events/stat.py
@@ -1717,7 +1717,9 @@ class ExpFitFgBgNormStatistic(PhaseTDStatistic,
         sngls = single_info[1]
         ln_noise_rate = sngls['snglstat']
         ln_noise_rate -= self.benchmark_lograte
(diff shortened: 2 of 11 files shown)


pycbc/events/stat.py
```diff
@@ -1715,13 +1715,16 @@ def rank_stat_single(self, single_info,
...
-        benchmark_logvol = sngls['benchmark_logvol']
-        network_logvol -= benchmark_logvol
+        if not np.isnan(sngls['benchmark_logvol']):
+            # Only include this term if the benchmark log volume is defined
+            # This should always be the case in the offline search, but may
+            # not be in the live search.
+            network_sigmasq = sngls['sigmasq']
+            network_logvol = 1.5 * numpy.log(network_sigmasq)
+            benchmark_logvol = sngls['benchmark_logvol']
+            network_logvol -= benchmark_logvol
```
maxtrevor, line 1727: This doesn't seem quite right. If the` benchmark_logvol` is nan, it seems that the `network_logvol` variable is never set to any value. I think this would cause an error a few lines down when calculating the `loglr`. I think we just need to add an 
```
else:
    network_logvol = 0
```

pycbc/events/significance.py
```diff
@@ -187,6 +187,14 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
...
     inputs
     """
+    # This nan to inf conversion is a safety shield; there shouldn't be any
+    # nan statistic values, but if there are, they should be considered as
+    # quiet as possible. nans are considered greater than any floats in an
+    # argsort
+    nanmask = np.isnan(back_stat)
+    bstat = copy.deepcopy(back_stat)
+    back_stat[nanmask] = -np.inf
+
```
maxtrevor, line 197: Can you add some logging for the case where there are nans? I like having this shield in place, but we need to know if there are nans making it in somehow.

maxtrevor (commented):
The changes made in rank_stat_single need to also be made in rank_stat_coinc

pycbc/events/stat.py
GarethCabournDavies: Good catch!

pycbc/events/significance.py
tdent: should this be `bstat[nanmask] = -np.inf` ? 
I'm not really seeing why the deepcopy is made either

GarethCabournDavies: That was a typo on line 195 rather than anything clever. The deepcopy is made so that back_stat is not modified in-place

pycbc/events/significance.py
```diff
@@ -187,6 +187,19 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
...
+    # nan statistic values, but if there are, they should be considered as
+    # quiet as possible. nans are considered greater than any floats in an
+    # argsort
+    nanmask = np.isnan(back_stat)
+    if any(nanmask):
+        logging.warning(
+            "Removing %d NaN background statistic values",
+            np.count_nonzero(nanmask),
+        )
+        back_stat = copy.deepcopy(back_stat)
```
tdent, line 200: I don't think this line does anything .. ?? 
At the end of the day I don't think there is any reason to not modify back_stat in place either

pycbc/events/stat.py
```diff
@@ -1715,13 +1715,18 @@ def rank_stat_single(self, single_info,
...
         sngls = single_info[1]
-
         ln_noise_rate = sngls['snglstat']
         ln_noise_rate -= self.benchmark_lograte
-        network_sigmasq = sngls['sigmasq']
-        network_logvol = 1.5 * numpy.log(network_sigmasq)
-        benchmark_logvol = sngls['benchmark_logvol']
-        network_logvol -= benchmark_logvol
+        if not np.isnan(sngls['benchmark_logvol']):
+            # Only include this term if the benchmark log volume is defined
```
tdent, line 1721: I think the code would be easier to understand with `if np.isnan()` first, as the chunk of comments is only relevant in the live / nan case. ATM you have the comments discussing the edge case just above the code that does the _normal_ calculation but then nothing above the actual edge case itself.

maxtrevor (commented):
I think these changes look good, but I see a few pipeline failures.

Anyways, I want to hold off on merging while I refresh myself on the process of cutting a new release including this.

pycbc/events/significance.py
GarethCabournDavies: the deepcopy will mean that the local variable points to a different memory than the input variable, so changes to it will not affect the input variable.

My reasoning for not modifying back_stat in-place is that I would not expect a function called `get_n_louder` to modify the background statistic in place, and I don't think it is sensible to do that

pycbc/events/significance.py
```diff
@@ -187,6 +187,19 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
...
     inputs
     """
+    # This nan to inf conversion is a safety shield; there shouldn't be any
+    # nan statistic values, but if there are, they should be considered as
+    # quiet as possible. nans are considered greater than any floats in an
+    # argsort
+    nanmask = np.isnan(back_stat)
+    if any(nanmask):
+        logging.warning(
+            "Removing %d NaN background statistic values",
```
tdent, line 197: should be 'Setting %d NaN background statistic values to -inf' ?

pycbc/events/significance.py
```diff
@@ -187,6 +187,19 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
...
+            np.count_nonzero(nanmask),
```
tdent, line 198: If all you want to do is get the number of True's in a Boolean array `nanmask.sum()` is simplest and fastest

pycbc/events/significance.py
```diff
@@ -187,6 +187,19 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
...
+        )
+        back_stat = copy.deepcopy(back_stat)
+        back_stat[nanmask] = -np.inf
```
tdent, line 201: seems like we want something like `back_stat_nonan = numpy.where(nanmask, -np.inf, back_stat)` ?

pycbc/events/significance.py
tdent: mmm .. regardless of whether it does what we want wrt modifying variables or not, `back_stat = copy.deepcopy(back_stat)` is deeply confusing code IMO.

pycbc/events/significance.py
```diff
@@ -187,8 +187,21 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
...
+    # quiet as possible. nans are considered greater than any floats in an
+    # argsort
+    nanmask = np.isnan(back_stat)
+
+    if any(nanmask):
+        logging.warning(
+            "Setting %d NaN background statistic values to -inf",
+            nanmask.sum(),
+        )
+    back_stat_nonan = np.where(nanmask, -np.inf, back_stat)
```
tdent, line 201: I guess it works equally well if nanmask is all False !

pycbc/events/stat.py
```diff
@@ -1715,13 +1715,20 @@ def rank_stat_single(self, single_info,
...
-        network_logvol = 1.5 * numpy.log(network_sigmasq)
-        benchmark_logvol = sngls['benchmark_logvol']
-        network_logvol -= benchmark_logvol
+        if not np.isnan(sngls['benchmark_logvol']):
+            network_sigmasq = sngls['sigmasq']
+            network_logvol = 1.5 * numpy.log(network_sigmasq)
+            benchmark_logvol = sngls['benchmark_logvol']
+            network_logvol -= benchmark_logvol
+        else:
+            # If the benchmark rate is nan, we don't want to propogate
```
tdent, line 1726:
```suggestion
            # If the benchmark rate is nan, we don't want to propagate
```

tdent (approved):
Looks pretty good now ..

PR #5050, opened 2025-02-19
Avoid nan values in network volume term - MASTER VERSION

GarethCabournDavies:
When testing live with the new ranking statistic, we were getting a lot of weird FARs - this is due to getting stat values which were nan, and these were considered louder than everything else in the count_n_louder.

This is a bug fix
This change affects the live search, and touches code of the offline search, but shouldn't affect anything

## Contents
We combat this problem in three ways, which should prevent it occurring as much as possible:
1. If median sigma from fits files for any benchmark-network detector is NaN, then we ignore it. As a result, e.g. NaNs in L1 should not affect H1 singles, which was the case before. Hopefully this will cut most cases, as it goes from being an "or" to an "and" for H1 and L1 
2. If the benchmark is NaN - i.e. both detectors have NaNs in the fits files - then we do not apply the log sensitivity term.
3. As a final fail-safe, in the significance calculation, move any background with statistic NaN to have statistic of `-np.inf`. This shouldn't happen due to the protections above

Downsides of this approach:
 - If there is a much-less-sensitive detector which has NaN values, then the down-ranking from this factor will not be as it applied
 - If there is a foreground event which has statistic NaN, then it would be ranked higher than all background, I don't think this should be possible given the protections 1 and 2 in place, but this is a concern about protection 3

#5049 

## Testing performed
None yet

### DIFF
--- a/pycbc/events/significance.py
+++ b/pycbc/events/significance.py
@@ -207,11 +207,10 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
             "Setting %d NaN background statistic values to inf",
             nanmask.sum(),
         )
-        back_stat = copy.deepcopy(back_stat)
-        back_stat[nanmask] = -np.inf
+    back_stat_nonan = np.where(nanmask, -np.inf, back_stat
 
     return _significance_meth_dict[method](
-        back_stat,
+        back_stat_nonan,
         fore_stat,
         dec_facs,
         **kwargs)
--- a/pycbc/events/significance.py
+++ b/pycbc/events/significance.py
@@ -197,18 +197,16 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
     Wrapper to find the correct n_louder calculation method using standard
     inputs
     """
-    # This nan to inf conversion is a safety shield; there shouldn't be any
-    # nan statistic values, but if there are, they should be considered as
-    # quiet as possible. nans are considered greater than any floats in an
-    # argsort
+    # Deal with edge case: we don't expect nan statistic values but if they
+    # exist they will ruin the n_louder count, as they are considered larger
+    # than floats in an argsort.
     nanmask = np.isnan(back_stat)
     if any(nanmask):
         logging.warning(
             "Setting %d NaN background statistic values to inf",
             nanmask.sum(),
         )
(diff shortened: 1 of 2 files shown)


pycbc/events/significance.py
```diff
@@ -197,8 +197,20 @@ def get_n_louder(back_stat, fore_stat, dec_facs,
...
+    # quiet as possible. nans are considered greater than any floats in an
+    # argsort
+    nanmask = np.isnan(back_stat)
+    if any(nanmask):
+        logging.warning(
+            "Setting %d NaN background statistic values to inf",
+            nanmask.sum(),
+        )
+    back_stat_nonan = np.where(nanmask, -np.inf, back_stat
+
```
tdent, line 211:
```suggestion
    # Deal with edge case: we don't expect nan statistic values but if they
    # exist they will ruin the n_louder count, as they are considered larger
    # than floats in an argsort.
    nanmask = np.isnan(back_stat)
    if any(nanmask):
        logging.warning(
            "Setting %d NaN background statistic values to inf",
            nanmask.sum(),
        )
    back_stat_nonan = np.where(nanmask, -np.inf, back_stat)
```
Compactifying the comments, and a bracket seems to have gone missing.

pycbc/events/stat.py
```diff
@@ -1501,18 +1501,25 @@ def sensitive_volume_factor(self, sngls):
...
+        # Get benchmark log volume as single-ifo information :
+        # benchmark_logvol for a given template is not ifo-dependent, so
+        # choose the first ifo for convenience
+        benchmark_logvol = sngls[0][1]["benchmark_logvol"]
+
+        if np.isnan(benchmark_logvol):
+            # This can be the case in pycbc live if there are no triggers
+            # from this template in the trigger fits file. In this case we
+            # just assume that the sigma of the trigger of interest is
+            # representative of the benchmark network.
```
tdent, line 1513:
```suggestion
            # from this template in the trigger fits file. If so, assume 
            # that sigma for the triggers being ranked is
            # representative of the benchmark network.
```

tdent (approved):

tdent:
We want np -> numpy in stat.py 👀

PR #5056, opened 2025-02-21
Added check on spin magnitude

MarcoCusinato:
Added small check to ensure the spin magnitude in the remnant mass calulation is positive. Just in case somebody puts negative magnitudes...

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -622,6 +622,8 @@ def remnant_mass_from_mass1_mass2_spherical_spin_eos(
     mass1, mass2, spin1_a, spin1_polar, spin2_a, spin2_polar, \
         input_is_array = \
         ensurearray(mass1, mass2, spin1_a, spin1_polar, spin2_a, spin2_polar)
+    assert numpy.all(spin1_a >= 0) and numpy.all(spin2_a >= 0), \
+        "Spin magnitude MUST be null or positive"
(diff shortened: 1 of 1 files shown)


pycbc/conversions.py
```diff
@@ -622,6 +622,8 @@ def remnant_mass_from_mass1_mass2_spherical_spin_eos(
     mass1, mass2, spin1_a, spin1_polar, spin2_a, spin2_polar, \
         input_is_array = \
         ensurearray(mass1, mass2, spin1_a, spin1_polar, spin2_a, spin2_polar)
+    assert numpy.all(spin1_a >= 0) and numpy.all(spin2_a >= 0), \
+        "Spin magnitude MUST be null or positive"
```
titodalcanton, line 626: I think you can shorten this:
```suggestion
    assert all(spin1_a >= 0) and all(spin2_a >= 0), \
        "Spin magnitude must be null or positive"
```

titodalcanton: Ah, but note that `all()` is probably much slower than `numpy.all()`, and this might be a performance-critical bit of code.

https://stackoverflow.com/questions/43382237/numpy-ndarray-all-vs-np-allndarray-vs-allndarray

Nevermind sorry!

MarcoCusinato: Ok let's stay with the numpy method then

titodalcanton (approved):

PR #5057, opened 2025-02-23
Update to ligo.skymap 2.1.1

lpsinger:
This version, and later versions, are compatible with both Numpy 1.x and 2.x.

This change changes: documentation, result presentation / plotting, scientific output

## Additional notes

### DIFF
--- a/companion.txt
+++ b/companion.txt
@@ -6,11 +6,7 @@ healpy
 
 # Needed for GraceDB uploads and skymap generation
 ligo-gracedb>=2.10.0
-#ligo.skymap!=1.1.0
-# Temporarily use fork that doesn't force numpy<2.0 incompatibilities
-# Remove when PR merged, or numpy > 2.0 is a hard requirement
-# https://github.com/lpsinger/ligo.skymap/pull/28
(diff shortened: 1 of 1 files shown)


lpsinger:
Right, this will fail now because ligo.skymap conforms to [SPEC 0](https://scientific-python.org/specs/spec-0000/) so it has dropped support for Python 3.9.

ahnitz:
@lpsinger That's something you and @titodalcanton will likely need to discuss. As far as I know, 3.9 is till required to be supported for the low latency analysis in pycbc. I'm assigning him to review.

ahnitz:
@titodalcanton What is the plan for moving pycbc live off of 3.9? Can we drop 3.9 in pycbc (at least for testing if not officially).

titodalcanton:
We found that moving away from 3.9 requires some changes related to MPI for PyCBC Live at CIT, and these potentially affect the latency, or at least, they lead to some results that we do not understand yet. So we decided to continue with 3.9 for the time being and revisit the transition to something like 3.12 once some of the current priorities start winding down.

PR #5073, opened 2025-03-05
skygrid single ifo extension

sebastiangomezlopez:

This is a: bug fix

This is a rehousing of changes proposed in a closed PR [#5048](https://github.com/gwastro/pycbc/pull/5048)

This change affects:  PyGRB

The proposed changes in this Pull request include:
- An expansion of the single IFO case at the `pycbc_make_sky_grid` level.


titodalcanton (changes requested):
Like I said in the previous PR, I am not a big fan of structuring code like this:

```
if some_condition:
    # do something short and simple
else:
    # do something very long and complicated
```

I would prefer something like this instead:

```
def make_single_det_grid(args):
    # do something short and simple

def make_multi_det_grid(args):
    # do something very long and complicated

if one_det_given:
    make_single_det_grid(args)
else:
    make_multi_det_grid()
```

bin/pycbc_make_sky_grid
```diff
@@ -42,6 +42,92 @@ def cart_to_spher(sky_points):
     spher[spher[:, 0] <0, 0] += 2 * np.pi
     return spher
 
+def make_single_det_grid(args):
+    '''
+    Compute a trivial skygrid object with a single point 
+    '''
```
titodalcanton, line 48:
```suggestion
    """Construct a sky grid for a single-detector analysis.
    Since a one-detector network has no sky localization capability,
    we just put a single point at the center of the input distribution.
    """
```

titodalcanton (approved):
Thanks, looks good. I just have a minor suggestion for a docstring.

pannarale:
@sebastiangomezlopez, before we merge, I suggest removing the disclaimer in the description of this PR.  I assume this is tested :-D

PR #5079, opened 2025-03-14
Remove un-needed try/except in pycbc_page_snrifar

### LINKED ISSUE #2178: page_snrifar try/except prevents diagnosis of missing foreground
There are foreground events in an hdf file I am trying to plot but the 'cumulative' plot can't find them: see eg https://www.atlas.aei.uni-hannover.de/~tdent/LSC/o2/rates/catalog/H1L1-SNRIFAR_CUMULATIVE_O2.png .
However because there is a try-except around the code that gets the foreground triggers I can't see why it doesn't find them.  I don't see why the try-except is needed anyway.  

Also (what I suspect is causing the issue) there is an unused variable 'fap_end' defined in the cumulative case foreground trigger retrieval.  Why is this 'fap_end' here at all?

GarethCabournDavies:
Closes #2178

### DIFF
--- a/bin/plotting/pycbc_page_snrifar
+++ b/bin/plotting/pycbc_page_snrifar
@@ -129,42 +129,38 @@ if h_inc_back_num > h_iterations:
 args.cumulative = not args.not_cumulative
 
 foreground_livetime = conv.sec_to_year(f.attrs['foreground_time'])
-try:
     
-    if args.cumulative:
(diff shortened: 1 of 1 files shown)


tdent (approved):
Happy to approve if this has been tested in a representative case (ie as used in the offline search).

titodalcanton:
@GarethCabournDavies could you please rebase on master to pick up the new tests?

PR #5083, opened 2025-03-14
Polarization problem with marginalized time model

aleynaakyuz:
The marginalized time model currently projects the signal onto the detector with the polarization fixed to zero. This causes issues for parameter estimation when polarization is measurable. This pull request introduces a polarization factor to the projection to address this limitation.

### DIFF
--- a/pycbc/inference/models/marginalized_gaussian_noise.py
+++ b/pycbc/inference/models/marginalized_gaussian_noise.py
@@ -360,6 +360,11 @@ class MarginalizedTime(DistMarg, BaseGaussianNoise):
                                                                  params['tc'])
             dtc = params['tc'] + dt
 
+            pol_phase = numpy.exp(-2.0j * params['polarization'])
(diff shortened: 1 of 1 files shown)


pycbc/inference/models/marginalized_gaussian_noise.py
```diff
@@ -360,6 +360,11 @@ def _loglr(self):
                                                                  params['tc'])
             dtc = params['tc'] + dt
 
+            pol_phase = numpy.exp(-2.0j * params['polarization'])
```
ahnitz, line 363: This should only be in the case of precalculated values. Otherwise, it will actuall mess up the polarization values.

ahnitz (approved):

ahnitz:
@aleynaakyuz Has done offline testing and shown that this now (as expected) reconstructs the expected results for polarization. 

@aleynaakyuz It may be useful to post for posterity a before / after plot from ones that you have to show what the problem looks like.

PR #5084, opened 2025-03-19
Refactoring pycbc_grb_trig_combiner

pannarale:
Now that PR #5016 enables shaping the single IFO SNR cut in PyGRB in a less aggressive, `pycbc_grb_trig_combiner` needs to be able to handle more triggers.  I have encountered cases in which it requires an excessive amount of memory and jobs do not complete, so I refactored this executable.

Rather than loading all data, combining it and writing it, it deals with one dataset at a time.

I also ran `flake8` and avoided repeatedly writing the same information about gating multiple times.

This is an efficiency update.

This change affects: PyGRB

## Testing performed

I ran the old code and the new one on the same input files and the only differences I am getting are for the `/gating/` datasets where I am reducing redundant output to a single copy.  To enable this comparison, I used `--seed` for the first time and noticed it needed fixing. 

## Additional notes

This PR enables comparisons reported [here](https://git.ligo.org/pycbc/pygrb-offline-analysis/-/wikis/Single-IFO-SNR-cut).  Prior to these changes, I was unable to complete all runs reported there, specifically ones in which I was applying the single IFO SNR cut to 1 or 0 detectors.


bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -54,68 +53,13 @@ TQDM_KW = {
...
 def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
     """Merge several HDF5 files into a single file
 
     Parameters
     ----------
     inputfiles : `list` of `str`
         the paths of the input HDF5 files to merge
-
-    outputfile : `str`
-        the path of the output HDF5 file to write
```
titodalcanton, line 118: The `outputfile` argument is still present, so you should keep this.

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -183,61 +127,85 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
...
-    network_event_id_names = {
-        "network/{}".format(x.replace('/', '_')): x for
+    network_event_id_names = [
+        "network/{}".format(x.replace('/', '_')) for
         x in sngl_event_id_names if not x.startswith("network")
-    }
+    ]
+    all_event_id_names = sngl_event_id_names + network_event_id_names
+    del sngl_event_id_names
+    del network_event_id_names
```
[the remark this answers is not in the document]
pannarale: Right, it's not the output item (which I removed as it was unused).

bin/pygrb/pycbc_grb_trig_combiner
pannarale: Right, it's not the output item (which I removed as it was unused).

bin/pygrb/pycbc_grb_trig_combiner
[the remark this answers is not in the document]
pannarale: Okay.  Just being tidy, since the other `del` on `data` was important.

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -183,61 +130,81 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
...
-                dataset_names.remove(search_datasets.pop())
 
+                # shift the event ids by the number of events gathered so far
+                n_events_h5in = len(h5in[dset][:])
+                data += [h5in[dset][:] + [n_events]*n_events_h5in]
+                n_events += n_events_h5in
+        data = numpy.concatenate(data)
+        for dset in all_event_id_names:
+            logging.info("Storing %s dataset", dset)
+            _ = h5out.create_dataset(dset, data=data, **compression_kw)
```
titodalcanton, line 192: Drop the `_ =` prefix.

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -486,10 +453,14 @@ logging.info(
...
-offsource_ntrials = [int(numpy.trunc(abs(b)/trialtime)) for b in bins["OFFSOURCE"]]
+offsource_ntrials = [int(numpy.trunc(abs(b)/trialtime))
+                     for b in bins["OFFSOURCE"]]
 
 # Seed random number generator
-numpy.random.seed(args.seed)
+if args.seed is None:
+    numpy.random.seed(args.seed)
+else:
+    numpy.random.seed(numpy.int64(args.seed))
```
titodalcanton, line 463: Can you explain this change? Why the explicit conversion to int64?

pannarale: It was taking the input as a string and failing.

pannarale: I/we kept it to the `None` default all along and it never failed, but when I actually used the option it complained that it was not an int64.

titodalcanton: A cleaner solution is to give the `--seed` argument a type of int. Then it will still be None if not given, and an int if given.

pannarale: Done!

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -183,61 +130,81 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
...
+                n_events_h5in = len(h5in[dset][:])
+                data += [h5in[dset][:] + [n_events]*n_events_h5in]
+                n_events += n_events_h5in
+        data = numpy.concatenate(data)
+        for dset in all_event_id_names:
+            logging.info("Storing %s dataset", dset)
+            h5out.create_dataset(dset, data=data, **compression_kw)
+        del data
+
+        # Handling template_id seprately
```
titodalcanton, line 195:
```suggestion
        # Handling template_id separately
```

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -183,61 +130,81 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
...
+        logging.info("Preparing template IDs", dset)
```
titodalcanton, line 196:
```suggestion
        logging.info("Preparing template IDs")
```

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -183,61 +130,81 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
...
         template_ids = template_hash_to_id(
             trigger_file=h5out, bank_path=args.bank_file
         )
+        logging.info("Storing template IDs", dset)
```
titodalcanton, line 200:
```suggestion
        logging.info("Storing template IDs")
```

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -486,7 +454,8 @@ logging.info(
 # there should always be two segments in the offsource bins,
 # one before and one after the onsource.
 # Nevertheless, simply rely on there being at least one stretch of offsource.
-offsource_ntrials = [int(numpy.trunc(abs(b)/trialtime)) for b in bins["OFFSOURCE"]]
+offsource_ntrials = [int(numpy.trunc(abs(b)/trialtime))
+                     for b in bins["OFFSOURCE"]]
```
titodalcanton, line 458: I am pretty sure this produces the same result given that all quantities are positive:
```suggestion
offsource_ntrials = [int(abs(b) / trialtime) for b in bins["OFFSOURCE"]]
```

bin/pygrb/pycbc_grb_trig_combiner
```diff
@@ -183,61 +130,81 @@ def merge_hdf5_files(inputfiles, outputfile, verbose=False, **compression_kw):
...
+            logging.info("Merging %s dataset", dset)
+            data = []
+            for filename in tqdm.tqdm(inputfiles, desc="Merging trigger files",
+                                      disable=not verbose, total=nfiles,
+                                      unit="files", **TQDM_KW):
+                with HFile(filename, 'r') as h5in:
+                    if len(h5in['network']['event_id']) == 0:
+                        continue
+                    data += [h5in[dset][:]]
+                    # read the search datasets and gating_datasets only once
```
titodalcanton, line 164: …but I do not see where they are actually read?

titodalcanton (changes requested):
See a few other minor corrections and a question in the comments. I also have a couple general questions.

First, this is basically switching from looping over files once, and over datasets for each file, to looping over datasets once and over files multiple times. I can understand how this ends up saving memory (you only need to keep one dataset in memory at once) though it is probably still quite inefficient in terms of filesystem I/O. I do not have a better solution on the spot, so if this works better, then fine, but I would not be surprised if this came to bite us again on a different cluster/node.

Second general question, you are slowly growing a (presumably large) Python list to accumulate each dataset, and then converting it to a Numpy array at the end. If you know the total size beforehand (and I am not thinking enough right now to judge if that is the case), then it would be more efficient to preallocate a Numpy array of that size, and simply put each dataset at the right spot in that array.

bin/pygrb/pycbc_grb_trig_combiner
pannarale: True

pannarale:
Hi @titodalcanton, 
 
> First, [...] I would not be surprised if this came to bite us again on a different cluster/node.

Yes, I have the same suspicion, i.e., this is an improvement (a necessary one right now) but not a firm solution.
 
> Second general question, you are slowly growing a (presumably large) Python list to accumulate each dataset, and then converting it to a Numpy array at the end. If you know the total size beforehand (and I am not thinking enough right now to judge if that is the case), then it would be more efficient to preallocate a Numpy array of that size, and simply put each dataset at the right spot in that array.

If I understand correctly what you are thinking, a (semi-)robust way to do this would be to read through all input files a dataset that is definitely in all files them, e.g., `ifo[0]/reweighted_snr`, the way it is done now.  This would tell us about the total size and then we could use that information to expedite things with all remaining datasets.

bin/pygrb/pycbc_grb_trig_combiner
pannarale: The are read the first time this loop is entered with the first file.  Then you hit the `break` and move on to the next dataset, but this is after the search or gating information was already read once.

pannarale: Perhaps it would be preferable to swap:
```
                    if len(h5in['network']['event_id']) == 0:
                        continue
                    data += [h5in[dset][:]]
                    # read the search datasets and gating_datasets only once
                    if dset in once_only_datasets:
                        break
```
to
Perhaps it would be preferable to swap:
```
                    # read the search datasets and gating_datasets only once
                    if dset in once_only_datasets:
                        break
                    if len(h5in['network']['event_id']) == 0:
                        continue
                    data += [h5in[dset][:]]
```
so that the search/gating information is not lost in the remote case in which all trigger files are empty.

titodalcanton: No I see it now, thanks. Let's keep as is.

titodalcanton (approved):
Thanks for the comments, let's merge this as is and revisit for further optimizations later if necessary.

PR #4955, opened 2024-11-23
Final version of `pycbc_pygrb_plot_snr_timeseries`

pannarale:
This PR is the sixth in the series started in PR #4929.  Much like PRs #4947 and #4950, this PR provides a version of `pycbc_pygrb_plot_snr_timeseries` that uses veto and segments files, as well as the utilities introduced recently to streamline the PyGRB results scripts.  Additionally, `pycbc_pygrb_plot_snr_timeseries` now has an option to activate plotting the onsource (by default the onsource is not shown) and this is handled in its dedicated workflow generator function.

## Standard information about the request (and the following ones that will be linked to this)

This is a: a new feature enabling veto definer file usage in PyGRB.  Utilities and scripts in results production are being  streamlined along the way.

This change affects: PyGRB

This change changes: result presentation / plotting and scientific output.

Should this change break the standard automated test running `--help` for PyGRB plotting scripts, I will add some workarounds to avoid this.  If needed, these will likely be empty functions: the plotting scripts will be progressively renovated in the whole series of PRs.

## Motivation
Now that the workflow generator passes the veto definer file to the jobs where needed, its usage in the PyGRB results scripts is possible.

## Testing performed
The totality of the changes that will be broken down in multiple PRs was tested on GRB 170817A data by producing a full results webpage (see [here](https://ldas-jobs.ligo.caltech.edu/~francesco.pannarale/LVC/pygrb_nov2024_1/)).

- [x] The author of this pull request confirms they will adhere to the [code of conduct](https://github.com/gwastro/pycbc/blob/master/CODE_OF_CONDUCT.md)

PR #5085, opened 2025-03-20
Implementation of building a Sky Grid with respect to a given covered probability criterion 

Thomas-JACQUOT:
This PR is a DRAFT.

This is a new feature for  `pycbc_make_sky_grid` script. 

This change affects PyGRB and may affect All-sky searches in the future when I will implement this technique to UniformSky class (still draft PR).

This change changes how the sky grid is generated and sky location classes. 

This change will have appropriate unit tests.

This change will enhance the sky grid generation by generating a sky grid with respect to the coverage that the user want.

## Motivation

This PR is motivated by having a flexible generation of sky grids in general, depending on how much probability you want to cover. Also, thanks to this PR, we can generate a sky grid with FITS map (in order to solve this issue : #4995)

## Contents
I have implemented a new function to the FisherSky and HealpixSky classes called `to_uniform_patch` that convert these distributions to a uniform patch, depending on the coverage that the user has to specify. This is what would be used to generate the sky grid.

#4955 

## Testing performed
Still on going...

### DIFF
--- a/pycbc/distributions/sky_location.py
+++ b/pycbc/distributions/sky_location.py
@@ -46,6 +46,86 @@ class UniformSky(angular.UniformSolidAngle):
     _default_azimuthal_angle = 'ra'
 
 
+class UniformCircleSky:
+
+    name = 'uniform_circle_sky'
+    _params = ['ra', 'dec']
+    
+    def __init__(self, **params):
+        self.mean_ra = angle_as_radians(params['mean_ra'])
+        self.mean_dec = angle_as_radians(params['mean_dec'])
+        self.sigma = angle_as_radians(params['sigma'])
+        if self.mean_ra < 0 or self.mean_ra > 2 * numpy.pi:
+            raise ValueError(
+                f'The mean RA must be between 0 and 2π, {mean_ra} rad given'
+            )
+        if self.mean_dec < -numpy.pi / 2 or self.mean_dec > numpy.pi / 2:
+            raise ValueError(
+                'The mean declination must be between '
+                f'-π/2 and π/2, {mean_dec} rad given'
+            )
+        if self.sigma < 0 or self.sigma > 2 * numpy.pi:
+            raise ValueError(
+                'Sigma must be non-negative and smaller than 2π '
+                '(preferably much smaller)'
+            )
+        self.rayleigh_scale = 0.66 * self.sigma
+        # Prepare a rotation that puts the North Pole at the mean position
+        self.rotation = Rotation.from_euler(
+            'yz', [numpy.pi / 2 - self.mean_dec, self.mean_ra]
+        )
+
+    @property
+    def params(self):
+        return self._params
+
+    @classmethod
+    def from_config(cls, cp, section, variable_args):
+        tag = variable_args
+        variable_args = variable_args.split(VARARGS_DELIM)
+        if set(variable_args) != set(cls._params):
+            raise ValueError(
+                "Not all parameters used by this distribution "
+                "included in tag portion of section name"
+            )
+        self.mean_ra = cp.get_opt_tag(section, 'mean_ra', tag)
+        self.mean_dec = cp.get_opt_tag(section, 'mean_dec', tag)
+        self.sigma = cp.get_opt_tag(section, 'sigma', tag)
+        return cls(
+            mean_ra=self.mean_ra,
+            mean_dec=self.mean_dec,
+            sigma=self.sigma,
+        )
+
+
+    def rvs(self, size):
+        # Draw samples from a distribution centered on the North pole
+        np_ra = numpy.random.uniform(low=0, high=(2 * numpy.pi), size=size)
+        np_dec = angular.SinAngle(
+            polar_bounds=(0,self.sigma)).rvs(
+            size=size).astype(numpy.float64)
+
+        # Convert the samples to intermediate cartesian representation
+        np_cart = numpy.empty(shape=(size, 3))
+        np_cart[:, 0] = numpy.cos(np_ra) * numpy.sin(np_dec)
+        np_cart[:, 1] = numpy.sin(np_ra) * numpy.sin(np_dec)
+        np_cart[:, 2] = numpy.cos(np_dec)
+
+        # Rotate the samples according to our pre-built rotation
+        rot_cart = self.rotation.apply(np_cart)
+
+        # Convert the samples back to spherical coordinates.
+        # Some unpleasant conditional operations are needed
+        # to get the correct angle convention.
+        rot_radec = FieldArray(size, dtype=[('ra', '<f8'), ('dec', '<f8')])
+        rot_radec['ra'] = numpy.arctan2(rot_cart[:, 1], rot_cart[:, 0])
+        neg_mask = rot_radec['ra'] < 0
+        rot_radec['ra'][neg_mask] += 2 * numpy.pi
+        rot_radec['dec'] = numpy.arcsin(rot_cart[:, 2])
+        return rot_radec
+
+
+
 class FisherSky:
     """A distribution that returns a random angle drawn from an approximate
     `Von_Mises-Fisher distribution`_. Assumes that the Fisher concentration
@@ -87,33 +167,33 @@ class FisherSky:
     _params = ['ra', 'dec']
 
     def __init__(self, **params):
-        mean_ra = angle_as_radians(params['mean_ra'])
-        mean_dec = angle_as_radians(params['mean_dec'])
-        sigma = angle_as_radians(params['sigma'])
-        if mean_ra < 0 or mean_ra > 2 * numpy.pi:
+        self.mean_ra = angle_as_radians(params['mean_ra'])
+        self.mean_dec = angle_as_radians(params['mean_dec'])
+        self.sigma = angle_as_radians(params['sigma'])
+        if self.mean_ra < 0 or self.mean_ra > 2 * numpy.pi:
             raise ValueError(
                 f'The mean RA must be between 0 and 2π, {mean_ra} rad given'
             )
-        if mean_dec < -numpy.pi / 2 or mean_dec > numpy.pi / 2:
+        if self.mean_dec < -numpy.pi / 2 or self.mean_dec > numpy.pi / 2:
             raise ValueError(
                 'The mean declination must be between '
                 f'-π/2 and π/2, {mean_dec} rad given'
             )
-        if sigma < 0 or sigma > 2 * numpy.pi:
+        if self.sigma < 0 or self.sigma > 2 * numpy.pi:
             raise ValueError(
                 'Sigma must be non-negative and smaller than 2π '
                 '(preferably much smaller)'
             )
-        if sigma > 0.35:
+        if self.sigma > 0.35:
             logger.warning(
                 'Warning: sigma = %s rad is probably too large for the '
                 'Fisher approximation to be valid',
-                sigma,
+                self.sigma,
             )
-        self.rayleigh_scale = 0.66 * sigma
+        self.rayleigh_scale = 0.66 * self.sigma
         # Prepare a rotation that puts the North Pole at the mean position
         self.rotation = Rotation.from_euler(
-            'yz', [numpy.pi / 2 - mean_dec, mean_ra]
+            'yz', [numpy.pi / 2 - self.mean_dec, self.mean_ra]
         )
 
     @property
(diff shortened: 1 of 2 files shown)


pycbc/distributions/sky_location.py
```diff
@@ -46,6 +46,86 @@ class UniformSky(angular.UniformSolidAngle):
...
+            raise ValueError(
+                'The mean declination must be between '
+                f'-π/2 and π/2, {mean_dec} rad given'
+            )
+        if self.sigma < 0 or self.sigma > 2 * numpy.pi:
+            raise ValueError(
+                'Sigma must be non-negative and smaller than 2π '
+                '(preferably much smaller)'
+            )
+        self.rayleigh_scale = 0.66 * self.sigma
```
titodalcanton, line 72: You can drop `rayleigh_scale`, it is not used in this distribution.

pycbc/distributions/sky_location.py
```diff
@@ -46,6 +46,86 @@ class UniformSky(angular.UniformSolidAngle):
...
 
+class UniformCircleSky:
+
+    name = 'uniform_circle_sky'
+    _params = ['ra', 'dec']
+    
+    def __init__(self, **params):
+        self.mean_ra = angle_as_radians(params['mean_ra'])
+        self.mean_dec = angle_as_radians(params['mean_dec'])
+        self.sigma = angle_as_radians(params['sigma'])
```
titodalcanton, line 57: I would rename sigma to radius in this class. Sigma is reminescent of a standard deviation, but in the case of a uniform disk, it makes more sense to talk about radius.

pycbc/distributions/sky_location.py
```diff
@@ -87,33 +167,33 @@ class FisherSky:
...
                 'Fisher approximation to be valid',
-                sigma,
+                self.sigma,
             )
-        self.rayleigh_scale = 0.66 * sigma
+        self.rayleigh_scale = 0.66 * self.sigma
         # Prepare a rotation that puts the North Pole at the mean position
         self.rotation = Rotation.from_euler(
-            'yz', [numpy.pi / 2 - mean_dec, mean_ra]
+            'yz', [numpy.pi / 2 - self.mean_dec, self.mean_ra]
```
titodalcanton, line 196: It seems these changes are not necessary. Simply add this at the end of the `__init__` method:
```python
# store the center position for `to_uniform_patch()`
self.mean_ra, self.mean_dec = mean_ra, mean_dec
```

pycbc/distributions/sky_location.py
```diff
@@ -129,13 +209,13 @@ def from_config(cls, cp, section, variable_args):
...
+        self.mean_ra = cp.get_opt_tag(section, 'mean_ra', tag)
+        self.mean_dec = cp.get_opt_tag(section, 'mean_dec', tag)
+        self.sigma = cp.get_opt_tag(section, 'sigma', tag)
         return cls(
-            mean_ra=mean_ra,
-            mean_dec=mean_dec,
-            sigma=sigma,
+            mean_ra=self.mean_ra,
+            mean_dec=self.mean_dec,
+            sigma=self.sigma,
```
titodalcanton, line 218: These are not necessary.

pycbc/distributions/sky_location.py
```diff
@@ -162,6 +242,10 @@ def rvs(self, size):
         rot_radec['dec'] = numpy.arcsin(rot_cart[:, 2])
         return rot_radec
 
+    def to_uniform_patch(self, coverage):
+        new_sigma = numpy.sqrt(self.rayleigh_scale**2 * (-2*numpy.log(1-coverage)))
+        return UniformCircleSky(mean_ra=self.mean_ra, mean_dec=self.mean_dec, sigma=new_sigma)
```
titodalcanton, line 247: See my previous comment, I would rename `new_sigma` to `radius` here.

pycbc/distributions/sky_location.py
```diff
@@ -243,6 +327,24 @@ def normalize_azimuth(self, phi):
...
+        prob = numpy.sort(prob)
+        ind_prob_rev_sorted = ind_prob_sorted[::-1]
+        list_ind = []
+        for i in range(len(ind_prob_sorted)):
+            list_ind.append(ind_prob_rev_sorted[i][0])
+        cum_sum_prob = numpy.cumsum(prob[::-1])
+        covered_sky = cum_sum_prob[cum_sum_prob<coverage]
+        self.pix_probs[list_ind[:len(covered_sky)]] = 1/len(covered_sky)
+        self.pix_probs[self.pix_probs!= 1/len(covered_sky)] = 0
+
```
titodalcanton, line 347: Not super necessary (maybe to be done in a later PR) but I would prefer this method to return a *new* HealpixSky instance, rather than modifying the existing instance, for consistency with the same method of the FisherSky class.

titodalcanton:
You will also need to add uniform_circle_sky to this list to prevent the tests from failing:
https://github.com/gwastro/pycbc/blob/master/test/test_distributions.py#L33-L39

bin/pycbc_make_sky_grid
```diff
@@ -62,92 +71,62 @@ def make_multi_det_grid(args):
...
-            number_of_points = int(2 * np.pi * i)
-            for j in range(number_of_points):
-                sky_points = np.row_stack(
-                    (
-                        sky_points,
-                        np.array([j / i, np.pi / 2 - i * angular_spacing]),
-                    )
+    input_dist = eval(args.input_dist)
+    sky_dist = eval(args.input_dist)
+    sky_dist = sky_dist.to_uniform_patch(args.coverage)
```
titodalcanton, line 76:
```suggestion
    sky_dist = input_dist.to_uniform_patch(args.coverage)
```

bin/pycbc_make_sky_grid
```diff
@@ -62,92 +71,62 @@ def make_multi_det_grid(args):
...
-    rota = r.apply(cart)
-
-    # Convert cartesian coordinates back to spherical coordinates
-    spher = cart_to_spher(rota)
-
-    sky_grid = SkyGrid(
-        spher[:, 0], spher[:, 1], args.instruments, args.trigger_time
-    )
-    return sky_grid
+                for i in range(len(detector_pairs))
```
titodalcanton, line 90: You can simplify this code if you iterate directly over `detector_pairs` instead of over its index:
```
a.time_delay_from_detector(b, …) for a, b in detector_pairs
```

bin/pycbc_make_sky_grid
```diff
@@ -62,92 +71,62 @@ def make_multi_det_grid(args):
...
+            ]
+            
+
+            # Calculate the light travel time between the detector pairs
+            light_travel_times = [
+                detector_pairs[i][0].light_travel_time_to_detector(detector_pairs[i][1])
+                for i in range(len(detector_pairs))
+            ]
```
titodalcanton, line 98: Same comment as above, iterate over `detector_pairs`, not over its index.

bin/pycbc_make_sky_grid
```diff
@@ -62,92 +71,62 @@ def make_multi_det_grid(args):
...
+
+            # Calculate the required angular spacing between the sky points
+            ang_spacings = [
+                (2 * args.timing_uncertainty)
+                / np.sqrt(light_travel_times[i] ** 2 - tds[i] ** 2)
+                for i in range(len(detector_pairs))
+            ]
```
titodalcanton, line 105: If you convert `light_travel_times` and `tds` to Numpy arrays, then you can do this operation without a list comprehension (simpler and faster).

bin/pycbc_make_sky_grid
```diff
@@ -62,92 +71,62 @@ def make_multi_det_grid(args):
...
+                grid = np.vstack((grid, input_dist.get_max_pixel_prob()))
+                continue
+            # FIXME it is not necessary to calculate *all* distances,
+            # we can reject as soon as we find one below threshold!
+            dist = angular_distance((prop_ra, prop_dec), grid)
+            if dist > angular_spacing:
+                # far enough from other points, accept
+                grid = np.vstack((grid, (prop_ra, prop_dec)))
+        num_new_accepted = grid.shape[0] - prev_size
+        logging.info(f'{num_new_accepted} points accepted, {grid.shape[0]} total')
```
titodalcanton, line 120: Logging functions are recommended to be used like this:
```suggestion
        logging.info('%d points accepted, %d total', num_new_accepted, grid.shape[0])
```
Not with f-strings.

bin/pycbc_make_sky_grid
```diff
@@ -62,92 +71,62 @@ def make_multi_det_grid(args):
...
+                / np.sqrt(light_travel_times[i] ** 2 - tds[i] ** 2)
+                for i in range(len(detector_pairs))
+            ]
+            angular_spacing = np.min(ang_spacings)
+
+            if grid.shape[0] == 0:
+                # first point, accept
+                # FIXME handling the first point can surely be moved before the loop!
+                grid = np.vstack((grid, input_dist.get_max_pixel_prob()))
+                continue
```
titodalcanton, line 112: This is a good opportunity to do what the FIXME says. Instead of doing
```
grid = np.empty((0,2))
```
and then entering the loop, you can just initialize `grid` with the first point before the loop.

bin/pycbc_make_sky_grid
```diff
@@ -156,12 +135,19 @@ parser.add_argument(
...
-    "radians are assumed.",
+    help="Input distribution of the sky map that you have, could "
+    "be HealpixSky or FisherSky. See sky location documentation for "
+    "more details",
+)
+parser.add_argument(
+    '--coverage',
+    type=float,
+    required=True,
+    help="Percentage of probability that you want to cover",
```
titodalcanton, line 150: Coverage will not make sense if the input distribution is `UniformSky`. I suggest you do the following:
1. Make `--input-dist` and `--coverage` optional arguments (remove `required=True`).
2. When instantiating `input_dist` and `sky_dist`, check if `--input-dist` and `--coverage` were given. If so, do what is being done now. Otherwise, instantiate both `input_dist` and `sky_dist` from `UniformSky`.

bin/pycbc_make_sky_grid
```diff
@@ -191,10 +177,11 @@ else:
     # network.
     sky_grid = make_multi_det_grid(args)
 
+
 extra_attributes = {
-    'trigger_ra': args.ra,
-    'trigger_dec': args.dec,
-    'sky_error': args.sky_error,
+    'triiger_ra': sky_grid[0][0],
```
titodalcanton, line 182:
```suggestion
    'trigger_ra': sky_grid[0][0],
```

pycbc/distributions/sky_location.py
```diff
@@ -243,6 +345,26 @@ def normalize_azimuth(self, phi):
         phi[phi >= 2 * numpy.pi] -= 2 * numpy.pi
         return phi
 
+    def to_uniform_patch(self, coverage):
+        """Apply the coverage criterion on the skymap
+        """
+        uniform_patch = self
```
titodalcanton, line 351: If you do this, you will modify the original distribution, which is not ideal. You probably need to do a "deep copy" here: https://docs.python.org/3/library/copy.html

pycbc/distributions/sky_location.py
```diff
@@ -46,6 +46,87 @@ class UniformSky(angular.UniformSolidAngle):
     _default_azimuthal_angle = 'ra'
 
 
+class UniformCircleSky:
+
```
titodalcanton, line 50: Add a docstring to this class to explain what it does. Use the docstring of `FisherSky` as a model.

titodalcanton (changes requested):
Here is another batch of comments. In addition to those specific ones, please rename the class `UniformCircleSky` to `UniformDiskSky` and the method `get_max_pixel_prob` to `get_max_prob_point`. I think those are more appropriate names.

Thomas-JACQUOT:
I just put back `--ra`, `--dec` and `--sky-error` arguments because they are needed in the `make_single_det_grid()` function

pycbc/distributions/sky_location.py
```diff
@@ -243,6 +364,26 @@ def normalize_azimuth(self, phi):
         phi[phi >= 2 * numpy.pi] -= 2 * numpy.pi
         return phi
 
+    def to_uniform_patch(self, coverage):
+        """Apply the coverage criterion on the skymap
+        """
```
titodalcanton, line 369:
```suggestion
        """Return a new HealpixSky object that represents a patch of uniform probability,
        defined as the region that encloses a given probability in the original map.
```

titodalcanton (changes requested):
See specific comments. It is probably a good idea to rebase on master as well while you are at it.

titodalcanton (changes requested):
Some comments remaining on `pycbc_make_sky_grid`. The rest is good to go, I think.

PR #5090, opened 2025-03-26
add trigger cuts in trigger fit

yi-fan-wang:
## Add trigger cuts in the single detector trigger fitting codes

This is a new feature

This change wouldn't affect anything if you don't use trigger cuts in the trigger fitting, .i.e., it's backward compatible. But offline search, live search, PyGRB, and any other pipeline that depends on trigger fitting could use this new feature.

This change follows style guidelines (See e.g. [PEP8](https://peps.python.org/pep-0008/)), has been proposed using the [contribution guidelines](https://github.com/gwastro/pycbc/blob/master/CONTRIBUTING.md). I'll try to add some unittests, but they don't exist at the moment

This change won't change any current pipelines, as it has backward compatibility.

## Motivation

We made some trigger cutting in the single trigger fitting and the coincidence job in 3-OGC and 4-OGC with an argument called "statistic-keywords", for example see [here](https://github.com/gwastro/4-ogc/blob/master/search/bank/bank_bbh.ini#L36). However I think it's never merged into the main branch. At the same time, there is a nice module in PyCBC to cut triggers using the codes in `events/cuts.py`. So I introduce these cuts into the trigger fitting codes. The coincidence codes already have it.

## Contents

Adding trigger cuts in [bin/all_sky_search/pycbc_fit_sngls_by_template](https://github.com/gwastro/pycbc/compare/master...yi-fan-wang:pycbc:triggerfitcut?expand=1#diff-3e27918d238d5d855742e1f2d3f6a875b61f2abc45a1ff1426067900c1fb5b66)

## Testing performed

To be performed

## Additional notes

### DIFF
--- a/bin/all_sky_search/pycbc_fit_sngls_by_template
+++ b/bin/all_sky_search/pycbc_fit_sngls_by_template
@@ -21,12 +21,13 @@ import pycbc
 from pycbc import events, init_logging
 from pycbc.events import triggers, trigger_fits as trstats
 from pycbc.events import stat as statsmod
+from pycbc.events import cuts
 from pycbc.types.optparse import MultiDetOptionAction
 from pycbc.io import HFile
 
 #### DEFINITIONS AND FUNCTIONS ####
 
-def get_stat(statname, trigs, threshold):
+def get_stat(args, trigs, threshold):
     # For now this is using the single detector ranking. If we want, this
     # could use the Stat classes in stat.py using similar code as in hdf/io.py
     # This requires additional options, so only change this if it's useful!
@@ -35,6 +36,7 @@ def get_stat(statname, trigs, threshold):
     select = []
     size = len(trigs['end_time'])
     s = 0
+    trigger_cut_dict, _ = cuts.ingest_cuts_option_group(args)
     while s < size:
         e = s + chunk_size if (s + chunk_size) <= size else size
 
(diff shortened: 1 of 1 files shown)


bin/all_sky_search/pycbc_fit_sngls_by_template
```diff
@@ -21,12 +21,13 @@ import pycbc
...
 from pycbc.events import triggers, trigger_fits as trstats
 from pycbc.events import stat as statsmod
+from pycbc.events import cuts
 from pycbc.types.optparse import MultiDetOptionAction
 from pycbc.io import HFile
 
 #### DEFINITIONS AND FUNCTIONS ####
 
-def get_stat(statname, trigs, threshold):
+def get_stat(args, trigs, threshold):
```
GarethCabournDavies, line 30: This looks good at first glance

I'm actually surprised that this line wasn't `args` before, as `statname` doesn't seem to be used at all.

yi-fan-wang: Looking at the code history, statname is a legacy from an old code used by single stat calculation. Then the single stat was much simplified and statname was left useless.

yi-fan-wang:
I split this PR to https://github.com/gwastro/pycbc/pull/5095 to separate the changes of get chisq function into that PR.

yi-fan-wang:
Hi @tdent I did some small tests with this PR on the mini search, which is one of the tests in the CI. Here is the search results with this PR [1], and the results using the v2.8.0 PyCBC [2]. The two should be the same, and indeed this is what I see.

I also tried to test the feature of this PR, by adding a trigger cutting in trigger fitting

> [fit_by_template]
> trigger-cuts = traditional_chisq:1.8:upper sg_chisq:3.5:upper
>
> [coinc]
> trigger-cuts = traditional_chisq:1.8:upper sg_chisq:3.5:upper

The result is shown in [3]. If looking at the background number 1 triggers, it's indeed different from the results of [1,2], as the old #1 trigger is screened now.

Overall I think this PR is ready to be merged.

[1]https://hypatia.aei.mpg.de/~yiwang/test/pycbc_minisearch_cuttrigs/html/
[2]https://hypatia.aei.mpg.de/~yiwang/test/pycbc_minisearch_cuttrigs/comparison_pycbcv2.8.0/html/
[3]https://hypatia.aei.mpg.de/~yiwang/test/pycbc_minisearch_cuttrigs/realcuttingtrigs/html/

tdent (approved):
Approve, assuming it does the things you want !

PR #5091, opened 2025-03-26
Cleanup stuff that handles the Docker image and CVMFS release

titodalcanton:
While working on #5089, I found the Docker and CVMFS stuff a bit hard to follow and redundant, so here I try to clean in up a bit.

This is mostly a cleanup of the Docker image and publication on CVMFS with no expected functional changes.

## Motivation

Mainly to simplify the upcoming PR that will upgrade the Python version shipped in the Docker image and in the CVMFS virtualenvs.

## Contents

First, I noticed that the CI jobs that build the Docker image and the CVMFS venv currently run the "setup Python" block. This does not seem necessary though, since everything in both jobs happens inside containers anyway. So I removed the "setup Python" stuff.

Second, I cleaned up the Dockerfile, mostly to break down long chains of commands into separate lines, to make further changes easier to review.

Third, I removed a couple of things in the Dockerfile that I *think* had no effect. Hopefully this can be reviewed carefully by others though.

Lastly, I reformatted a couple shell scripts  associated with Docker and CVMFS to make them easier to read and edit.

Again, I tried hard not to change the functionality. I think there are further things to improve, for example we should probably clear the dnf and pip caches after making the Docker image, to reduce the size. But I prefer to propose those in a followup PR, which should be easier to review after this one.

N/A

## Testing performed

None, relying on the CI.

## Additional notes

N/A

### DIFF
--- a/.github/workflows/build_venv.yml
+++ b/.github/workflows/build_venv.yml
@@ -13,11 +13,6 @@ jobs:
       -
         uses: actions/checkout@v1
       -
-        name: "Set up Python"
-        uses: actions/setup-python@v5
-        with:
-          python-version: 3.9
-      -
         env:
           OSG_ACCESS: "${{secrets.OSG_ACCESS}}"
         name: "Preparing ssh stuff"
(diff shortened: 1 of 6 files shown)


titodalcanton:
Another thing I would like to remove is forcing pip<22.0 when creating the Docker image. Can anyone remember why we do that?

ahnitz (approved):

PR #5099, opened 2025-04-01
Get LAL aux data from Zenodo and cache it

titodalcanton:
Edit the basic CI tests to download the LAL auxiliary data (currently only the files for SEOBNRv4_ROM) from Zenodo rather than git.ligo.org, and use a GitHub cache for it. While the speed gain is negligible, hopefully this reduces the load on git.ligo.org and helps make the CI tests more robust against web site outages.

### DIFF
--- a/.github/workflows/basic-tests.yml
+++ b/.github/workflows/basic-tests.yml
@@ -26,23 +26,34 @@ jobs:
         sudo apt-get -o Acquire::Retries=3 update
         sudo apt-get -o Acquire::Retries=3 install *fftw3* mpi intel-mkl* graphviz
         pip install tox pip setuptools --upgrade
-    - name: installing auxiliary data files
+    - name: Cache LAL auxiliary data files
+      id: cache-lal-aux-data
+      uses: actions/cache@v4
+      with:
+        key: lal-aux-data
+        path: ~/lal_aux_data
+    - if: ${{ steps.cache-lal-aux-data.outputs.cache-hit != 'true' }}
+      name: Download LAL auxiliary data files
       run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/basic-tests.yml
```diff
@@ -26,23 +26,34 @@ jobs:
         sudo apt-get -o Acquire::Retries=3 update
         sudo apt-get -o Acquire::Retries=3 install *fftw3* mpi intel-mkl* graphviz
         pip install tox pip setuptools --upgrade
-    - name: installing auxiliary data files
+    - name: Cache LAL auxiliary data files
```
GarethCabournDavies, line 29: Would you be able to add a link to the instructions on how to do this into this PR? Just so we have it on-hand in case something breaks in the future

GarethCabournDavies: It looks like https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/caching-dependencies-to-speed-up-workflows, but if you used something else, do say

.github/workflows/basic-tests.yml
```diff
@@ -26,23 +26,34 @@ jobs:
...
+      id: cache-lal-aux-data
+      uses: actions/cache@v4
```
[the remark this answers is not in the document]
titodalcanton: Good question. `actions/cache` is just a repository on GitHub (https://github.com/actions/cache), and v4 is just a tag on it, so I assume the same argument applies here as for any other action (though tags in official actions are probably a bit safer than some random repository out there).

.github/workflows/basic-tests.yml
titodalcanton: Yes that is what I used, though I also looked at what Duncan did in `mac-test.yml`.

PR #5110, opened 2025-04-24
Dont use pylab, use pyplot directly

GarethCabournDavies:
I had some import issues which I thought were caused by using pylab rather than pyplot directly. Turns out they aren't, but this should mean that we aren't using this when it is deprecated

This is a deprecation fix
This change touches all areas of the codebase, but should have no effect

## Motivation
From [matplotlib documentation](https://matplotlib.org/3.1.1/tutorials/introductory/usage.html#matplotlib-pyplot-and-pylab-how-are-they-related)
```pylab is a convenience module that bulk imports [matplotlib.pyplot](https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.pyplot.html#module-matplotlib.pyplot) (for plotting) and [numpy](https://docs.scipy.org/doc/numpy/reference/index.html#module-numpy) (for mathematics and working with arrays) in a single namespace. pylab is deprecated and its use is strongly discouraged because of namespace pollution. Use pyplot instead.```

## Contents
mainly a find-and-replace style change for pylon - one bit required actual thought about whether pylab was wrapping numpy or pyplot

### DIFF
--- a/bin/bank/pycbc_bank_verification
+++ b/bin/bank/pycbc_bank_verification
@@ -27,7 +27,7 @@ import logging
 import numpy
 import matplotlib
 matplotlib.use('Agg')
-import pylab
+from matplotlib import pyplot as plt
 
 from igwn_ligolw import lsctables, utils as ligolw_utils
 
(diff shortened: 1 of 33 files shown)


ahnitz (approved):
@GarethCabournDavies This seems pretty straightforward. Thanks for cleaning this up.

PR #5114, opened 2025-04-28
Expand multi_inspiral examples and tests

titodalcanton:
Refactoring and improvement of the `pycbc_multi_inspiral` examples and tests.

This is an extension of the examples and CI tests.

This change affects PyGRB.

This change changes the examples and tests, but no analysis code.

## Motivation

The current examples and CI tests for `pycbc_multi_inspiral` only exercise it with the HLV detector network. However, its code takes several branches depending on how many detectors are being used. We would like to do some cleanup/refactoring of that code in the near future, so we want to first expand the scopes of the CI tests accordingly.

## Contents

First, I edited `run.sh` to avoid the explicit download of frame files, and instead use PyCBC's download capability. This requires enabling autogating in `pycbc_multi_inspiral` because we have to gate the L1 glitch ourselves; I think this makes for a better test anyway. Moreover, getting frame files like this has the extra benefit of leaving the frame files in Astropy's download cache, and I will set up a GitHub Actions cache for that in another PR, so we will avoid further downloads of those frame files from GWOSC.

Next, I replaced the old script `run.sh` with three new scripts `gw170817_X.sh`, with `X` in {`hlv`, `hl`, `h`}. The scripts are identical except for the parameters that control which detectors are used. I could have used some sort of loop in a single script, but since these scripts also act as examples, I wanted to keep them simple to hopefully encourage people to look at them.

Finally, I had to edit `check_gw170817_trigs.py` because the network SNR will now depend significantly on which detectors were analyzed. I also added some basic sanity checks to all triggers in that script.

N/A

## Testing performed

Relying on GitHub Actions.

## Additional notes

N/A


ahnitz:
@titodalcanton, All, would it be possible to add these to the documentation while you are at it?

sebastiangomezlopez:
Thanks @titodalcanton this is very useful, I'm happy to merge this changes once you answer @ahnitz request.

sebastiangomezlopez (approved):

titodalcanton:
@ahnitz are you thinking about a page dedicated to pycbc_multi_inspiral, for example under https://pycbc.org/pycbc/latest/html/searches.html? That makes sense, though I would prefer to make a separate PR for it.

ahnitz:
@titodalcanton Yeah, and fair enough. It can be a separate PR

examples/multi_inspiral/clean.sh
```diff
@@ -1 +1 @@
-rm -f bank_veto_bank.xml *.hdf *.gwf
```
pannarale, line 1: Are the frame files deleted in another way now?

titodalcanton: Deleting them is not necessary because they are not even saved in the current directory. They are cached under `~/.astropy/cache/`.

PR #5115, opened 2025-04-30
PyGRB: handling the 2 IFO case with the coherent search methods

pannarale:
`pycbc_multi_inspiral` treats the case with 2 interferometers as a coincident search, rather than a coherent search.  With the assumption of a left/right circularly polarised signal, this should not be the case.  See https://inspirehep.net/literature/1323266, Section V A.

This is a: new feature.

This change is supported by the CI tests in PyCBC.

## Contents
Anywhere `pycbc_multi_inspiral` separated the 1 IFO, the 2 IFO and more than 2 IFO cases, I simply reduced the branches to 1 IFO vs more than 1 IFO.

## Testing performed
Tests with full runs are in progress.

## Additional notes

### DIFF
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -793,10 +793,10 @@ with ctx:
                         coinc_triggers = {}
                         logging.debug("No coincident triggers were found")
                     # If there are triggers above coinc threshold and more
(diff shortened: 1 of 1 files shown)


titodalcanton (approved):
Looks good, mostly trusting the CI here.

PR #5117, opened 2025-05-05
Add support for time and sky-marginalization when using the gated gaussian models

cdcapano:
# Add support for time and sky-marginalization when using the gated Gaussian models

This PR (finally) adds the code needed to be able to marginalize over sky and time uncertainties when using the GateGaussian models to excise part of a signal in the time domain, such as when doing black hole spectroscopy. The methods used here were described in [arXiv:2312.15146](https://arxiv.org/abs/2312.15146) and [arXiv:2312.14118](https://arxiv.org/abs/2312.14118). Much of the code in this pull request was used for those papers.

To do the marginalization, it is necessary to recalculate the normalization of the likelihood, which is based on the determinant of the covariance matrix. Since the covariance matrix changes as you change the gate, the determinant needs to be recalculated on the fly. Fully calculating the determinant is too computationally intractable. However, as shown in [arXiv:2312.15146](https://arxiv.org/abs/2312.15146), the determinant as a function of gate width can be approximated to high precision using linear interpolation. This patch adds support for that. If normalization is turned on, the determinant is calculated for 3 gate sizes using the full scipy determinant solver. A linear interpolator is then created to interpolate between the gate sizes. When sampling, the interpolator is used to quickly calculate the normalization.

It was also found in [arXiv:2312.15146](https://arxiv.org/abs/2312.15146) that marginalizing over time requires a signal model for all times that a signal is observable, even those that you wish to ignore. To accomplish this, we used the hierarchical model in that paper, with an IMR template used for the pre-merger model and a QNM model for the post merger. In testing and using that, we discovered some bugs in the hierarchical model in the way it stores data. This patch also fixes that.

We've added unit tests to test these changes (in `test/test_bhspec.py`). The added tests use the maxL parameters we found doing a marginalized QNM analysis of GW150914 in [arXiv:2312.14118](https://arxiv.org/abs/2312.14118) as a reference point (the maxL values were retrieved from the [data release for that paper](https://zenodo.org/records/10493247)). The reference values are added as a json file, along with the config file from that paper necessary to replicate the analysis. The unit tests use that config file to setup the hierarchical analysis, set the maxL values, then check that the loglikelihood obtained is the same as what was obtained in that paper. We also check that toggling normalization on and off yields the same value as the unnormalized loglikelihood +/- the log normalization value. Finally, we also test that we can load the model using the config file, data, and psds stored in a checkpoint file and get back the same loglikelihood.

The commits in this branch were written by a combination of @acorreia61201 and myself. This patch supersedes PR #4521.


pycbc/types/timeseries.py
```diff
@@ -563,6 +563,31 @@ def psd(self, segment_duration, **kwds):
                            seg_stride=seg_stride,
                            **kwds)
 
+    def get_gate_indices(self, time, window):
```
ahnitz, line 566: Is this meant to match up to the 'gate' method? Is there a risk that they could diverge from each other? 

This raises a couple questions. (1) Is this really tied to gating or is this somewhat more general and could be articulated as such? (2) if it really is this specific to gating, does it make sense to have it more closely tied to the actual gate application function?

cdcapano: @ahnitz This replaces the same code in the `gate` method. See the line removals at lines 608-609 below. I moved these lines to their own method because we need to be able to get the gate indices in the GatedGaussian models. Having a stand alone method to do that serves as a single point of truth.

I think that answers your question? Or did you mean another `gate` function?

cdcapano:
@ahnitz This is ready now.

ahnitz (approved):
@cdcapano This looks great. I'm really happy to see people add proper unit tests in with code changes.

PR #5118, opened 2025-05-06
PyGRB single IFO case fixes

pannarale:
- This PR fixes a bug in the PyGRB post-processing: when auto-gating occurs, the `gating` entry of results files should not be treated as a numpy array pertaining to the triggers found.  This was never encountered before because auto-gating was never triggered in the tests we ran so far.
- This PR also makes sure that the coincident SNR cut is applied in the single IFO case applies, in line with what happens when 2 or more IFOs are available for the analysis.  This keeps the amount of data under control, and is the strategy adopted previously with PyGRB+coh_PTF.
- Finally, this PR fixes two comments: one for syntax and one missing an input parameter description.

This change affects: PyGRB

## Testing
With the first of these changes, I was able to produce plots that would otherwise through an error when trying to use index triggers on, e.g., `L1/gating/` in the results files.

The output files of inspiral jobs of a single IFO run are seen to be reduced by a factor 10 in size with the cut on coincident SNR (which is effectively a second cut on single IFO SNR) in place.

### DIFF
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -757,9 +757,9 @@ with ctx:
                         % len(snr_dict[ifo])
                         for ifo in args.instruments
                     }
-                    # Calculate the coincident and coherent SNR.
-                    # First check there is enough data to compute the SNRs.
-                    if len(coinc_idx) != 0 and nifo > 1:
+                    # Calculate the coincident SNR, regardless of how many IFOs
+                    # are available, only if there are triggers so far.
+                    if len(coinc_idx) != 0:
                         # Find coinc SNR at trigger times and apply coinc SNR
                         # threshold (which depopulates coinc_idx accordingly)
                         (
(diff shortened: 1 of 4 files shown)


bin/pycbc_multi_inspiral
```diff
@@ -757,9 +757,9 @@ with ctx:
                         % len(snr_dict[ifo])
                         for ifo in args.instruments
                     }
-                    # Calculate the coincident and coherent SNR.
-                    # First check there is enough data to compute the SNRs.
-                    if len(coinc_idx) != 0 and nifo > 1:
+                    # Calculate the coincident SNR, regardless of how many IFOs
+                    # are available, only if there are triggers so far.
+                    if len(coinc_idx) != 0:
```
titodalcanton, line 762: It is strange to talk about "coincident SNR" in a one-detector analysis. I think a more correct term is incoherent network SNR, which is valid regardless of the number of detectors. If you agree, please update the comment.

pannarale: Okay, I will update the comment, but I will figure out a way to leave the term "coincident" given the `coinc` in variable names all over the place.

titodalcanton:
Please rebase on master to pick up the CI fix.

titodalcanton (approved):

PR #5119, opened 2025-05-07
Workaround for git.ligo.org authentication change

titodalcanton:
The authentication method used by git.ligo.org changed recently. This breaks downloading files from repositories on git.ligo.org via `pycbc.workflow.resolve_url()`. Here is one way to fix it.

This is a new feature required by an LVK infrastructure change.

This change affects all workflows that use `resolve_url()`, in particular the offline all-sky and PyGRB searches.

This change changes basic workflow infrastructure.

## Motivation

We have historically relied on `ciecplib` to download files from repositoried hosted on git.ligo.org via Shibboleth authentication. Due to changes recently made to git.ligo.org, and outside our control, this does not work anymore. The recommended approach now is to use a GitLab Personal Access Token which each user has to get, store somewhere, and give to each HTTP request for authorization. The URL scheme to download a given file has also changed; the previous scheme, which you still get when you click the "download" button on the web interface, does not work anymore.

## Contents

I refactored part of `resolve_url()`, keeping the functionality as similar as possible, and added a special case for URLs that point to git.ligo.org. That special case assumes the presence of a file at `$XDG_CONFIG_HOME/pycbc/git-ligo-org.pat` with GitLab's PAT in it, reads it, and passes its content as a header to the `ciecplib` call.

Note that the previous code already had git.ligo.org and code.pycbc.phy.syr.edu as special cases to invoke the Shibboleth authentication, though it was much simpler before. code.pycbc.phy.syr.edu does not resolve anymore, so I dropped it.

None.

## Testing performed

I tried running the new `resolve_url()` (after setting up my PAT file) and it downloads the file correctly.

## Additional notes

Everyone using a PyCBC workflow that downloads files from a git.ligo.org repository will now have to set up a GitLab PAT via the web interface and store it on the cluster they want to use at `$XDG_CONFIG_HOME/pycbc/git-ligo-org.pat`.

### DIFF
--- a/pycbc/workflow/configuration.py
+++ b/pycbc/workflow/configuration.py
@@ -27,6 +27,7 @@ workflow construction. This module is described in the page here:
 https://ldas-jobs.ligo.caltech.edu/~cbc/docs/pycbc/ahope/initialization_inifile.html
 """
 
+import re
 import os
 import logging
 import stat
@@ -34,7 +35,6 @@ import shutil
 import subprocess
 from shutil import which
 import urllib.parse
-from urllib.parse import urlparse
 import hashlib
 
 from pycbc.types.config import InterpolatingConfigParser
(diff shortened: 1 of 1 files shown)


titodalcanton:
I tested this with several URLs we commonly use in offline analyses (different projects and different refs), and they all work.

spxiwh (approved):
I think I would prefer to use the gitlab API (and don't see the added dependancy as any sort of issue). But I am happy to approve any change that allows us to access configuration files!

GarethCabournDavies:
This looks good - just checking regarding the path to the token and environment variable - is this straightforward enough/very basic instructions to set up?

PR #5122, opened 2025-05-09
Add waveform transform to apply hypertriangle map

acorreia61201:
This PR adds a waveform transform that applies hypertriangulation to a set of indistinguishable parameters. The code is based on a scheme by Buscicchio et al. (https://doi.org/10.1103/PhysRevD.100.084041) to prevent label-switching when doing PE.

This is a new feature. Changes are applied to `pycbc.conversions` to add the transform itself and to allow for calls during inference or I/O.

## Motivation

Label switching can occur when performing Bayesian PE with a set of indistinguishable parameters. If the N parameters are unlabeled, there will be N! modes in the posterior, which can greatly slow down convergence for high numbers of parameters. A simple example of this is primary/secondary mass in CBC systems, which can be remedied by enforcing $m_1 \geq m_2$ via the existing `primary_mass` and `secondary_mass` waveform transforms. Another example of this is QNM parameters in post-merger analyses.

A more general way to approach this is to apply hypertriangulation as proposed by Buscicchio et al. This method transforms a set of indistinguishable parameters assumed to be sampled on identical uniform priors. The transformation produces a set of parameters sorted in ascending order (which are _not_ the same as the inputs; i.e. this algorithm is not simply sorting the inputs).

To use during PE, the parameters to enforce labels by can be sampled over identical uniform priors. The waveform transform can then be applied using the `custom_multi` method. The user must provide the range of the uniform prior as a kwarg; otherwise prior is treated as a unit hypercube (i.e. each parameter is assumed to vary over [0,1]).

For example, consider a waveform approximant that takes in frequency parameters `f1`, `f2`, and `f3`, which all vary over the same uniform prior. We can apply to the hypertriangle map as follows in the configuration file:

```
[variable_params]
f1_samp =
f2_samp =
f3_samp =

[prior-f1_samp]
name = uniform
min-f1_samp = 20
max-f1_samp = 256

[prior-f2_samp]
name = uniform
min-f2_samp = 20
max-f2_samp = 256

[prior-f3_samp]
name = uniform
min-f3_samp = 20
max-f3_samp = 256

[waveform-transforms_f1+f2+f3]
name = custom_multi
inputs = f1_samp, f2_samp, f3_samp
f1, f2, f3 = hypertriangle(f1_samp, f2_samp, f3_samp, bounds=(20, 256))
```

### DIFF
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -110,20 +110,17 @@ def hypertriangle(*params, bounds=(0, 1)):
     posterior. Adapted from Buscicchio et al (PhysRevD.100.084041).
 
     This transformation assumes that the input parameters were sampled
-    from the same uniform prior distribution.
+    from the same uniform prior distribution defined by bounds.
 
     Parameters
     ----------
     params : float(s) or array(s)
         The input parameter values. Parameters are assumed to be
-        sampled from the same uniform prior distribution. Values can
-        be input as an arbitrary number of non-keyword arguments, or
-        as a list using the following syntax:
+        sampled from the same uniform prior distribution. Each
+        argument is treated as one parameter; input arrays will
+        be treated as multiple values for that one parameter.
 
-        >>> param_list = [x, y, z, ...]
-        >>> hypertriangle(*param_list)
-
-    bounds : tuple
+    bounds : tuple (optional)
         The lower and upper bounds of the input parameter priors.
         Default (0, 1) for a unit hypercube.
 
@@ -132,6 +129,11 @@ def hypertriangle(*params, bounds=(0, 1)):
     array
         The mapped parameters. Output values are in ascending order.
     """
+    # check inputs all have the same shape
+    ref_shape = numpy.shape(params[0])
+    assert numpy.all([numpy.shape(params[i]) == ref_shape for i in range(len(params))]), \
+        "All inputs must have the same number of elements"
+    
     # map to numpy array
     params, input_is_array = ensurearray(params)
 
@@ -144,18 +146,20 @@ def hypertriangle(*params, bounds=(0, 1)):
     
     # hypertriangulate
     try:
-        K, num_pts = numpy.shape(scaled_params)
+        K, num_pts = scaled_params.shape
     except ValueError:
         K = numpy.size(scaled_params)
         num_pts = 1
-    idx = numpy.repeat(numpy.arange(K), repeats=num_pts, axis=0)
+    idx = numpy.repeat(numpy.arange(K), repeats=num_pts)
     scaled_params.resize(K, num_pts)
     idx.resize(K, num_pts)
-    fac = numpy.power(1 - scaled_params, 1./(K - idx))
+    fac = numpy.power(1 - scaled_params, 1/(K - idx))
     out_scaled_params = 1 - numpy.cumprod(fac, axis=0)
 
     # rescale to prior bounds
     out_params = (out_scaled_params * (bounds[1] - bounds[0])) + bounds[0]
+    if num_pts == 1:
+        out_params = [out_params[i][0] for i in range(K)]
     return out_params
 
 #
--- a/pycbc/conversions.py
+++ b/pycbc/conversions.py
@@ -102,6 +102,62 @@ def sec_to_year(sec):
     return sec / lal.YRJUL_SI
 
 
+def hypertriangle(*params, bounds=(0, 1)):
+    """
+    Apply a hypertriangle map to a series of input parameter values.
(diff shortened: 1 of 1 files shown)


WuShichao:
This might also be useful for the LISA Global fit.

ahnitz:
@acorreia61201 To clarify, I think you mean by this not being a simple sorting that it is not the case that say

f1_samp, f2_samp, f3_samp  = (30, 70, 10) ----> f1, f2, f3 = (70, 30, 10)

But rather the values of f*_samp get maps to specific (but numerically different values) within a space where f1,f2,f3 is ordered (e.g. a hypertriangle). This mapping is 1-1 in the sence that your original cube is mapped to the hypertriangle and maintains that the prior is has the same sense of uniform. 

The fact that you are setting the same bound "ranges" is more of an aid rather than implying that f1_samp of 30 actually means that any of f1/f2/f3 would actually be 30 (e.g. it could correspond to one of them being say 52.33). The guarantee here is simply that the f1/f2/f3 will have the same bounds, a uniform prior (but without the indistinguishable swapping removed by restricting to a triangle); essentially they cover the full space but implicitly correspond go a unimodal version of the same space.

ahnitz:
@acorreia61201 To check that you've implemented this correct, have you done a test where you simply draw some samples and then do the mapping to see that the points can mapped to the expected distribution?

RiccardoBuscicchio:
> @acorreia61201 To check that you've implemented this correct, have you done a test where you simply draw some samples and then do the mapping to see that the points can mapped to the expected distribution?

Just to clarify. When you define $f_1, f_2, f_3$  over the cube, it is very important to not think of them as physical $f$ 's. 
They are "virtual" variables (say, $f^{\star}_1, f^{\star}_2, f^{\star}_3$. The sampler "sees" the hypercube HC, but inside the likelihood one must apply the map $\phi(f^\star_1,f^\star_2,f^\star_3)$ to obtain the physical ones $f_1, f_2,f_3$, which will be in the HT, with a prior identical to that over HC. 

As a consequence, the posterior samples obtained by the sampler must be post processed with the same map, as they are  $(f^{\star}_1, f^{\star}_2,f^{\star}_3)$.

ahnitz:
> > @acorreia61201 To check that you've implemented this correct, have you done a test where you simply draw some samples and then do the mapping to see that the points can mapped to the expected distribution?
> 
> Just to clarify. When you define f 1 , f 2 , f 3 over the cube, it is very important to not think of them as physical f 's. They are "virtual" variables (say, f 1 ⋆ , f 2 ⋆ , f 3 ⋆ . The sampler "sees" the hypercube HC, but inside the likelihood one must apply the map ϕ ( f 1 ⋆ , f 2 ⋆ , f 3 ⋆ ) to obtain the physical ones f 1 , f 2 , f 3 , which will be in the HT, with a prior identical to that over HC.
> 
> As a consequence, the posterior samples obtained by the sampler must be post processed with the same map, as they are ( f 1 ⋆ , f 2 ⋆ , f 3 ⋆ ) .

@RiccardoBuscicchio I skimmed your paper and I think I follow you. I think we are saying the same thing (or I intended to).  In the configuration that @acorreia61201 shows for pycbc inference, the parameters f1_samp / f2_samp / f3_samp are the virtual variables that the sampler sees and so do not correspond to the actual frequencies. The mapping to the actual frequencies is done inside the likelihood function and that is indicated by the 'waveform_transforms' section. So whatever the signal is in the example should be expecting frequencies f1/f2/f3. 

What I'd expect is that if @acorreia61201 draws from the virtual parameters and then applies the mapping that the transformed samples covers a hypertriangle in the physical space that ensures only a single mode of the label switching symmetry is represented.

ahnitz:
@acorreia61201 You'll need to rebase, but otherwise, I think this looks reasonable.

ahnitz (approved):
Rebase, and check that all tests pass. Then I think this is good to go. Please add a configuration file and example to the distributions documentation as a follow-up PR after this is merged. Feel free to merge once the tests pass here.

RiccardoBuscicchio:
@acorreia61201 @ahnitz another subtle point that I'll let you decide whether it is relevant to assert. 
The map is applied to any *one* parameter of the indistinguishable objects, but the prior must be symmetric across all parameters of said objects. 
So maybe when HT is applied it is worth asserting that all parameters are identical across sources.

I am not familiar enough with the whole infrastructure, but I thought it was useful to raise this point here.

acorreia61201:
@ahnitz I've rebased and passed the unittests.

ahnitz (approved):

PR #5123, opened 2025-05-09
updated for depcreated fillcolor value

ahnitz:

This is a: bug fix

## Motivation
The most recent matplotlib release May 8th, dropped the ability to pass None as a facecolor to indicated that it should not be visible. Instead it requires passing the literal string 'none'. 

## Additional notes


cdcapano (approved):
Thanks for fixing this! Odd that they changed it to 'none' (if anything, you think they'd go the opposite way).

PR #5124, opened 2025-05-10
Optimize volume_montecarlo performance

fvngs:
Improved array operations by pre-allocating arrays and reducing concatenations

A standard efficiency upgrade for the volume_montecarlo function.
It doesn't change the output of the function at all.

This change affects: the offline search, PyGRB

## Motivation
I saw the function was using a lot of concat calls which could be reduced, as it can be a substantial performance bottleneck.

## Contents
All-round performance optimizations to the function were made, primarily improving the efficiency of the array operations. The performance improvements made were quite substantial, as can be seen from the testing below. 

## Testing performed
|      Size |    Original (s) |   Optimized (s) |  Speedup|
|---------|-----------------|----------------|---------------------------|
|       100 |        0.000651 |        0.000357 |     1.82x|
|     1000 |        0.001938 |        0.000629 |     3.08x|
|     10000 |        0.023510 |        0.007141 |     3.29x|
|    100000 |        0.323696 |        0.057062 |     5.67x|

### DIFF
--- a/pycbc/sensitivity.py
+++ b/pycbc/sensitivity.py
@@ -136,12 +136,15 @@ def volume_montecarlo(found_d, missed_d, found_mchirp, missed_mchirp,
     volume_error: float
         The standard error in the volume
     """
+    # Powers for different distributions
     d_power = {
         'log'             : 3.,
         'uniform'         : 2.,
         'distancesquared' : 1.,
         'volume'          : 0.
     }[distribution]
+    
+    # Powers for chirp mass scaling
     mchirp_power = {
             'log'             : 0.,
             'uniform'         : 5. / 6.,
@@ -149,13 +152,14 @@ def volume_montecarlo(found_d, missed_d, found_mchirp, missed_mchirp,
             'volume'          : 15. / 6.
     }[distribution]
 
-    # establish maximum physical distance: first for chirp distance distribution
+    # Establish maximum physical distance: first for chirp distance distribution
     if limits_param == 'chirp_distance':
+        # Standard BNS chirp mass for conversion
         mchirp_standard_bns = 1.4 * 2.**(-1. / 5.)
         all_mchirp = numpy.concatenate((found_mchirp, missed_mchirp))
         max_mchirp = all_mchirp.max()
         if max_param is not None:
-            # use largest injected mchirp to convert back to distance
+            # Use largest injected mchirp to convert back to distance
             max_distance = max_param * \
                                   (max_mchirp / mchirp_standard_bns)**(5. / 6.)
         else:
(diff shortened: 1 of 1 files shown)


pannarale:
This does not affect PyGRB, as far as I know.

pycbc/sensitivity.py
```diff
@@ -164,49 +168,43 @@ def volume_montecarlo(found_d, missed_d, found_mchirp, missed_mchirp,
...
-    # volume of sphere
+    # Calculate volume of sphere out to maximum distance
     montecarlo_vtot = (4. / 3.) * numpy.pi * max_distance**3.
 
-    # arrays of weights for the MC integral
+    # Calculate weights based on distribution parameters
     if distribution_param == 'distance':
-        found_weights = found_d ** d_power
-        missed_weights = missed_d ** d_power
+        found_weights = numpy.power(found_d, d_power)
```
tdent, line 182: I find this in the documentation for numpy power: 
`The ** operator can be used as a shorthand for np.power on ndarrays`

Hence, I do not believe this change to use numpy.power can have any effect on performance, and it reduces readability IMO. Please revert these specific changes unless they are shown to significantly improve performance.

pycbc/sensitivity.py
```diff
@@ -164,49 +168,43 @@ def volume_montecarlo(found_d, missed_d, found_mchirp, missed_mchirp,
...
     else:
         raise NotImplementedError("%s is not a recognized parameter"
                                   % limits_param)
 
-    # volume of sphere
+    # Calculate volume of sphere out to maximum distance
     montecarlo_vtot = (4. / 3.) * numpy.pi * max_distance**3.
 
-    # arrays of weights for the MC integral
+    # Calculate weights based on distribution parameters
```
tdent, line 180:
```suggestion
    # Calculate weights for the MC integral
```
'MC integral' is useful scientific/statistical context to understand what is happening

pycbc/sensitivity.py
```diff
@@ -164,49 +168,43 @@ def volume_montecarlo(found_d, missed_d, found_mchirp, missed_mchirp,
...
-                                  % distribution_param)
-
-    all_weights = numpy.concatenate((found_weights, missed_weights))
-
-    # measured weighted efficiency is w_i for a found inj and 0 for missed
-    # MC integral is volume of sphere * (sum of found weights)/(sum of all weights)
-    # over injections covering the sphere
-    mc_weight_samples = numpy.concatenate((found_weights, 0 * missed_weights))
-    mc_sum = sum(mc_weight_samples)
+        found_weights = (numpy.power(found_d, d_power) *
```
tdent, line 185: The comment has been removed here, please reinstate it.

pycbc/sensitivity.py
```diff
@@ -164,49 +168,43 @@ def volume_montecarlo(found_d, missed_d, found_mchirp, missed_mchirp,
...
-    # measured weighted efficiency is w_i for a found inj and 0 for missed
-    # MC integral is volume of sphere * (sum of found weights)/(sum of all weights)
-    # over injections covering the sphere
-    mc_weight_samples = numpy.concatenate((found_weights, 0 * missed_weights))
-    mc_sum = sum(mc_weight_samples)
+        found_weights = (numpy.power(found_d, d_power) * 
+                        numpy.power(found_mchirp, mchirp_power))
+        missed_weights = (numpy.power(missed_d, d_power) * 
+                         numpy.power(missed_mchirp, mchirp_power))
+
```
tdent, line 189: The final `else` statement with NotImplementedError has been removed.  I don't see why.  We want the code to raise prompt and informative errors ..

pycbc/sensitivity.py
```diff
@@ -164,49 +168,43 @@ def volume_montecarlo(found_d, missed_d, found_mchirp, missed_mchirp,
...
+    # Pre-allocate array and store found weights
+    total_size = len(found_weights) + len(missed_weights)
+    mc_weight_samples = numpy.zeros(total_size)
+    mc_weight_samples[:len(found_weights)] = found_weights
+
+    # Calculate Monte Carlo statistics
```
tdent, line 195: The explanation of the calculation in the previous calculation has been removed and replaced by a generic non-informative comment.  Please reinstate the previous comment.

pycbc/sensitivity.py
```diff
@@ -164,49 +168,43 @@ def volume_montecarlo(found_d, missed_d, found_mchirp, missed_mchirp,
...
+    mc_sum = numpy.sum(mc_weight_samples)
+    mc_sum_squares = numpy.sum(mc_weight_samples ** 2)
+    mc_sample_variance = (mc_sum_squares / len(mc_weight_samples) - 
+                         (mc_sum / len(mc_weight_samples)) ** 2)
 
     if limits_param == 'distance':
-        mc_norm = sum(all_weights)
+        mc_norm = sum(numpy.concatenate((found_weights, missed_weights)))
```
tdent, line 202: If numpy.concatenate is inefficient, why is it being used here?

pycbc/sensitivity.py
```diff
@@ -164,49 +168,43 @@ def volume_montecarlo(found_d, missed_d, found_mchirp, missed_mchirp,
...
+
+    # Pre-allocate array and store found weights
+    total_size = len(found_weights) + len(missed_weights)
+    mc_weight_samples = numpy.zeros(total_size)
+    mc_weight_samples[:len(found_weights)] = found_weights
+
+    # Calculate Monte Carlo statistics
+    mc_sum = numpy.sum(mc_weight_samples)
+    mc_sum_squares = numpy.sum(mc_weight_samples ** 2)
+    mc_sample_variance = (mc_sum_squares / len(mc_weight_samples) -
```
tdent, line 198: This calculation does not seem to reproduce the previous calculation using 'Ninj' in all cases. In addition, `len(mc_weight_samples)` is just the same as `total_size` so I don't see why the existing variable is note used.

tdent (commented):
I do not see what is the overall strategy here.  In one place a concatenate operation has been removed but the operation is added in more than one other place.  Arbitrary changes from `**` to `numpy.power` which should have no effect on performance are made.  In addition several nontrivial comments have been removed.

tdent:
What specific example function call was made to obtain the advertised speedups?

PR #5129, opened 2025-05-14
Fix incorrect OSDF URL resolution

titodalcanton:
`resolve_url()` does not seem to work for OSDF URLs. This should fix it.

This is a bug fix.

This change affects offline workflows.

## Motivation

#5001 changed the way we resolve `osdf://` URLs to use `pelican object get <source> <destination>`. However, the code there actually calls `pelican object\ get <source> <destination>`, i.e. "object get" is given as a single argument made of two words separated by a space, which `pelican` does not like:
```
Error: unknown command "object get" for "pelican"
Run 'pelican --help' for usage.
```

## Contents

This replaces the "object get" arg with two args "object" and "get". I also noticed that the OSDF branch of the code returned immediately after calling `pelican`, which is not what the other branches do, so I removed the return as well.

#5001

## Testing performed

None so far, be back shortly.

## Additional notes

This would be needed on the 2.3 and 2.8 release branches if we want to get files from the OSDF.

### DIFF
--- a/pycbc/workflow/configuration.py
+++ b/pycbc/workflow/configuration.py
@@ -205,9 +205,10 @@ def resolve_url(
         # OSDF will require a scitoken to be present and stashcp to be
         # available. Thanks Dunky for the code here!
         cmd = [
-            which("pelican") or "pelican", 
-            "object get", 
-            u.scheme + "://" + u.path, 
+            which("pelican") or "pelican",
+            "object",
+            "get",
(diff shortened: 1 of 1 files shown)


spxiwh (approved):
Though I worry about https://github.com/gwastro/pycbc/issues/5020

titodalcanton:
I just generated a test workflow that gets a file from OSDF, and this works.

PR #5132, opened 2025-05-21
Pygrb segplot bug

pannarale:

This is a: bug fix.

This change affects: PyGRB

This change changes: result plotting whenever no data is available, allowing the workflow generator to end gracefully.

## Motivation
When trying to generate a PyGRB workflow at a time when no data of any detector is available in the offsource and the onsource, the desired behaviour is for the workflow generation script to generate an empty plot, warn the user there is no data and exit.  However, this was not the case.

## Contents
Tracing back the error, all that needs to be done is to check whether a specific dictionary contains any meaningful information (i.e., segments) or not, rather than checking if it has keys.  If each key has empty lists, there are no segments, and the empty plot has to be produced.

[](https://github.com/gwastro/pycbc/issues/5131)

## Testing performed
Running this branch on data-empty times, the correct behaviour is seen.

### DIFF
--- a/pycbc/results/pygrb_plotting_utils.py
+++ b/pycbc/results/pygrb_plotting_utils.py
@@ -71,7 +71,7 @@ def make_grb_segments_plot(wkflow, science_segs, trigger_time, trigger_name,
     from pycbc.results.color import ifo_color
 
     ifos = wkflow.ifos
-    if len(science_segs.keys()) == 0:
+    if len(sum(science_segs.values(), [])) == 0:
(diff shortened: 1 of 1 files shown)


pannarale:
This is the only plot made if the analysis does not start, while the `pycbc_pygrb_plot_snr_timeseries` plotting issue was in post-processing the results of a GRB that could be analysed.

Thomas-JACQUOT (approved):

titodalcanton (approved):

PR #5134, opened 2025-06-04
Make `pycbc_pygrb_grb_info_table` more flexible 

Thomas-JACQUOT:
Adding a new argument to `pycbc_pygrb_grb_info_table` to make it work with #5085.

## Standard information about the request.

This is a new implementation to make `pycbc_pygrb_grb_info_table`  work with #5085.

This change affects PyGRB workflows.

This change changes plotting grb information.

## Motivation
PyGRB workflow will not be able to work if this change is not merged, `pycbc_pygrb_grb_info_table` only takes as input ra dec and sky-error, where with #5085 you can generate a sky grid without giving ra dec and sky error but only giving an input distribution. We need to add an argument that takes input distribution in account. 

## Contents
Only few Ifs statements and a new argument input_dist.

### DIFF
--- a/bin/pygrb/pycbc_pygrb_grb_info_table
+++ b/bin/pygrb/pycbc_pygrb_grb_info_table
@@ -35,6 +35,7 @@ import lal
 from pycbc import add_common_pycbc_options, init_logging
 import pycbc.version
 import pycbc.results
+import pycbc.distributions
 from pycbc.detector import Detector, ppdets
 from pycbc.results.pygrb_postprocessing_utils import get_antenna_dist_factor
 from pycbc.types import angle_as_radians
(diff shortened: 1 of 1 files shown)


bin/pygrb/pycbc_pygrb_grb_info_table
```diff
@@ -88,14 +90,23 @@ utc_time = datetime(*lal.GPSToUTC(opts.trigger_time)[0:6]).strftime("%Y-%m-%d %H
 data[0].append(utc_time)
 headers.append('UTC Time')
 
-data[0].append(f'{math.degrees(opts.ra):.3f}')
-headers.append('R.A. (deg)')
+if opts.input_dist != None:
```
titodalcanton:
```suggestion
if opts.input_dist is not None:
```

bin/pygrb/pycbc_pygrb_grb_info_table
```diff
@@ -88,14 +90,23 @@ utc_time = datetime(*lal.GPSToUTC(opts.trigger_time)[0:6]).strftime("%Y-%m-%d %H
...
 
-data[0].append(f'{math.degrees(opts.dec):.3f}')
-headers.append('Dec (deg)')
+    data[0].append(f'{math.degrees(opts.dec):.3f}')
+    headers.append('Dec (deg)')
 
-data[0].append(f'{math.degrees(opts.sky_error):.3f}')
-headers.append('Sky Error (deg)')
+    data[0].append(f'{math.degrees(opts.sky_error):.3f}')
+    headers.append('Sky Error (deg)')
```
titodalcanton: Try to minimize the amount of duplicated code inside the two branches of the "if" statement. In the two branches, assign two variables `ra` and `dec`. Then, after the "if" statement, use the `ra` and `dec` variables to fill in the `data` and `headers`.

PR #5135, opened 2025-06-11
add LFS into resolve_url https helper

GarethCabournDavies:
We noticed that files under git-lfs did not work with the reworked resolve_url

Adding `&lfs=true` seems to make the gitlab download helper work for lfs files _as well as_ non-lfs files still working


spxiwh (approved):
Looks good, but also needs to go onto the main branch

PR #5139, opened 2025-06-12
Fixing `pycbc_pygrb_grb_info_table` typos

Thomas-JACQUOT:

This is a bug fix.

This change affects PyGRB.

This change changes GRB info table script

## Motivation

This change will avoid this error : 
```shell
/home/thomas.jacquot/.conda/envs/test_workflow_sky_grid/lib/python3.12/site-packages/pykerr/qnm.py:2: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as
          import pkg_resources
        /home/thomas.jacquot/.conda/envs/test_workflow_sky_grid/bin/pycbc_pygrb_grb_info_table:104: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing th
          data[0].append(f'{math.degrees(ra):.3f}')
        /home/thomas.jacquot/.conda/envs/test_workflow_sky_grid/bin/pycbc_pygrb_grb_info_table:107: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing th
          data[0].append(f'{math.degrees(dec):.3f}')
          File "/home/thomas.jacquot/.conda/envs/test_workflow_sky_grid/bin/pycbc_pygrb_grb_info_table", line 119, in <module>
            factor = get_antenna_dist_factor(
                     ^^^^^^^^^^^^^^^^^^^^^^^^
          File "/home/thomas.jacquot/.conda/envs/test_workflow_sky_grid/lib/python3.12/site-packages/pycbc/results/pygrb_postprocessing_utils.py", line 476, in get_antenna_dist_factor
            fp, fc = antenna.antenna_pattern(ra, dec, 0, geocent_time)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
          File "/home/thomas.jacquot/.conda/envs/test_workflow_sky_grid/lib/python3.12/site-packages/pycbc/detector/ground.py", line 331, in antenna_pattern
            gha = self.gmst_estimate(t_gps) - right_ascension
                  ~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~
        TypeError: unsupported operand type(s) for -: 'float' and 'NoneType'
``` 

and this one too 
```shell
/home/thomas.jacquot/.conda/envs/test_workflow_sky_grid/lib/python3.12/site-packages/pykerr/qnm.py:2: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as
          import pkg_resources
          File "/home/thomas.jacquot/.conda/envs/test_workflow_sky_grid/bin/pycbc_pygrb_grb_info_table", line 93, in <module>
            if opts.input_dist is not None:
               ^^^^^^^^^^^^^^^
        AttributeError: 'Namespace' object has no attribute 'input_dist'
``` 

## Contents
I corrected the name of the input distribution argument in order to be consistent with the other scripts, and changed the shape of ra and dec in the case of a given input distribution. 

## Testing performed
I ran a PyGRB workflow that showed the fix was working. 

### DIFF
--- a/bin/pygrb/pycbc_pygrb_grb_info_table
+++ b/bin/pygrb/pycbc_pygrb_grb_info_table
@@ -97,7 +97,8 @@ if opts.input_dist is not None:
         ) 
     input_dist = eval('pycbc.distributions.' + opts.input_dist)
     ra, dec = input_dist.get_max_prob_point()
-    
+    sky_error = None
+
 else:
     ra = opts.ra
     dec = opts.dec
@@ -109,7 +110,7 @@ headers.append('R.A. (deg)')
 data[0].append(f'{math.degrees(dec):.3f}')
 headers.append('Dec (deg)')
 
-if sky_error != 0:
+if sky_error is not None:
     data[0].append(f'{math.degrees(sky_error):.3f}')
     headers.append('Sky Error (deg)')
 


bin/pygrb/pycbc_pygrb_grb_info_table
```diff
@@ -92,8 +92,8 @@ headers.append('UTC Time')
 
 if opts.input_dist is not None:
     input_dist = eval('pycbc.distributions.' + opts.input_dist)
-    ra = input_dist.get_max_prob_point()[0]
-    dec = input_dist.get_max_prob_point()[1]
+    ra = input_dist.get_max_prob_point()[0][0]
+    dec = input_dist.get_max_prob_point()[1][0]
```
titodalcanton: I do not like this, this should not be necessary in principle. Have you tested that this works both with HealpixSky and FisherSky input distributions? I bet it only works with HealpixSky. In that case, you should fix HealpixSky so that it returns a tuple of two floats and not a tuple of iterables.

Thomas-JACQUOT: Good catch, I modified it

titodalcanton:
While you are at it, I suggest one more edit: replace
```
ra = input_dist.get_max_prob_point()[0]
dec = input_dist.get_max_prob_point()[1]
```
with
```
ra, dec = input_dist.get_max_prob_point()
```
Calling `get_max_prob_point()` twice is wasteful, and with this change you only call it once.

PR #5143, opened 2025-06-25
Substituting scatterplots with hexbins where possible

MarcoCusinato:
In cases where many points are displayed as crosses the plots become crowded. This change those scatter plots into hexbins.

This change affects: PyGRB plotting

This change changes: plotting

Please see [this](https://ldas-jobs.ligo.caltech.edu/~marco.cusinato/LVK/GRB231117A_3/2._signal_consistency/2.01_chi_squared_tests/)

### DIFF
--- a/pycbc/results/pygrb_plotting_utils.py
+++ b/pycbc/results/pygrb_plotting_utils.py
@@ -191,15 +191,48 @@ def pygrb_plotter(trigs, injs, xlabel, ylabel, opts,
                   colors=None, vert_spike=False, cmd=None):
     """Master function to plot PyGRB results"""
     from matplotlib import pyplot as plt
+    import matplotlib
 
     # Set up plot
     fig = plt.figure()
     cax = fig.gca()
+    # Axes: labels and limits
+    cax.set_xlabel(xlabel)
+    cax.set_ylabel(ylabel)
+    if opts.x_lims:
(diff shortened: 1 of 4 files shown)


pycbc/results/pygrb_plotting_utils.py
```diff
@@ -191,15 +191,48 @@ def pygrb_plotter(trigs, injs, xlabel, ylabel, opts,
...
+        scales = ['log', 'log'] if opts.use_logs else ['linear', 'linear']
+        ## Necessary trick to avoid artifacts
+        xmin, xmax = cax.get_xlim()
+        if not isinstance(trigs[0], numpy.ndarray):
+            trigs[0] = numpy.array(trigs[0])
+        if not isinstance(trigs[1], numpy.ndarray):
+            trigs[1] = numpy.array(trigs[1])
+        mask = (trigs[0] >= xmin) & (trigs[0] <= xmax)
+        norm = matplotlib.colors.LogNorm()
+        gsize = 300
```
titodalcanton: I think the bins are a bit too small here, isolated single triggers are hard to see. I suggest using 150 here. I also noticed the logic below to choose the bins based on the number of samples. I think if you use 150 it will be good in all cases. We can complicate the logic once we have looked at several cases later.

Issue #3660, opened 2021-03-08
PyGRB results webpage


pannarale:
We have a first take on the PyGRB results webpage in the standard PyCBC format [here](https://ligo.gravity.cf.ac.uk/~francesco.pannarale/LVC/new_pygrb_webpage_1/).  Right now it has most of what was on the old webpages, but there is a lot more that can and should be done.  So use this issue to bring up bugs, desiderata, etc.

pannarale:
Known requests 
- [ ] Include GPS times of injections in tables
- [ ] Follow up of loudest missed injections
- [ ] Follow up of loudest offsource events
- [ ] Include in tables information regarding the spins of templates ringing with triggers/injections (this is postponed to after the switch to hdf5)
- [ ] Sky maps with antenna responses

titodalcanton:
@pannarale could you also consider the following requests:

* [x] I have a hard time reading found/missed injection plots. I propose using higher-resolution plots (try `dpi=150` or `dpi=200`, or double the figure size), smaller symbols, consistent use of the makers (x = missed, o = found) and no mix of blue and black symbols.
* [x] Found/missed plots with some kind of optimal SNR on the y axis. Not sure which one gives the sharpest separation between found and missed in a coherent search though, so this may require comparing a few different choices (e.g. optimal network SNR or decisive optimal SNR). [PR #5335 ]
* [x] Try a log10 FAP color scale, I suspect it would be more useful: 0.8 and 0.2 basically indicate the same conclusion, while 0.05 and 0.0005 imply very different results.
* [x] Add a tiny and very brief legend on the bottom left or bottom right of the figure, indicating what the various symbols mean.

pannarale:
@jakeb245, per the discussion we had today about `extract_basic_trig_properties`, please add the argument `opts` in `pycbc_pygrb_efficiency` when the `extract_basic_trig_properties` function call happens.

pannarale:
Moving more webpage/post-processing tasks here from issue [#4419](https://github.com/gwastro/pycbc/issues/4419).

Comparing the [latest results webpage](https://ldas-jobs.ligo.caltech.edu/~francesco.pannarale/LVC/pygrb_dev_webpage2/) to the [most complete one](https://ldas-jobs.ligo.caltech.edu/~francesco.pannarale/LVC/py3_pygrb_webpage_3/) generated when using old xml results files, the items missing are
- [ ] tables of missed and quiet injections (A31 and A32)
[...]

PR #5145, opened 2025-06-26
Adding injections and bank corner plots to pyGRB workflows

MarcoCusinato:
This change affects: PyGRB

Summary: 

- A corner plots for the template bank has been added to the results summary page
- A corner plot for each set of injections has been added to the Injections summary page 

Please see [this](https://ldas-jobs.ligo.caltech.edu/~marco.cusinato/LVK/GRB231117A_4/) for the current implementation.

This also addresses one of the tasks in #3660

### DIFF
--- a/bin/pygrb/pycbc_pygrb_results_workflow
+++ b/bin/pygrb/pycbc_pygrb_results_workflow
@@ -144,6 +144,12 @@ parser.add_argument(
     help="The location of the full template bank file",
 )
 parser.add_argument(
+    "--full-inj-files",
+    action="store",
+    nargs="+",
+    help="Location(s) of the generated injections file(s)",
+)
+parser.add_argument(
     "--segment-dir",
     action="store",
     help="The location of the segment files",
@@ -219,6 +225,7 @@ if not set(inj_sets).issubset(eff_secs):
     logging.error(err_msg, inj_sets, eff_secs)
 inj_sets = [i.upper() for i in inj_sets]
 inj_files = labels_in_files_metadata(inj_sets, start_rundir, args.inj_files)
+full_inj_files = labels_in_files_metadata(inj_sets, start_rundir, args.full_inj_files)
 
 # File instance of the veto file
 veto_file = args.veto_file
@@ -273,9 +280,15 @@ plot_node, skygrid_plot = _workflow.make_pygrb_plot(
     wflow, 'pygrb_plot_skygrid', out_dir, trig_file=offsource_file
 )
 plotting_nodes.append(plot_node)
+# Plot the template bank
+plot_node, bank_plot = _workflow.make_pygrb_plot(
+    wflow, 'pycbc_plot_bank_corner', out_dir, bank_file=bank_file
+)
+
 # Retrieve the segments plot (produced in the preprocessing stage)
 seg_plot = display_seg_plot(out_dir, args.segment_dir)
-summary_layout = [(grb_info_table[0],), (seg_plot, skygrid_plot[0])]
+summary_layout = [(grb_info_table[0],), (seg_plot, skygrid_plot[0]),
+                  (bank_plot[0],)]
 layout.two_column_layout(out_dir, summary_layout)
 
 #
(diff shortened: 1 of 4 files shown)


bin/pygrb/pycbc_pygrb_results_workflow
```diff
@@ -468,6 +481,25 @@ layout.two_column_layout(coinc_out_dir, coinc_files)
...
 out_dir = rdir['injections']
 _workflow.makedir(out_dir)
+## Create corner plot
+files = _workflow.FileList([])
+for inj_set in inj_sets:
+    # Retrieve the injections full files
+    inj_file = full_inj_files.find_output_with_tag(
+        inj_set, fail_if_not_single_file=True
+    )
+    # Genereta a corner bank plot for each injection set
```
titodalcanton:
```suggestion
    # Generate a corner bank plot for each injection set
```

bin/pygrb/pycbc_pygrb_results_workflow
```diff
@@ -468,6 +481,25 @@ layout.two_column_layout(coinc_out_dir, coinc_files)
 #
 out_dir = rdir['injections']
 _workflow.makedir(out_dir)
+## Create corner plot
+files = _workflow.FileList([])
```
titodalcanton: I recommend a more specific name for this variable, as this variable name is already used above. I think `inj_corner_plot_files` is good.

bin/pygrb/pycbc_pygrb_results_workflow
```diff
@@ -273,9 +280,15 @@ plot_node, skygrid_plot = _workflow.make_pygrb_plot(
     wflow, 'pygrb_plot_skygrid', out_dir, trig_file=offsource_file
 )
 plotting_nodes.append(plot_node)
+# Plot the template bank
+plot_node, bank_plot = _workflow.make_pygrb_plot(
+    wflow, 'pycbc_plot_bank_corner', out_dir, bank_file=bank_file
+)
+
```
titodalcanton: I think you also need to append `plot_node` to `plotting_nodes`, as done in the lines above for the sky grid plot.

pycbc/workflow/grb_utils.py
```diff
@@ -566,6 +575,15 @@ def make_pygrb_plot(workflow, exec_name, out_dir,
...
                 node.add_opt('--'+log_flag)
+    elif exec_name == 'pycbc_plot_bank_corner':
+        node.add_opt('--no-suptitle')
+        if inj_file:
+            node.add_opt('--title', f'\"{tags[0]} injections\"')
+            params = workflow.cp.get_opt_tags(exec_name, 'parameters', ['injs'])
+        else:
+            node.add_opt('--title', f'\"Template bank\"')
+            params = workflow.cp.get_opt_tags(exec_name, 'parameters', ['bank'])
+        node.add_opt('--parameters', params)
```
titodalcanton: I do not think you need to handle the parameter list explicitly here. Any option given in the executable's section via the config file will be passed straight as a command line option by the workflow base code.

titodalcanton:
As a broader comment, I suggest not using `pycbc_plot_bank_corner` for the internal executable name. Use just `plot_corner` instead.

pycbc/workflow/grb_utils.py
MarcoCusinato: It is needed for the subsections (i.e. `pycbc_plot_bank_corner_NSBH`). And we need to pass the parameters separately because injections and bank plotting have different parameters.

bin/pygrb/pycbc_pygrb_results_workflow
MarcoCusinato: I think you are right; I wonder why this works, though.

titodalcanton: I guess in your test, the job may have completed early enough to hide the missing dependency. But we might not always be so lucky [1], so it best to do this properly.

[1] https://en.wikipedia.org/wiki/Race_condition

pycbc/workflow/grb_utils.py
```diff
@@ -566,6 +575,15 @@ def make_pygrb_plot(workflow, exec_name, out_dir,
             if workflow.cp.has_option_tags(exec_name, log_flag,
                                            tags=[subsection]):
                 node.add_opt('--'+log_flag)
+    elif exec_name == 'pycbc_plot_bank_corner':
+        node.add_opt('--no-suptitle')
+        if inj_file:
+            node.add_opt('--title', f'\"{tags[0]} injections\"')
+            params = workflow.cp.get_opt_tags(exec_name, 'parameters', ['injs'])
+        else:
+            node.add_opt('--title', f'\"Template bank\"')
```
titodalcanton:
```suggestion
            node.add_opt('--title', '\"Template bank\"')
```

pycbc/workflow/grb_utils.py
titodalcanton: As far as I am aware, the functionality you describe comes "for free" already. In the config file you will have two sections
```
[plot_corner-bank]
parameters = mass1 mass2 spin1z spin2z
title = "Template bank"
no-suptitle =

[plot_corner-bnsinj]
parameters = mass1 mass2 distance inclination ra dec
title = "Injections"
no-suptitle =
```
When you instantiate `PlotExecutable` and call `create_node()` above, this should *automatically* pick up the right section and all the arguments therein. I do not think you need to explicitly call `add_opt()`.

bin/pygrb/pycbc_pygrb_results_workflow
MarcoCusinato: Done.

titodalcanton (approved):
Thanks @MarcoCusinato this looks ready enough. Have you had a chance to test it with the latest changes?

PR #5146, opened 2025-06-26
Fixes bug with filter.matchedfilter.optimized_match 

### LINKED ISSUE #5144: Extreme discontinuities in optimized_match that are not present in match
When calculating the mismatch between waves, discontinuities in the match value can occur when using pycbc.filter.matchedfilter.optimized_match that are not present when using match. Specifically, the research I'm doing is investigating lensed vs unlensed waves and certain combinations of chirp mass and time delay cause a spike in the mismatch (1-match). An example is shown in the following graph:

Image

This is not an isolated issue, it occurs in multiple regions of phase-space, as seen by the sharp red-orange regions in the following contour plot:

Image

Or  zoomed-in:

Image

[Here's a link](http://nextcloud.benmckallip.com/s/DcwYTdNx5SrKwjE) to a pkl file containing the waveform strain for the range of waves from the first plot for testing purposes. Usage is as follows:
```
with open('data.pkl', 'rb') as file:
    td_list, lensed_waveform_list, unlensed_waveform_list = pickle.load(file)
```

Each item in `lensed_waveform_list` and `unlensed_waveform_list` is a FrequencySeries object of the strain of the wave. The PSD is calculated using the following function:

```
def Sn(f_arr, f_min=20, delta_f=0.25, frequencySeries=True):
    """
    Calculates the power spectral density of the aLIGO noise curve based on arXiv:0903.0338.

    Parameters
    ----------
    f_arr : np.ndarray
        The frequency array.
    f_min : float, optional
        The minimum frequency. Defaults to 20 Hz.
    delta_f : float, optional
        The frequency step size. Defaults to 0.25 Hz.
    frequencySeries : bool, optional
        If True, returns a FrequencySeries object. Defaults to True.

    Returns
    -------
    np.ndarray or FrequencySeries
        The power spectral density of the aLIGO noise curve.
    """

    Sn_val = np.zeros_like(f_arr)
    for i in range(len(f_arr)):
        if f_arr[i] < f_min:
            Sn_val[i] = np.inf
        else:
            S0 = 1e-49
            f0 = 215
            Sn_temp = (
                np.power(f_arr[i] / f0, -4.14)
                - 5 * np.power(f_arr[i] / f0, -2)
                + 111
                * (
                    (1 - np.power(f_arr[i] / f0, 2) + 0.5 * np.power(f_arr[i] / f0, 4))
                    / (1 + 0.5 * np.power(f_arr[i] / f0, 2))
                )
            )
            Sn_val[i] = Sn_temp * S0

    if frequencySeries:
        return FrequencySeries(Sn_val, delta_f=delta_f)
    return Sn_val
```

bmck039:
This PR fixes a bug in filter.matchedfilter.optimized_match that causes the return of incorrect match values. The code originally did not respect the minimization bounds, causing local minima to be incorrectly returned in some rare instances.  

This is a: bug fix

This change will: Fix a bug described in [#5144](https://github.com/gwastro/pycbc/issues/5144)

## Motivation

Incorrect match values were encountered when calculating the match over different wave parameters. In these cases, the match discontinuously jumped as described in [#5144](https://github.com/gwastro/pycbc/issues/5144).

## Contents

 * matchedfilter.py: `optimized_match()`: Changes scipy minimize_scalar method to one that respects minimization bounds
 * test_matchedfilter.py: `test_optimized_match_valid()`: Adds to existing automated tests by using a known-tricky waveform to ensure `optimized_match()` calculates the correct value

Fixes [#5144](https://github.com/gwastro/pycbc/issues/5144)

## Testing performed

Extremely simple change, no testing required.

## Additional notes


bmck039:
Thanks for the reminder, it should be updated now to reflect that.

jacopok:
I don't exactly understand why the test is failing and don't really have the bandwidth to look into it at the moment.

I think the way to go here could be to relax the requirement in the test suite to 3 decimal places and perhaps allow the user to pass `**kwargs` to the `optimized_match` function, which then could be passed to the optimizer call, as @ahnitz originally suggested.

titodalcanton:
If someone wants to proceed here, try rebasing the PR to master, this may help with tests that failed for unrelated reasons.

From a very quick glance, I also recommend using a CSV file, or even better a HDF5 file, to store the data in `test/testing_data.py`.

bmck039:
It seems the only remaining failed test is related to a repeated 404 on the virtual environment image mirror. Not sure what's causing that, but I think it's safe to say that this PR doesn't affect anything related to that.

jacopok:
This looks good to me!

ahnitz (changes requested):
The failed test is unrelated to this PR. Otherwise I think this is largely ok. Please consider this approved but with one requested change. 

I'd ask that we not commit a binary file to the repository. Can we get this from elsewhere? If needed, I'm fine with having a separate repo for this, but if we commit it to the main code repo we are stuck with a binary file in the history for a while (these tend to accumulate as differences require full new versions in the future).

PR #5147, opened 2025-06-27
Marginalized Total Mass Model. 

kkacanja:
This pull request introduces a new inference model that performs marginalization over total mass by rescaling a reference waveform across a specified total mass grid. This model is designed to apply brute-force marginalization on top of an existing base model using waveform rescaling, making it simple and modular to include in existing analyses.

### DIFF
--- a/pycbc/inference/models/brute_marg.py
+++ b/pycbc/inference/models/brute_marg.py
@@ -24,13 +24,9 @@ from pycbc.pool import BroadcastPool as Pool
 from scipy.special import logsumexp
 
 from .gaussian_noise import BaseGaussianNoise
-from .tools import draw_sample, marginalize_likelihood, DistMarg
+from .tools import draw_sample, DistMarg
 
-
-from pycbc.waveform import td_taper
-from pycbc.waveform import get_waveform_filter_length_in_time
 from pycbc.conversions import mass1_from_mtotal_q, mass2_from_mtotal_q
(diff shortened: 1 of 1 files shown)


WuShichao:
Does it work with different mass ratio `q`?

kkacanja:
> Does it work with different mass ratio `q`?

Yea, you can set a prior on q. You need mass ratio in order to get mass1 and mass2 for the waveform generation

WuShichao:
> > Does it work with different mass ratio `q`?
> 
> Yea, you can set a prior on q. You need mass ratio in order to get mass1 and mass2 for the waveform generation

So, for a (2,2)-mode BBH, in principle, you can only sample spin parameters and marginalize over all other parameters?

kkacanja:
> > > Does it work with different mass ratio `q`?
> > 
> > 
> > Yea, you can set a prior on q. You need mass ratio in order to get mass1 and mass2 for the waveform generation
> 
> So, for a (2,2)-mode BBH, in principle, you can only sample spin parameters + mass ratio, and marginalize over all other parameters?

Yup. One way I am using this is to marginalize over mtotal, tc, sky location, polarization and phase (so you specify the base model as marginalized time) and only sampling over eccentricity q and spins. This model may be beneficial for very slow waveform models

ahnitz:
@kkacanja Before merging this, can you add an example and unittest for the model.

PR #5157, opened 2025-07-17
Avoid unnecessary SNR (re)normalization in `multi_inspiral`

titodalcanton:
`pycbc_multi_inspiral` is unnecessarily un-normalizing the SNRs in preparation for the chi^2 calculations.

This is an efficiency update.

This change affects PyGRB.

This change should not affect any result, just code speed.

## Motivation

In #5148, @spxiwh realized that `pycbc_multi_inspiral` is spending a relatively large amount of time doing array divisions. Indeed, in the inner loop over sky points, the SNRs are un-normalized before being passed to the chi^2 calculations, *even when those calculations are disabled*. This is not a huge problem compared to other inefficiencies, but of the various optimizations discussed in #5148, this is the most straightforward one by far.

## Contents

On a close inspection of the chi^2 codes, I realized that the corresponding `value()` methods have really no reason to take the un-normalized SNRs and the normalization factor separately, since all they do is to simply multiply them. So here I just pass the normalized SNRs and a dummy normalization of 1.

I also took the opportunity to fix a likely typo in the initialization of `ifo_out_vals` which has been on my radar for a while.

#5148

## Testing performed

None yet.

## Additional notes

Out of curiosity, I had a fresh peek at how `pycbc_inspiral` calls the chi^2 calculations. It seems to me that, if one were to enable all three types of chi^2, `pycbc_inspiral` would be normalizing the same SNRs three times. Perhaps it would be more efficient to refactor the code such that the normalization is done once, before calling the chi^2 codes, and have the chi^2 codes simply take normalized SNRs.

### DIFF
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -1030,8 +1030,8 @@ with ctx:
                                 template,
                                 stilde[ifo].psd,
                                 stilde[ifo],
-                                coherent_ifo_trigs[ifo] / norm_dict[ifo],
-                                norm_dict[ifo],
+                                coherent_ifo_trigs[ifo],
+                                1,  # coherent_ifo_trigs already normalized
                                 coinc_idx_det_frame[ifo]
                                 + stilde[ifo].analyze.start,
                             )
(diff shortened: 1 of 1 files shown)


spxiwh (approved):
Change looks fine, if you've tested it and are happy, free to merge.

titodalcanton:
I tested through a comparison of the resulting trigger datasets between various runs of the GW170817 example (after enabling auto and bank chi^2). It is remarkably difficult to get identical results even between two runs of the same code, on the same machine, ran only a few minutes apart. But I consistently find differences in auto and bank chi^2 of order 1e-3 max between the original code and this PR, and I find the same across different runs of the original code, so I am happy.

PR #5159, opened 2025-07-29
fix a bug in the mutli loglikelihood method of the gated gaussian noise class

yi-fan-wang:
This PR fixes a one-line bug in the `multi_loglikelihood` method of `GatedGaussianNoise` class. I think this is critical as otherwise two waveforms with different lengths can't be summed in the next line. 

In particular, this PR wouldn't affect the previous ringdown black hole spectroscopy runs, as ringdown analysis only uses `hierarchical` model, not `multi_signal`. 


cdcapano (approved):
Looks good. Thanks for catching!

PR #5160, opened 2025-07-29
Reading hdf files with search

aleynaakyuz:
Currently, search only accepts GWF files as input. This update adds support for reading HDF files in GWOSC format if `hdf-store` is used instead of `frame-files.`

### DIFF
--- a/pycbc/frame/store.py
+++ b/pycbc/frame/store.py
@@ -46,30 +46,21 @@ def read_store(fname, channel, start_time, end_time):
 
     """
     fhandle = HFile(fname, 'r')
-    if channel not in fhandle:
-        raise ValueError('Could not find channel name {}'.format(channel))
 
-    # Determine which segment data lies in (can only read contiguous data now)
-    starts = fhandle[channel]['segments']['start'][:]
-    ends = fhandle[channel]['segments']['end'][:]
+    starts = fhandle['meta']['GPSstart'][()]
+    duration = fhandle['meta']['Duration'][()]
+    ends = starts + duration
 
-    diff = start_time - starts
-    loc = numpy.where(diff >= 0)[0]
-    sidx = loc[diff[loc].argmin()]
+    if starts > start_time:
+        raise ValueError("Cannot read data segment before {}".format(starts))
 
-    stime = starts[sidx]
-    etime = ends[sidx]
+    if ends < end_time:
+        raise ValueError("Cannot read data segment past {}".format(ends))
 
-    if stime > start_time:
-        raise ValueError("Cannot read data segment before {}".format(stime))
+    data = fhandle['strain']['Strain'][:]
+    sample_rate = len(data) / (ends - starts)
 
-    if etime < end_time:
-        raise ValueError("Cannot read data segment past {}".format(etime))
-
-    data = fhandle[channel][str(sidx)]
-    sample_rate = len(data) / (etime - stime)
-
-    start = int((start_time - stime) * sample_rate)
-    end = int((end_time - stime) * sample_rate)
-    return TimeSeries(data[start:end], delta_t=1.0/sample_rate,
+    start_inx = int((start_time - starts) * sample_rate)
+    end_inx = int((end_time - starts) * sample_rate)
+    return TimeSeries(data[start_inx:end_inx], delta_t=1.0/sample_rate,
                       epoch=start_time)


ahnitz:
@aleynaakyuz It would probably make sense to have a specific function that reads the hdf format and returns the appropriate time series (and also a mirror for writing) so that it can be called from the appropriate places. The 'store' format is a bit different of an idea an has a different format than the GWOSC format.

ahnitz:
@aleynaakyuz However, it's good to go forward with testing, and we can talk about perhaps the most sensible modularization when I get back.

aleynaakyuz:
@ahnitz can you review the functions

aleynaakyuz:
To test these functions, I used the example search given [here](https://github.com/gwastro/pycbc/tree/master/examples/search) with minor changes to the template [bank](https://sugwg-web.syr.edu/~aakyuz/search_tests/default/search/bank.sh).

All tests bellow performed by using the same data but in different formats

1. running the test without making any changes. 

- [configuration](https://sugwg-web.syr.edu/~aakyuz/search_tests/default/search/)

- [results](https://sugwg-web.syr.edu/~aakyuz/search_tests/default/search/html/)

2.  changing the data to GWOSC hdf format

- [configuration](https://sugwg-web.syr.edu/~aakyuz/search_tests/GWOSC/)

- [results](https://sugwg-web.syr.edu/~aakyuz/search_tests/GWOSC/html/) 

3.  changing the data to PyCBC time series saved hdf format 

- [configuration](https://sugwg-web.syr.edu/~aakyuz/search_tests/type2_search/)

- [results](https://sugwg-web.syr.edu/~aakyuz/search_tests/type2_search/html/)

4. changing the data to GwPy time series write hdf5 format

- [configuration](https://sugwg-web.syr.edu/~aakyuz/search_tests/type3_search/)

- [results](https://sugwg-web.syr.edu/~aakyuz/search_tests/type3_search/html/)

5. changing the data to cache gwf file

- [configuration](https://sugwg-web.syr.edu/~aakyuz/search_tests/gwf_lcf/)
- [results](https://sugwg-web.syr.edu/~aakyuz/search_tests/gwf_lcf/html/)

6. changing the data to cache hdf file

- [configuration](https://sugwg-web.syr.edu/~aakyuz/search_tests/hdf_lcf/)
- [results](https://sugwg-web.syr.edu/~aakyuz/search_tests/hdf_lcf/html/)

All these tests give the same results.

pycbc/strain/strain.py
```diff
@@ -225,11 +225,23 @@ def from_cli(opt, dyn_range_fac=1, precision='single',
                     start_time=opt.gps_start_time-opt.pad_data,
                     end_time=opt.gps_end_time+opt.pad_data,
                     sieve=sieve)
-        elif opt.frame_files or opt.frame_cache:
-            strain = pycbc.frame.read_frame(
-                    frame_source, opt.channel_name,
-                    start_time=opt.gps_start_time-opt.pad_data,
-                    end_time=opt.gps_end_time+opt.pad_data,
+        elif opt.frame_files:
```
ahnitz: @aleynaakyuz Would it make sense to have the logic in some more general read_gw_data or similar function so that it can be used more easily in other codes as well?

aleynaakyuz: @ahnitz Do you mean writing a function that calls `read_frame`, `read_hdf5_frame` or `read_frame_cache`?
Which codes do you mean by other codes? Right now both search and inference can support hdf files without making any changes to the configuration.

pycbc/frame/frame.py
```diff
@@ -33,6 +33,7 @@
 
 import pycbc
 from pycbc.types import TimeSeries, zeros
+from pycbc.io.hdf import HFile
```
ahnitz: We probably don't need to use this unless you use some of the extra functionality

pycbc/frame/frame.py
```diff
@@ -263,7 +275,253 @@ def read_frame(location, channels, start_time=None,
...
+    elif len(locations) > 1:
+        channels = [channels]*len(locations)
+    else:
+        channels = [channels]
+
+    for i, file_path in enumerate(locations):
+        _, file_name = os.path.split(file_path)
+        _, file_extension = os.path.splitext(file_name)
+
+        if file_extension in [".lcf", ".cache"]:
```
ahnitz: The hdf / gwf functions should understand cache files and be able to support mixes between cache files / individual files.

ahnitz:
@aleynaakyuz Looks like you accidentally closed this. You should have the option to reopoen once you add a new commit.

PR #5165, opened 2025-08-13
Add functionality to sample coalescence time in detector frame

acorreia61201:
This PR adds the ability to sample the coalescence time parameter `tc` in the detector frame. Currently, `tc` is sampled in the geocentric frame and converted to detector frames to calculate likelihoods. The changes here preserve this behavior by default, but add the ability to sample in one of the detectors specified in the inference configuration.

This is a new feature. This change mainly affects inference when using `FDomainDetFrameGenerator`. Any code that uses the `generate` method to generate waveforms in the detector frame can also make use of this. Current functionality (generating using `tc` in geocentric frame) is preserved by default.

## Motivation
This change is being made as a potential improvement to convergence when conducting PE, especially when implementing sky location and time marginalization in PE (see PR #5117). Sampling over the full sky can result in a ~30 ms variation in arrival time. If we additionally sample over `tc` directly, this can lead to a large posterior range. Additionally, there is a degeneracy between shifting sky location and `tc` that can lead to complicated posterior distributions, which can lead to convergence issues in the results. Sampling in one of the detector frames should alleviate convergence issues by removing the degeneracy between `tc` and sky location and reducing the sampling uncertainty in at least one of the detectors.

## Contents
All changes are made to the `generate` method of `FDomainDetFrameGenerator`. This functionality can be activated by adding an option `tc_ref_frame` to the configuration, e.g.:
```
[static-params]
tc_ref_frame = L1
```
During waveform generation in PE, time shifts will be applied in each detector relative to the specified frame. If `tc_ref_frame = geocentric` is specified (or if the option is omitted from the config), the same `time_delay_from_earth_center` delay is applied to all detectors, mimicking current behavior. If the specified frame matches one of the detectors specified in the `[data]` section, no time shift will be applied in that detector, and `time_delay_from_detector` is used instead to compute the time shift from the sampling detector to all other detectors.

### DIFF
--- a/pycbc/detector/ground.py
+++ b/pycbc/detector/ground.py
@@ -475,8 +475,8 @@ class Detector(object):
                                              declination,
                                              t_gps)
     
-    def convert_tc(self, ref_tc, ra, dec, ref_frame='geocentric'):
-        """Convert tc from reference frame to another detector.
+    def arrival_time(self, ref_tc, ra, dec, ref_frame='geocentric'):
+        """Compute the arrival time in this detector.
         
         Parameters
         ----------
@@ -487,14 +487,13 @@ class Detector(object):
         dec : float
             Declination.
         ref_frame : str (optional)
-            The detector to convert from, in which tc is sampled. Default
+            The detector to convert from, in which ref_tc is sampled. Default
             'geocentric'.
             
         Returns
         -------
         float : 
-            The coalescence time converted to the detector specified by
-            current_frame.
+            The coalescence time converted to the current detector frame.
         """
         if ref_frame == 'geocentric':
             # from geocenter
--- a/pycbc/detector/ground.py
+++ b/pycbc/detector/ground.py
@@ -474,6 +474,45 @@ class Detector(object):
                                              right_ascension,
                                              declination,
                                              t_gps)
+    
+    def convert_tc(self, ref_tc, ra, dec, ref_frame='geocentric'):
+        """Convert tc from reference frame to another detector.
+        
+        Parameters
+        ----------
+        ref_tc : {float, lal.LIGOTimeGPS}
+            The coalescence time to convert, defined in ref_frame
+        ra : float
+            Right ascension.
+        dec : float
+            Declination.
+        ref_frame : str (optional)
+            The detector to convert from, in which tc is sampled. Default
+            'geocentric'.
+            
+        Returns
+        -------
+        float : 
+            The coalescence time converted to the detector specified by
+            current_frame.
+        """
+        if ref_frame == 'geocentric':
+            # from geocenter
+            tc = ref_tc + \
+                self.time_delay_from_earth_center(ra, dec, ref_tc)
+        elif ref_frame == self.name:
+            # no time shift; sampling in current det
+            tc = ref_tc
+        elif ref_frame in get_available_detectors():
+            # from sampling det
+            refdet = Detector(ref_frame)
+            tc = ref_tc + \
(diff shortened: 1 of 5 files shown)


pycbc/waveform/generator.py
```diff
@@ -671,16 +671,30 @@ def generate(self, **kwargs):
...
-                            self.current_params['polarization'],
-                            self.current_params['tc'])
+                refframe = self.current_params.get('tc_ref_frame', 'geocentric')
+                ra = self.current_params['ra']
+                dec = self.current_params['dec']
+                ref_tc = self.current_params['tc']
+                pol = self.current_params['polarization']
+                # apply response function
+                fp, fc = det.antenna_pattern(ra, dec, pol, ref_tc)
                 thish = fp*hp + fc*hc
```
cdcapano, line 684: The tc at the detector should be used for the antenna pattern rather than the `ref_tc` (this is actually a minor bug in current master). To do that, move lines 680 and 681 to line 697; i.e., just after the time shift conditional block, but before `apply_fd_time_shift` is run.

cdcapano (changes requested):
Thanks for doing this. See my comment about switching around the antenna pattern calculation.

Also, please add the same functionality to `FDomainDetFrameTwoPolGenerator` and `FDomainDetFrameModesGenerator` so we have consistent features. And add some documentation about `tc_ref_frame` to the doc string under `location_args` in each class. (Don't add it to the `location_args` set -- those are the required arguments; just add something about it in the doc string.)

ahnitz:
@cdcapano @acorreia61201 This is probably a reasonable way to add this. I think we had an example floating around at some point of sampling in the detector frame just using waveform transforms, but I imagine this is probably faster. 

I think the only question is the format for specifying the reference frame. I think what you've chosen is probably fine and should be  also implementable in models not based on the generator framework. The alternative I can think of is to put this information in the  prior configuration instead (maybe as some kind of extra metadata field?). 

e.g. 

```
[prior-tc]
frame = L1
```

That way at least the parameters get grouped with specific information about them. I'm not sure if this works in all cases though. What do you think?

pycbc/waveform/generator.py
```diff
@@ -561,7 +561,45 @@ def generate(self, **kwargs):
     def select_rframe_generator(self, approximant):
         """Method to select waveform generator based on an approximant."""
         pass
-
+    
+    def convert_tc(self, ref_tc, ra, dec, current_frame,
```
ahnitz: Would this make more sense then as just a method of the detector class directly?

acorreia61201: You're right; I was debating either putting it there or in the waveform generator as it is now. It'll probably be useful to have this as a standalone detector method in hindsight

pycbc/detector/ground.py
```diff
@@ -474,6 +474,45 @@ def time_delay_from_detector(self, other_detector, right_ascension,
                                              right_ascension,
                                              declination,
                                              t_gps)
+    
+    def convert_tc(self, ref_tc, ra, dec, ref_frame='geocentric'):
```
ahnitz: @acorreia61201 Is is there a clearer name for this? It seems like it is always giving you the time *at* the current detector. Is that not right? 'convert_tc' seems a bit vague. 

This seems more like a function should be called something like 'arrival_time'  (e.g. det.arrival_time(...)) or something similarly description so the function is somewhat clear in the name.

cdcapano (approved):
Looks good!

PR #5166, opened 2025-08-13
Expanding help string of --filter-string

khunsang:
Expanding the help string of the option `--filter-string`

### DIFF
--- a/bin/plotting/pycbc_plot_singles_vs_params
+++ b/bin/plotting/pycbc_plot_singles_vs_params
@@ -51,7 +51,10 @@ parser.add_argument('--veto-file', type=str,
 parser.add_argument('--segment-name', default=None, type=str,
                     help='Optional, name of segment list to use for vetoes')
 parser.add_argument('--filter-string', default=None, type=str,
-                    help='Optional, boolean expression for filtering triggers')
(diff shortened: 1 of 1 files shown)


bin/plotting/pycbc_plot_singles_vs_params
```diff
@@ -51,7 +51,10 @@ parser.add_argument('--veto-file', type=str,
 parser.add_argument('--segment-name', default=None, type=str,
                     help='Optional, name of segment list to use for vetoes')
 parser.add_argument('--filter-string', default=None, type=str,
-                    help='Optional, boolean expression for filtering triggers')
+                    help='Optional, boolean expression for filtering triggers'
+                         'Example: "self.snr >=6"'
+                         'Parameter in "self.parameter" should be either from'
+                         'trigger params or bank params.')
```
GarethCabournDavies:
```suggestion
                    help='Optional, boolean expression for filtering triggers.'
                         'Example: "self.snr >=6"'
                         'Parameter in "self.parameter" should be either from'
                         'trigger or bank params.')
```

GarethCabournDavies (approved):
Comment for minor pedantry. Thanks for the change!

PR #5167, opened 2025-08-14
Add script for calculating Savage-Dickey ratio

acorreia61201:
This PR adds a script that estimates the Bayes factor for inference outputs using the Savage-Dickey ratio.

This is a new feature. No existing code is being changed; the script is completely separate from existing executables. 

## Motivation
The Savage-Dickey ratio can be used to approximate the Bayes factor as a method of model selection after conducting inference jobs. This is useful when, e.g., determining the detectability of quasinormal modes in CBC ringdowns. The SD ratio can be calculated by taking the ratio of the prior to the posterior at a given value signifying the null hypothesis. (In the case of QNM analyses, the null hypothesis is signified by a QNM amplitude of zero.)

## Contents
This PR adds a new `bin/inference` script, `pycbc_inference_savage_dickey`, that calculates the SD ratio for at least one input file. This is done by:

1. loading the input HDF files with standard `io` procedures to get the posterior samples and prior distributions,
2. generating KDEs for the posterior and prior, and 
3. taking the ratio of the (normalized) prior to the (normalized) posterior at a given null value for the given parameter.

Currently, the resulting Bayes factors for each file are printed to the command line.


bin/inference/pycbc_inference_savage_dickey
```diff
@@ -0,0 +1,210 @@
...
+      f"at null value {null}")
+print("======================================================================")
+
+priors = []
+posteriors = []
+for file in opts.input_file:
+    fp = loadfile(file, 'r')
+    cp = fp.read_config_file()
+    # read in the prior of given parameter
+    if opts.prior_name is not None:
```
kkacanja, line 134: You should also have it for the opposite as well. So if you specify --prior-min or max you require --prior-name. I made the mistake of specify the log distribution when choosing my bounds and I got the wrong answer out

bin/inference/pycbc_inference_savage_dickey
```diff
@@ -0,0 +1,210 @@
...
+            plt.xscale('log')
+        f = os.path.basename(file)
+        plt.title(f"Posterior and Prior of {param} from {f}")
+        plt.legend()
+        name, _ = os.path.splitext(f)
+        plt.savefig(f'plot-{name}.png')
+
+    ### FIXME: For now I'm just printing this; should we save this to file 
+    ### somewhere?
+    print(f"File: {file}")
```
kkacanja, line 248: I think printing it is fine unless you want to update the input file to have an attribute that stores the SDDR

bin/inference/pycbc_inference_savage_dickey
```diff
@@ -0,0 +1,210 @@
...
+                    "sampling issues at the edges of a histogram. This is not "
+                    "done by default.")
+parser.add_argument("--reflect-kde-on-right", action="store_true", default=False,
+                    help="Specify whether to reflect samples on the right bound "
+                    "when generating KDEs. This is done to prevent KDE "
+                    "sampling issues at the edges of a histogram. This is not "
+                    "done by default.")
+parser.add_argument("--plot", action='store_true', default=False,
+                    help="Specify whether to plot the samples and KDEs of the "
+                         "prior and posterior as a consistency check.")
```
cdcapano: Have this take a file name output.

bin/inference/pycbc_inference_savage_dickey
```diff
@@ -0,0 +1,210 @@
...
+
+opts = parser.parse_args()
+pycbc.init_logging(opts.verbose)
+
+# read in the posterior of given parameter
+### FIXME: ideally the parameters argument only reads in one parameter
+if len(opts.parameters) > 1:
+    raise ValueError("Multiple parameters specified. Only one parameter is "
+                     "accepted at a time.")
```
cdcapano: I suppose you could have it be able to calculate the ratio separately for each provided parameter, but that would be a pretty rare scenario. I think just raising an error works.

kkacanja (approved):
Looks good, feel free to merge

PR #5168, opened 2025-08-14
Add functionality to combine posteriors via evidence sampling in `extract_samples`

acorreia61201:
This PR adds functionality to combine posteriors by sampling according to their evidence values when running `pycbc_inference_extract_samples` with multiple input files.

This is a new feature. Current functionality is preserved by default; a new command line argument activates the new functionality. This change affects posterior extraction from raw inference outputs.

## Motivation
Currently, running `pycbc_inference_extract_samples` with multiple input files will combine all posterior samples into one output file, discarding any metadata related to evidence. If, for example, one wishes to split one of the priors across multiple inference jobs (to improve convergence, parallelize the process, etc.), the ideal way to combine posteriors would be to sample points from each run according to their individual evidence values rather than saving all points in one file. A sub-posterior with an evidence value twice as high as another should have twice as many points in the full posterior.

## Contents
A new command-line input, `--combine-via-sampling`, has been added to `pycbc_inference_extract_samples` to allow for posterior file combination where samples are selected according to each file's log evidence values. Each input posterior's samples are assigned a common weight based on that posterior's log evidence value. Points are then sampled for the full posterior by drawing points randomly based on these weights. The resulting posterior will contain as many points as the smallest input posterior, and will contain logZ and dlogZ values in the metadata computed from the `logsumexp` of each logZ and the quadrature sum of each dlogZ, respectively. The original functionality is preserved simply by not specifying `combine-via-sampling` when running `extract_samples`.

### DIFF
--- a/bin/inference/pycbc_inference_extract_samples
+++ b/bin/inference/pycbc_inference_extract_samples
@@ -228,19 +228,17 @@ for fp in fps:
                 current_val = format_attr(out.attrs[key])
             if not isthesame(current_val, val):
                 if key == 'remapped_params':
-                    # ...unless it's remapped_params; just combine unique items
-                    # brute-force if they don't match exactly
-                    cat_params = True
-                    val = numpy.concatenate((val, current_val))
-                    val_set = set(tuple(entry) for entry in val)
-                    val = [list(entry) for entry in val_set]
-                    out.attrs[key] = val
+                    # ...unless it's remapped_params; just save first file's
+                    # entries if they don't match between files
+                    warnings.warn("WARNING: remapped_params metadata does not "
+                                 "match between files; saving metadata from "
+                                 "first file")
                 else:
                     raise ValueError("cannot combine all files; file attr {} is "
                                      "not the same across all files ({} vs {})"
                                      .format(key, current_val, val))
 
-# store what parameters were renamed (if not already concatenated above)
+# store what parameters were renamed (if not already saved above)
 if not cat_params:
     out.attrs['remapped_params'] = list(labels.items())
 


bin/inference/pycbc_inference_extract_samples
```diff
@@ -83,6 +85,18 @@ parser.add_argument("--skip-groups", default=None, nargs="+",
...
+                         "By default, pycbc_inference_extract_samples will"
+                         "dump all samples from multiple inputs into one output"
+                         "file. If this option is specified, the samples will"
+                         "be randomly sampled with weighting based on the evidence"
+                         "in each input. For example, if the log evidence"
+                         "of one input is twice that of the other, the resulting"
+                         "posterior file with this option specified will have"
+                         "twice as many points from input 1 than input 2."
+                         "The output will have the same number of samples as the"
+                         "smallest input file.")
```
cdcapano: Add spaces before the ending double quote on each line so that the help message is formatted correctly.

bin/inference/pycbc_inference_extract_samples
```diff
@@ -158,6 +212,11 @@ for fp in fps:
 # store what parameters were renamed
 out.attrs['remapped_params'] = list(labels.items())
 
+# write combined evidence and dlog evidence if combining via sampling
+if opts.combine_via_sampling:
+    out.attrs['log_evidence'] = logsumexp(logz_list)
+    out.attrs['dlog_evidence'] = numpy.sqrt(sum([i**2 for i in dlogz_list]))
+
```
cdcapano, line 260: There's an issue with this: you only want to add evidence in this way if the priors are mutually exclusive. If someone was combining two files that have the same prior, the evidence would be averaged. (And if the priors are partially overlapping, they shouldn't be combining at all.) In principle, combine via sampling could be done in either case. Maybe add another option called `mutually-exclusive-priors`, and if that's set, add the evidences; otherwise average them. Make sure to put in the help message that if `mutually-exclusive-priors` is not set it's assumed that the priors are the same. Also, as stated above, only do all of this if  log evidence exists in the file.

bin/inference/pycbc_inference_extract_samples
```diff
@@ -101,14 +116,49 @@ if len(opts.input_file) == 1:
...
+                   for p in params}
+        logz_list = []
+        dlogz_list = []
+        len_list = []
+        raw_samps_list = []
+        weights_list = []
+        # get samples and parameters from each file
+        for file in opts.input_file:
+            fp = loadfile(file, 'r')
+            logz, dlogz = fp.log_evidence
```
cdcapano: Not all samplers write out evidence (`emcee`, for example). If not, trying to get the evidence will raise a KeyError (see [here](https://github.com/gwastro/pycbc/blob/e88c2725d0bc87b9ae977e29f3069c4abf310032/pycbc/inference/io/base_hdf.py#L437-L442) ). Put a try/except around this. In the case that there's no log evidence you'll need to have if statements below whenever you deal with evidence.

cdcapano (changes requested):
There's a couple potential issues with the way the evidence is being handled. See in-line comments.

bin/inference/pycbc_inference_extract_samples
```diff
@@ -83,6 +85,47 @@ parser.add_argument("--skip-groups", default=None, nargs="+",
...
+                    help="If specifying --combine-via-sampling, specify "
+                         "whether to treat priors as overlapping. By "
+                         "default, the provided input files are assumed to "
+                         "have identical priors, and their evidences will be "
+                         "averaged. If this option is specified, the priors "
+                         "are assumed to be different and overlapping, and "
+                         "the evidences will not be combined at all. This "
+                         "option is mutually exclusive with "
+                         "--mutually-exclusive-priors; an error will be "
+                         "thrown if both are specified.")
```
cdcapano: Remove this option; we can't handle partially overlapping priors. It's on the user to know that they shouldn't try to combine partially overlapping priors.

bin/inference/pycbc_inference_extract_samples
```diff
@@ -83,6 +85,47 @@ parser.add_argument("--skip-groups", default=None, nargs="+",
...
+                    help="If specifying --combine-via-sampling, specify "
+                         "whether to treat priors as mutually exclusive. By "
+                         "default, the provided input files are assumed to "
+                         "have identical priors, and their evidences will be "
+                         "averaged. If this option is specified, the priors "
+                         "are assumed to be different and non-overlapping, "
+                         "and the evidences will be added together. This "
+                         "option is mutually exclusive with "
+                         "--overlapping-priors; an error will be thrown if "
+                         "both are specified.")
```
cdcapano: Modify the help message here once you remove overlapping-priors. Also, I wouldn't say "different priors", just "non-overlapping".

bin/inference/pycbc_inference_extract_samples
```diff
@@ -156,8 +241,37 @@ for fp in fps:
...
+                         "--mutually-exclusive-priors; select one based "
+                         "on input priors.")
+    elif opts.mutually_exclusive_priors:
+        # add together the evidences; quadrature sum the residuals
+        out.attrs['log_evidence'] = logsumexp(logz_list)
+        out.attrs['dlog_evidence'] = numpy.sqrt(sum([i**2 for i in dlogz_list]))
+    elif opts.overlapping_priors:
+        # do not combine the evidences; insert numerical placeholders
+        out.attrs['log_evidence'] = numpy.float64(-numpy.inf)
+        out.attrs['dlog_evidence'] = numpy.float64(-numpy.inf)
```
cdcapano: As stated above, remove the overlapping_priors option.

acorreia61201:
@cdcapano @kkacanja I've addressed Collin's comments. Feel free to take a look when you get the chance.

I've made another change besides this. When combining multiple files, the code checks whether the metadata value `remapped_params` is the same between the posterior files. This can be an issue if, e.g. two files use two different waveform transforms to generate a parameter with the same name. I've made a change (indicated by a `FIXME`) that just saves the first file's `remapped_params` as a placeholder. This could cause some confusion if someone were to dig into the metadata of combined posterior files, but on the other hand I'm not sure if it would be okay to outright remove the `remapped_params` attribute from the combined posterior. I'm curious what your opinions are on this; if what's currently there is fine I can just remove the `FIXME` before merging.

kkacanja (approved):
The two tests I suggested passed, as well as writing a warning for a user when using combine-via-sampling. Once this passes the checks, feel free to merge

PR #5175, opened 2025-09-01
Fix a small error in the construction of the circular sky grid

titodalcanton:
**Note: this PR targets the 2.8 release branch.**

Fix a small bug in the placement of the circular sky grid used by PyGRB 2.8.

This is a bug fix.

This change affects PyGRB 2.8.

This change changes scientific output.

## Motivation

Answering some review questions for PyGRB 2.8 reminded me that the circular grid implemented in `pycbc_make_sky_grid` has a couple minor issues:
1. It leaves a "seam" at the end of each ring due to incorrect rounding.
2. In calculating the number of points per ring, it assumes that the error radius is small. This approximation is mostly ok considering the typical uncertainties of Fermi/GBM localizations, but the exact calculation is not much more complicated, so the approximation is pointless.

## Contents

I reimplemented the loop that places the rings using the correct calculation, and I added comment to explain why the calculation differs from what stated in Williamson et al 2014.

N/A

## Testing performed

I built a few different sky grids with the old and new code and plotted them together:

image

image

image

image

## Additional notes

### DIFF
--- a/bin/pycbc_make_sky_grid
+++ b/bin/pycbc_make_sky_grid
@@ -8,7 +8,7 @@ gravitational wave triggers and calculate the coherent SNRs and related
 statistics. 
 
 The grid is constructed following the method described in Section V of
-https://arxiv.org/abs/1410.6042.
+Williamson et al 2014, https://arxiv.org/abs/1410.6042.
 
 Please refer to help(pycbc.types.angle_as_radians) for the recommended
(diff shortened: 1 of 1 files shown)


titodalcanton:
To fix the tests, it is probably best to wait for #5176 to be merged first, and then rebase this one.

pannarale:
Looks good, thank you for catching this.

PR #5178, opened 2025-09-04
Fixes to PyGRB found-missed plot

titodalcanton:
Fix a couple issues in PyGRB's found-missed plot code.

This is a bug fix and result visualization improvement.

This change affects PyGRB.

## Motivation

PyGRB's found-missed plots have a few issues:
* GPS time axes are incorrectly labeled as "Time since {external trigger time}` where in fact it is just the GPS time.
* The `--colormap` option is ignored and a hardcoded choice is made instead.
* The colorbar for p-values is linear, but p-values between ~0.1 and 1 essentially tell the same story, so most of the colorbar range is useless. A log scale is more appropriate here.
* The choice of plotting order is implemented through significant code duplication.
* The variable choices `longitude` and `latitude` do not work as they have been replaced by `ra` and `dec` in PyGRB.

## Contents

* Fix the GPS time labeling.
* Fix the `longitude` and `latitude` variable choices.
* Do not ignore the `--colormap` option.
* Introduce a `--linear-colormap` option and default to a log colormap.
* Increase the resolution of plots in PNG format and tweak the size of some symbols.
* Add a legend.
* Add/fix units in axis labels.
* Implement support for `eff_dist`, though I implemented the inverse of the sum of the inverses, not literally the inverse of the sum (like the labels say, which I think makes no sense… Happy to discuss this).
* Minor code cleanup and simplification. I removed the site name conversion code, as we can assume H1, L1 etc is clear to every analyst.

This addresses some of my own requests in #3660.

## Testing performed

A few comparisons here: https://ldas-jobs.ligo.caltech.edu/~tito.canton/pycbc/test_pr5178/new/.

## Additional notes

None.

### DIFF
--- a/bin/pygrb/pycbc_pygrb_plot_injs_results
+++ b/bin/pygrb/pycbc_pygrb_plot_injs_results
@@ -189,7 +189,6 @@ def complete_inj_data(injs, keys, tag, ifos=[]):
 # =============================================================================
 
 # FIXME: not all of these work
-# TODO: effective distance
 admitted_vars = easy_keys + ['mtotal', 'q', 'mchirp',
                              'spin1a', 'spin2a',
                              'incl', 'cos_incl', 'abs_incl', 'cos_abs_incl',
@@ -410,7 +409,7 @@ axis_labels_dict = {'mchirp': "Chirp Mass (solar masses)",
                     'q': "Mass ratio",
                     'distance': "Distance (Mpc)",
                     'eff_site_dist': f"{opts.ifo} effective distance (Mpc)",
-                    'eff_dist': "Inverse sum of effective distances (Mpc)",
+                    'eff_dist': "Inverse sum of inverse effective distances (Mpc)",
                     'end_time': "GPS time (s)",
                     'sky_error': "Rec. sky error (rad)",
                     'coa_phase': "Phase of complex SNR (rad)",
@@ -582,7 +581,7 @@ if plot_title is None:
                   'mtotal': "total mass",
                   'q': "mass ratio",
                   'distance': "distance (Mpc)",
-                  'eff_dist': "inverse sum of effective distances",
+                  'eff_dist': "inverse sum of inverse effective distances",
                   'eff_site_dist': "site specific effective distance",
                   'end_time': "time",
                   'coa_phase': "phase of complex SNR",


bin/pygrb/pycbc_pygrb_plot_injs_results
```diff
@@ -174,23 +187,26 @@ def complete_inj_data(injs, keys, tag, ifos=[]):
...
-parser = ppu.pygrb_initialize_plot_parser(description=__doc__)
-parser.add_argument("--found-missed-file",
-                    help="The hdf injection results file", required=True)
-parser.add_argument("--trig-file",
-                    help="The hdf offsource trigger file", required=True)
-parser.add_argument("--trigger-time",
-                    help="The GPS time of the external trigger", required=True)
+
 # FIXME: not all of these work
 # TODO: effective distance
```
pannarale: You can remove this TODO

bin/pygrb/pycbc_pygrb_plot_injs_results
```diff
@@ -386,30 +404,25 @@ if found_quieter[x_qty].size:
...
-                    'longitude': "Longitude (degrees)",
-                    'incl': "Inclination (iota)",
-                    'abs_incl': 'Magnitude of inclination (|iota|)',
+                    'end_time': "GPS time (s)",
+                    'sky_error': "Rec. sky error (rad)",
+                    'coa_phase': "Phase of complex SNR (rad)",
+                    'dec': "Declination (rad)",
+                    'ra': "Right ascension (rad)",
+                    'incl': "Inclination iota (deg)",
+                    'abs_incl': 'Magnitude of inclination |iota| (deg)',
```
pannarale, line 419: Did we really keep this single angle in degrees?  I have to double check, but I am hoping you did already :-)

bin/pygrb/pycbc_pygrb_plot_injs_results
```diff
@@ -386,30 +404,25 @@ if found_quieter[x_qty].size:
...
-
 # Take care of axes labels
 axis_labels_dict = {'mchirp': "Chirp Mass (solar masses)",
                     'mtotal': "Total mass (solar masses)",
                     'q': "Mass ratio",
                     'distance': "Distance (Mpc)",
-                    'eff_site_dist': f"{sitename.get(opts.ifo)} effective "
-                                     "distance (Mpc)",
+                    'eff_site_dist': f"{opts.ifo} effective distance (Mpc)",
                     'eff_dist': "Inverse sum of effective distances (Mpc)",
```
pannarale: Please correct this label since you caught the mistake.

pannarale (changes requested):

bin/pygrb/pycbc_pygrb_plot_injs_results
titodalcanton: `incl` and `abs_incl` are explicitly converted to degrees in the `complete_incl_data()` function. And if you look at plots, the inclination values are indeed plotted in degrees.

PR #5181, opened 2025-09-17
Update pkg_resources (to be deprecated) to use importlib for plugins

GarethCabournDavies:
pkg_resources is deprecated and will error in coming releases of setup tools. This looks to emulate the same plugin functionality but using importlib

This is an update given a deprecation warning
This change affects inference sampler and waveform plugins
This change changes no outputs, I hope
This change will - I think - be seamless

## Motivation
deprecation warnings state that pkg_resources will be removed, possibly as early as November

## Contents
remove pkg_resources iter_entry_points, and replace with importlib.metadata entry_points. with the group keyword, and using load() rather than resolve(), this _should_ then be identical

Part of https://github.com/gwastro/pycbc/issues/5179

## Testing performed
`tox -e py-unittest` works, and I think that uses the plugin waveform generator from BBHx

## Additional notes
🎵 

### DIFF
--- a/.github/workflows/bank-compress-workflow.yml
+++ b/.github/workflows/bank-compress-workflow.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: '3.10'
+        python-version: '3.11'
     - name: install condor
       run: |
         wget -qO - https://research.cs.wisc.edu/htcondor/ubuntu/HTCondor-Release.gpg.key | sudo apt-key add -
--- a/.github/workflows/basic-tests.yml
+++ b/.github/workflows/basic-tests.yml
@@ -13,7 +13,7 @@ jobs:
       max-parallel: 60
       matrix:
         os: [ubuntu-22.04]
-        python-version: ['3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.11', '3.12']
         test-type: [unittest, search, docs]
     steps:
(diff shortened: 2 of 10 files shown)


pycbc/inference/models/__init__.py
```diff
@@ -242,9 +243,22 @@ def add_plugins(self):
...
+        # Check if the Python version is 3.10 or newer
+        if sys.version_info >= (3, 10):
+            # Use the modern syntax with the 'group' keyword argument
+            def get_entry_points(group_name):
+                return entry_points(group=group_name)
+        else:
+            # Use the older syntax for Python 3.9 and earlier
+            all_entry_points = entry_points()
+            def get_entry_points(group_name):
+                return all_entry_points.get(group_name, ())
```
titodalcanton: Given that this is implemented identically twice, and that at some point we may want to drop the Python 3.9 hack, I suggest putting it into a small helper function in one of the generic modules. Also, can the version check decision happen inside `get_entry_points()`?

GarethCabournDavies: I've put this in waveform plugins and the sampler plugin imports that one - I have tried to make it obvious where this can be remove when we drop 3.9 support

Version check is inside get_entry_points now

GarethCabournDavies: I will remove this hack, as we are dropping 3.9 support now anyway

titodalcanton (approved):

PR #5182, opened 2025-09-17
first pass to drop 3.9 and 3.10 support

ahnitz:

This drops support for python 3.9 and 3.10

No changes to analysis code, but the testing will no longer be run for 3.9 and 3.10. 

## Motivation
Python 3.9 and 3.10 are past normal support guidelines for scientific python. While maintaining broader compatibility is fine, we are increasingly hitting library deprecations and having to write code to support simultaneous versions. 

## Additional notes

### DIFF
--- a/.github/workflows/bank-compress-workflow.yml
+++ b/.github/workflows/bank-compress-workflow.yml
@@ -15,7 +15,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: '3.10'
+        python-version: '3.11'
     - name: install condor
       run: |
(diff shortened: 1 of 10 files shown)


titodalcanton:
After this, we can probably merge #5057?

titodalcanton:
Is it too early to start testing on 3.13?

ahnitz:
@titodalcanton I think we should add python 3.13 immediately after this in a separate PR assuming this is quick to do.

ahnitz:
@titodalcanton @spxiwh Looks like this works.

GarethCabournDavies (approved):
I don't see anywhere that this has missed

spxiwh (approved):
Thanks!

PR #5183, opened 2025-09-17
Expand PyGRB tables 

pannarale:

This is a: new feature in PyGRB tables.
- A column for p-values is added in injections tables (both "quiet found" and "missed found")
- The table that used to only display "missed-found" injections now also covers missed injections

This change affects: PyGRB

## Contents
- I tracked the p-value so that it is shown in the output tables (perhaps useful for "quiet" found injections).
- Given the frequent request to be able to easily read the injection parameters of missed injections, I extended the purpose of the "missed-found" table: it now also reports missed injections.
- Small dictionary improvements
- Fields missing when grouping injections together (e.g., a missed injection does not have a recovered chirp mass) are filled with `-1`s so that the formatting of the table is preserved (which is not the case using `None`, or strings such as `'-'`).

## Testing performed
I shared expanded tables with the PyGRB developers.

## Additional notes

### DIFF
--- a/bin/pygrb/pycbc_pygrb_page_tables
+++ b/bin/pygrb/pycbc_pygrb_page_tables
@@ -648,7 +648,7 @@ if found_missed_file is not None:
                            '##.##', '##.##', '##.##',
                            '##.##', '##.##'])
     format_strings.extend(['##.##' for ifo in ifos])
-    format_strings.extend(['##.##', '##.######',
+    format_strings.extend(['##.##', '#.######',
                            '##.##', '##.##', '##.##',
                            '##.##', '##.##', '##.##',
                            '##.##', '##.##'])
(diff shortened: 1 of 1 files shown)


bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -647,7 +648,7 @@ if found_missed_file is not None:
                            '##.##', '##.##', '##.##',
                            '##.##', '##.##'])
     format_strings.extend(['##.##' for ifo in ifos])
-    format_strings.extend(['##.##',
+    format_strings.extend(['##.##', '##.######',
```
titodalcanton: A single digit should be sufficient before the point.

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -696,21 +697,27 @@ if found_missed_file is not None:
...
+        if key not in cut_injs:
+            cut_injs[key] = np.full(len(cut_injs['mass1']), -1)
+        if key not in missed_injs:
+            missed_injs[key] = np.full(len(missed_injs['mass1']), -1)
+    t_missed = [np.concatenate((missed_injs[key], vetoed[key], cut_injs[key]))
+                for key in keys]
     t_missed = list(zip(*t_missed))
     t_missed.sort(key=lambda elem: elem[0])
-    logging.info("Writing %d missed-found injections to html file.",
+    logging.info("Writing %d missed and missed-found injections to html file.",
```
titodalcanton: Can we just say "Writing %d injections"?

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -696,21 +697,27 @@ if found_missed_file is not None:
...
     t_missed = zip(*t_missed)
     t_missed = [np.asarray(d) for d in t_missed]
     html_table = pycbc.results.html_table(t_missed, th,
                                           format_strings=format_strings,
                                           page_size=20)
-    kwds = {'title': "Missed found injections",
-            'caption': "Recovered parameters and statistic values of \
-            injections that are recovered, but with reweighted SNR \
-            below threshold or vetoed.",
+    kwds = {'title': "Missed and found but missed injections",
```
titodalcanton: "Found but missed" sounds like a confusing jargon. I think "missed", "cut" and "vetoed" (like the variable names) are more self-explanatory. What do you think of "Missed, cut and vetoed injections" as the title?

bin/pygrb/pycbc_pygrb_page_tables
```diff
@@ -696,21 +697,27 @@ if found_missed_file is not None:
...
+            'caption': "Parameters of missed injections and \
+            recovered parameters and statistic values of \
+            injections that are recovered but then missed because \
```
titodalcanton: I would replace "missed" with "cut" on this line.

bin/pygrb/pycbc_pygrb_page_tables
pannarale: This was to distinguish the message from the other one (`"Writing %d quiet-found injections to h5 and html files."`).  I will sprinkle the "missed", "cut" and "vetoed" suggestion around, and you can see check if things read better.

PR #5184, opened 2025-09-17
Add python 3.13

ahnitz:
Add support for python 3.13


ahnitz:
A surprising amount just works out of the box. The only error I see atm is because of a change in how escape characters are handled in python 3.13. These are now a different type of warning/error which causes the documentation to fail to build. 

e.g. errors like 
```
<unknown>:7: SyntaxWarning: invalid escape sequence '\i'
```

Issue #5179, opened 2025-09-08
Recent deprecation warnings


titodalcanton:
The "unittest" action currently raises this bunch of deprecation warnings:
```
.tox/py-unittest/lib/python3.12/site-packages/astropy/cosmology/core.py:30
  /home/runner/work/pycbc/pycbc/.tox/py-unittest/lib/python3.12/site-packages/astropy/cosmology/core.py:30: DeprecationWarning: The module `astropy.cosmology.core` is deprecated since v7.1 and will be removed in a future version. Import from `astropy.cosmology` instead.
    warnings.warn(

.tox/py-unittest/lib/python3.12/site-packages/pykerr/qnm.py:2
  /home/runner/work/pycbc/pycbc/.tox/py-unittest/lib/python3.12/site-packages/pykerr/qnm.py:2: UserWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html. The pkg_resources package is slated for removal as early as 2025-11-30. Refrain from using this package or pin to Setuptools<81.
    import pkg_resources

.tox/py-unittest/lib/python3.12/site-packages/pkg_resources/__init__.py:3146
  /home/runner/work/pycbc/pycbc/.tox/py-unittest/lib/python3.12/site-packages/pkg_resources/__init__.py:3146: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('ligo')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

.tox/py-unittest/lib/python3.12/site-packages/pkg_resources/__init__.py:3146
  /home/runner/work/pycbc/pycbc/.tox/py-unittest/lib/python3.12/site-packages/pkg_resources/__init__.py:3146: DeprecationWarning: Deprecated call to `pkg_resources.declare_namespace('sphinxcontrib')`.
  Implementing implicit namespace packages (as specified in PEP 420) is preferred to `pkg_resources.declare_namespace`. See https://setuptools.pypa.io/en/latest/references/keywords.html#keyword-namespace-packages
    declare_namespace(pkg)

test/test_bhspec.py::TestBHSpecModel::test_file_io
[...]

PR #5186, opened 2025-09-19
Deprecation fixes

GarethCabournDavies:
This fixes a few of the deprecation warnings in #5179

This is a: deprecation fix
This change changes: documentation, result presentation / plotting, scientific output

## Motivation
Things are deprecated; if we don't remove them now, they will be failures

## Contents
Generally, this is just going through the suggested fixes for the deprecations.

I'm not sure on one or two places:
- in `pycbc/coordinates/space.py`, could this ever actually be an array, is it only a single-value array in the tests? This is also a warning in `pycbc/events/coinc.py`, which I need to check

#5179 

## Testing performed
`tox-e py-unittest` runs with fewer deprecation warnings

## Additional notes

### DIFF
--- a/pycbc/inference/models/relbin.py
+++ b/pycbc/inference/models/relbin.py
@@ -223,7 +223,7 @@ class Relative(DistMarg, BaseGaussianNoise):
         for ifo in data:
             # store data and frequencies
             d0 = self.data[ifo]
-            self.f[ifo] = numpy.array(d0.sample_frequencies)
+            self.f[ifo] = numpy.asarray(d0.sample_frequencies)
             self.df[ifo] = d0.delta_f
             self.end_time[ifo] = float(d0.end_time)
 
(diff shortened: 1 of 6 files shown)


pycbc/inference/models/relbin.py
```diff
@@ -223,7 +223,7 @@ def __init__(
         for ifo in data:
             # store data and frequencies
             d0 = self.data[ifo]
-            self.f[ifo] = numpy.array(d0.sample_frequencies)
+            self.f[ifo] = numpy.asarray(d0.sample_frequencies)
```
titodalcanton: Could this just be `d0.sample_frequencies.numpy()`?

GarethCabournDavies:
```suggestion
            self.f[ifo] = d0.sample_frequencies.numpy()
```

GarethCabournDavies:
tested, and this passes the tox py -unittest once #5195 is applied

titodalcanton (approved):

PR #5188, opened 2025-09-22
Bugfix to `plot_posterior` when plotting extracted posterior files

acorreia61201:
This PR fixes a bug when running `pycbc_inference_plot_posterior` with an extracted posterior file and wildcard parameters. 

This is a bug fix. This change affects posterior plotting via `plot_posterior`. This change changes the behavior of `plot_posterior` when using a wildcard parameter argument to call only the parameters present in the file.

## Motivation
Currently, if using a wildcard (i.e. `--parameters '*'`), this script calls the `variable_params` section of the file, which assumes that the parameters to be plotted are the same as the parameters that were sampled over during PE. However, if the user provides a posterior extracted using `pycbc_inference_extract_samples`, the parameters stored in the file are not guaranteed to be the same as the `variable_params` (e.g., a waveform transform may have been applied during extraction). Thus, the script will attempt to call parameter names that may not be in the posterior and throw an error. Ideally, the wildcard should call all the parameters present in the file regardless of what was sampled over during PE.

## Contents
This PR changes the behavior of `plot_posterior` (via the `ResultsArgumentParser` call) to plot all parameters in the file rather than the `variable_params` section when using the `'*'` wildcard. All other functionality remains unchanged; i.e. if the user specifies all of the parameters manually, this PR has no effect.

## Additional notes
This largely produces the same expected functionality of the wildcard parameter. There is one additional side effect: if the posterior contains likelihood parameters (e.g. `loglikelihood`, `logwt`), those will be plotted as their own scatter plots. Ideally, these wouldn't be plotted, but they have no effect on the plotting otherwise. Additional changes can be made to this PR to exclude these parameters if that is the optimal behavior.

### DIFF
--- a/pycbc/inference/io/__init__.py
+++ b/pycbc/inference/io/__init__.py
@@ -446,11 +446,13 @@ class ResultsArgumentParser(argparse.ArgumentParser):
         to not be included. May also specify sampler-specific arguments. Note
         that ``input-file``, ``file-help``, and ``parameters`` are always
         added.
-    defaultparams : {'variable_params', 'all'}, optional
+    defaultparams : {'variable_params', 'samples', 'all'}, optional
         If no ``--parameters`` provided, which collection of parameters to
-        load. If 'all' will load all parameters in the file's
-        ``samples_group``. If 'variable_params' or None (the default) will load
-        the variable parameters.
+        load. If 'samples' will load all parameters in the file's
+        ``samples_group``, excluding likelihood stats (i.e. loglikelihood,
+        logwt). If 'all' will load all parameters in the file's
+        ``samples_group`` including likelihood stats. If 'variable_params' or 
+        None (the default) will load the variable parameters.
     autoparamlabels : bool, optional
         Passed to ``add_results_option_group``; see that function for details.
     \**kwargs :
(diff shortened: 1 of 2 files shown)


ahnitz:
@aleynaakyuz Can you please review this?

pycbc/inference/io/__init__.py
```diff
@@ -462,11 +464,15 @@ def __init__(self, skip_args=None, defaultparams=None,
...
         self.no_input_file_err = False
+        self.skip_params = []
         if skip_args is None:
             skip_args = []
         self.skip_args = skip_args
         if defaultparams is None:
             defaultparams = 'variable_params'
+        if defaultparams == 'samples':
+            defaultparams = 'all'
+            self.skip_params = ['loglikelihood', 'logwt']
```
aleynaakyuz, line 475: You’re only skipping two parameters here, but there are other ones that could also be plotted. I plotted result_extracted.hdf [here](https://github.com/sugwg/GW250114/tree/main/pe/results/spin/aligned_spin_imr) and the result looks like [this](https://sugwg-web.syr.edu/~aakyuz/pr-extracted.png)

PR #5190, opened 2025-09-24
Revert to regular BBHx install for tests

titodalcanton:
A while ago we started using a fork of BBHx as a workaround for some issues upstream. Let's see if we can drop that now.

#4857 

## Testing performed

None, relying on CI.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -21,7 +21,7 @@ conda_deps =
 
 [bbhx]
 deps =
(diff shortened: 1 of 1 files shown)


spxiwh:
This won't work. BBHx has changed the API and the current version is not compatible. Good news is that it will install nicely and everything, but we need to update the interface code!

PR #5191, opened 2025-09-30
Fix taper issue

aleynaakyuz:
Tapering time-domain waveforms using LAL currently produces a sharp cutoff in the phase posterior, as shown below:

example-sharp-cut

This PR introduces the following changes:

- Moves taper_timeseries from utils to the TimeSeries class.

- Adds tapering as a built-in option of TimeSeries.

- Introduces support for constant window tapering.

- Adds two new options:
`taper_window`
`taper_method`

Using the new options as `taper=start`, `taper_window=0.04`, `taper_method=constant ` produces a phase prior without the sharp cutoff, as shown below:
example-constant-window

### DIFF
--- a/pycbc/strain/strain.py
+++ b/pycbc/strain/strain.py
@@ -226,7 +226,7 @@ def from_cli(opt, dyn_range_fac=1, precision='single',
                     end_time=opt.gps_end_time+opt.pad_data,
                     sieve=sieve)
         elif opt.frame_files or opt.frame_cache:
-            strain = pycbc.frame.read_gw_data(
+            strain = pycbc.frame.read_frame(
                     frame_source, opt.channel_name,
                     start_time=opt.gps_start_time-opt.pad_data,
                     end_time=opt.gps_end_time+opt.pad_data,


bin/pycbc_banksim
```diff
@@ -134,8 +133,8 @@ def get_waveform(approximant, phase_order, amplitude_order, spin_order,
                                  f_lower=start_frequency,
                                  amplitude_order=amplitude_order)
         if hasattr(wf_params, 'taper'):
-            hp = taper_timeseries(hp, wf_params.taper)
-            hc = taper_timeseries(hc, wf_params.taper)
+            hp.taper_timeseries(location=wf_params.taper)
```
ahnitz: Tapering shouldn't be in-place, so who does thsi work? Shouldn't it be 

hp = hp.taper_timeseries.....

pycbc/types/timeseries.py
```diff
@@ -562,7 +563,72 @@ def psd(self, segment_duration, **kwds):
         return welch(self, seg_len=seg_len,
                            seg_stride=seg_stride,
                            **kwds)
+    
+    # map between tapering string in sim_inspiral table or inspiral
+    # code option and lalsimulation constants
 
+    def taper_timeseries(self, location=None, tapermethod='lal', return_lal=False, gate_params=None):
```
ahnitz: This doesn't seem to really simplify how to taper no? It's just a rename of the gate method then? Shouldn't the options make it easier to apply the intentded tapering?

ahnitz: Also, shouldn't locatoin still be respected when using a constant time width?

pycbc/waveform/generator.py
```diff
@@ -276,12 +276,21 @@ def _postgenerate(self, res):
...
-            hc = taper_timeseries(hc, tapermethod=self.current_params['taper'])
-        except KeyError:
-            pass
-        return hp, hc
+
+        if self.current_params.get('taper_method') == 'constant':
+            if self.current_params['taper_window'] is None:
+                raise ValueError("If taper_method is 'constant', taper_window must be set")
+            else:
+                gate_params_hp = [(hp.start_time, self.current_params['taper_window']/2, self.current_params['taper_window'])]
```
ahnitz: See previous comment as this shouldn't be so complicated here if you simplify the method arguments.

pycbc/waveform/generator.py
```diff
@@ -306,8 +315,8 @@ def _postgenerate(self, res):
             tapermethod = self.current_params['taper']
             for mode in res:
                 ulm, vlm = res[mode]
-                ulm = taper_timeseries(ulm, tapermethod=tapermethod)
-                vlm = taper_timeseries(vlm, tapermethod=tapermethod)
+                ulm = ulm.taper_timeseries(location=tapermethod)
```
ahnitz: We might want to clarify between the method and the location now in the naming. Also, how does this allow using the constant window? It would seem to not allow it.

pycbc/waveform/waveform.py
```diff
@@ -1317,9 +1317,9 @@ def get_two_pol_waveform_filter(outplus, outcross, template, **kwargs):
         # taper the time series hp if required
         if 'taper' in input_params.keys() and \
                 input_params['taper'] is not None:
-            hp = wfutils.taper_timeseries(hp, input_params['taper'],
+            hp = hp.taper_timeseries(location=input_params['taper'],
```
ahnitz: Again, no ability to using the constant window here.

pycbc/inject/inject.py
```diff
@@ -1262,7 +1262,7 @@ def apply(self, strain, detector_name, f_lower=None, distance_scale=1):
 
             # compute the detector response, taper it if requested
             # and add it to the strain
-            strain = wfutils.taper_timeseries(strain, inj.taper)
+            strain = strain.taper_timeseries(location=inj.taper)
```
ahnitz: No ability to use constant window.

pycbc/inject/inject.py
```diff
@@ -104,8 +104,8 @@ def projector(detector_name, inj, hp, hc, distance_scale=1):
 
     # taper the polarizations
     try:
-        hp_tapered = wfutils.taper_timeseries(hp, inj.taper)
-        hc_tapered = wfutils.taper_timeseries(hc, inj.taper)
+        hp_tapered = hp.taper_timeseries(location=inj.taper, tapermethod=inj.get('taper_method'))
```
ahnitz: How does one use the constant window here?

ahnitz:
@acorreia61201 I've provided a few initial comments, but can you then do a thorough review after the initial comments are addressed.

aleynaakyuz:
The parallel PR for tutorial tests are [here](https://github.com/gwastro/PyCBC-Tutorials/pull/35)

aleynaakyuz:
Also, with recent changes, I had to adjust the `taper_window`. New test parameters are `taper = start`,` taper_window = 0.085`, `taper_method = constant`

pycbc/types/timeseries.py
```diff
@@ -562,7 +563,98 @@ def psd(self, segment_duration, **kwds):
...
+            'TAPER_STARTEND': sim.SIM_INSPIRAL_TAPER_STARTEND,
+            'startend'      : sim.SIM_INSPIRAL_TAPER_STARTEND}
+
+        taper_func_map = {
+            _numpy.dtype(float32): sim.SimInspiralREAL4WaveTaper,
+            _numpy.dtype(float64): sim.SimInspiralREAL8WaveTaper}
+
+        tsdata = self
+
+        if location is None:
```
acorreia61201, line 610: I'm not sure if this condition is necessary. If location is None, it should fulfill both this and the next condition.

pycbc/types/timeseries.py
```diff
@@ -562,7 +563,98 @@ def psd(self, segment_duration, **kwds):
...
+            raise ValueError("Must specify a tapering method (function was called"
+                            "with tapermethod=None)")
+        if location not in taper_map.keys():
+            raise ValueError("Unknown tapering method %s, valid methods are %s" % \
+                            (tapermethod, ", ".join(taper_map.keys())))
```
acorreia61201: If this is meant to catch bad location args, this should print "Unknown tapering location {location}"; right now this prints the tapering method and lists valid location args.

bin/pycbc_single_template
```diff
@@ -293,8 +293,7 @@ with ctx:
         if len(td_template) > (7 * tlen // 8):
             new_start_idx = len(td_template) - (7 * tlen // 8)
             td_template = td_template[new_start_idx:]
-            taper_func = pycbc.waveform.utils.taper_timeseries
-            td_template = taper_func(td_template, tapermethod='TAPER_START')
+            td_template = td_template.taper_timeseries(location='TAPER_START')
```
acorreia61201, line 296: Should this be modified to use the constant window? It would require hard-coding some window length...on the other hand, if the results here don't have to be as accurate as full-fledged PE, it's probably fine as-is.

pycbc/waveform/generator.py
```diff
@@ -306,8 +305,8 @@ def _postgenerate(self, res):
             tapermethod = self.current_params['taper']
             for mode in res:
                 ulm, vlm = res[mode]
-                ulm = taper_timeseries(ulm, tapermethod=tapermethod)
-                vlm = taper_timeseries(vlm, tapermethod=tapermethod)
+                ulm = ulm.taper_timeseries(location=tapermethod, tapermethod=self.current_params.get('taper_method'), taper_window=self.current_params.get('taper_window'))
```
acorreia61201: It might be beneficial to change the variable names a bit here to avoid confusion. You could change line 305 to something like `location = self.current_params['taper']`, then specify `location=location`; or you could remove 305 and just specify the current_params directly in the call signature.

pycbc/types/timeseries.py
```diff
@@ -562,7 +563,98 @@ def psd(self, segment_duration, **kwds):
...
+
+        Parameters
+        ----------
+        tsdata : TimeSeries
+            Series to be tapered, dtype must be either float32 or float64
+        location : string
+            Should be one of ('TAPER_NONE', 'TAPER_START', 'TAPER_END',
+            'TAPER_STARTEND', 'start', 'end', 'startend') - NB 'TAPER_NONE' will
+            not change the series!
+        tapermethod : string
```
acorreia61201, line 583: Might be good to specify this is an optional arg that defaults to 'lal'

pycbc/types/timeseries.py
```diff
@@ -562,7 +563,98 @@ def psd(self, segment_duration, **kwds):
...
+        if location is None:
+            raise ValueError("Must specify a tapering method (function was called"
+                            "with tapermethod=None)")
+        if location not in taper_map.keys():
+            raise ValueError("Unknown tapering method %s, valid methods are %s" % \
+                            (tapermethod, ", ".join(taper_map.keys())))
+        if tsdata.dtype not in (float32, float64):
+            raise TypeError("Strain dtype must be float32 or float64, not "
+                        + str(tsdata.dtype))
+        if tapermethod == 'lal' or tapermethod == None:
```
acorreia61201: The way this is written, the function allows for the user to specify None, whereas the function should take in only 'lal' or 'constant' according to the docstring. This is a minor thing (i.e. I don't think there will be an instance where someone explicitly specifies None here), but if you wanted to prevent this behavior you could remove the None condition here and replace the instances in the other files where you have `self.current_params.get('taper_method')` with `self.current_params.get('taper_method', 'lal')`. This preserves the function defaulting to 'lal' if taper_method isn't specified, without allowing for an explicit None input.

bin/pycbc_banksim
```diff
@@ -134,8 +133,8 @@ def get_waveform(approximant, phase_order, amplitude_order, spin_order,
                                  f_lower=start_frequency,
                                  amplitude_order=amplitude_order)
         if hasattr(wf_params, 'taper'):
-            hp = taper_timeseries(hp, wf_params.taper)
-            hc = taper_timeseries(hc, wf_params.taper)
+            hp = hp.taper_timeseries(location=wf_params.taper, tapermethod=wf_params.get('taper_method'), taper_window=wf_params.get('taper_window'))
```
ahnitz: Check that this will actually work

bin/pycbc_banksim_skymax
```diff
@@ -135,8 +134,8 @@ def get_waveform(approximant, phase_order, amplitude_order, spin_order,
                                  f_lower=start_frequency,
                                  amplitude_order=amplitude_order)
         if hasattr(wf_params, 'taper'):
-            hp = taper_timeseries(hp, wf_params.taper)
-            hc = taper_timeseries(hc, wf_params.taper)
+            hp = hp.taper_timeseries(location=wf_params.taper, tapermethod=wf_params.get('taper_method'), taper_window=wf_params.get('taper_window'))
```
ahnitz: Check this

pycbc/types/timeseries.py
```diff
@@ -562,7 +563,98 @@ def psd(self, segment_duration, **kwds):
...
+                raise ValueError("If taper_method is 'constant', taper_window must be set")
+            elif location=='start' or location=='TAPER_START':
+                first_nonzero = _numpy.nonzero(tsdata)[0][0]
+                nonzero_starttime = tsdata.start_time + first_nonzero * tsdata.delta_t
+                gate_params = [(nonzero_starttime, 0, taper_window)]
+            elif location=='end' or location=='TAPER_END':
+                last_nonzero = _numpy.nonzero(tsdata)[0][-1]
+                nonzero_endtime = tsdata.end_time - last_nonzero * tsdata.delta_t
+                gate_params = [(nonzero_endtime - taper_window, 0, taper_window)]
+            elif location=='startend' or location=='TAPER_STARTEND':
```
ahnitz: Remove some redundancy

bin/pycbc_banksim
aleynaakyuz: @ahnitz I am not sure how to check this. There is an [example](https://github.com/gwastro/pycbc/tree/master/examples/banksim) about this part of the code however, it is outdated and I couldn't make it work. Are we using this script?

bin/pycbc_single_template
aleynaakyuz: I think it’s best to leave it as is for now since it’s already hard-coded. We can fix it in a separate PR.

ahnitz (approved):
@aleynaakyuz Do two things then this can be considered approved. 

(1) Clean up the PR so it doesn't mix other PRs into it. 
(2) after fixing that check that the code performs as expected in the examples.

acorreia61201:
@aleynaakyuz Looks like that was just a transient XLAL error. Everything seems to be passing after a rerun (barring the tutorials, but you have the parallel PR to fix that)

PR #5198, opened 2025-10-07
Quadtratic modes

nishkalrao20:
# Implementation of Quadratic modes in Ringdown

This is a: new feature

This change will: affect `waveform/ringdown.py`, with major additions to the `multimode_base` function. Supplemental edits to the `conversions.py`.

## Motivation
In general relativity, perturbed black holes formed in binary mergers settle to a stationary state by emitting gravitational waves (GWs) through a discrete set of quasi-normal modes (QNMs). While the linear QNMs have been extensively studied, recent numerical simulations and theoretical investigations (References: [1](https://arxiv.org/abs/2208.07380), [2](https://arxiv.org/abs/2208.07374), [3](https://arxiv.org/abs/2410.14529)) suggest that nonlinear (quadratic) modes may also be excited during the merger process. Implemented and tested quadratic mode injections and analyzed their detectability.

Beyond the linear approximation, the nonlinear nature of Einstein's equations where first-order perturbations will act as effective sources for second-order (quadratic) modes. At this level, the total waveform can be expressed as a sum of linear and quadratic contributions. Schematically, we have $$h_{+}-ih_{\times}=\frac{M}{r}\sum_{k\equiv\ell m n}A_{k}\exp\bigl[i\omega_{k}t+\Phi_{k}\bigr] Y_{k}$$ where the index $k$ runs over both linear modes and their nonlinear (quadratic) counterparts.

For quadratic modes generated by the coupling of two linear modes denoted $k_1\equiv(\ell_1 m_1 n_1)$ and $k_2\equiv(\ell_2 m_2 n_2)$, the following relations are expected:
$\omega_{\ell m n} = \omega_{\ell_1 m_1 n_1}+\omega_{\ell_2 m_2 n_2}$
$\frac{1}{\tau_{\ell m n}} = \frac{1}{\tau_{\ell_1 m_1 n_1}}+\frac{1}{\tau_{\ell_2 m_2 n_2}}$
$A_{\ell m n} \propto A_{\ell_1 m_1 n_1}A_{\ell_2 m_2 n_2}$
$\Phi_{\ell m n} = \Phi_{\ell_1 m_1 n_1}+\Phi_{\ell_2 m_2 n_2}+\text{const.}$

Implementation of the amplitudes and phase is done through the couplings prescribed, with an extra parameter for the coefficient of proportionality. The frequencies and damping times serve the agnostic runs, through the above relations.

## Contents
Syntax for the quadratic modes required is `qlmns=` _lmn=l1m1n1 x l2m2n2_ (e.g, 440=220x220).
An example of the relevant snippet of the configuration file:
```
[variable_params]
logqref_amp = 
qref_phi = 

[static_params]
lmns = 211 221 331 441
qlmns = 441=221x221

[prior-logqref_amp]
name = uniform
min-logqref_amp = 18
max-logqref_amp = 24

[waveform_transforms-qref_amp]
name = custom
inputs = logqref_amp
qref_amp = 10**logqref_amp

[prior-qref_phi]
name = uniform_angle
```

## Testing performed
The code has been rigorously tested in a companion repository: [quadratic_bbh](https://github.com/nishkalrao20/quadratic_bbh) with injections, PE runs, and verifying the detectability.

## Additional notes

### DIFF
--- a/pycbc/waveform/ringdown.py
+++ b/pycbc/waveform/ringdown.py
@@ -973,6 +973,7 @@ def multimode_base(input_params, domain, freq_tau_approximant=False):
                 # skip
                 continue
             if domain == 'td':
+                print('{}: Frequency: {}, Tau: {}'.format(lmn, freqs[lmn], taus[lmn]))
                 hplus, hcross = td_damped_sinusoid(
                     freqs[lmn], taus[lmn], amps[lmn], phis[lmn], sample_times,
                     l=int(lmn[0]), m=int(lmn[1]), n=int(lmn[2]),
@@ -984,6 +985,7 @@ def multimode_base(input_params, domain, freq_tau_approximant=False):
                 outplus += hplus
                 outcross += hcross
             elif domain == 'fd':
+                print('{}: Frequency: {}, Tau: {}'.format(lmn, freqs[lmn], taus[lmn]))
                 hplus, hcross = fd_damped_sinusoid(
                     freqs[lmn], taus[lmn], amps[lmn], phis[lmn], sample_freqs,
                     l=int(lmn[0]), m=int(lmn[1]), n=int(lmn[2]),
@@ -999,6 +1001,7 @@ def multimode_base(input_params, domain, freq_tau_approximant=False):
                 # skip
                 continue
             if domain == 'td':
+                print('{}: Frequency: {}, Tau: {}'.format(qlmn, qfreqs[qlmn], qtaus[qlmn]))
                 hplus, hcross = td_damped_sinusoid(
                     qfreqs[qlmn], qtaus[qlmn], q_amps[qlmn], q_phis[qlmn],
                     sample_times, l=int(qlmn[0]), m=int(qlmn[1]), n=int(qlmn[2]),
@@ -1010,6 +1013,7 @@ def multimode_base(input_params, domain, freq_tau_approximant=False):
                 outplus += hplus
                 outcross += hcross
             elif domain == 'fd':
+                print('{}: Frequency: {}, Tau: {}'.format(qlmn, qfreqs[qlmn], qtaus[qlmn]))
                 hplus, hcross = fd_damped_sinusoid(
                     qfreqs[qlmn], qtaus[qlmn], q_amps[qlmn], q_phis[qlmn],
                     sample_freqs, l=int(qlmn[0]), m=int(qlmn[1]), n=int(qlmn[2]),


pycbc/waveform/ringdown.py
```diff
@@ -119,6 +119,50 @@ def format_lmns(lmns):
...
+    else:
+        raise ValueError('Format of parameter qlmns not recognized. See '
+                         'approximant documentation for more info.')
+
+    out = []
+    # Cycle over the qlmns to ensure that we get back a list of strings that
+    # are three digits long, and that nmodes!=0
+    for qlmn in qlmns:
+        # Quadratic modes are given as qlmn=lmn1xlmn2
+        qlmn, lmn12 = qlmn.split('=')
```
cdcapano, line 157: I wouldn't use an `=` sign here. That could cause issues. How about using a colon instead; i.e., `qlmn:lmn1xlmn2`?

pycbc/waveform/ringdown.py
```diff
@@ -119,6 +119,50 @@ def format_lmns(lmns):
...
+    # Catch case of qlmns given as float (as int injection values are cast
+    # to float by pycbc_create_injections), cast to int, then string
+    if isinstance(qlmns, float):
+        qlmns = str(int(qlmns))
+    # Case 1: the qlmns are given as a string, e.g. '221 331'
+    if isinstance(qlmns, str):
+        qlmns = qlmns.split(' ')
+    # Case 2: the qlmns are given as strings in a list, e.g. ['221', '331']
+    elif isinstance(qlmns, list):
+        pass
```
cdcapano, line 147: Why an `elif` that leads to a `pass`? It'd make more sense to just add a second `if` that leads to the ValueError. In other words:
```suggestion
    if not isisntance(qlmns, list):
        raise ValueError(...)
```

pycbc/waveform/ringdown.py
```diff
@@ -179,6 +223,70 @@ def lm_amps_phases(**kwargs):
...
+    ref_dphi = kwargs.pop('dphi', 0.)
+    if isinstance(ref_amp, str) and ref_amp.startswith('amp'):
+        # assume a mode was provided; check if the mode exists
+        ref_mode = ref_amp.replace('amp', '')
+        try:
+            ref_amp = kwargs.pop(ref_amp)
+            amps[ref_mode] = ref_amp
+        except KeyError:
+            raise ValueError("Must provide an amplitude for the reference "
+                             "mode {}".format(ref_amp))
```
cdcapano, line 261: This block is copied from `lm_amps_phases`. Copying code is usually a bad idea. I'd move this to a stand-alone function (maybe, `get_ref_mode_params`), then have both `lm_amps_phases` and `qlm_amps_phases` call it.

pycbc/waveform/ringdown.py
```diff
@@ -179,6 +223,70 @@ def lm_amps_phases(**kwargs):
...
+            except KeyError:
+                raise ValueError('Must provide reference amplitude for quadratic modes, and \
+                                    amp{} and amp{} are required'.format(mode1, mode2))
+            try:
+                phis[modeq] = kwargs['phi' + mode1] + kwargs['phi' + mode2] + qref_phi
+            except KeyError:
+                raise ValueError('Must provide reference phi for quadratic modes, and \
+                                    phi{} and phi{} are required'.format(mode1, mode2))
+            dphis[modeq] = kwargs.pop('dphi'+qlmn, ref_dphi)
+            dbetas[modeq] = kwargs.pop('dbeta'+qlmn, ref_dbeta)
```
cdcapano, line 287: Does it make sense to have a dphi and dbeta for the quadratic modes? These parameters allow the phases and amplitudes of the -m modes to differ from the +m modes, as you would expect if equatorial symmetry is broken. Would you also expect that freedom for the quadratic modes, or would that be sourced from linear modes?

pycbc/waveform/ringdown.py
```diff
@@ -119,6 +119,50 @@ def format_lmns(lmns):
 
     return out
 
+def format_qlmns(qlmns):
+    """Checks if the format of the parameter qlmns is correct, returning the
+    appropriate format if not, and raise an error if nmodes=0.
```
cdcapano, line 124: Here and in other doc strings, please initially define the qlmns as the quadratic modes. It took me awhile to realize what `qlmn` stood for.

pycbc/waveform/ringdown.py
```diff
@@ -119,6 +119,50 @@ def format_lmns(lmns):
...
+
+    The required format for the ringdown approximants is a list of qlmn modes
+    as a single whitespace-separated string, with n the number
+    of overtones desired. Alternatively, a list object may be used,
+    containing individual strings as elements for the qlmn modes.
+    For instance, qlmns = '221 331' are the modes 220, 221, 222, and 330.
```
cdcapano, line 130: Wouldn't `221 331` give `220 330`? I think for `220, 221, 222`, you'd need to specify `223`. 

I regret that we made the N be the number of modes rather than the overtone number. It's made it confusing on how to specify overtones. At some point I would like to switch to explicitly setting overtones (i.e., instead of setting `lmns = 222` for 220+221, you would do `lmns = 220 221`). However, that will break backward compatibility, so would require some thought to implement. This PR isn't the place for that.

That said, does it make sense to follow the same syntax for quadratic modes? It might add to the confusion, but maybe for the qlmns the n should be the overtone number rather than the number of modes. The issue I could see with the current format is you wouldn't be able to form a quadratic mode from different overtones, although I don't know if that's possible or not.

pycbc/conversions.py
```diff
@@ -1260,6 +1198,47 @@ def get_lm_f0tau_allmodes(mass, spin, modes):
             tau[key.format(l, abs(m), n)] = tmp_tau
     return f0, tau
 
+def get_qlm_f0tau_allmodes(mass, spin, qmodes):
+    """Returns a dictionary of all of the frequencies and damping times for the
+    requested modes.
```
cdcapano, line 1203: Again, make explicit what the `q` stands for in the doc string.

pycbc/conversions.py
```diff
@@ -1260,6 +1198,47 @@ def get_lm_f0tau_allmodes(mass, spin, modes):
...
+    ----------
+    mass : float or array
+        Mass of the black hole (in solar masses).
+    spin : float or array
+        Dimensionless spin of the final black hole.
+    qmodes : list of str
+        The modes to get. Each string in the list should be formatted
+        'qlmN', where l (m) is the l (m) index of the
+        harmonic and N is the number of overtones to generate (note, N is not
+        the index of the overtone).
```
cdcapano, line 1215: As I said above, I think it more sense with the quadratic lmns to explicitly specify the overtone number, rather than make N the number of modes.

pycbc/waveform/ringdown.py
```diff
@@ -749,63 +901,130 @@ def multimode_base(input_params, domain, freq_tau_approximant=False):
...
-                pol=pols[lmn], polnm=polnms[lmn])
-            outplus[kmin:] += hplus
-            outcross[kmin:] += hcross
+    if input_params['lmns'] is not None:
+        for lmn in freqs:
+            if amps[lmn] == 0.:
+                # skip
+                continue
+            if domain == 'td':
+                print('{}: Frequency: {}, Tau: {}'.format(lmn, freqs[lmn], taus[lmn]))
```
cdcapano, line 976: Debugging print statements should be removed.

pycbc/waveform/ringdown.py
nishkalrao20: I agree. So far, it hasn't caused issues in the runs, but I agree that it is better if we shift to a colon.

pycbc/waveform/ringdown.py
nishkalrao20: Yeah, I agree. I realized a similar syntax has been used in `format_lmns` too. Do I change that too?

pycbc/waveform/ringdown.py
nishkalrao20: I agree. I couldn't help but copy the base functions, but unifying them as a bigger function is helpful.

pycbc/waveform/ringdown.py
nishkalrao20: Oh, I am sorry. I will fix the same.

pycbc/waveform/ringdown.py
nishkalrao20: Sorry, the doc string was copied from the `lmn` mode. I will fix the same.

Yeah, I agree that $n$ being the number of modes is confusing. Yeah, it should be hard to fix that here, since we don't want to mess with the older versions!

I agree. In the present syntax, it is wrong if we want a quadratic mode from different overtones, because we specify: `441=221x221` `442=222x222`, I believe it messes up since `222` is 220+221. So far, for the tests, I have used only 221. Any suggestions on how to generate quadratic modes from different overtones? It would be helpful if the lmn syntax is changed.

pycbc/conversions.py
nishkalrao20: I agree, as I pointed out there. Should I just change the quadratic mode overtone number? So if we had 442=222x222, I would have to change the code such that 222 corresponds to 221 in the new syntax, and not 220+221.

pycbc/waveform/ringdown.py
nishkalrao20: I am really sorry. This change had been removed locally, I somehow am unsure how this got pushed.

PR #5204, opened 2025-10-14
Correction of `pygrb_plot_skygrid` script to plot legacy case

Thomas-JACQUOT:

This is a bug fix.

This change affects PyGRB.

## Motivation
`pygrb_plot_skygrid`  script was not covering the legacy case for sky grids, i.e. in the ini file there are `ra`, `dec` and `sky-error`. This is something that we still want to do in case for some reason we don't have FITS file for a GRB.

## Contents
Small if statement to distinguish both cases.

## Testing performed
I've generated a circular grid by giving `ra=45deg`, `dec=45deg` and `sky-error=3deg` and this is the output with H, HL,  HLV and HLVK network
circular_H
circular_HL
circular_HLV
circular_HLVK

### DIFF
--- a/bin/pycbc_make_sky_grid
+++ b/bin/pycbc_make_sky_grid
@@ -85,7 +85,7 @@ def make_multi_det_grid(args):
     if (args.ra and args.dec and args.sky_error) is not None:
         # --ra/--dec/--sky-error arguments
         assert args.coverage is None
-        args.input_dist = f"pycbc.distributions.UniformDiskSky(mean_ra={args.ra}, mean_dec={args.dec}, radius={args.sky_error})"
+        args.input_dist = f'UniformDiskSky(mean_ra="{args.ra}", mean_dec="{args.dec}", radius="{args.sky_error}")'
         sky_dist = dist_from_str(args.input_dist)
         grid = np.vstack((np.empty((0, 2)), (args.ra, args.dec)))
     elif args.input_dist == 'UniformSky()':


bin/pygrb/pycbc_pygrb_plot_skygrid
```diff
@@ -83,9 +83,13 @@ for outdir in outdirs:
...
-    ra , dec = f['ra'][:], f['dec'][:]
-    dist = f.attrs['input_distribution']
-    input_dist = eval("pycbc.distributions."+dist)
+    ra, dec = f['ra'][:], f['dec'][:]
+    if "input_distribution" in f.attrs.keys():
+        dist = f.attrs['input_distribution']
+        input_dist = eval("pycbc.distributions."+dist)
+    else:
+        mean_ra, mean_dec, sky_error = f.attrs['trigger_ra'], f.attrs['trigger_dec'], f.attrs['sky_error']
+        input_dist = pycbc.distributions.FisherSky(mean_ra=mean_ra, mean_dec=mean_dec, sigma=sky_error)
```
titodalcanton: When we build the sky grid with the `--ra`/`--dec`/`--sky-error` arguments, `pycbc_make_sky_grid` assumes a uniform disk distribution, not a Fisher distribution. We should be consistent here. Perhaps a better way to achieve this is to simply store the `input_distribution` attribute even when `pycbc_make_sky_grid` is called with `--ra`/`--dec`/`--sky-error`.

bin/pycbc_make_sky_grid
```diff
@@ -238,6 +238,7 @@ elif (args.ra and args.dec and args.sky_error) == None and "Uniform" in args.inp
     }
 else:
     extra_attributes = {
+        'input_distribution': f"UniformDiskSky(mean_ra={args.ra}, mean_dec={args.dec}, radius={args.sky_error})"
```
titodalcanton: This works, though we are still running against the idea of [not repeating things](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself). I propose the following instead. In line 88, instead of instantiating `sky_dist` directly, set `args.input_dist` using a string, as you are doing here in line 241. Then instantiate `sky_dist = dist_from_str(args.input_dist)` as is done for the other two cases. In this way, `args.input_dist` will be set correctly in all cases, and in the code down here you can simply set `extra_attributes['input_distribution']` once for all cases.

bin/pycbc_make_sky_grid
```diff
@@ -85,9 +85,8 @@ def make_multi_det_grid(args):
     if (args.ra and args.dec and args.sky_error) is not None:
         # --ra/--dec/--sky-error arguments
         assert args.coverage is None
-        sky_dist = pycbc.distributions.UniformDiskSky(
-            mean_ra=args.ra, mean_dec=args.dec, radius=args.sky_error
-        )
+        args.input_dist = f"pycbc.distributions.UniformDiskSky(mean_ra={args.ra}, mean_dec={args.dec}, radius={args.sky_error})"
```
titodalcanton: The `pycbc.distributions.` prefix should not be included here as it is already added by `dist_from_str()`. Also, quotes are needed around the arguments, since they will be strings when the units are included (e.g. "1.234rad").
```suggestion
        args.input_dist = f'UniformDiskSky(mean_ra="{args.ra}", mean_dec="{args.dec}", radius="{args.sky_error}")'
```

titodalcanton (changes requested):

bin/pycbc_make_sky_grid
titodalcanton: Argh I was actually wrong in my second statement. Those are floats, because we are using `angle_as_radians()` as type for those options. Well, in the end it does not really make a difference because `UniformDiskSky` calls `angle_as_radians()` anyways.

PR #5205, opened 2025-10-14
Remove polarization_samples deprecationwarning and actually deprecate it

GarethCabournDavies:
This is my attempt to remove the DeprecationWarning in setup_marginalization.

Ive added a marginalise_polarization kwarg which is True by default, and does some of the needed stuff that causes a crash later.

This is a deprecation
This change affects inference

This change might warrant a new release, but I'm not sure

## Motivation
Deprecate the thing that's had a DeprecationWarning for a while

## Contents
Remove the if statement that had the warning, add a True-by-default kwarg, and add in the stuff that was needed.

I may need to rethink this as it could break when using False, but I'm not sure. We could just do the marginalisation as default without any kwarg? This was already broken if polarization_samples wasn't in the kwargs before anyway so that could be a future fix

#5179 #4093 

## Testing performed
Passes the automated tests

## Additional notes
🗒️ 


pycbc/inference/models/tools.py
```diff
@@ -63,6 +62,8 @@ def setup_marginalization(self,
                               marginalize_vector_params=None,
                               marginalize_vector_samples=1e3,
                               marginalize_sky_initial_samples=1e6,
+                              marginalize_polarization=True,
```
ahnitz, line 65: See below

pycbc/inference/models/tools.py
```diff
@@ -126,20 +127,15 @@ def pop_prior(param):
         self.marginalize_sky_initial_samples = \
             int(float(marginalize_sky_initial_samples))
 
+        if marginalize_polarization:
```
ahnitz, line 130: I don't think you want this section at all. The point was to just use the more general marginalize_vector setup. I think this change should just be removing options, not renaming them.

GarethCabournDavies: This was added in as self.marginalize_vector_params must have 'polarization' as a key later. I'm happy to simply put this in as a no-if-statement thing if that is better?

PR #5206, opened 2025-10-14
Add option for high frequency cutoff in pycbc_inspiral

yi-fan-wang:
Adding options for high frequency cutoff in pybc_insiral. The motivation here is that: when downsampling, the antialising filter would only work well until ~0.85 Nyquist Frequency. This is not a problem when the gravitational wave frequency doesn't approach to that high, but may be an issue if so.

Also fix a typo in the pycbc_insiral:  args.require_valid_checkpoint --> opt.require_valid_checkpoint, because args doesn't exist.

Tested with the default search example:
 - The example search with this PR: https://hypatia.aei.mpg.de/~yiwang/pub/ssm/fupper_search/html/
 - Without this PR: https://hypatia.aei.mpg.de/~yiwang/pub/ssm/search/html/
 It turns out the ranking statistic changed a little bit

### DIFF
--- a/pycbc/vetoes/chisq.py
+++ b/pycbc/vetoes/chisq.py
@@ -35,22 +35,20 @@ def power_chisq_bins_from_sigmasq_series(sigmasq_series, num_bins, kmin, kmax):
 
     Parameters
     ----------
-
     sigmasq_series: FrequencySeries
         A frequency series containing the cumulative power of a filter template
         preweighted by a psd.
     num_bins: int
         The number of chisq bins to calculate.
     kmin: int
-        DOCUMENTME
+        Index corresponding to the lowest frequency to use in chisq calculation.
     kmax: int
-        DOCUMENTME
+        Index for the highest (ie maximum) frequency to use in chisq.
 
     Returns
     -------
-
     bins: List of ints
-        A list of the edges of the chisq bins is returned.
+        A list of the edges of the chisq bins.
     """
     sigmasq = sigmasq_series[kmax - 1]
     edge_vec = numpy.arange(0, num_bins) * sigmasq / num_bins
@@ -65,14 +63,13 @@ def power_chisq_bins(htilde, num_bins, psd, low_frequency_cutoff=None,
 
     Parameters
     ----------
-
     htilde: FrequencySeries
         A frequency series containing the template waveform
     num_bins: int
         The number of chisq bins to calculate.
     psd: FrequencySeries
         A frequency series containing the psd. Its length must be commensurate
-        with the template waveform.
+        with the template waveform. 
     low_frequency_cutoff: {None, float}, optional
         The low frequency cutoff to apply
     high_frequency_cutoff: {None, float}, optional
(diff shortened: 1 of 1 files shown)


bin/pycbc_inspiral
```diff
@@ -385,7 +389,7 @@ with ctx:
             tnum_start, event_mgr = events.EventManager.restore_state(checkpoint_file)
             checkpoint_exists = True
         except Exception as e:
-            if args.require_valid_checkpoint:
+            if opt.require_valid_checkpoint:
```
tdent, line 392: Can you put a note in the PR description for this please, eg 'Also fix checkpoint exception handling'

yi-fan-wang: @tdent Yes, done

tdent:
Concerning the example test search, there are very small changes in the single detector triggers, see the binned triggers and binned histogram plots.

pycbc/waveform/waveform.py
```diff
@@ -1167,7 +1167,8 @@ def td_fd_waveform_transform(approximant):
         # We can also make a td version of this
         # This will override any existing approximants with the same name
         # (ex. IMRPhenomXX)
-        cpu_td[approximant] = get_td_waveform_from_fd
+        if approximant not in td_apx:
```
ahnitz: @yi-fan-wang I don't think we can approve this. In particular, the lalsim internal conversions don't preserve phase / timing in as exact a way. This has caused several problems in the past with inconsistencies between templates and injections. If you want to be able to select between the path, I think having a way to do that would be fine, but please do not change the default here.

ahnitz (changes requested):

ahnitz (commented):
@yi-fan-wang One thing broadly here, is how does this compare to setting the f_final in the template bank? That should be similar, no? If not, you may want to check to see if there are places where the high_frequency_cutoff is not currently getting supplied that it would need to be. Naively I think setting f_final is more guaranteed to affect all the right places (think all the signal consistency tests, etc, not just MF).

pycbc/waveform/waveform.py
yi-fan-wang: @ahnitz Got it. I'll revert to remove this. (BTW I haven't understood the purpose of this. It's cherry-picked from v23_RELEASE_BRANCH and may be useful for some runs in some specific settings, though it's never in the main branch probably for the reasons you mentioned).

ahnitz: @yi-fan-wang Yeah, I think addresses if you want to allow either path is probably outside the scope of this PR, so if needed consider what to do separately.

tdent:
@ahnitz considering your question: In the FilterBank code, it looks like there is no way to specify a high frequency cutoff directly.  So if the MatchedFilterControl can take f_upper, why can't FilterBank as well?  (Or vice versa, why does MatchedFilterControl advertise an option if setting it doesn't consistently affect the template?)

I'll also note that power_chisq_bins() has a kwarg `high_frequency_cutoff=None` ... but this isn't set by any of the other functions in vetoes that call it. 

So at present it looks like _neither_ the bank f_final nor MatchedFilterControl can influence the actual setup of power chisq ... A bit of a mess no?

PR #5209, opened 2025-10-20
Add binary script to generate mock data

sum33it:

This is a: new feature and an efficiency update. This PR adds the features to the generation of mock data for third-generation ground-based detectors (@sum33it ) and LISA (@WuShichao). We make use of already existing functions and create a top-level binary file `pycbc_generate_mock_data` which shall contain i) noise channel, ii) signal channel, and iii) glitch channel separately and together. 

This change affects: None. Since most of the code is addition to the current codes, it does not change previous codes.

This change will: not break current functionality, does not require additional dependencies, do require a new release.

## Motivation

The mock data generation for next-generation detectors (or for the current generation) for testing PE/search pipelines is an important tool. Even though there are publicly available mock data challenges, the developers might need to generate mock data with their own choice of noise, signal, and even glitches. This PR aims to provide these tools for PyCBC users.

## Contents

This PR contains a top-level binary code for generating mock data for a network of ground-based detectors, LISA detector, and example script to generate the mock data. 

None

## Testing performed

The example scripts for tests are to be put in `examples/mdc_generation`. We plan to do tests of mock data generation for a few hours with a chosen binary population and glitch distribution.

## Additional notes

Co-authored-by: WuShichao <[shichao.wu@aei.mpg.de](mailto:shichao.wu@aei.mpg.de)>


ahnitz (commented):
@sum33it @WuShichao Can you explain why you wouldn't just use pycbc_condition_strain directly? What does this do that it doesn't? And if there is something, why not improve that script? I think some explanation of the use case is needed here. A

bin/pycbc_generate_mock_data
```diff
@@ -0,0 +1,133 @@
+#!/Users/Kumar020/ve/python3p11p9/pycbc_development/bin/python
```
ahnitz: This should be fixed

sum33it: Sure! Fixing it in the next commit.

ahnitz:
@sum33it Shouldn't glitch generation be a generic injection capability though? Hence it would automatically work with say pycbc_condition_strain as specified in an injection file?

ahnitz:
@sum33it The other points make more sense assuming you want to have some of the population builtin (maybe some default config file defining the populations??).

sum33it:
@ahnitz regarding glitch: yes it can be a in a generic injection capability. 

Regarding population: Yes I plan to put a default config file and if someone choses default population, this binary script first create a population injection file and then feed it to the `pycbc_condition_strain` like the script does it now. We can also put few variation of the default population  e.g. at higher redshift. 

Lastly, I wanted to add an option for the people who just wanted to generate frame file with one injection in a detector network with/without some generic glitch/overlapping signals. These requirements came out from discussion within ET collaboration for the groups which are testing their pipelines and want to generate quick mock data and need a tool for this. So I want to package all these generic options in a single top level script.

ahnitz:
@sum33it I think the documentation should just be in this PR if that's what you mean.

WuShichao:
> @sum33it Shouldn't glitch generation be a generic injection capability though? Hence it would automatically work with say pycbc_condition_strain as specified in an injection file?
> 
> The same for options for earth rotation, etc. Those should be fundamental capabilities, not high level ones. Also, that is somewhat an option already in the injection configuration, though only as a static parameter that applies to all the injections.

We will also make this compitable with `LISA Glitch` package https://lisa-simulation.pages.in2p3.fr/glitch/

docs/mock_data_generation.rst
```diff
@@ -0,0 +1,193 @@
...
+an independent random realization.
+
+---------------------------------------------------------
+Signals embedded in noise
+---------------------------------------------------------
+
+To generate time series data for a detector network, that contain the signal(s)
+and noise realization, one need to provide an ``hdf`` injection file containing
+the list of injection parameters. The following command generate the projected
+signals in each detector of the network, without the noise::
```
WuShichao: "without the noise" should be "with the noise"

WuShichao:
I add support of `--fake-strain-from-file`.

ahnitz (approved):

PR #5213, opened 2025-10-27
Fixed Constraints, Modified Brute Bank to enable optimized match calculation

kkacanja:
****Fixed Constraints****

Constraints were not working for config files used in brute bank especially for waveform transformed parameters. The constraints such as 

[constraint-1]
name = custom
inputs = mass1, mass2
constraint_arg = (mass1 <= 2.0) & (mass1 >= 0.1) & (mass2 <= 1.0) & (mass2 >= 0.1)

Were not recognized. The issue resulted since these are waveform transform parameters. To resolve this, I made sure the waveform transforms are being called first.

****Brute Bank Changes****

The changes in brute bank allow the user to enable an optimized match calculation that uses the head and tail of the waveform, essentially looking over a small time offset. This optimization works well for smaller parameter spaces. When exploring larger parameter spaces, there may be better matches at further offsets in time so using the optimized match option may result with a  bank with fewer matches or holes.

### DIFF
--- a/pycbc/distributions/constraints.py
+++ b/pycbc/distributions/constraints.py
@@ -57,54 +57,30 @@ class Constraint(object):
     def __call__(self, params):
         """Evaluates constraint.
         """
-        # cast to FieldArray
+
         if isinstance(params, dict):
             params = record.FieldArray.from_kwargs(**params)
         elif not isinstance(params, record.FieldArray):
             raise ValueError("params must be dict or FieldArray instance")
 
-        # Force transform execution before constraint attempt
-        if self.transforms:
-            # transforms.apply_transforms returns a NEW FieldArray object
-            params = transforms.apply_transforms(params, self.transforms)
-
-        # try to evaluate; this will assume that all of the needed parameters
-        # for the constraint exists in params
         try:
             out = self._constraint(params)
-        except Exception as e:
+        except (NameError, AttributeError, TypeError):
 
-            raise e
+            if self.transforms:
+                params = transforms.apply_transforms(params, self.transforms)
+
+            out = self._constraint(params)
+
 
         if isinstance(out, record.FieldArray):
             out = out.item() if params.size == 1 else out
         return out
 
     def _constraint(self, params):
-        """ Evaluates constraint function. """
-        eval_scope = {}
-        eval_scope.update(numpy.__dict__)
-
-        if params.size == 0:
-            return numpy.array([], dtype=bool)
-
-        for name in params.dtype.names:
-            data = params[name]
-
-            if data is None:
-                data = numpy.full(params.shape, -numpy.inf)
-            elif data.dtype == object:
-                none_mask = numpy.array([x is None for x in data])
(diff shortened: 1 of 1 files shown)


bin/bank/pycbc_brute_bank
```diff
@@ -81,6 +81,11 @@ parser.add_argument('--tau0-cutoff-frequency', type=float, default=15.0)
 parser.add_argument('--nprocesses', type=int, default=1,
     help='Number of processes to use for waveform generation parallelization. If not given then only a single core will be used.')
 pycbc.psd.insert_psd_option_group(parser)
+parser.add_argument('--optimized-match', action='store_true',
```
ahnitz: Use a more descriptive name of what is being done. Optimized might imply many different things.

bin/bank/pycbc_brute_bank
```diff
@@ -320,8 +325,13 @@ class GenUniformWaveform(object):
...
         self.ifft = pycbc.fft.IFFT(qtilde, q)
-        self.md = q._data[-100:]
-        self.md2 = q._data[0:100]
+
+        if args.optimized_match:
+            self.md = q._data[-100:]
+            self.md2 = q._data[0:100]
+        else:
+            self.md = q
+            self.md2 = numpy.zeros(20)
```
ahnitz: This was a bit of a hack, what should be done is in the place where this is used, to just use the first one. That way the second path wouldn't be checked at all rather than be set to a nonsense value.

pycbc/distributions/constraints.py
```diff
@@ -57,22 +57,22 @@ def __init__(self, constraint_arg, static_args=None, transforms=None,
...
-        # try to evaluate; this will assume that all of the needed parameters
-        # for the constraint exists in params
         try:
             out = self._constraint(params)
-        except NameError:
-            # one or more needed parameters don't exist; try applying the
-            # transforms
-            params = transforms.apply_transforms(params, self.transforms) \
-                     if self.transforms else params
+        except (NameError, AttributeError, TypeError):
```
ahnitz, line 68: Is the main issue here that a different error was being produce that wasn't being caught to go through this path?

kkacanja: It seems like the original issue triggers NameError tries to execute the waveform transforms to define what mass1 and mass2 are before applying the constraints

hence the error 1 "ValueError: no field of name (mass1 <= 2"

Tries again:

Error 2: "ValueError: no field of name (mass1 <= 2.0) & (mass1 >= 0.1) & (mass2 <= 1.0) & (mass2 >= 0.1)"

kkacanja: After the third attempt I get a new error:

TypeError: 'NoneType' object is not subscriptable

So it was completely ignoring the transform

ahnitz (approved):
@kkacanja Please confirm that the constraints are now properly applied in this version. Otherwise, I think this is good to go.

PR #5214, opened 2025-10-28
Upgrade action dependencies to fix PyPI mixed-case deprecation

titodalcanton:
Release 2.10.0 failed to upload the source tarball because the mixed-case tarball name is no longer accepted:
```
Uploading PyCBC-2.10.0.tar.gz
WARNING  Error during upload. Retry with the --verbose option for more details. 
ERROR    HTTPError: 400 Bad Request from https://upload.pypi.org/legacy/        
         Filename 'PyCBC-2.10.0.tar.gz' is invalid, should be                   
         'pycbc-2.10.0.tar.gz'. 
```
Here I try to figure out what is happening and how to fix it properly.

For the record, I get the right lowercase tarball name when running `python setup.py sdist` from a virtualenv with Python 3.12 and setuptools 79.0.0.

### DIFF
--- a/.github/workflows/distribution.yml
+++ b/.github/workflows/distribution.yml
@@ -18,7 +18,7 @@ jobs:
       - uses: actions/checkout@v4
         with:
           fetch-depth: 0
-      - uses: actions/setup-python@v6
+      - uses: actions/setup-python@v5
         with:
           python-version: '3.11'
       - name: Install cibuildwheel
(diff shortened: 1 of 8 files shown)


ahnitz:
@titodalcanton Is some dependency forcing an old version maybe?

.github/workflows/distribution.yml
```diff
@@ -53,6 +53,8 @@ jobs:
         path: dist/
     - name: Build source distribution
       run: |
+        python -c 'import setuptools ; print("setuptools version", setuptools.__version__)'
```
ahnitz, line 56: Given the annoyance here, I think we can leave this print statement in the final PR to aid our future selves.

GarethCabournDavies (approved):
This looks good to me. The upgrade to packaging is going onto my todo list, but if anyone else gets to it, that would be great

PR #5216, opened 2025-11-01
Brute Bank Input File Error

kkacanja:
This patch initializes match if it is not defined. When using a seed bank in some instance I received "AttributeError: 'FrequencySeries' object has no attribute 'matches'. Did you mean: 'match'?"

### DIFF
--- a/bin/bank/pycbc_brute_bank
+++ b/bin/bank/pycbc_brute_bank
@@ -237,6 +237,16 @@ class TriangleBank(object):
                 return False
 
             hc = self[j]
+
+            # Defensive initialization if matches/indices are missing
+            if not hasattr(hc, 'matches'):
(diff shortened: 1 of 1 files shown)


bin/bank/pycbc_brute_bank
```diff
@@ -237,6 +237,16 @@ class TriangleBank(object):
                 return False
 
             hc = self[j]
+
+            # Defensive initialization if matches/indices are missing
+            if not hasattr(hc, 'matches'):
+                hc.matches = numpy.empty(len(self))
```
ahnitz, line 243: This is ok, but only if it's not really getting to this point often. It shouldn't be though.

ahnitz (approved):

PR #5217, opened 2025-11-03
Move the pycbc array type and some constants away from lalsuite

GarethCabournDavies:
One thing which is cumbersome about installing pycbc is the sheer size of the package.
Quite a big proportion of this is in lalsuite, and so this is work to start moving away from `lal` for a few bits.

There are two changes here; in the pycbc array's epoch, which used LIGOTimeGPS but is now a float64, this is probably the most invasive change. We also update the constants to have a fallback option if lal is not installed

This is an efficiency update (for efficiency in dependencies)
This change affects: the offline search, the live search, inference, PyGRB through the array type
This should change nothing in outputs

## Contents
The main things here are:

 - The change of the array epoch to a float64

 - EDITED: We also have inclusion of a `pycbc.constants` module. When imported, this checks for an environment variable, either `default` or `lal`, and then chooses between `astropy` / `numpy` or `lal` equivalent constants. I have started to update to use that module rather than lal in a few places, but that work will be ongoing.

 - I've changed from using lal's gpstime module, and GPSToUTC to use astropy times

## Testing performed
The array type has existing unit tests, which I believe cover this change

I have added a module to test that the constants are consistent between one another. This does require that `lal` is installed when the test itself is performed.

### DIFF
--- a/examples/multi_inspiral/faceon_faceaway.sh
+++ b/examples/multi_inspiral/faceon_faceaway.sh
@@ -32,8 +32,8 @@ PAD=8
 BANK_FILE=gw170817_single_template.hdf
 BANK_VETO_FILE=bank_veto_bank.xml
 CHANNEL=SIMULATED_GW170817
-RA="5.016076270234897 rad"
-DEC="-0.408407044967 rad"
+RA="'5.016076270234897 rad'"
+DEC="'-0.408407044967 rad'"
 
 echo -e "\\n\\n>> [`date`] Getting template bank"
 wget -nv -nc ${CONFIG_URL}/${BANK_FILE}
--- a/pycbc/events/eventmgr.py
+++ b/pycbc/events/eventmgr.py
@@ -836,8 +836,8 @@ class EventManagerCoherent(EventManagerMultiDetBase):
         starts = []
         ends = []
         for seg in self.segments[self.ifos[0]]:
-            starts.append(seg.start_time.gpsSeconds)
-            ends.append(seg.end_time.gpsSeconds)
+            starts.append(int(seg.start_time))
+            ends.append(int(seg.end_time))
         f['search/segments/start_times'] = starts
         f['search/segments/end_times'] = ends
         # Individual ifo stuff
--- a/pycbc/filter/matchedfilter.py
+++ b/pycbc/filter/matchedfilter.py
@@ -240,7 +240,7 @@ class MatchedFilterControl(object):
         else:
             raise ValueError("Invalid downsample factor")
 
-    def full_matched_filter_and_cluster_symm(self, segnum, template_norm, window, epoch=None):
+    def full_matched_filter_and_cluster_symm(self, segnum, template_norm, window, epoch=""):
         """ Returns the complex snr timeseries, normalization of the complex snr,
         the correlation vector frequency series, the list of indices of the
         triggers, and the snr values at the trigger locations. Returns empty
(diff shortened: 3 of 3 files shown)


pycbc/constants.py
```diff
@@ -0,0 +1,102 @@
...
+
+    raise NotImplementedError(
+        'Contact the PyCBC team, this should never happen. You are '
+        'trying to use a LAL constant which doesnt have an astropy/numpy '
+        'fallback defined.'
+    )
+
+# Expose the constants as attributes of the module
+
+C_SI = get_constant('C_SI')
```
ahnitz: @GarethCabournDavies All constants should be gotten from or derived from astropy where it it exists.

ahnitz: Or thinks like PI should just be gotten from say numpy / scipy as appropriate.

ahnitz (changes requested):

pycbc/constants.py
GarethCabournDavies: That's what this will do 'under the hood' - if I remember right there is some reason (I can't remember exactly what) that lal defines its own PI (and TWOPI) that needs to exact match something on their end. I will see if I can find out why

Or are you saying we should use astropy/numpy/scipy constants as preference over LAL?

spxiwh: @ahnitz While I agree that it is better in general to use `astropy` for constants to be consistent, both of the main GW collaborations (IGWN+3G (lumping these together)) and LISA, define constants separately from `astropy`, and sometimes this can matter. (Cosmology choices tend to be the most obvious gotcha, but LISA positioning might also be quite sensitive as well).

I think it would be good to consider a mechanism where `astropy` is the default, but you should be able to say "I want constants from LAL" or "I want constants from `lisa_constants`" ... We could also try to persuade both those communities to be more consistent, but I'm not sure that's going to happen ..

GarethCabournDavies: I will have a look at how to implement that. Maybe the `pip install pycbc[igwn]`-style mechanism may be useful there?

Note: I found that LISA constants for all the bits I could find that were defined by astropy calls straight through anyway

WuShichao: > @ahnitz While I agree that it is better in general to use `astropy` for constants to be consistent, both of the main GW collaborations (IGWN+3G (lumping these together)) and LISA, define constants separately from `astropy`, and sometimes this can matter. (Cosmology choices tend to be the most obvious gotcha, but LISA positioning might also be quite sensitive as well).
> 
> I think it would be good to consider a mechanism where `astropy` is the default, but you should be able to say "I want constants from LAL" or "I want constants from `lisa_constants`" ... We could also try to persuade both those communities to be more consistent, but I'm not sure that's going to happen ..

I'm also thinking about using the latest `lisa_constants` for `Mojito Light` in PyCBC.

ahnitz: @spxiwh I think I largely agree. That makes sense only when the constants are unique / specialized / specific variations (as you say cosmology is a good example or things specific to an actual detector).

There should be a preference here though to get constants from more standard sources when they do exist. There are a lot of cases where the constants aren't even different numerically and in those cases we should just switch to keep the set of nonstandard constants limited.

pycbc/constants.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+import numpy as np
+from astropy import (
+    constants as aconstants,
+    units as aunits
+)
+# first, do mappings for constants which the fallback is directly in astropy/numpy
+# All are in SI units
+_FALLBACK_MAPPING = {
```
ahnitz: @GarethCabournDavies @spxiwh 

In fact, almost everything in this mapping is something we should consider just not using lal for. A few of these may require checking (earth_si, I could imagine some specific value was chosen that is not known to different precision). Most of these though don't need to come from lal in any situation I can imagine. 

I'd even prefer we don't have fallbacks (at least not without understanding the potential use case a bit more), because that means there could be unexpected behavior if say lal is or is not the source.

GarethCabournDavies:
I'm going to close this in favour of opening separate PRs for:
- the constants change
- the change to time conversions
- the array epoch type change, which is a lot more invasive and is causing headaches in terms of testing

PR #5218, opened 2025-11-03
add scale option to bank compression along with higher order interpolation functions (2nd, 3rd, and 4th order). 

ahnitz:


ahnitz:
This means that I'd suggest using the highest order interpolant for long signals. I haven't tested with very short ones to see where the balance there lies. It's possible lower order is fine for that. 

The main purpose here is to minimize the number of points required to extrapolate each template. For the TaylorF2 case with a  low mass signal we are observing the following.

linear ~ 400 points
quadratic ~ 100 points
quartic ~ 60 points. 

This yields *storage* savings for large template banks. The reconstruction cost appears to be within ~ 10% of the original inline linear method (~ 2.8 vs 2.5 ms in this example).

PR #5226, opened 2025-11-14
Changes run_dir in pycbc_make_offline_grb_workflow to allow proper output-dir config

ETVincent:

This is a bug fix for gitlab issue described here: https://git.ligo.org/pycbc/pygrb-offline-analysis/-/issues/48 wherein the method through which pycbc_make_offline_grb_workflow wrote the output directory was hard coded to be just the GRB##### tag. It will now (hopefully) be whatever is entered as the output-dir config. 

This change affects: PyGRB

This change changes: directory location for PyGRB run

This change should not break current functionality

## Motivation

The PyGRB search attempted to use an --output ${GRB} argument that caused issues. The name of the output directory never changed. Hopefully this solves that. 

https://git.ligo.org/pycbc/pygrb-offline-analysis/-/issues/48

### DIFF
--- a/bin/pygrb/pycbc_make_offline_grb_workflow
+++ b/bin/pygrb/pycbc_make_offline_grb_workflow
@@ -54,10 +54,11 @@ logging.info("Generating %s workflow" % workflow_name)
 
 # Setup run directory
 triggername = str(wflow.cp.get("workflow", "trigger-name"))
-logging.info("Workflow will be generated in %s" % args.output_dir)
-if not os.path.exists(args.output_dir):
-    os.makedirs(args.output_dir)
-os.chdir(args.output_dir)
+output_dir = os.path.abspath(args.output_dir)
+logging.info("Workflow will be generated in %s", output_dir)
+if not os.path.exists(output_dir):
+    os.makedirs(output_dir)
+os.chdir(output_dir)
 
 # SEGMENTS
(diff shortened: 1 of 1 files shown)


titodalcanton (changes requested):
Based on what various people suggested, I think the scope of this change should be a bit broader than it currently is.
1. Drop lines 56-59.
2. Drop line 61 (what is currently being changed).
3. Replace every occurrence of `run_dir` with `args.output_dir`.
4. In line 353, replace `base_dir` with `args.output_dir`.

This will remove the need to have the config option `workflow:output-directory`, on the other hand it will require `--output-dir` to explicitly include the trigger name (if that is the convention we choose, but we are now able to choose a different one if wanted).

We will need to make a test workflow after these changes, as I am not 100% sure the logic in line 353 will necessarily work like this.

pracchia:
I tested this branch and the `--output-dir ${folder_name}` argument for `pycbc_make_offline_grb_workflow ` is not pointing to an absolute path but to the relative path in the folder where the command is run, giving a traceback.

The workflow runs by setting the argument as`--output-dir ${PWD}/${folder_name}`

Would it be better to have `--output-dir` to use an absolute path?

titodalcanton:
@pracchia tested my latest commit and it generates a workflow even with `--output-dir XYZ`, so I am happy to merge.

PR #5227, opened 2025-11-20
Use time-domain waveforms for bank generation

Kanchan-05:
## A support for generating template banks using time-domain waveforms

This feature extends pycbc_brute_bank to allow users to generate template banks using time-domain (TD) waveform models. 

This change changes: documentation, scientific output

## Motivation

The main motivation for this feature is that several waveform families either lack faster frequency-domain implementations or are substantially more accurate when evaluated in the time domain. Because the pycbc_brute_bank code computes matches in the frequency domain, TD waveforms must be converted before overlaps can be calculated. Using TD waveforms requires an additional step where the generated time-series waveform is converted to the frequency domain using the get_fd_waveform_from_td function in waveform.py, before the match calculation is done. This step conversion step introduces additional computational cost, and in some cases (eg. SEOBNRv5HM) may dominate the overall runtime of the bank generation process.  

This PR implements the TD-to-FD conversion internally using a straightforward FFT-based calculation. 

### DIFF
--- a/bin/bank/pycbc_brute_bank
+++ b/bin/bank/pycbc_brute_bank
@@ -53,6 +53,8 @@ parser.add_argument('--approximant',  required=False,
 parser.add_argument('--minimal-match', default=0.97, type=float)
 parser.add_argument('--buffer-length', default=2, type=float,
     help='size of waveform buffer in seconds')
+parser.add_argument('--use-td-waveform', action='store_true',
+    help='Generate waveform in the time domain (default is frequency domain).')
 parser.add_argument('--full-resolution-buffer-length', default=None, type=float,
     help='Size of the waveform buffer in seconds for generating time-domain signals at full resolution before conversion to the frequency domain.')
 parser.add_argument('--max-signal-length', type= float,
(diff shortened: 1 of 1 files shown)


bin/bank/pycbc_brute_bank
```diff
@@ -53,6 +53,8 @@ parser.add_argument('--approximant',  required=False,
 parser.add_argument('--minimal-match', default=0.97, type=float)
 parser.add_argument('--buffer-length', default=2, type=float,
     help='size of waveform buffer in seconds')
+parser.add_argument('--use-td-waveform', action='store_true',
```
kkacanja, line 56: I think it would be better to remove the option and have the code handle the waveform generation internally. So if the approximant is TD use the time domain waveform, if the approximant is in fd_approximants use the frequency domain handling.

Kanchan-05: @kkacanja The pycbc_brute_bank code already handles this automatically. The --use-td-waveform option would simply force the code to use time-domain waveforms exclusively, which can be useful for faster bank generation.

PR #5235, opened 2025-11-28
Make changes to constants, use numpy/astropy but use LAL when requested

GarethCabournDavies:
Use numpy / astropy constants instead of lal. Allow use of lal constants if a environment variable is set.

This is a change to the source of truth for certain values
This change affects: the offline search, the live search, inference, PyGRB (but shouldn't actually affect anything if values are the same)
This change changes: documentation, result presentation / plotting, scientific output (but shouldn't actually affect anything if values are the same)

## Motivation
I am worried about bloat of the pycbc package, and removing lal dependency (when we will use it for non-LIGO searches in future) should make the package significantly smaller

## Contents
- add a new pycbc/constants.py module which sets constants according to numpy/astropy but can be changed for lal constants if wanted
- add a test so that these can be compared

This was part of #5217 but was getting to be hard to manage along with with the changes to the epoch

## Testing performed
Testing according to the CI should pass. all constants are currently with np.isclose between numpy/astropy and lal

### DIFF
--- a/pycbc/constants.py
+++ b/pycbc/constants.py
@@ -81,7 +81,7 @@ def get_constant(name):
     NotImplementedError
         If the constant is not found in any of the available packages.
     """
-    if _CONSTANTS == 'lal': # Allow lal, or LAL, or whatever
+    if _CONSTANTS.lower() == 'lal': # Allow LAL
         if lal is None:
             raise ImportError(
                 "PYCBC_CONSTANT_SOURCE is set to 'lal', but the 'lal' module is not installed. "
--- a/pycbc/waveform/SpinTaylorF2.py
+++ b/pycbc/waveform/SpinTaylorF2.py
@@ -355,7 +355,7 @@ def spintaylorf2(**kwds):
     IM_SBfac3= -2.*sin(thetaJ)
     IM_SBfac4= cos(thetaJ)
 
-    #####Calculate the PN terms # FIXME replace with functions in lalsimulation #####
+    #####Calculate the PN terms#####
     theta = -11831./9240.
     lambdaa = -1987./3080.0
     pfaN = 3.0/(128.0 * eta)


pycbc/constants.py
```diff
@@ -0,0 +1,102 @@
...
+    -------
+    float or object
+        The value of the constant.
+
+    Raises
+    ------
+    NotImplementedError
+        If the constant is not found in any of the available packages.
+    """
+    if _CONSTANTS == 'lal': # Allow lal, or LAL, or whatever
```
spxiwh: `LAL` ... Change to `# Allow LAL`

pycbc/constants.py
```diff
@@ -0,0 +1,102 @@
...
+        return _DEFAULT_MAPPING[name]
+
+    raise NotImplementedError(
+        'Contact the PyCBC team, this should never happen. You are '
+        'trying to use a constant which is not defined.'
+    )
+
+# Expose the constants as attributes of this module:
+for const_name in _DEFAULT_MAPPING:
+    globals()[const_name] = get_constant(const_name)
```
spxiwh, line 102: I'm assuming that all this is evaluated at initial import of this file? There's no issue with repeated function calls every time I need a low-level constant?

pycbc/vetoes/chisq_cupy.py
```diff
@@ -25,11 +25,12 @@
...
 import numpy
 import cupy as cp
-import lal
 from mako.template import Template
 
+import pycbc.constants
+
 LALARGS = {
-    'TWOPI': lal.TWOPI,
+    'TWOPI': pycbc.constants.TWOPI,
```
spxiwh: This *must* be a constant at the point this file is imported? I think that's what happens (see question in constants.py), but if it's not it will be an issue in this kernel.

pycbc/waveform/pycbc_phenomC_tmplt.py
```diff
@@ -16,18 +16,17 @@
 #  MA  02111-1307  USA
```
spxiwh, line 16: Is there a reason for this module to still exist? PhenomC is now outdated, this is not how we do GPU waveforms for templates ...

spxiwh: (Doesn't need to be resolved here, just wondering out loud)

pycbc/waveform/spa_tmplt_cuda.py
```diff
@@ -66,8 +64,8 @@
             break;
     }
     phasing *= pfaN / v5;
-    phasing -=  LAL_PI_4;
-    phasing -= int(phasing / (LAL_TWOPI)) * LAL_TWOPI;
+    phasing -= (PI/4.0);
+    phasing -= int(phasing / (2.0*PI)) * 2.0*PI;
```
spxiwh: It might make more sense for PI to be replaced with the actual constant *before* compilation, rather than leaving it as a variable ... I'm hoping PI doesn't vary within a single executable!

spxiwh: Actually this is the pyCUDA code and I think you do what I ask for already in the cupy code, so I'm happy (though why did you choose to do this differently in the two cases??)

pycbc/waveform/pycbc_phenomC_tmplt.py
spxiwh: I think you did miss the same thing in `SpinTaylorF2.py` though ... Again though that one might well be better removed at this point.

spxiwh (approved):
I think this mostly does what it says. I have a few questions, but mostly it looks good to me.

Are there no Cython codes doing c-imports of LAL constants? `spa_tmplt_cpu.pyx` does seem to import lal, but then uses libc.math for `PI` and doesn't use lal at all.

pycbc/constants.py
GarethCabournDavies: I think this was from when I was doing a `.lower()` on the string, now it just looks passive-aggressive 😟

pycbc/constants.py
GarethCabournDavies: unless someone monkey-patches, I don't think its possible that these will change.

For safety, I will check over where pycbc.constants values are used and move that to the import at the start of the file. At least then the value won't change for us

pycbc/waveform/pycbc_phenomC_tmplt.py
GarethCabournDavies: I don't think I missed anything in SpinTaylorF2?

pycbc/constants.py
GarethCabournDavies: Updated now

pycbc/constants.py
GarethCabournDavies: checked, and where you flagged it in chisq_cupy was actually the only usage

pycbc/waveform/spa_tmplt_cuda.py
GarethCabournDavies: Fixed now. I think it was just that this wasn't done in one go, so my style changed in the interim!

PR #5236, opened 2025-11-28
Change time conversion routines to use astropy rather than lal

GarethCabournDavies:
In a series of PRs for removing lal dependency in pycbc, this one is probably the most straightforward, and uses astropy time routines where lal was used before.

This probably should never have used lal before, but was just a bit easier / what we were used to

This is a: change for reducing reliance on lal
This change affects: the offline search, the live search, PyGRB
This change changes: nothing except which dependency is used

## Motivation
Reducing dependence on lal

## Contents
Change time conversion routines to use astropy rather than lal. This is mainly changing the gps_to_utc function to one which uses satrapy's Time object

split off from #5217 

## Testing performed
Used ipython to check that the routines give same output the same given the same input

### DIFF
--- a/bin/pygrb/pycbc_pygrb_minifollowups
+++ b/bin/pygrb/pycbc_pygrb_minifollowups
@@ -27,6 +27,7 @@ import os
 import argparse
 import logging
 
+from igwn_segments import segmentlist
 import pycbc.workflow as wf
 from pycbc.workflow.core import FileList, resolve_url_to_file
 import pycbc.workflow.minifollowups as mini
@@ -181,11 +182,17 @@ if veto_file:
     veto_file = os.path.join(start_rundir, args.veto_file)
     veto_file = wf.resolve_url_to_file(veto_file)
 
+seg_file_abs_paths = [os.path.join(start_rundir, f) for f in args.seg_files]
+
 # Convert the segments files to a FileList
-seg_files = wf.FileList([
-    wf.resolve_url_to_file(os.path.join(start_rundir, f))
-    for f in args.seg_files
-])
+seg_files = wf.FileList(
+    [wf.resolve_url_to_file(f) for f in seg_file_abs_paths]
+)
+
+# The Q-scan jobs need to be told the valid science segments to deal with
+# triggers near the boundaries correctly. We use the offsource segment as a
+# proxy for the valid science segments, and ignore vetoed time inside it.
+valid_segs = segmentlist([ppu._read_seg_files(seg_file_abs_paths)['off']])
 
 # (Loudest) off/on-source events are on time-slid data so the
 # try will succeed, as it finds the time shift columns.
@@ -220,7 +227,8 @@ for num_event in range(num_events):
         for ifo in ifos:
             files += mini.make_qscan_plot(workflow, ifo, gps_time,
                                           args.output_dir,
-                                          tags=tags)
+                                          tags=tags,
+                                          data_segments=valid_segs)
     # Handle off/on-source loudest triggers follow-up (which may be on slid
     # data in the case of the off-source)
     else:
@@ -235,7 +243,8 @@ for num_event in range(num_events):
                                           tags=tags)
             files += mini.make_qscan_plot(workflow, ifo, ifo_time,
                                           args.output_dir,
-                                          tags=tags)
+                                          tags=tags,
+                                          data_segments=valid_segs)
 
     layouts += list(layout.grouper(files, 2))
 
--- a/pycbc/workflow/minifollowups.py
+++ b/pycbc/workflow/minifollowups.py
@@ -793,7 +793,7 @@ def make_qscan_plot(workflow, ifo, trig_time, out_dir, injection_file=None,
                     data_segments=None, time_window=100, tags=None):
     """ Generate a make_qscan node and add it to workflow.
 
-    This function generates a single node of the singles_timefreq executable
+    This function generates a single node of the plot_qscan executable
     and adds it to the current workflow. Parent/child relationships are set by
     the input/output files automatically.
 
@@ -812,7 +812,7 @@ def make_qscan_plot(workflow, ifo, trig_time, out_dir, injection_file=None,
         plot.
     data_segments: igwn_segments.segmentlist (optional, default=None)
         The list of segments for which data exists and can be read in. If given
-        the start/end times given to singles_timefreq will be adjusted if
+        the start/end times given to plot_qscan will be adjusted if
         [trig_time - time_window, trig_time + time_window] does not completely
         lie within a valid data segment. A ValueError will be raised if the
         trig_time is not within a valid segment, or if it is not possible to
@@ -820,7 +820,7 @@ def make_qscan_plot(workflow, ifo, trig_time, out_dir, injection_file=None,
         trigger. This **must** be coalesced.
     time_window: int (optional, default=None)
         The amount of data (not including padding) that will be read in by the
-        singles_timefreq job. The default value of 100s should be fine for most
+        plot_qscan job. The default value of 100s should be fine for most
         cases.
     tags: list (optional, default=None)
         List of tags to add to the created nodes, which determine file naming.
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
 # requirements for most basic library use
 astropy>=2.0.3,!=4.2.1,!=4.0.5
 Mako>=1.0.1
-scipy>=0.16.0
+scipy>=0.16.0,<1.17.0
 matplotlib>=2.0.0
-numpy>=1.16.0,!=1.19.0,!=2.2.2
+numpy>=1.16.0,!=1.19.0,!=2.2.2,<2.4
 pillow
 h5py>=3.0.0,!=3.7.0
 jinja2
--- a/setup.py
+++ b/setup.py
@@ -32,8 +32,8 @@ requires = []
 setup_requires = ['numpy>=1.16.0']
 install_requires = setup_requires + [
     'cython>=0.29',
-    'numpy>=1.16.0,!=1.19.0,!=2.2.2',
-    'scipy>=0.16.0',
+    'numpy>=1.16.0,!=1.19.0,!=2.2.2,<2.4',
+    'scipy>=0.16.0,<1.17.0',
     'astropy>=2.0.3,!=4.2.1,!=4.0.5',
     'matplotlib>=1.5.1',
     'mpld3>=0.3',


pycbc/vetoes/chisq_cupy.py
titodalcanton, line 1: This change seems unrelated, revert?

pycbc/live/__init__.py
titodalcanton, line 1: This should be reverted.

pycbc/time.py
```diff
@@ -0,0 +1,148 @@
+"""
+Module to contain time conversions use in pycbc
```
titodalcanton: use -> used

pycbc/time.py
```diff
@@ -0,0 +1,148 @@
...
+        The datetime object to strip the time from.
+
+    Returns
+    -------
+    datetime
+    """
+    return date.replace(hour=0, minute=0, second=0, microsecond=0)
+
+
+def strip_time_from_gps(gps, format="%Y-%m-%d %H:%M:%S"):
```
titodalcanton: The `format` argument is not used, should it be? Or should it be removed?

pycbc/time.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    Return the GPS time at the start of the day.
```
titodalcanton:
```suggestion
    Round a GPS time to the midnight of the same UTC day, and return the date of that day as a string.
```

pycbc/time.py
```diff
@@ -0,0 +1,148 @@
...
+
+    Parameters
+    ----------
+    gps : float
+        The GPS time to strip the time from.
+
+    Returns
+    -------
+    float
```
titodalcanton:
```suggestion
    float, str
```

test/test_time.py
titodalcanton, line 1: As a general comment on this file, I think having Astropy compute the expected answer is not very useful; what if a bug is introduced in a new Astropy release? I would rather hardcode the expected answer directly, e.g. verify that 1451917960 translates to 2026-01-08T14:32:22+0000 and that 2026-01-08T14:32:22+0000 translates to 1451917960, and do this for a small list of times across the entire extent from the GPS epoch to now, keeping the list hardcoded. For the "true" conversions I would use something like https://heasarc.gsfc.nasa.gov/cgi-bin/Tools/xTime/xTime.pl or just `lal_tconvert`, but I would avoid Astropy for the same reason as above.

test/test_time.py
```diff
@@ -0,0 +1,126 @@
...
+
+    def test_datetime_to_str_formats(self):
+        from datetime import datetime
+        dt = datetime(2020, 2, 29, 15, 4, 5)
+        self.assertEqual(pycbc.time.datetime_to_str(dt, format="%Y-%m-%d"), "2020-02-29")
+        self.assertEqual(pycbc.time.datetime_to_str(dt, format="%H:%M:%S"), "15:04:05")
+
+    def test_gps_now_and_gmst(self):
+        # gps_now should be close to astropy Time.now().gps
+        now_astropy = Time.now()
```
titodalcanton: Again this is not a very powerful test because `gps_now()` just does exactly the same thing that the test does (so if Astropy breaks, the test will still pass). You could instead convert `gps_now` to a datetime using the new APIs, and then compare the result to `datetime.datetime.utcnow()`.

test/test_time.py
```diff
@@ -0,0 +1,126 @@
...
+        gps_now = pycbc.time.gps_now()
+        # Allow a couple of seconds tolerance for the call overhead
+        self.assertAlmostEqual(gps_now, float(now_astropy.gps), delta=2.0)
+
+        # gmst_accurate should match astropy sidereal_time('mean').rad
+        gps_sample = Time('2020-01-01T00:00:00', format='isot', scale='utc').gps
+        gmst_expected = Time(gps_sample, format='gps', scale='utc', location=(0, 0)).sidereal_time('mean').rad
+        gmst = pycbc.time.gmst_accurate(gps_sample)
+        # compare within a small tolerance (radians)
+        self.assertAlmostEqual(gmst, gmst_expected, places=9)
```
titodalcanton: I would make this a separate test method from `test_gps_now()`. Going back to what I wrote at the top, I suggest using a couple different times and hardcoding the expected results from an independent calculation (e.g. https://www.iditect.com/tools/sidereal-time.html) rather than using Astropy for the truth.

titodalcanton (changes requested):
Thanks, here are a few more requests.

Note that the tests broke because `gps_now()` has moved. This is effectively an API breakage, so make sure this is noted in the top description.

pycbc/time.py
GarethCabournDavies: Good catch!

pycbc/vetoes/chisq_cupy.py
GarethCabournDavies: yes, not sure how that one happened

test/test_time.py
GarethCabournDavies: Updated to use datetime.now() (utcnow is deprecated)

test/test_time.py
GarethCabournDavies: done

test/test_time.py
GarethCabournDavies: I ended up using lal_tconvert in the end as apparently that also does GMST conversions

test/test_time.py
```diff
@@ -0,0 +1,182 @@
...
+    def test_midnight_is_before_or_equal_and_within_day(self):
+        gps = 1262304000 + 3600 * 12  # midday
+        midnight, _ = pycbc.time.strip_time_from_gps(gps)
+        # midnight should be <= provided time and within 24 hours difference
+        self.assertLessEqual(midnight, gps)
+        self.assertLess(gps - midnight, 86400)
+
+# Truth conversions for certain times around leap seconds.
+# These are converted using lal_tconvert,
+# rather than astropy conversions used internally
```
titodalcanton, line 43: Worth adding a comment to explain what the three values mean, i.e. `{utc_string: (gps_seconds, gmst)}`.

You can also drop all the trailing zeroes from GPSs and GMSTs.

test/test_time.py
```diff
@@ -0,0 +1,182 @@
...
+    '2005-12-31 16:17:18': (820081051.0, 6.012923000000),
+    '2006-01-01 17:18:19': (820171113.0, 0.013905000000),
+    '2008-12-31 19:20:21': (914786435.0, 0.535335000000),
+    '2009-01-01 20:21:22': (914876497.0, 0.819502000000),
+    '2012-06-30 22:23:24': (1025130219.0, 4.454637000000),
+    '2012-07-01 23:24:25': (1025220281.0, 4.738804000000),
+    '2015-06-30 01:02:03': (1119661339.0, 5.119072000000),
+    '2015-07-01 02:03:04': (1119751401.0, 5.403239000000),
+    '2016-12-31 04:05:06': (1167192323.0, 2.815130000000),
+    '2017-01-01 05:06:07': (1167282385.0, 3.099297000000),
```
titodalcanton: Add a datetime in 2026 as well just for good measure. In fact, is there any reason not to test dates ~10 years in the future? Might be useful for LISA/3G.

test/test_time.py
```diff
@@ -0,0 +1,182 @@
...
+        gps_now = pycbc.time.gps_now()
+        dt_from_gps = pycbc.time.gps_to_utc_datetime(gps_now)
+        # Ensure dt_from_gps is timezone-aware in UTC for a fair comparison
+        if dt_from_gps.tzinfo is None:
+            dt_from_gps = dt_from_gps.replace(tzinfo=timezone.utc)
+
+        now_dt = datetime.now(timezone.utc)
+        # difference in seconds
+        delta_secs = abs((dt_from_gps - now_dt).total_seconds())
+        self.assertLessEqual(delta_secs, 2.0)
```
titodalcanton: I bet we can make the tolerance 1 s, what do you think? This should protect against incorrect handling of one leap second. Maybe move the calls to `gps_now()` and `datetime.now()` next to each other to minimize the delay.

titodalcanton (changes requested):
A few last minor things remaining from me.

test/test_time.py
GarethCabournDavies: My reason for setting this so low was to actually ensure correct handling of leap seconds

test/test_time.py
GarethCabournDavies: Will do - it will mean that if a leap second is added, we will need to update the truth list though

titodalcanton: Yup, I bet we will not have a leap second [any time soon](https://en.wikipedia.org/wiki/Leap_second#/media/File:Leapsecond.ut1-utc.svg) but it would be a good reminder when we do!

GarethCabournDavies: I added a date in 2039, which passes, however this does cause warnings as I don't think the astropy data is accurate after that

test/test_time.py
titodalcanton: Meaning have the test pass even when run a few seconds away from a leap second? Hmm ok, though I think routinely verifying <=1 s accuracy far from leap seconds is generally more important… Up to you.

test/test_time.py
spxiwh:
```
 NO leap second will be introduced at the end of June 2026.
 The difference between Coordinated Universal Time UTC and the
 International Atomic Time TAI is :

     from 2017 January 1, 0h UTC, until further notice : UTC-TAI = -37 s

 Leap seconds can be introduced in UTC at the end of the months of December
 or June,  depending on the evolution of UT1-TAI. Bulletin C is mailed every
 six months, either to announce a time step in UTC, or to confirm that there
 will be no time step at the next possible date.


                                            Christian BIZOUARD
                                            Director
                                            Earth Orientation Center of IERS
                                            Observatoire de Paris, France
```

test/test_time.py
GarethCabournDavies: sorry I missed which part this was on - yes we definitely need to reduce it!

test/test_time.py
```diff
@@ -0,0 +1,186 @@
...
+        """confirm gps_now converts to a UTC datetime close to the system UTC
+        time"""
+
+        gps_now = pycbc.time.gps_now()
+        now_dt = datetime.now(timezone.utc)
+
+        dt_from_gps = pycbc.time.gps_to_utc_datetime(gps_now)
+        # Ensure dt_from_gps is timezone-aware in UTC for a fair comparison
+        if dt_from_gps.tzinfo is None:
+            dt_from_gps = dt_from_gps.replace(tzinfo=timezone.utc)
```
titodalcanton: Is this not already ensured in `gps_to_utc_datetime()`?

GarethCabournDavies: you are right - this is a hangover from when `gps_to_utc_datetime` did not ensure that

PR #5197, opened 2025-10-07
Work towards removing `pycbc_submit_dax`

GarethCabournDavies:
Here I am working towards getting rid of `pycbc_submit_dax`. I have included the commits from #4945 (so that @alishahassan gets credit).

This is work in progress, but I am trying to make it so that it can be merged any time

## Standard information about the request

This is the start of removing an unwanted code
This change affects: the offline search, inference, (will affect) PyGRB

This change (will) change documentation eventually
This change has been proposed using the [contribution guidelines](https://github.com/gwastro/pycbc/blob/master/CONTRIBUTING.md)

## Contents
Finding any references to `pycbc_submit_dax` and fixing them as appropriate. I wanted to get the PR in sooner rather than later so that the CI tests can run nicely

## Links to any issues or associated PRs
#4945 #5080

- [x] The author of this pull request confirms they will adhere to the [code of conduct](https://github.com/gwastro/pycbc/blob/master/CODE_OF_CONDUCT.md)

PR #5237, opened 2025-11-28
Change timeseries, frequencyseries epoch to be float64 rather than ligotimegps

GarethCabournDavies:
In the series of removing lal from pycbc, here I have made it so that the epoch of the array, timeseries, and frequencyseries types use float64 rather than ligotimegps

This is a: change to a core data type
This change changes: hopefully nothing

## Motivation
Reduce dependency on LAL

## Contents
Change the epoch data type to float64, allowing LIGOTimeGPS to be used to set this if wanted

Split off from #5217 

## Testing performed
CI tests should work to test this, I will mark as draft so that I can make sure they all pass

### DIFF
--- a/pycbc/types/frequencyseries.py
+++ b/pycbc/types/frequencyseries.py
@@ -22,6 +22,7 @@ import h5py
 import numpy as _numpy
 
 from pycbc.types.array import Array, _convert, zeros, _noreal
+from pycbc.types.utils import determine_epoch
 from pycbc.types import float64
 from pycbc.libutils import import_optional
 
@@ -54,35 +55,10 @@ class FrequencySeries(Array):
                 raise TypeError('must provide either an initial_array with a delta_f attribute, or a value for delta_f')
         if not delta_f > 0:
             raise ValueError('delta_f must be a positive number')
-        # We gave a nonsensical default value to epoch so we can test if it's been set.
-        # If the user passes in an initial_array that has an 'epoch' attribute and doesn't
-        # pass in a value of epoch, then our new object's epoch comes from initial_array.
-        # But if the user passed in a value---even 'None'---that will take precedence over
-        # anything set in initial_array.  Finally, if the user passes in something without
-        # an epoch attribute *and* doesn't pass in a value of epoch, it becomes 'None'
-
-        # If epoch is already a float64, we can directly add it, so
-        # don't do this conversion
-        if not isinstance(epoch, float64):
-            if epoch == "":
-                # epoch is not given
-                try:
-                    # inherit epoch from initial array
-                    epoch = initial_array._epoch
-                except AttributeError:
-                    # nothing given, and can't grab the epoch from initial
-                    # array - fall back to zero
-                    epoch = float64(0)
-            elif epoch is not None: # If it is passed None, we do allow this
-                # epoch is given but is not already a float64 - convert it
-                is_ltg = _lal is not None and isinstance(epoch, _lal.LIGOTimeGPS)
-                if not is_ltg and not _numpy.isscalar(epoch):
-                    raise TypeError("epoch must be a number, not array-like")
-                epoch = float64(epoch)
 
         Array.__init__(self, initial_array, dtype=dtype, copy=copy)
         self._delta_f = delta_f
-        self._epoch = epoch
+        self._epoch = determine_epoch(epoch, initial_array)
 
     def _return(self, ary):
         return FrequencySeries(ary, self._delta_f, epoch=self._epoch, copy=False)
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -24,6 +24,7 @@ import numpy as _numpy
 from scipy.io.wavfile import write as write_wav
 
 from pycbc.types.array import Array, _convert, complex_same_precision_as, zeros
+from pycbc.types.utils import determine_epoch
 from pycbc.types.array import _nocomplex
 from pycbc.types.frequencyseries import FrequencySeries
 from pycbc.types import float32, float64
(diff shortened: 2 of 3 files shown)


GarethCabournDavies:
This has loads of stuff that Im doing for debugging, that will be removed when I have worked out what is up, so isn't ready for review yet

pycbc/results/versioning.py
```diff
@@ -18,10 +18,13 @@
 import subprocess
 import urllib.parse
 
-import lal
-import lalframe
-
 import pycbc.version
+from pycbc.libutils import import_optional
+
+lal = import_optional('lal')
```
GarethCabournDavies, line 24: These changes maybe don't belong in here, but its as good a place as any I think

pycbc/tmpltbank/lattice_utils.py
```diff
@@ -121,12 +120,16 @@ def generate_anstar_3d_lattice(maxv1, minv1, maxv2, minv2, maxv3, minv3, \
     v3s : numpy.array
         Array of positions in the second dimension
     """
-    # Lalpulsar not a requirement for the rest of pycbc, so check if we have it
+    # Lal/Lalpulsar are not a requirement for the rest of pycbc, so check if we have it
```
GarethCabournDavies, line 123: Again not really part of the epoch changes, but just bundled in

pycbc/types/array.py
```diff
@@ -33,14 +33,17 @@
...
 import numpy as _numpy
 from numpy import float32, float64, complex64, complex128, ones
 from numpy.linalg import norm
 
 import pycbc.scheme as _scheme
 from pycbc.scheme import schemed, cpuonly
 from pycbc.opt import LimitedSizeDict
+from pycbc.libutils import import_optional
+
+_lal = import_optional('lal')
```
GarethCabournDavies, line 46: This is only used in the lal() method, so its fairly safe to assume that someone using that would want lal installed!

pycbc/types/frequencyseries.py
```diff
@@ -50,30 +54,26 @@ def __init__(self, initial_array, delta_f=None, epoch="", dtype=None, copy=True)
...
-                    else:
-                        epoch = _lal.LIGOTimeGPS(epoch)
-                except:
-                    raise TypeError('epoch must be either None or a lal.LIGOTimeGPS')
+                    # inherit epoch from initial array
+                    epoch = initial_array._epoch
+                except AttributeError:
+                    # nothing given, and can't grab the epoch from initial
+                    # array - fall back to zero
+                    epoch = float64(0)
```
GarethCabournDavies: I'm not 100% sure what the default should be. Was None intended as the "epoch doesn't matter" signifier before? Maybe we should fall back to that

GarethCabournDavies: This is also repeated pretty much verbatim for both frequencyseries and timeseries, maybe there should be a function in array.py or similar?

test/fft_base.py
```diff
@@ -467,7 +471,7 @@ def setUp(self):
         self.in_c2c_rev = [3.0-1.0j,-1.0+3.0j]
         self.out_c2c_rev = [2.0+2.0j,4.0-4.0j]
         # For Time/FrequencySeries, we want to test with a non-trivial epoch
-        self.epoch = LTG(3,4)
+        self.epoch = 3 + 4 * 1e-9
```
GarethCabournDavies, line 474: This is the only change in the file that is actually related to the epoch change - the rest is general cleanup

GarethCabournDavies:
I think that after this and #5197, we should bump the minor release number as it will significantly change how we interact with the code

pycbc/filter/matchedfilter.py
```diff
@@ -240,7 +240,7 @@ def __init__(self, low_frequency_cutoff, high_frequency_cutoff, snr_threshold, t
         else:
             raise ValueError("Invalid downsample factor")
 
-    def full_matched_filter_and_cluster_symm(self, segnum, template_norm, window, epoch=None):
```
ahnitz, line 243: What's the reason for this particular change?

pycbc/types/array.py
ahnitz: Seems reasonable here.

pycbc/types/frequencyseries.py
```diff
@@ -50,30 +54,26 @@ def __init__(self, initial_array, delta_f=None, epoch="", dtype=None, copy=True)
...
-        # pass in a value of epoch, then our new object's epoch comes from initial_array.
-        # But if the user passed in a value---even 'None'---that will take precedence over
-        # anything set in initial_array.  Finally, if the user passes in something without
-        # an epoch attribute *and* doesn't pass in a value of epoch, it becomes 'None'
-        if not isinstance(epoch,_lal.LIGOTimeGPS):
+
+        # If epoch is already a float64, we can directly add it, so
+        # don't do this conversion
+        if not isinstance(epoch, float64):
             if epoch == "":
```
ahnitz: I'm not so sure about the "" thing. Why is this chosen rather than None?

pycbc/types/frequencyseries.py
```diff
@@ -334,7 +335,7 @@ def lal(self):
             LAL frequency series object containing the same data as self.
             The actual type depends on the sample's dtype. If the epoch of
             self was 'None', the epoch of the returned LAL object will be
-            LIGOTimeGPS(0,0); otherwise, the same as that of self.
+            LIGOTimeGPS(0,0); otherwise, convert it to LIGOTimeGPS
```
ahnitz, line 320: Do we want these functions to not be float64 too though? Is that a later change, or what is the thinking here?

GarethCabournDavies: In this case the epoch needs converting as the returned object is a LAL FrequencySeries, which uses LIGOTimeGPS for the epoch

pycbc/types/frequencyseries.py
GarethCabournDavies: I was trying to mimic current behaviour, introduce in 4eb823f.

I will add back in the comment that I removed above

pycbc/types/frequencyseries.py
```diff
@@ -50,30 +54,26 @@ def __init__(self, initial_array, delta_f=None, epoch="", dtype=None, copy=True)
...
         if not delta_f > 0:
             raise ValueError('delta_f must be a positive number')
-        # We gave a nonsensical default value to epoch so we can test if it's been set.
-        # If the user passes in an initial_array that has an 'epoch' attribute and doesn't
-        # pass in a value of epoch, then our new object's epoch comes from initial_array.
-        # But if the user passed in a value---even 'None'---that will take precedence over
-        # anything set in initial_array.  Finally, if the user passes in something without
-        # an epoch attribute *and* doesn't pass in a value of epoch, it becomes 'None'
-        if not isinstance(epoch,_lal.LIGOTimeGPS):
+
```
GarethCabournDavies:
```suggestion
        # We gave a nonsensical default value to epoch so we can test if it's been set.
        # If the user passes in an initial_array that has an 'epoch' attribute and doesn't
        # pass in a value of epoch, then our new object's epoch comes from initial_array.
        # But if the user passed in a value---even 'None'---that will take precedence over
        # anything set in initial_array.  Finally, if the user passes in something without
        # an epoch attribute *and* doesn't pass in a value of epoch, it becomes 'None'

```

pycbc/filter/matchedfilter.py
GarethCabournDavies: Change no longer being made - there was some logic in the epoch calculation which needed updated

pycbc/types/frequencyseries.py
GarethCabournDavies: Explanation added in new function

PR #5239, opened 2025-12-03
PyGRB: make sure Q-scans stay within science time

### LINKED ISSUE #5238: PyGRB minifollowup qscans sometimes extend outside science time and fail
See error message in https://git.ligo.org/pycbc/pygrb-offline-analysis/-/issues/31.

titodalcanton:
Q-scan plots in PyGRB minifollowups can occasionally try to go outside of science time and then fail.

This is a bug fix.

This change affects PyGRB.

## Motivation

Problem raised in #5238.

## Contents

`make_qscan_plot()` takes an optional segment list of science segments to check against. This is currently not provided by PyGRB's minifollowup code, so I add it. There is a caveat that I do not have the actual science segments readily available, so I take the "offsource" segment as a proxy for them, ignoring any vetoed time inside it. I expect this to be good enough to avoid the crashes we have seen, but it is probably not ideal.

I also fix the docstring for `make_qscan_plot()` which still refers to a previous executable.

Fixes #5238.

## Testing performed

None yet.

## Additional notes

N/A

### DIFF
--- a/bin/pygrb/pycbc_pygrb_minifollowups
+++ b/bin/pygrb/pycbc_pygrb_minifollowups
@@ -182,19 +182,17 @@ if veto_file:
     veto_file = os.path.join(start_rundir, args.veto_file)
     veto_file = wf.resolve_url_to_file(veto_file)
 
+seg_file_abs_paths = [os.path.join(start_rundir, f) for f in args.seg_files]
+
 # Convert the segments files to a FileList
-seg_file_abs_paths = [
-    wf.resolve_url_to_file(os.path.join(start_rundir, f))
-    for f in args.seg_files
-]
-seg_files = wf.FileList(seg_file_abs_paths)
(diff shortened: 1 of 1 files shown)


titodalcanton:
Test workflow completed at https://ldas-jobs.ligo.caltech.edu/~tito.canton/pycbc/pygrb/test_GRB230904C_qscan_fix/results_webpage/ after fixing a mistake. I will compare it with the one based on 2.8.x next.

titodalcanton:
I did a manual rerun of the `pycbc_pygrb_minifollowups` job in the original run, and then I did a diff of the resulting DAX with respect to the original. Indeed I can see the GPS end times of the Q-scan jobs being shifted earlier, and staying within the science time. So I think this is ready.

This will need a rebase after the failing tests are fixed, however.

PR #5243, opened 2025-12-15
Plot injs mod

sebastiangomezlopez:

This is a: new feature

This change affects: PyGRB

## Motivation

This PR is still a draft. 

It adds further capabilities for the plot_injs_results script related to single ifo effective distances. Additionally adds the "mollweide" projection to display the injections in the sky plane.

### DIFF
--- a/bin/pygrb/pycbc_pygrb_plot_injs_results
+++ b/bin/pygrb/pycbc_pygrb_plot_injs_results
@@ -47,6 +47,9 @@ __program__ = "pycbc_pygrb_plot_injs_results"
 # =============================================================================
 # Functions
 # =============================================================================
+def ra_to_ra_mollweide(ra):
+    return (ra + np.pi) % (2*np.pi) - np.pi
+
 def process_var_strings(qty):
     """Add underscores to match HDF column name conventions"""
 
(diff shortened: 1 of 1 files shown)


bin/pygrb/pycbc_pygrb_plot_injs_results
```diff
@@ -212,7 +216,8 @@ admitted_vars = easy_keys + ['mtotal', 'q', 'mchirp',
                              'cosincl', 'absincl', 'cosabsincl',
                              'sky_error', 'skyerror', 'end_time', 'endtime',
                              'dec', 'ra', 'coaphase', 'coa_phase',
-                             'eff_site_dist', 'eff_dist',
+                             'eff_site_dist', 'eff_dist', 'eff_dist_H1',
+                             'eff_dist_L1', 'eff_dist_V1', 'eff_dist_sqr_quadsum',
```
titodalcanton, line 220: Note that we already have "eff_site_dist" to plot the effective distance for a specific site. What is the motivation for adding the explicit per-site ones?

bin/pygrb/pycbc_pygrb_plot_injs_results
```diff
@@ -448,9 +457,13 @@ axis_labels_dict = {'mchirp': "Chirp Mass (solar masses)",
                     'spin2y': "Spin y-component of 2nd binary component",
                     'spin2z': "Spin z-component of 2nd binary component",
                     'chi_p': "Effective precession spin"}
-
 fig = plt.figure()
-ax = fig.gca()
+if 'dec' in y_qty and 'ra' in x_qty:
+    ax = fig.add_subplot(111, projection="mollweide")
+    missed_inj['ra'] = ra_to_ra_mollweide(missed_inj['ra'])
+    found_inj['ra'] = ra_to_ra_mollweide(found_inj['ra'])
```
titodalcanton, line 464: There are some cases where the all-sky projection plot is detrimental, typically when the input skymap is very precise, as all points are bunched together into a single blob. Perhaps we could have different variable names to control the projection, e.g. "ra"/"dec" and "ra_mollweide"/"dec_mollweide". Another way (probably my preferred one actually) would be to add a command line argument like `--sky-projection mollweide` that is only used when "ra" and "dec" are being plotted.

PR #5250, opened 2026-01-12
Fix numpy scalar array test failures

GarethCabournDavies:
CI failing in a few places - I think these should fix it

This is a bug fix

## Motivation
Get the tests to pass

## Contents
Where numpy scalar arrays are being used, use `.item()` or just don't make them arrays in the first place

Many recent PRs are failing - I won't list them here

## Testing performed
CI tests will do the needed testing

- [x The author of this pull request confirms they will adhere to the [code of conduct](https://github.com/gwastro/pycbc/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/pycbc/events/coinc.py
+++ b/pycbc/events/coinc.py
@@ -949,8 +949,8 @@ class LiveCoincTimeslideBackgroundEstimator(object):
 
                 # Check that a coinc exists
                 if 'foreground/ifar' in result:
-                    ifar = result['foreground/ifar']
-                    stat = result['foreground/stat']
+                    ifar = result['foreground/ifar'].item()
+                    stat = result['foreground/stat'].item()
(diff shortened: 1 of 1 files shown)


pycbc/io/gracedb.py
```diff
@@ -159,6 +159,8 @@ def __init__(self, coinc_ifos, ifos, coinc_results, **kwargs):
                      if f'foreground/{ifo}' in n]
             for name in names:
                 val = coinc_results[f'foreground/{ifo}/{name}']
+                if isinstance(val, (numpy.ndarray, numpy.generic)):
+                    val = val.item()
```
titodalcanton, line 163: I do not think this is needed. All I need to get Live working is to ensure line 140 produces a Python scalar. Does this come from somewhere else than Live?

Issue #5105, opened 2025-04-09
Find a way to use a GitHub Actions cache to avoid downloading the frame files


titodalcanton:
We download a lot of GW data from GWOSC as part of the CI tests. This fails more often than I would like - it just did. Ideally we could use a GitHub Actions cache to keep the GW data, as we now do with the ROM files for LALSimulation. One difficulty is that the scripts that download the frame files (and as a result, the frame files themselves) are scattered in a few different places, and they would need to be told not to download a frame file that already exists.

titodalcanton:
Addressed, at least partially, in #5256.

titodalcanton:
This is now probably addressed, to a reasonable extent, by #5441.

PR #5256, opened 2026-01-16
CI: cache example GW data

titodalcanton:
Try to use a GitHub Actions cache for some of the GW data downloaded during the CI tests.

This is an efficiency update.

This change affects the CI tests.

## Motivation

The CI tests download quite a bit of GW data and auxiliary files. These downloads fail much more often than I like, wasting a lot of time and computing power. Let's try to use a GitHub Actions cache for at least some of the data. This is basically a continuation of #5099.

While poking at this, I also noticed some sub-optimality in how the GW data was being downloaded, so that is addressed too.

## Contents

This sets up a new cache for the files downloaded by the inference examples. These files are scattered throughout the example subdirectories, and some of them are even duplicates, so I had to explicitly list a bunch of paths. Is there a better way? Maybe we should rethink how we download these files in a way that is more organized and makes it easier to cache everything, but I could hear some suggestions.

Other than that, this fixes `examples/inference/lisa_smbhb_ldc/get.sh` to behave correctly when files exist already, and improves a bunch of similar scripts elsewhere (they are working fine, but might break if the web servers start doing URL redirects).

#5105

## Testing performed

None.

## Additional notes

None.

### DIFF
--- a/.github/workflows/basic-tests.yml
+++ b/.github/workflows/basic-tests.yml
@@ -78,12 +78,6 @@ jobs:
       with:
         name: documentation-page
         path: _gh-pages
-    - name: List GW data files
-      run: |
-        echo "Current working directory:"
-        pwd
-        echo "GW data files:"
-        find . -type f -name \*.gwf
   deploy_documentation:
     runs-on: ubuntu-latest
     needs: build


examples/inference/lisa_smbhb_ldc/get.sh
```diff
@@ -1,16 +1,24 @@
 set -e
 
+download_if_absent() {
```
ahnitz, line 3: I'm confused why the change to this file is different than the others. Shouldn't the caching already accomplish this, or no? e.g. why the special download_if_absent function where the other scripts don't have anything similar?

titodalcanton: The other scripts download a single file per loop, so their current approach (`test -f && continue`) works. For the LDC example, however, there are two files downloaded per loop, so if for some reason the first file exists but the second does not, then we have a problem.

But admittely this is ugly. I wish `curl` had an option like `--ignore-if-file-exists`, which does not seem the case.

examples/inference/lisa_smbhb_ldc/get.sh
```diff
@@ -1,16 +1,24 @@
...
+    local URL="$1"
+    local FILENAME=$(basename "$URL")
+    if [ ! -f "$FILENAME" ]; then
+        echo "Downloading $FILENAME"
+        curl -O -L --show-error --silent "$URL"
```
GarethCabournDavies, line 8: Should we take this opportunity to add retries if the download fails due to transient issues as well? e.g. add `--retry 2 --retry-delay 5` to curl?

titodalcanton: As far as I have noticed, the failures are typically brief downtimes of DCC or GWOSC rather than random network glitches. So retrying, even for a few minutes, would probably not help much. I propose we merge this as is, and then look into retries if we still have problems.

GarethCabournDavies: makes sense

PR #5258, opened 2026-01-19
Remove all the old cache-related options

spxiwh:
This patch removes all the OS-related CACHE_SIZE and similar stuff in `opt.py`

This is a: Modernization

This change changes: Efficiency options in some Cython code.

This change will: not break any functionality

## Motivation

PyCBC has always made use of blocks of C-code in places where efficiency is crucial. Currently this is done via Cython for CPU code. In the olden days we used the now-defunct `weave` module in scipy. One major difference is that `weave` relied on compilation at runtime. This meant that we needed to ship valid compilers around with us, but meant that we could compile code for the hardware it is running on.

With Cython the code is compiled at install time. As most people [sensibly] use wheels, the compilation happens on some GitHub CI machine. Therefore it makes no sense to try to optimize code based on what it's being compiled on ... Because that's a very different machine to what you're running on ... One can still compile PyCBC from source if you want to do this, but largely any options for optimization of code based on the machine has been removed with the weave removal.

ANYWAY, the issue is that there is still a bunch of old code in `opt.py` that we used to use when doing "at runtime" compilation using `weave`. Most of this is now not used, but there were two places (both event processing for offline code) where it is still in use. These use cache_sizes to determine block sizes for parallel use cases. These cache sizes are an argument to the precompiled code, so still have an affect. However, the default values are within a factor of 2 of "modern" CPU cores on CIT.

Adding to this, is the observation that the `getconf` executable we use to identify cache sizes is known to fail (see #5193). A more complicated solution is proposed there using a different function, but I don't know how general purpose that is, or if we just end up in this situation again down the line. By far the *worst* thing that can happen is that this returns garbage, and we run production code using "really bad" values (e.g. "1" or "0" or "-1").

Therefore, I propose to remove that functionality entirely and instead use the "sensible default" value that was identified for both of these codes. I also allow the user to directly override this using an environment variable, though I deliberately do not document that, as I do not intend that this would be used by more than a handful of people (if anyone).

## Contents

I change `simd_correlate` and `threshold_cpu` so that they don't try to use the identified the cache size when running. I move the environment-value-override (which was always there) into these functions, and remove the large block in `opt.py` where this was done.

This would fix #5193.

## Testing performed

The code [hopefully] runs on the test suite, and I tested that the default values are close to realistic values that the code might currently run with.

## Additional notes

### DIFF
--- a/pycbc/filter/simd_correlate.py
+++ b/pycbc/filter/simd_correlate.py
@@ -57,7 +57,7 @@ if l2_cache_size is not None:
     # Since we need 3 vectors fitting in L2 cache, divide by 3
     # We find the nearest power-of-two that fits, and the length
     # of the single-precision complex array that fits into that size.
-    pow2 = int(_np.log(l2_cache_size/3.0)/_np.log(2.0))
+    pow2 = int(_np.log2(l2_cache_size / 3.0))
     default_segsize = pow(2, pow2) // _np.dtype(_np.complex64).itemsize
 else:
     # Seems to work for Sandy Bridge/Ivy Bridge/Haswell, for now?


pycbc/events/threshold_cpu.py
```diff
@@ -23,14 +23,15 @@
...
 from .eventmgr import _BaseThresholdCluster
-from .. import opt
 
 logger = logging.getLogger('pycbc.events.threshold_cpu')
 
-if opt.HAVE_GETCONF:
-    default_segsize = opt.LEVEL2_CACHE_SIZE / numpy.dtype('complex64').itemsize
+if os.environ.get("_PYCBC_L2_CACHE_SIZE", None):
+    l2_cache_size = int(os.environ["_PYCBC_L2_CACHE_SIZE"])
+    default_segsize = l2_cache_size / numpy.dtype('complex64').itemsize
```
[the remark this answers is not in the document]
spxiwh: I think int conversion happens later anyway, but sure, this makes sense.

pycbc/events/threshold_cpu.py
```diff
@@ -23,14 +23,15 @@
...
+import os
 from .simd_threshold_cython import parallel_thresh_cluster, parallel_threshold
 from .eventmgr import _BaseThresholdCluster
-from .. import opt
 
 logger = logging.getLogger('pycbc.events.threshold_cpu')
 
-if opt.HAVE_GETCONF:
-    default_segsize = opt.LEVEL2_CACHE_SIZE / numpy.dtype('complex64').itemsize
+if os.environ.get("_PYCBC_L2_CACHE_SIZE", None):
```
titodalcanton: It may still be a good idea to have a single function or piece of code in pycbc.opt that gets the L2 cache size from the env var. So the name of the env var is defined in a single place.

pycbc/filter/simd_correlate.py
```diff
@@ -52,12 +52,13 @@ def correlate_simd(ht, st, qt):
...
 
-if opt.HAVE_GETCONF:
+l2_cache_size = opt.get_l2_cache_size()
+if l2_cache_size is not None:
     # Since we need 3 vectors fitting in L2 cache, divide by 3
     # We find the nearest power-of-two that fits, and the length
     # of the single-precision complex array that fits into that size.
-    pow2 = int(_np.log(opt.LEVEL2_CACHE_SIZE/3.0)/_np.log(2.0))
-    default_segsize = pow(2, pow2)/_np.dtype(_np.complex64).itemsize
+    pow2 = int(_np.log(l2_cache_size/3.0)/_np.log(2.0))
```
titodalcanton: Is there a reason why this is not simply `pow2 = int(_np.log2(l2_cache_size / 3.0))`?

spxiwh: Not that I'm aware of, changing this ...

PR #5267, opened 2026-01-26
CI: single source of truth

GarethCabournDavies:
This is an attempt to have single sources of truth defined for the CI workflows. As a few of the workflows use the same information, it would be good to reuse it without the repetition across files.

## Motivation
Single source of truth

## Contents
Add a new GitHub actions, to be called by the workflows. This defines a single source of truth which is then inherited by the other workflows

## Testing performed
this seems to work in the act tests performed, but these have not been extensive

### DIFF
--- a/.github/actions/install-system-deps/action.yml
+++ b/.github/actions/install-system-deps/action.yml
@@ -1,0 +1,16 @@
+name: Install system packages
+description: "Composite action to install common system packages (fftw3, intel-mkl, mpi, graphviz) on Ubuntu runners."
+
+runs:
+  using: 'composite'
+  steps:
+    - name: Update apt
+      shell: bash
+      run: |
+        sudo apt-get -o Acquire::Retries=3 update
+
+    - name: Install system packages
+      shell: bash
+      run: |
+        set -euxo pipefail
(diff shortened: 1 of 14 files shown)


.github/actions/install-system-deps/action.yml
```diff
@@ -0,0 +1,14 @@
...
+  using: 'composite'
+  steps:
+    - name: Update apt
+      run: |
+        sudo apt-get -o Acquire::Retries=3 update
+
+    - name: Install system packages
+      run: |
+        set -euxo pipefail
+        sudo apt-get -o Acquire::Retries=3 install -y *fftw3* intel-mkl* mpi graphviz
```
GarethCabournDavies: Note that here I have included a few things which weren't used everywhere, but I don think it is too expensive to run these install as well?

.github/workflows/basic-tests.yml
```diff
@@ -13,67 +13,37 @@ jobs:
       max-parallel: 60
       matrix:
         os: [ubuntu-24.04]
-        python-version: ['3.11', '3.12', '3.13']
+        python_version: ['3.11', '3.12', '3.13']
```
GarethCabournDavies, line 16: The `-` in  GitHub variables can sometimes be misconstrued as negation, so underscores are safer

PR #5275, opened 2026-02-02
Add support for two new search parameters eccentricity and rel_anomaly in the postprocessing stage to display them

yi-fan-wang:

This update adds support for two new search parameters, "eccentricity" and "rel_anomaly", which are utilized by the SEOBNRv5E(HM) waveform approximant. These parameters have been integrated into all relevant postprocessing displays, including single triggers, background coincidences, foreground coincidences, and injections.

### DIFF
--- a/bin/minifollowups/pycbc_page_coincinfo
+++ b/bin/minifollowups/pycbc_page_coincinfo
@@ -253,7 +253,7 @@ for ifo in files.keys():
         data[-1].append('%5.2f' % bank['eccentricity'][tid])
         data[-1].append('%5.2f' % bank['rel_anomaly'][tid])
         headers.append("eccentricity")
-        headers.append("rel_anomaly")
+        headers.append("anomaly")
     except:
         pass
 
--- a/bin/minifollowups/pycbc_page_injinfo
+++ b/bin/minifollowups/pycbc_page_injinfo
@@ -72,7 +72,7 @@ labels = {
     'chieff': '&chi;<sub>eff</sub>',
     'chip': '&chi;<sub>p</sub>',
     'eccentricity': 'eccentricity',
-    'rel_anomaly': 'relanomaly',
+    'rel_anomaly': 'anomaly',
 }
 
 params += ['tc']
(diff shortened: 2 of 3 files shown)


bin/minifollowups/pycbc_page_coincinfo
```diff
@@ -248,6 +248,14 @@ for ifo in files.keys():
     headers.append("s<sub>1z</sub>")
     headers.append("s<sub>2z</sub>")
     headers.append("Duration")
+    # display eccentricity and relative anomaly if they exist
+    try:
+        data[-1].append('%5.2f' % bank['eccentricity'][tid])
+        data[-1].append('%5.2f' % bank['rel_anomaly'][tid])
```
ahnitz, line 254: Different models uses different names here. Any thoughts on how to generalize? Or simply report everything that's in the file?

yi-fan-wang: I think this would be more relevant when we have more than one eccentric waveform models in a single injection set/template bank. Displaying all the parameters probably mean there would be more than one anomaly parameters, while only one of them take effects a time for a specific waveform model. I don't think it's ideal.

I prefer to set up a unified naming convention for anomaly and eccentricity from the PyCBC side, then map them to the specific names of each waveform. This can be done in a waveform plugin which ideally is able to call all eccentric waveform models. But the downside would be a parameter would mean different things in different models (not necessary a downside because that's just the fact?)

yi-fan-wang: Hi @ahnitz Getting back to this comment, I have concerns about reporting every parameter in the bank file. In principle, I have no objections, but there are parameters like approximants and f_lower that people may not need to display. Also, this seems like a big change that would impact the webpages of all current search pipelines. I would prefer not to make such a large change without thoroughly understanding different people's opinions.

If the issue is limited to naming conventions for eccentricity, I don't know enough about other waveform models at the moment, so need to study them and then come up with a unified convention. How about we merge this PR now? It should work for any eccentric EOB waveform. I'll try to generalize it when we need to deal with other models in the future.

yi-fan-wang:
Changed the label of rel_anomaly to "anomaly". 

Then I did a small test search run with this PR, the results are https://hypatia.aei.mpg.de/~yiwang/projects/test-pycbc-addecchtml/example_search/html/. The labels are correctly shown, so I'll merge

PR #5276, opened 2026-02-03
New effective eccentric mchirp parameter

rahuldhurkunde:
Added a new eccentric mchirp parameter as defined in [arXiv:2107.14736](https://arxiv.org/pdf/2107.14736).

This is a: new feature

This change affects: the offline search.

This change changes:result presentation / plotting

## Motivation
A new effective mchirp-e parameter might be better for eccentric searches, i.e. may lead to better bank placement. 

## Contents
Added a new method in `conversions.py`

## Testing performed
Reproduced the Figure 3. in the [paper](https://arxiv.org/pdf/2107.14736). 

ecc_mchirp_scatter_parallel

## Additional notes


pycbc/conversions.py
```diff
@@ -478,6 +478,47 @@ def mass2_from_tau0_tau3(tau0, tau3, f_lower):
     return mass2_from_mtotal_eta(mtotal, eta)
 
 
+def ecc_mchirp_from_mchirp_ecc(mchirp, ecc):
```
ahnitz: @rahuldhurkunde Nice. I might suggest two things. 

(1) (minor) is there a better name than 'ecc mchirp'? 
(2) Is it possible to write an inverse function? e.g. give back the mchirp from the 'ecc mchirp' + ecc.

rahuldhurkunde: @ahnitz The inverse function requires a root solver. So to avoid adding a scipy dependency, I had to include a new bisection root solver function: `solve_root`.

Simple testing seems to be working:

```
>>> conversions.ecc_mchirp_from_mchirp_ecc(20.0, 0.2)
20.668174895333745
>>> conversions._mchirp_from_ecc_mchirp_ecc(20.668174895333745, 0.2)
19.999999648528163
```

Sorry I can't think of any better name for the new parameter. Happy to change if you have any suggestions.

rahuldhurkunde: @ahnitz I've dropped the e^6 term, simplified the inversion to a quadratic one and renamed the new parameter to `Emchirp`.

rahuldhurkunde: I've bring back the cubic term, and performed some quick testing on 50000 points. In short, both the functions work well and this looks ready to merge from my end. 

```
--- Testing Conversions for 50000 points ---
Using ranges from plot_ecc_mchirp.py:
  ecc: [0.0, 0.6]
  ecc_mchirp: [5.0, 80.0]

Timing (total for 50000 points):
  mchirp_from_ecc_mchirp_ecc (Inverse): 0.003358 s
  ecc_mchirp_from_mchirp_ecc (Forward): 0.002018 s

Validity Check (Round-trip):
  Number of NaNs in recovered mchirp: 0
  Max absolute difference in mchirp: 2.13e-14
  Mean absolute difference in mchirp: 6.26e-16

  RESULT: SUCCESS - The inverse function is valid in this range.
```

yi-fan-wang:
@rahuldhurkunde I have a PR https://github.com/gwastro/pycbc/pull/5328, where I try to enable fitting the noise trigger rate with respect to eccentricity. I think it'd be good to think about adding eccentricity mchirp as well

rahuldhurkunde:
This was implemented as a part of https://github.com/gwastro/pycbc/pull/5344. Closing.

PR #5277, opened 2026-02-04
Two PyGRB plot fixes

pannarale:
This PR fixes two bugs of the PyGRB plotting infrastructure.
1. The function selecting data from a specific slide-id was never being called.
2. The background distribution plot was unchanged when showing SNR or SNR after cuts on the horizontal axis.

**Update**: this PR also improves performance with respect to [these original lines](https://github.com/pannarale/pycbc/blob/7012e541cc3d3097cbf5c761dfcba74e5b54653b/pycbc/results/pygrb_postprocessing_utils.py#L333-L339).

This is a: bug fix.

This change affects: PyGRB.

This change changes: result plotting.

This change will be included in the v2.8.6 release.  It does not affect final results, but one intermediate plot and plots that the user may want to generate to dig through results, but are generally not produced by default.

## Motivation
https://github.com/gwastro/pycbc/issues/5274 for the first point.  The other item was first noted by @Thomas-JACQUOT and once I verified that it affect all test analyses available, I looked into uncovering the origin of the bug.

## Contents

1. The call to `_slide_filter` was wrapped in an `if` that could never be true.  It is intended to compare array sizes, but on of the terms of comparison had a cut applied to it, unlike the other one, so they were never the same length.
2. The selection of triggers after the reweighted SNR cut was incorrect.  The reweighted SNR cut happens in each post-processing script by treating triggers with reweighted SNR below a user-defined threshold as though their reweighted SNR were 0; however, the `pycbc_pygrb_plot_stats_distribution` was trying to identify cut triggers by looking for triggers with reweighted SNR  identical to 0, rather than below threshold.

**Update**: rather than searching for slide_ids in an MxN fashion ([original code](https://github.com/pannarale/pycbc/blob/7012e541cc3d3097cbf5c761dfcba74e5b54653b/pycbc/results/pygrb_postprocessing_utils.py#L333-L339)) the code now exploits sorting prior to searching and it determine the array of sought ids only once.

https://github.com/gwastro/pycbc/issues/5274

## Testing performed
I checked that plotting the time series with `--slide-id all` and `--slide-id 0` (or another number) now produces different plots, the latter being less populated.

I checked that the background distributions produced with `--x-variable snr` and `--x-variable snruncut` (as well as `--x-variable bestnr`) are now indeed different.  The `snr` distribution generally has a larger coverage of the horizontal axis as it has all triggers, i.e., also those that do not pass the cut on reweighted SNR.

Tests were performed on ongoing analyses so they will be shared privately with the reviewer.

**Update**: the changes for performance improvement do not alter the output of the timeseries plotting job, but they reduce runtime by about 45%.

## Additional notes

### DIFF
--- a/bin/pygrb/pycbc_pygrb_plot_stats_distribution
+++ b/bin/pygrb/pycbc_pygrb_plot_stats_distribution
@@ -51,7 +51,7 @@ parser = ppu.pygrb_initialize_plot_parser(description=__doc__)
 parser.add_argument("-F", "--trig-file", action="store", required=True,
                     help="Location of off-source trigger file")
 parser.add_argument("-x", "--x-variable", required=True,
-                    choices=["bestnr", "snr", "snruncut"],
+                    choices=["bestnr", "snr", "snraftercuts"],
                     help="Quantity to plot on the horizontal axis.")
 ppu.pygrb_add_bestnr_cut_opt(parser)
 ppu.pygrb_add_slide_opts(parser)
@@ -114,13 +114,13 @@ trig_data = ppu.extract_trig_properties(
 # Calculate SNR and BestNR values and maxima
 time_veto_max_snr = {}
 time_veto_max_bestnr = {}
-time_veto_max_snr_uncut = {}
+time_veto_max_snr_aftercuts = {}
 
 for slide_id in slide_dict:
     num_slide_segs = len(trial_dict[slide_id])
     time_veto_max_snr[slide_id] = np.zeros(num_slide_segs)
     time_veto_max_bestnr[slide_id] = np.zeros(num_slide_segs)
-    time_veto_max_snr_uncut[slide_id] = np.zeros(num_slide_segs)
+    time_veto_max_snr_aftercuts[slide_id] = np.zeros(num_slide_segs)
 
 for slide_id in slide_dict:
     for j, trial in enumerate(trial_dict[slide_id]):
@@ -138,12 +138,12 @@ for slide_id in slide_dict:
         sbv_cut = trig_data[keys[2]][slide_id][:] >= opts.newsnr_threshold
         if not (trial_cut & sbv_cut).any():
             continue
-        time_veto_max_snr_uncut[slide_id][j] =\
+        time_veto_max_snr_aftercuts[slide_id][j] =\
             max(trig_data[keys[1]][slide_id][trial_cut & sbv_cut])
 
 # This is the data that will be plotted
 full_time_veto_max_snr = ppu.sort_stat(time_veto_max_snr)
-full_time_veto_max_snr_uncut = ppu.sort_stat(time_veto_max_snr_uncut)
+full_time_veto_max_snr_aftercuts = ppu.sort_stat(time_veto_max_snr_aftercuts)
 _, _, full_time_veto_max_bestnr = \
     ppu.max_median_stat(slide_dict, time_veto_max_bestnr, trig_data[keys[2]],
                         total_trials)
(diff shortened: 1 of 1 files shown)


pannarale:
I will note that lines [333-339](https://github.com/pannarale/pycbc/blob/two_plot_fixes/pycbc/results/pygrb_postprocessing_utils.py#L333-L339) are slow.  We will see if optimising them could be part of this PR or a new issue to be opened.

pycbc/results/pygrb_postprocessing_utils.py
```diff
@@ -361,7 +361,13 @@ def load_data(input_file, ifos, rw_snr_threshold=None, data_tag=None,
...
             if 'network/slide_id' in trigs.keys():
-                if trigs_dict[path].size == trigs['network/slide_id'][:].size:
+                # The slide selection is applied to datasets that contain
+                # properties of surviving triggers. These datasets are
+                # identified knowing that each trigger has a slide id, so
+                # they must have as many entries as the 'network/slide_id'
+                # dataset once triggers below threshold are removed from it.
+                if trigs_dict[path].size == \
+                        trigs['network/slide_id'][above_thresh].size:
                     trigs_dict[path] = _slide_filter(trigs, trigs_dict[path],
```
titodalcanton: Do you need to also replace `trigs` with `trigs_dict` in the call to `_slide_filter()`? If you pass `trigs`, `_slide_filter()` will use the original `network/slide_id` dataset without the `above_thresh` cut, no?

bin/pygrb/pycbc_pygrb_plot_stats_distribution
titodalcanton: So a variable named "uncut" is holding results with the cut? Just checking that this is not a mistake.

pannarale: This caught my attention too.  It's inherited from pylal and is meant to denote triggers that are not cut.  Happy to change it since it confused both of us.

pycbc/results/pygrb_postprocessing_utils.py
pannarale: I need to provide `_slide_filter` with an additional option for an extra mask (`above_thresh` in this case).  `trigs` is the whole block of trigger data, so passing `trigs_dict` will not work because it is being filled up in this loop.

titodalcanton: Perhaps it is worth rethinking the logic? E.g. would it make sense to combine the RW SNR cut and the slide ID cut into a single mask, and then go through the datasets and just apply that one mask?

pannarale: Okay, so rescoping (and renaming) `_slide_filter` to something like `_apply_mask_to_triggers`.  Actually, maybe I can just remove the function: this is the only place where it is called.

pannarale: Having said all that, the discussion above implies that `trigs_dict[path]` was being masked with a mask that is too long and that would not work.  I have to think this through more.

bin/pygrb/pycbc_pygrb_plot_stats_distribution
pannarale: Since I am at this, shall also relabel bestNR and its variants to reweightedSNR?

pannarale: I performed the relabelling of `uncut` to `aftercuts`.  I decided not to carry out the other relabelling because several other PyGRB lines of code stick to `bestnr` so I would rather have an overall PR to handle that organically.

PR #5278, opened 2026-02-04
V2315 changes

spxiwh:
Need to make what is hopefully one final release on the v23 release branch to include the new pegasus 5.1.2 release.

I think this should now be ready to merge. It basically contains:

* Some changes to the various github tests to make them work
  * The test workflows now run with 5.1.X (matching the main branch, which actually still runs with 5.1.1)
  * The tutorial tests are disabled ... The tutorials require main PyCBC and shouldn't be tested on release branches like this one
* Merging the "support for pegasus 5.1.X" patch #5142 ... No further changes are needed for 5.1.1
* Force pegasus 5.1.2 dependencies
* Restrict a bunch of other dependencies to "less than whatever it was at the end of 2024"
* Added a SINGULARITY requirement for OSG as recommended by James

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,8 @@ install_requires = setup_requires + [
     'pegasus-wms.api == 5.1.2',
     'python-ligo-lw >= 1.7.0',
     'ligo-segments',
+    'lalsuite!=7.2,<7.25',
+    'lscsoft-glue>=1.59.3',
     'pykerr',
 ]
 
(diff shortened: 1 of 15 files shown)


tdent (approved):
Approving as a formality if needed (no actual changes check)

setup.py
```diff
@@ -47,11 +47,9 @@
     'tqdm',
     'setuptools',
     'gwdatafind',
-    'pegasus-wms.api >= 5.0.6',
+    'pegasus-wms.api == 5.1.2',
     'python-ligo-lw >= 1.7.0',
     'ligo-segments',
-    'lalsuite!=7.2,<7.25',
```
titodalcanton, line 53: Why is this removed from here but not from `requirements.txt`?

spxiwh: I think this was a rebase typo ... It won't matter for a release like this, but I'll fix it.

PR #5281, opened 2026-02-04
Improvements to pycbc_brute_bank

ahnitz:

This is a feature update for pycbc brute bank. 

This affects bank generation, so could affect use in all searches, though only indirectly.

## Motivation
The main motivation is to improve the computational (both wall clock time and memory use) of pycbc_brute_bank.

## Contents
There are three features changes. 

1. The options to do match calculations / proposal point checking in parallel. This does affect the behavior and is not an entirely drop-in option. It enables parallelization by *assuming* that proposals do not need to be checked against each other. That means if two proposals were identical points, they would both end up getting added to the bank, since no check between them is done. However,  I believe that in most cases, this not really a problem. Without this option, requesting multiple cores will *only* parallelize bank generation as is currently the case. It think that is the safest default for the general case. 

2. Option to limit the amount of store information on the inter-bank matches. (through option --max-connections XX). By default this is set to infinite which matches the old behavior. If set to some number, only the most important matches are actually stored for each template (in the way we do things these are actually the 'low' matches). This means that if you get a particularly large bank, you can avoid being dominated in memory usage by the connections themselves instead of the cached waveforms. There are cases, where this can result in 5-10x memory reduction (again only for a large bank region) without noticeable change to result bank size. It  could potentially increase the number of matches required though if set too small. 

3. The option to save the bank even with an early termination. If you send a sigterm or ctrl-c (interrupt) signal to the job, it will now try to intercept this and do an emergency save of the bank. 

This should be merged first https://github.com/gwastro/pycbc/pull/5227

## Testing performed
I've verified that by default banks are not modified. The defaults resort to the old behavior.

### DIFF
--- a/bin/bank/pycbc_brute_bank
+++ b/bin/bank/pycbc_brute_bank
@@ -106,7 +106,8 @@ dists_joint = prior_from_config(cp=config_parser)
 
 def get_stable_match(p_inj, p_tmplt):
     """Iteratively doubles buffer length until the match stabilizes."""
-    current_buflen = args.buffer_length if args.full_resolution_buffer_length is None else args.full_resolution_buffer_length
+    frbl = args.full_resolution_buffer_length
+    current_buflen = args.buffer_length if frbl is None else frbl
     last_match = -1.0
     
     # We loop until convergence or a safety cap (e.g., 64s)


bin/bank/pycbc_brute_bank
```diff
@@ -229,23 +236,33 @@ class TriangleBank(object):
...
+                msort = matches[r].argsort()
+                
+                msorted = matches[r][msort]
+                rsorted = r[msort]
+                keep = numpy.ones(len(msorted), dtype=bool)
+                if args.max_connections < len(keep):
+                    #keep[args.max_connections//2: -args.max_connections//2] = False
+                    keep[args.max_connections:] = False
+                
+                hp.matches = msorted[keep].copy()
```
GarethCabournDavies, line 254: Does this array copy not now increase the memory footprint, particularly in the  max_connections infinite / default case?

bin/bank/pycbc_brute_bank
```diff
@@ -229,23 +236,33 @@ class TriangleBank(object):
...
                               % (mmax, len(self), msig, mtau, mnum))
+                hp.max_match = mmax
                 return False
 
             hc = self[j]
 
             # Defensive initialization if matches/indices are missing
             if not hasattr(hc, 'matches'):
-                hc.matches = numpy.empty(len(self))
+                hc.matches = numpy.empty(2)
```
GarethCabournDavies: Im not sure I understand this change

bin/bank/pycbc_brute_bank
```diff
@@ -81,6 +85,8 @@ parser.add_argument('--tau0-end', type=float)
 parser.add_argument('--tau0-cutoff-frequency', type=float, default=15.0)
 parser.add_argument('--nprocesses', type=int, default=1,
     help='Number of processes to use for waveform generation parallelization. If not given then only a single core will be used.')
+parser.add_argument('--parallel-check', action='store_true', help="Do bank checking parallel, note that this means that proposals WILL NOT be checked against each other.")
```
GarethCabournDavies, line 76: Could this line be wrapped to match maximum line length requirements?

bin/bank/pycbc_brute_bank
```diff
@@ -265,34 +282,39 @@ class TriangleBank(object):
...
+                    hp.gen = gen
+                    hp.checked = hp not in self
+ 
+                if hp.checked:
+                    self.max_matches.append(hp.max_match)
+                
+                if hp.checked or force_add:
                     num_added += 1
-                    self.insert(hp)
+                    self.insert(hp)
```
GarethCabournDavies: whitespace

bin/bank/pycbc_brute_bank
```diff
@@ -265,34 +282,39 @@ class TriangleBank(object):
...
+                
+                if hp.checked or force_add:
                     num_added += 1
-                    self.insert(hp)
+                    self.insert(hp)    
             else:
                 logging.info("Waveform generation failed!")
                 continue
-
+
```
GarethCabournDavies, line 311: whitespace

bin/bank/pycbc_brute_bank
ahnitz: Actually it's the opposite, otherwise loose reference could keep something in memory when we don't really need it anymore.

bin/bank/pycbc_brute_bank
ahnitz: In truth, I'd like to remove these lines. I can't really figure out why they are needed. The only guess I have is that perhaps there is some boundary issue that this is masking. However, to the point, there is no reason to add these with the full size of the existing bank. Since the information added is meaningless, it ends up being a waste of memory. Putting it at 2 was simply a lazy way to force the correct dimensions.

ahnitz:
@GarethCabournDavies I'll address the style issues before merging.

bin/bank/pycbc_brute_bank
ahnitz: A further clarification may help here. If you just return a view, even if that's all you have, the *full* array is always kept. I explicitly want the reduced form to be what is stored and not the full array, hence explicitly asking for a new array that contains a copy of what the view pointed to. This means that the original values can be cleaned up by python's garbage collector.

examples/search/bank.sh
```diff
@@ -9,13 +13,10 @@ pycbc_brute_bank \
 --buffer-length 2 \
 --sample-rate 2048 \
 --tau0-threshold 0.5 \
---approximant IMRPhenomD \
 --tau0-crawl 5 \
 --tau0-start 0 \
---tau0-end 50 \
+--tau0-end 15 \
```
GarethCabournDavies, line 18: Why change tau0-end?

bin/bank/pycbc_brute_bank
```diff
@@ -37,22 +40,16 @@ parser = argparse.ArgumentParser(description=__doc__)
 pycbc.add_common_pycbc_options(parser)
 parser.add_argument('--output-file', required=True,
     help='Output file name for template bank.')
-parser.add_argument('--input-file',
+parser.add_argument('--input-file', nargs='*', default=[],
     help='Bank to use as an initial set of starting samples.')
 parser.add_argument('--keep-entire-input-file', help="All of the templates in the input file will be kept in the output file, even if they do not pass normal mismatch criteria.", action='store_true')
 parser.add_argument('--input-config',
```
GarethCabournDavies: This is now required, right?

```suggestion
parser.add_argument('--input-config', required=True,
```

bin/bank/pycbc_brute_bank
```diff
@@ -92,30 +86,47 @@ args = parser.parse_args()
...
+    waveform_transforms = transforms.read_transforms_from_config(
+            config_parser, 'waveform_transforms')
+else:
+    waveform_transforms = None
+
+dists_joint = prior_from_config(cp=config_parser)
+
+def get_stable_match(p_inj, p_tmplt):
+    """Iteratively doubles buffer length until the match stabilizes."""
+    current_buflen = args.buffer_length if args.full_resolution_buffer_length is None else args.full_resolution_buffer_length
```
GarethCabournDavies: Wrap (somehow!)

GarethCabournDavies:
```suggestion
    frbl = args.full_resolution_buffer_length
    current_buflen = args.buffer_length if frbl is None else frbl
```

Maybe something like this?

bin/bank/pycbc_brute_bank
```diff
@@ -416,90 +473,76 @@ gen = GenUniformWaveform(args.buffer_length,
     args.sample_rate, args.low_frequency_cutoff)
 bank = TriangleBank()
 
-def wf_wrapper(p):
+def silent_child_exit(signum, frame):
+    # The child workers execute this branch
+    # os._exit(0) ensures they die instantly and quietly
```
GarethCabournDavies, line 479: 'ensure the child workers die instantly and quietly' - out of context this sentence is worrying

examples/search/bank.sh
ahnitz: It wasn't actually using that range if you look at the config file. I could revert, but it would not affect the bank. It just reduces the runtime slightly while it figures out that it can't place templates.

bin/bank/pycbc_brute_bank
ahnitz: Wow, I actually didn't notice that reading...

PR #5282, opened 2026-02-06
add new requirements to have singularity for OSG site

yi-fan-wang:

This is a fix to add a new condor requirement `HAS_CVMFS_singularity_opensciencegrid_org=True` in the Pegasus workflow when using OSG sites. This is caused by a upstream OSG configuration change shown [here](https://github.com/opensciencegrid/osg-flock/commit/914b2cc5874ac377a3c6f160c5b247e6cb6aa706).  Hence the attribute `HAS_SINGULARITY=True` no longer implies `HAS_CVMFS_singularity_opensciencegrid_org=True`.  So the old requirements don't actually require the CVMFS singularity image repository and one can land places where there's a CVMFS problem.

This solution is given by James Clark in this [LVK computing issue](https://git.ligo.org/computing/helpdesk/-/issues/8359#note_1664585)(LVK account required to check the page)


spxiwh:
But @yi-fan-wang this is duplicating what is already in #5278 .... It would be good to put this on the main branch as well though, so perhaps we can repurpose this PR for that.

PR #5284, opened 2026-02-12
Make HGroup more robust to changes in h5py.Group

GarethCabournDavies:
There were some changes to h5py's create_dataset which I didn't see for a while. This change will make HGroup more robust to changes in h5py.Group as a result

Ive extended the checking for which situations need flecther32 checksums enabled, I dont think this is necessary for the situations we have now, but it is for safety in the future.

I've also added a couple of unit testing modules (largely written by copilot to be fair) which tests many different methods of creation, whether the datasets are the same after being loaded etc., and whether fletcher32 is enabled or not. The wider hdf.py checks are for things like the bool and indexing selections.

This change touches all areas which use h5py for saving files (which is everywhere), but actually affects nothing

## Motivation
I didn't want to rely on me getting automated emails of whether a file has changed any more (by the evidence already, this doesn't work)

extends #4831

## Testing performed
unit tests added so that the testing is done automatically

### DIFF
--- a/test/test_io_hgroup.py
+++ b/test/test_io_hgroup.py
@@ -84,7 +84,7 @@ class TestIOHFile(unittest.TestCase):
                 f.create_dataset(
                     "data_np_str_array", data=np.array(str_list, dtype="S1")
                 )
-                should_not_be_checksummed.append("data_np_str_array")
+                should_be_checksummed.append("data_np_str_array")
 
                 # Make a dataset from some string data
                 f.create_dataset(


pycbc/io/hdf.py
```diff
@@ -31,34 +31,91 @@
...
+                # Non-tuple shapes are treated as non-scalar
+                is_scalar = False
+        elif data is not None:
+            try:
+                arr = np.asarray(data)
+                is_scalar = (arr.ndim == 0)
+            except Exception:
+                is_scalar = False
+
+        # Only set fletcher32 for numeric, non-scalar datasets.
```
ahnitz: Is the only reason we need the complicated logic above to determine if the checksum can be used or not? If there no way to check this within h5py itself to avoid having the logic be here?

GarethCabournDavies: It is a bit unwieldy, I will replace with a try/except (EAFP)

pycbc/io/hdf.py
```diff
@@ -39,58 +39,19 @@ def create_group(self, *args, **kwargs):
...
+        if 'fletcher32' in kwds:
+            del kwds['fletcher32']
 
-        # Only set fletcher32 for numeric, non-scalar datasets.
-        if dtype_numeric and not is_scalar:
-            kwds['fletcher32'] = True
+        try:
+            return super().create_dataset(name, *args, **kwds, fletcher32=True)
+        except ValueError:
+            return super().create_dataset(name, *args, **kwds)
```
ahnitz, line 57: Is any debug level logging worth having here?

test/test_io_hgroup.py
```diff
@@ -0,0 +1,250 @@
...
+                    "HFile-created dataset should have fletcher32 checksum"
+                )
+
+                # Assert that the h5py.File-created dataset _doesnt_
+                # have fletcher32 enabled (If this fails, h5py has updated its
+                # default behaviour, and our wrapper isn't needed)
+                self.assertFalse(
+                    _dataset_has_fletcher32(d_h5py),
+                    "h5py.File-created dataset should not have fletcher32 checksum"
+                )
```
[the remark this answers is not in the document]
GarethCabournDavies: I want to make sure that we know if h5py changes the default to include fletcher32, as the changes made there will be more robust that any we could sensibly implement. Failing at this point means that we would know straight away

PR #5285, opened 2026-02-13
Remove gwpy

### LINKED ISSUE #5257: Remove dependency on gwpy
In #5255 @ahnitz and I were discussing the dependency on gwpy. We do in fact depend on it, see `companion.txt`. However, the only place where we use it is `pycbc_live_collated_dq_trigger_rates`, and from a quick glance it looks like we could achieve the same functionality just with APIs from PyCBC itself and ligo.segments. We should try that and drop the gwpy requirement.

GarethCabournDavies:
Update `pycbc_live_collated_dq_trigger_rates` to use pycbc-native stuff rather than gwpy. This means that we can remove gwpy from companion.txt

Closes #5257 

This is a change to use already-included code rather than external processes
This change will: remove a companion dependency

## Testing performed
I would like to run an example of this with code before/after, but trying to find an example with still-existing files is not obvious!

### DIFF
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -1235,7 +1235,7 @@ class TimeSeries(Array):
             plot2 = pyplot.plot(self.sample_times, self.imag(), **kwds)
             return plot1, plot2
 
-    def bool_to_segmentlist(self, epoch=None):
+    def bool_to_segmentlist(self):
         """
         Convert a boolean pycbc TimeSeries (this must be bool or integer) to
         an igwn_segments.segmentlist of (start, end) in GPS seconds.
@@ -1251,7 +1251,7 @@ class TimeSeries(Array):
             )
 
         segs = segmentlist([])
-        
+
         if arr.size == 0:
             return segs.coalesce()
 


bin/live/pycbc_live_collated_dq_trigger_rates
```diff
@@ -57,6 +57,39 @@ def find_frames(ifo, frame_dir, frame_type, start, end):
     return files
 
 
+def timeseries_to_segmentlist(ts_bool, epoch=None):
```
titodalcanton: I think this would fit well as a method of the TimeSeries class (but reading further down, I doubt `ts_bool` is actually a PyCBC TimeSeries object? If not, ignore this) In that case we probably want to verify at the beginning that the dtype of the timeseries is bool and raise a ValueError otherwise.

bin/live/pycbc_live_collated_dq_trigger_rates
```diff
@@ -57,6 +57,39 @@ def find_frames(ifo, frame_dir, frame_type, start, end):
...
+        if ts_bool[i]:  # True -> inside flagged region
+            if not in_seg:
+                in_seg = True
+                curr_start = start_time + i * dt
+        else:
+            if in_seg:
+                # close current seg; end is start of this index
+                curr_end = start_time + i * dt
+                segs.append(segment(curr_start, curr_end))
+                in_seg = False
```
titodalcanton: I fear that this might be quite inefficient if the timeseries has hundreds of thousands of items or more.

@maxtrevor is probably best placed to judge if that is a realistic expectation.

bin/live/pycbc_live_collated_dq_trigger_rates
```diff
@@ -119,24 +153,33 @@ for seg in observing_segs:
...
-        end=seg[1],
-        pad=0,
-    )
+    try:
+        ts_dq_channel, ts_dq_ok_channel = read_frame(
+            frames, [dq_channel, dq_ok_channel],
+            start_time=seg[0], end_time=seg[1]
+        )
+    except Exception:
+        raise
```
titodalcanton: What is the point of catching and reraising here?

bin/live/pycbc_live_collated_dq_trigger_rates
```diff
@@ -119,24 +153,33 @@ for seg in observing_segs:
...
-    dq_ok_flag = (tsdict[dq_ok_channel] == 1).to_dqflag()
-    dq_flag = (tsdict[dq_channel] <= args.dq_thresh).to_dqflag()
+    # Is the channel above threshold?
+    dq_mask = (ts_dq_channel <= args.dq_thresh)
+    dq_segs = timeseries_to_segmentlist(dq_mask)
 
-    valid_bad_dq = dq_flag & dq_ok_flag
+    # these are the times where dq channel is working, and
+    # the dq channel is above threshold
+    valid_bad_dq = (dq_segs & dq_ok_segs)
```
titodalcanton: Parentheses not needed.

bin/live/pycbc_live_collated_dq_trigger_rates
maxtrevor: iDQ is in the frames with a sampling rate of 128 Hz.  I can't remember if this only processes 8 seconds at a time, or if we hold some extra time in memory, but there should be O(1000) entries in the timeseries.

bin/live/pycbc_live_collated_dq_trigger_rates
GarethCabournDavies: There wasn't one - I think I was doing so sensibly before, but have removed this now

The gw_datafind version wasn't actually working before, so there has been a bit of jiggery-pokery here anyway

bin/live/pycbc_live_collated_dq_trigger_rates
GarethCabournDavies: You are right, we do need the start time and delta t which we get from the teimseries object - I have reimplemented as a method of timeseries.

bin/live/pycbc_live_collated_dq_trigger_rates
GarethCabournDavies: Ive converted to a vectorised approach, which should be faster

bin/live/pycbc_live_collated_dq_trigger_rates
```diff
@@ -26,12 +26,12 @@ import glob
 import numpy as np
 
 import gwdatafind
```
titodalcanton: Import no longer needed.

GarethCabournDavies:
```suggestion

```

bin/live/pycbc_live_collated_dq_trigger_rates
```diff
@@ -85,30 +84,36 @@ pycbc.init_logging(args.verbose)
...
+    ar_flag_name = args.analysis_flag_name.format(ifo=args.ifo)
+    observing_flag = pycbc.dq.query_flag(
+        args.ifo, ar_flag_name, args.gps_start_time, args.gps_end_time
+    )
+else:
+    # If we are not querying the database for segments, just assume that the
+    # full time is in observing mode
+    observing_flag = segmentlist(
+        [segment(args.gps_start_time, args.gps_end_time)]
+    )
```
maxtrevor, line 98: I see that the analysis flag name is required now, so is there any reason to keep the if/else structure? Seems like the else will never be reached

GarethCabournDavies: The analysis-flag-name _was_ required, but is now not required, so this is needed for simple testing

pycbc/types/timeseries.py
```diff
@@ -1234,6 +1235,48 @@ def plot(self, **kwds):
...
+        """
+        Convert a boolean pycbc TimeSeries (Truth-like when condition holds) to
+        an igwn_segments.segmentlist of (start, end) in GPS seconds.
+        """
+
+        # Is the data truthlike?
+        # bools or numbers are OK, but we require finite values
+        arr = self.numpy()
+        if arr.dtype.kind != 'b' and not _numpy.issubdtype(arr.dtype, _numpy.number):
+            raise RuntimeError(
```
titodalcanton: This should probably raise a TypeError.

pycbc/types/timeseries.py
```diff
@@ -1234,6 +1235,48 @@ def plot(self, **kwds):
...
+                'is truthlike'
+            )
+
+        segs = segmentlist([])
+        
+        if arr.size == 0:
+            return segs.coalesce()
+
+        # Convert to bool Treat NaNs as zeros
+        b = _numpy.nan_to_num(arr, nan=0).astype(bool)
```
titodalcanton: This looks risky, I would avoid it and require up above that the dtype be either bool or integer (i.e. float is not permitted).

GarethCabournDavies: okay, makes sense

pycbc/types/timeseries.py
GarethCabournDavies: updated

PR #5286, opened 2026-02-16
Inference workflow small_test failure

GarethCabournDavies:
I think it's to do with setting the pegasus site. Lets copy this from the search example and see if that works


GarethCabournDavies:
Fixes failure exposed by #5270

examples/workflow/inference/small_test/workflow.ini
```diff
@@ -66,7 +66,15 @@ page_versioning = ${which:pycbc_page_versioning}
 
 [pegasus_profile]
 condor|request_memory = 1G
-pycbc|primary_site = condorpool_shared
+pycbc|primary_site = condorpool_symlink
```
ahnitz: One shouldn't change the primary site here. This workflow in general *should* use the shared site by default.

PR #5293, opened 2026-02-27
Add options to fix PSD dynamic range issues when inpainting

acorreia61201:

This is a new feature, intended to fix a bug that can occur in the gating and in-painting code. The added functionality affects how the low frequency portions of PSDs are handled, although these are intended to be optional; the base functionality should be unaffected.

## Motivation
We have encountered a numerical bug when performing PE using the gated gaussian noise models with PSDs supplied from text files. Specifically, calculating likelihoods on different machines can produce wildly different log likelihood values, some of which can even be large positive values. These have led to different potentially singularly-concentrated posteriors depending on the machine we run PE on.

We've traced the issue to the `solve_toeplitz` method in the gating and in-painting code, which becomes numerically unstable when passing in covariance matrices with large condition numbers. We found that this can occur when supplying PSDs from text files and specifying the `psd-inverse-length` option in the data section. With the latter option specified, the inverse PSD values below the low frequency cutoff are set to zero, which dramatically widens the dynamic range of the PSD and causes the numerical instabilities. The same behavior is not seen when `psd-inverse-length` is not specified, in which case the values below `kmin` are set to the value of the inverse PSD at `kmin`.

## Contents
This PR adds two options to the PSD argument parser that control how the PSD is handled when `psd-inverse-length` is specified (i.e. when a PSD is passed to `inverse_spectrum_truncation`). Specifically, an option `--invpsd-trunc-low-freq-fill-value` is added that allows the user to specify what value the inverse PSD below `kmin` is set to. The default behavior - setting the inverse PSD to 0 - is preserved; any float can be used as an input. The option also allows for input `kmin` to set the low-frequency values to the value at `kmin`, which matches the behavior when `psd-inverse-length` is not specified.

An additional option `--invpsd-trunc-which-spectrum` is added that allows the user to specify which spectrum - the inverse ASD or inverse PSD - to use when truncating. Currently, the inverse ASD is used. The option to use the inverse PSD is added to be consistent with the rest of `psd.from_cli`.

## Testing performed
Three models were loaded using a zero noise injection and PSD text files obtained from a previous GW150914-like PE run [here](https://zenodo.org/records/10493247). All three models were identical barring their data sections; one did not specify `psd-inverse-length`, one specified `psd-inverse-length = 8`, and the third specified `psd-inverse-length = 8`, `invpsd-trunc-low-freq-fill-value = kmin`, and `invpsd-trunc-which-spectrum = invpsd`.

Below is a plot of the inverse L1 PSDs from each of the three models. The black dashed line shows the PSD with no inverse PSD handling, blue shows the PSD with just `psd-inverse-length` specified, and orange shows the PSD with the new options specified. Just specifying `psd-inverse-length` leads to a large dynamic range below the low frequency cutoff; the new options correct this.

image

The condition numbers of the resulting Toeplitz matrices used in gating and in-painting were calculated for each model assuming a 1-second (2048-sample) gate. For the PSDs without inverse PSD options and with the new inverse PSD options, the condition number was ~1e4, whereas the PSD with just `psd-inverse-length` specified gives a condition number of ~1e17 (notably above the threshold 1e16 assumed for numerical stability).

### DIFF
--- a/pycbc/psd/__init__.py
+++ b/pycbc/psd/__init__.py
@@ -71,7 +71,6 @@ def from_cli(opt, length, delta_f, low_frequency_cutoff,
         f_low = opt.psd_low_frequency_cutoff
     else:
         f_low = low_frequency_cutoff
-    print(f_low)
     sample_rate = (length -1) * 2 * delta_f
 
     try:
@@ -219,14 +218,12 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
                              "impulse response of the overwhitening "
                              "filter (s)")
     psd_options.add_argument("--psd-low-frequency-cutoff", type=float,
-                             default=None,
                              help="(Optional) The low frequency cutoff for the "
                                   "PSD. If not specified, the low frequency "
                                   "cutoff of the matched filter/likelihood "
                                   "integral will be used.")
     # Truncation options if specifying psd-inverse-length
-    psd_options.add_argument("--invpsd-trunc-method", default=None,
-                             choices=["hann"],
+    psd_options.add_argument("--invpsd-trunc-method", choices=["hann"],
                              help="(Optional) What truncation method to use "
                                   "when applying psd-inverse-length. If not "
                                   "provided, a hard truncation will be used.")
@@ -295,7 +292,7 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
                                  help="(Required for --psd-estimation) "
                                  "The separation between consecutive "
                                  "segments (s)")
-        psd_options.add_argument("--psd-num-segments", type=int, default=None,
+        psd_options.add_argument("--psd-num-segments", type=int,
                                  help="(Optional, used only with "
                                  "--psd-estimation). If given, PSDs will "
                                  "be estimated using only this number of "
@@ -360,7 +357,6 @@ def insert_psd_option_group_multi_ifo(parser):
                           help="(Required for --psd-estimation) The separation"
                                " between consecutive segments (s)")
     psd_options.add_argument("--psd-num-segments", type=int, nargs="+",
-                          default=None,
                           action=MultiDetOptionAction, metavar='IFO:NUM',
                           help="(Optional, used only with --psd-estimation). "
                                "If given PSDs will be estimated using only "
@@ -373,8 +369,7 @@ def insert_psd_option_group_multi_ifo(parser):
                           action=MultiDetOptionAction, metavar='IFO:LENGTH',
                           help="(Optional) The maximum length of the impulse"
                           " response of the overwhitening filter (s)")
-    psd_options.add_argument("--invpsd-trunc-method", default=None,
-                             choices=["hann"],
+    psd_options.add_argument("--invpsd-trunc-method", choices=["hann"],
                              help="(Optional) What truncation method to use "
                                   "when applying psd-inverse-length. If not "
                                   "provided, a hard truncation will be used.")
@@ -387,6 +382,7 @@ def insert_psd_option_group_multi_ifo(parser):
                                   "PSD.")
     psd_options.add_argument("--invpsd-trunc-low-freq-fill-value", default=0.,
                              action=MultiDetOptionAction, metavar='IFO:VALUE',
+                             nargs="+",
                              help="(Optional) Value to set the inverse PSD to "
                                   "for frequencies below the low frequency "
                                   "cutoff when applying psd-inverse-length. "


pycbc/psd/__init__.py
```diff
@@ -202,11 +220,31 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
                              help="(Optional) The maximum length of the "
                              "impulse response of the overwhitening "
                              "filter (s)")
+    psd_options.add_argument("--psd-low-frequency-cutoff", default=None,
+                             help="(Optional) The low frequency cutoff for the "
+                                  "PSD. By default, uses the low frequency "
+                                  "cutoff given by the model if using a model.")
```
cdcapano: These options are used for both inference and searches (which do not have "models"). I suggest changing the second sentence to:

```suggestion
                                  "PSD. If not specified, the low-frequency cutoff "
                                  "of the matched filter/likelihood integral will be "
                                  "used.")
```

pycbc/psd/__init__.py
```diff
@@ -202,11 +220,31 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
...
+                                  "truncation when applying psd-inverse-length. "
+                                  "The default ('invasd') truncates the inverse "
+                                  "ASD, while 'invpsd' truncates the inverse "
+                                  "PSD.")
+    psd_options.add_argument("--invpsd-trunc-low-freq-fill-value", default=0,
+                             help="(Optional) Value to set the inverse PSD to "
+                                  "for frequencies below the low frequency "
+                                  "cutoff when applying psd-inverse-length. "
+                                  "Default 0; accepts any float. Also accepts "
+                                  "'kmin'; this sets values below the cutoff "
```
cdcapano: Change 'kmin' to 'fmin', it's a little more descriptive.

pycbc/psd/__init__.py
```diff
@@ -306,6 +344,11 @@ def insert_psd_option_group_multi_ifo(parser):
                           help="Measure PSD from the data, using given "
                           "average method. Choose from "
                           "mean, median or median-mean.")
+    psd_options.add_argument("--psd-low-frequency-cutoff", nargs="+",
+                             action=MultiDetOptionAction, metavar='IFO:FREQ',
+                             help="(Optional) The low frequency cutoff for the "
+                                  "PSD. By default, uses the low frequency "
+                                  "cutoff given by the model if using a model.")
```
cdcapano: Change the second sentence to what I suggest above.

pycbc/psd/__init__.py
```diff
@@ -333,6 +376,21 @@ def insert_psd_option_group_multi_ifo(parser):
...
+                                  "truncation when applying psd-inverse-length. "
+                                  "The default ('invasd') truncates the inverse "
+                                  "ASD, while 'invpsd' truncates the inverse "
+                                  "PSD.")
+    psd_options.add_argument("--invpsd-trunc-low-freq-fill-value", default=0,
+                             help="(Optional) Value to set the inverse PSD to "
+                                  "for frequencies below the low frequency "
+                                  "cutoff when applying psd-inverse-length. "
+                                  "Default 0; accepts any float. Also accepts "
+                                  "'kmin'; this sets values below the cutoff "
```
cdcapano: Again, change kmin to fmin.

pycbc/psd/__init__.py
```diff
@@ -202,11 +220,31 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
...
                                   "when applying psd-inverse-length. If not "
                                   "provided, a hard truncation will be used.")
+    psd_options.add_argument("--invpsd-trunc-which-spectrum", type=str,
+                             default='invasd', choices=["invasd", "invpsd"],
+                             help="(Optional) Which spectrum to use to perform "
+                                  "truncation when applying psd-inverse-length. "
+                                  "The default ('invasd') truncates the inverse "
+                                  "ASD, while 'invpsd' truncates the inverse "
+                                  "PSD.")
+    psd_options.add_argument("--invpsd-trunc-low-freq-fill-value", default=0,
```
cdcapano: Make the default `0.` (i.e., a float, not an int).

pycbc/psd/estimate.py
```diff
@@ -244,23 +255,44 @@ def inverse_spectrum_truncation(psd, max_filter_len, low_frequency_cutoff=None,
...
+    if low_frequency_fill_value != 0:
+        if low_frequency_fill_value == 'kmin':
+            low_val = 1/psd[kmin]
+            inv_spectrum[:kmin] = low_val
+        elif type(low_frequency_fill_value) == float:
+            inv_spectrum[:kmin] = 1/low_frequency_fill_value
+        else:
+            raise ValueError(f'Invalid low frequency fill value '
+                             f'{low_frequency_fill_value}; '
+                             f'input must be either a float or "kmin"')
```
cdcapano: The help for the command-line option says that low-frequency-fill-value is what value the inverse PSD should be set to, but this is taking it to be the value of the PSD. I think you want the former. The logic could be cleaned up a bit too. I wouldn't try to check types either, that gets messy. Just assume that the user has passed something sensible if it's not 'fmin'. Let the array setting code handle raising errors if non-sensible values are provided. Suggest:

```suggestion
    if low_frequency_fill_value == 'fmin':
        low_frequency_fill_value = 1./psd[kmin]
    inv_spectrum[:kmin] = low_frequency_fill_value
```

pycbc/psd/estimate.py
```diff
@@ -244,23 +255,44 @@ def inverse_spectrum_truncation(psd, max_filter_len, low_frequency_cutoff=None,
...
+
+    inv_spectrum[kmin:N//2] = (1.0 / psd[kmin:N//2])
```
cdcapano, line 271: What happens at N//2? Is there a chance it could still be 0 there? If so, we could still have the same dynamic range problem.

pycbc/psd/estimate.py
```diff
@@ -244,23 +255,44 @@ def inverse_spectrum_truncation(psd, max_filter_len, low_frequency_cutoff=None,
...
+
+    # if truncating asd, take sqrt
+    if which_spectrum != 'invpsd':
+        if which_spectrum == 'invasd':
+            inv_spectrum[:N//2] = inv_spectrum[:N//2]**0.5
+        else:
+            raise ValueError(f'Invalid which_spectrum input {which_spectrum}; '
+                             f'input must be either "invpsd" or "invasd"')
```
cdcapano: I'd avoid nested if statements where possible; they can be confusing to parse. Suggest:

```suggestion
    # if truncating asd, take sqrt
    if which_spectrum == 'invasd':
        inv_spectrum[:N//2] = inv_spectrum[:N//2]**0.5
    elif which_spectrum != 'invpsd':
        raise ValueError(f'Invalid which_spectrum input {which_spectrum}; '
                             f'input must be either "invpsd" or "invasd"')
```

pycbc/psd/__init__.py
```diff
@@ -333,6 +376,21 @@ def insert_psd_option_group_multi_ifo(parser):
...
+                             help="(Optional) Value to set the inverse PSD to "
+                                  "for frequencies below the low frequency "
+                                  "cutoff when applying psd-inverse-length. "
+                                  "Default 0; accepts any float. Also accepts "
+                                  "'kmin'; this sets values below the cutoff "
+                                  "to the inverse PSD value at the low "
+                                  "frequency cutoff.")
```
cdcapano: Which cutoff is used? The psd-low-frequency-cutoff or the low-frequency-cutoff? It should probably be psd-low-frequency-cutoff if it isn't. Either way, this should be clarified in the help message.

acorreia61201: I've updated the docs to reflect that this will use the input `psd-low-frequency-cutoff`, or whatever's supplied by the matched filter/likelihood model otherwise.

pycbc/psd/estimate.py
acorreia61201: The value at N//2 isn't manually set, so it is zero here. This is true for all the plots above though, so it doesn't seem to affect the dynamic range. Alternatively, if we want to be safe, perhaps we could change this to set the spectrum values all the way to last sample?

pycbc/psd/__init__.py
```diff
@@ -202,11 +220,33 @@ def insert_psd_option_group(parser, output=True, include_data_options=True):
                              help="(Optional) The maximum length of the "
                              "impulse response of the overwhitening "
                              "filter (s)")
+    psd_options.add_argument("--psd-low-frequency-cutoff", default=None,
+                             help="(Optional) The low frequency cutoff for the "
+                                  "PSD. If not specified, the low frequency "
+                                  "cutoff of the matched filter/likelihood "
+                                  "integral will be used.")
```
titodalcanton: I recommend you remove the `default=None` kwarg (not necessary) and add `type=float` to indicate that real values are required. Then you can drop the `float()` casts above.

pycbc/psd/__init__.py
```diff
@@ -306,6 +346,12 @@ def insert_psd_option_group_multi_ifo(parser):
                           help="Measure PSD from the data, using given "
                           "average method. Choose from "
                           "mean, median or median-mean.")
+    psd_options.add_argument("--psd-low-frequency-cutoff", nargs="+",
+                             action=MultiDetOptionAction, metavar='IFO:FREQ',
+                             help="(Optional) The low frequency cutoff for the "
+                                  "PSD. If not specified, the low frequency "
+                                  "cutoff of the matched filter/likelihood "
+                                  "integral will be used.")
```
titodalcanton: Similar comment here, pass the `type=float` kwarg.

pycbc/psd/__init__.py
```diff
@@ -333,6 +379,22 @@ def insert_psd_option_group_multi_ifo(parser):
...
+                                  "PSD.")
+    psd_options.add_argument("--invpsd-trunc-low-freq-fill-value", default=0.,
+                             help="(Optional) Value to set the inverse PSD to "
+                                  "for frequencies below the low frequency "
+                                  "cutoff when applying psd-inverse-length. "
+                                  "Default 0; accepts any float. Also accepts "
+                                  "'fmin'; this sets values below the cutoff "
+                                  "to the inverse PSD value specified by "
+                                  "`psd-low-frequency-cutoff` (or the matched "
+                                  "filter/likelihood integral otherwise).")
```
titodalcanton, line 393: I think these also need the `action=MultiDetOptionAction` and `metavar='IFO:LENGTH'` kwargs, like the nearby arguments.

pycbc/psd/__init__.py
acorreia61201: Addressed this here and elsewhere. There were other options besides those I added that specified `default=None`. I removed those to be consistent; to my knowledge and based on what you pointed out this shouldn't affect other parts of the code

pycbc/psd/estimate.py
cdcapano: No, let's leave it alone then. This would have occurred in the past when we sampled the psd down to zero Hz, and it didn't affect things.

PR #5302, opened 2026-03-16
Allow td-> fd automatic conversion more often, if taper settings are chosen

ahnitz:

This is a new feature

Available waveforms through get_fd_waveform

## Motivation

This allow additional waveforms to work through the get_fd_waveform interface that are auto-converted from td waveforms. This makes additional approximants functional for codes that expect or require this interface.

## Contents

Currently, td-> fd conversion only occurs under certain conditions, e.g. a valid length estimator. However, this enables td-> fd conversion to work where this is not available and the user manually sets the tapering options. 

If you explicilty set the 'taper method' option it will apply the taper as directed and proceed with the conversion. Otherwise, it will use the existing logic which requires the existence of a length estimator to determine the tapering settings. 

## Additional notes

### DIFF
--- a/pycbc/types/timeseries.py
+++ b/pycbc/types/timeseries.py
@@ -19,18 +19,14 @@ Provides a class representing a time series.
 """
 import os as _os
 import h5py
-
-import numpy as _numpy
-from scipy.io.wavfile import write as write_wav
-
 from pycbc.types.array import Array, _convert, complex_same_precision_as, zeros
-from pycbc.types.utils import determine_epoch
 from pycbc.types.array import _nocomplex
 from pycbc.types.frequencyseries import FrequencySeries
 from pycbc.types import float32, float64
-from pycbc.libutils import import_optional
+import lal as _lal
+import numpy as _numpy
+from scipy.io.wavfile import write as write_wav
 
-_lal = import_optional('lal')
 
 class TimeSeries(Array):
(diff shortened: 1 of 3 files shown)


ahnitz:
@cdcapano @GarethCabournDavies Poke

GarethCabournDavies (approved):
Yeah this makes sense, and doesn't look like it would affect anything which doesn't have the taper_method parameter.

yi-fan-wang:
This PR breaks `pycbc_plot_waveform` in the followup page of an offline search. With this PR, a standalone time domain waveform is not supported any more without a length estimator, unless I provide a set of three parameters "taper", "taper_method" and "taper_window". The infrastructure to support these three parameters doesn't exist in `pycbc_plot_waveform`

yi-fan-wang:
The solution can be implementing this PR's idea till the end by adding support for "taper", "taper_method", and "taper_window" to pycbc_plot_waveform. But I have concerns about the idea to convert every time domain waveform to the frequency domain. We should check that it should either have a length estimator or have a valid taper method. But this seems hard to implement in the add_custom_waveform stage, because the conversion is determined without looking at its input parameters.

I'd also like to raise a more general discussion that the middle ground of `pycbc` and waveform libraries should be coordinated, for example in this case of this PR both `pyseobnr` and `pycbc` are trying to do waveform tapering.

PR #5316, opened 2026-04-01
Fixing pygrb coincidences

pannarale:
Fix incomplete and buggy handling of slides in looking for coincidence within PyGRB.

This is a: bug fix.

This change affects: the PyGRB background.

## Motivation
It was noticed that the PyGRB timeslides contain fewer and fewer triggers as the sliding progresses.  When plotted as a function of time, the triggers in the slides were showing gaps that widened with sliding.  Here is a plot of trigger times and slide IDs.

[time_index_vs_slide_id_skygrid.pdf](https://github.com/user-attachments/files/28021403/time_index_vs_slide_id_skygrid.pdf)

## Contents
This needs a fix on two levels as far as I can tell.

1. `pycbc_multi_inspiral` has the `time_delay_idx` dictionary to store the indexing "corrections" needed at a given slide, sky position and detector to convert indices from geocenter to detector or vice versa.  `pycbc_multi_inspiral` checks that the indices, after correction, are in a meaningful range.  However, this is designed thinking of detector-geocenter zerolag delay, while `time_delay_idx` is also going to shift things according to the slide.  Effectively, this does the correct trimming for the first IFO (typically H1), but as the others are timeslid it drops triggers.  The first part of the fix is to perform this check  on `slide=0` (zerolag) so that only index+skyposition combinations that do not suit our purposes are thrown away.

2. At this point the indices may be slid AND wrapped around so that `get_coinc_indices` can convert them to geocenter and look for repeated indices.  In the current implementation we are looking for repeated indices without a wraparound, so coincidences are lost.  The second part of the fix carries out the wrap around in a consistent way.  This is done also in the cythonized function for the 2 IFO case.

In summary, the current implementation throws away too many triggers prior when searching for coincidences and then loses coincidences by not wrapping around time slides.  After the fix, there are no more gaps in a plot like the one shown above; further a couple additional slide IDs can be populated, whereas they were effectively empty prior to the fix.

## Testing performed
Other than the diagnostic plot above being resolved, the p-value distribution of the background in a number of test boxes has smaller uncertainties, indicating that the background is now populated more fully.

## Additional notes

### DIFF
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -638,7 +638,7 @@ with ctx:
     # Loop over templates
     for t_num, template in enumerate(bank):
         # Loop over segments
-        for s_num in range(len(segments[args.instruments[0]]))
+        for s_num in range(len(segments[args.instruments[0]])):
             stilde = {ifo: segments[ifo][s_num] for ifo in args.instruments}
             # Checks the 'inj_filter_rejector' options to determine whether
             # to filter this template/segment if injections are present
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -638,7 +638,7 @@ with ctx:
     # Loop over templates
     for t_num, template in enumerate(bank):
         # Loop over segments
-        for s_num, segment_ifo0 in enumerate(segments[args.instruments[0]]):
+        for s_num in range(len(segments[args.instruments[0]]))
             stilde = {ifo: segments[ifo][s_num] for ifo in args.instruments}
             # Checks the 'inj_filter_rejector' options to determine whether
             # to filter this template/segment if injections are present
--- a/bin/pycbc_multi_inspiral
+++ b/bin/pycbc_multi_inspiral
@@ -638,7 +638,7 @@ with ctx:
     # Loop over templates
     for t_num, template in enumerate(bank):
         # Loop over segments
-        for s_num, stilde in enumerate(segments[args.instruments[0]]):
+        for s_num, segment_ifo0 in enumerate(segments[args.instruments[0]]):
             stilde = {ifo: segments[ifo][s_num] for ifo in args.instruments}
             # Checks the 'inj_filter_rejector' options to determine whether
             # to filter this template/segment if injections are present
@@ -694,6 +694,8 @@ with ctx:
             #   Typically only a tiny fraction will be populated.
             power_chisq_arrays = dict.fromkeys(args.instruments)
             power_chisq_dof_arrays = dict.fromkeys(args.instruments)
+            # For convenience also store the lengths of the snr_dict values
+            wraparound_dict = dict.fromkeys(args.instruments)
             for ifo in args.instruments:
                 logging.debug(
                     "Filtering template %d/%d, ifo %s", t_num + 1, n_bank, ifo
@@ -725,8 +727,9 @@ with ctx:
                 snr_dict[ifo] = (
                     snr_ts[matched_filter[ifo].segments[s_num].analyze] * norm
                 )
+                wraparound_dict[ifo] = len(snr_dict[ifo])
                 assert (
-                    len(snr_dict[ifo]) > 0
+                    wraparound_dict[ifo] > 0
                 ), f'SNR time series for {ifo} is empty'
                 norm_dict[ifo] = norm
                 corr_dict[ifo] = corr.copy()
@@ -734,10 +737,10 @@ with ctx:
                 snr[ifo] = snrv * norm
                 # Initialize the cache for power chi^2
                 power_chisq_arrays[ifo] = np.full(
-                    len(snr_dict[ifo]), np.nan, dtype=np.float32
+                    wraparound_dict[ifo], np.nan, dtype=np.float32
                 )
                 power_chisq_dof_arrays[ifo] = np.zeros(
-                    len(snr_dict[ifo]), dtype=np.int32
+                    wraparound_dict[ifo], dtype=np.int32
                 )
                 del snr_ts, norm, corr, ind, snrv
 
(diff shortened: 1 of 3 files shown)


bin/pycbc_multi_inspiral
```diff
@@ -638,7 +638,7 @@ with ctx:
     # Loop over templates
     for t_num, template in enumerate(bank):
         # Loop over segments
-        for s_num, stilde in enumerate(segments[args.instruments[0]]):
+        for s_num, segment_ifo0 in enumerate(segments[args.instruments[0]]):
```
pannarale: This was harmless because **this** `stilde` was never used, but I am against having variables with the same name in different scopes.

titodalcanton: Use `_` if the variable is never used.

titodalcanton: …or even better, iterate over `range(len(segments[args.instruments[0]]))`

titodalcanton:
@pannarale is this ready for me to review it, or should I wait for more tests/changes?

pannarale:
Hi @titodalcanton, I edited the description and removed the WIP label.  Also, there were two FIXMEs I had noted down that now got their dedicated issue for posterity rather than being buried as comments in `pycbc_multi_inspiral` and forgotten.

bin/pycbc_multi_inspiral
```diff
@@ -694,6 +694,8 @@ with ctx:
             #   Typically only a tiny fraction will be populated.
             power_chisq_arrays = dict.fromkeys(args.instruments)
             power_chisq_dof_arrays = dict.fromkeys(args.instruments)
+            # For convenience also store the lengths of the snr_dict values
+            wraparound_dict = dict.fromkeys(args.instruments)
```
titodalcanton, line 698: Can `snr_dict` actually have a different length for different detectors? Do we intend to support that scenario? If the answer is yes, all good. Otherwise, it may be best to make `wraparound_dict` an integer variable that is initially set to None, then set to `len(snr_dict)` for the first detector, and for the remaining detectors is used to assert that the length remains unchanged.

bin/pycbc_multi_inspiral
```diff
@@ -756,22 +759,28 @@ with ctx:
...
+                    # store trigger indices list in a dictionary; when there
+                    # are no triggers, the dictionary is empty. Indices are
+                    # kept only if they remain within the time being processed
+                    # after the time delay correction from detector to
+                    # geocenter due to sky location is applied, i.e., require
+                    # them to be between 0 and len(snr_dict[ifo]).
                     # idx[ifo] - time_delay_idx[slide][position_index][ifo]
-                    # (which applies the necessary time shift by operating on
-                    # indices) to be between 0 and len(snr_dict[ifo]).
+                    # operates on indices to apply a time shift from detctor
```
titodalcanton, line 770: Should the code line above be edited to replace `slide` with 0?

bin/pycbc_multi_inspiral
```diff
@@ -782,27 +791,30 @@ with ctx:
...
-                    # tracks (at the geocent) triggers above threshold in
+                    # tracks (at the detectors) triggers above threshold in
                     # individual detectors. If only one IFO is available, then
                     # the method returns trigger indices for that IFO. Given
                     # that we are looping over timeslides and sky positions,
                     # the appropriate dictionary of time delays wrt the
-                    # geocenter must be passed to operate at the geocenter.
+                    # geocenter must be passed to operate at the geocenter,
+                    # as well as the dictionary to enable the wrap around at
+                    # each detecotor prior to searching for coincidences.
```
titodalcanton:
```suggestion
                    # each detector prior to searching for coincidences.
```

bin/pycbc_multi_inspiral
```diff
@@ -1122,8 +1135,11 @@ with ctx:
                         network_out_vals['my_network_chisq'] = np.real(
                             network_chisq_dict
                         )
+                        # coinc_idx has the geocenter indices of triggers.
+                        # The wrap around happened in get_coinc_indices.
                         network_out_vals['time_index'] = (
-                            coinc_idx + stilde[ifo].cumulative_index
+                            coinc_idx
+                            + stilde[args.instruments[0]].cumulative_index
```
titodalcanton, line 1143: This change should have no effect if I understand correctly, but the older code looks simpler. What motivates this?

pycbc/events/eventmgr_cython.pyx
titodalcanton, line 1: The change to this file seems to change the scaling of the coinc cost from N1 + N2 to N1 N2. Of course correctness is the priority here, but I would like another round of profiling at some point soon to make sure we are not going backwards in terms of performance.

titodalcanton (changes requested):
I mostly have questions at this point, with a disclaimer that I still find this part of the algorithm very non-obvious and hard to review, and I wonder if it even _can_ be implemented in a way that is easy to review.

Also note that the tests are failing due to an unrelated (and nontrivial) problem, which we are looking into.

bin/pycbc_multi_inspiral
titodalcanton: And `wraparound_len` would be a better name in the latter case.

bin/pycbc_multi_inspiral
pannarale: It was meant to explain the sense of `idx[ifo] - time_delay_idx[slide][position_index][ifo]`, so that the need to set `slide` to 0 below would be clear ("The trimming here is carried out on the zerolag indices, i.e., setting slide to 0").  Let me add a clause and some words to qualify my intention.

bin/pycbc_multi_inspiral
pannarale: As far as I understand, they should have the same length, but everything IFO-related is handled as a dictionary so I stuck to the same spirit/style with the wraparound.  I also kept the switch from ` len(snr_dict[ifo])` minimal, but we can try this change you propose.

bin/pycbc_multi_inspiral
pannarale: The loop on `ifo` ends about 15 lines above, so using `args.instruments[0]` is a cleaner/safer keyword choice than "whatever `ifo` represents at this point."

pycbc/events/eventmgr_cython.pyx
pannarale: Yes, I definitely hurt the scaling here, but the old approach was not catching everything and I opted for correctness over speed.  Do we open a new issue or expand the purpose of https://github.com/gwastro/pycbc/issues/5336?

bin/pycbc_multi_inspiral
pannarale: This is emulates the call made for `network_out_vals['coherent_snr']` a few lines above.

titodalcanton: Ah! Good point, I had missed the variable reuse. All good.

pycbc/events/eventmgr_cython.pyx
titodalcanton: As you prefer.

bin/pycbc_multi_inspiral
titodalcanton: Let's keep it like this for now.

titodalcanton (approved):
Happy to merge, with two caveats:
* For correctness, I am relying heavily on the testing you have done. I am not really able to see that the code is correct just by following through it.
* You will need to rebase when we have fixed the broken CI (probably with #5341).

pycbc/events/eventmgr_cython.pyx
pannarale: I extended the issue to cover this.

titodalcanton:
Please rebase on the current master to get the tests to pass.

PR #5318, opened 2026-04-08
Fixed rounding bug in pycbc_dtphase

SamInsley:
When binning samples into integer bins, the current code truncates the samples rather than consistently rounding. This results in the 0 bin having approximately twice as many samples as the others. This fix will round the samples to the nearest integer instead.

This is a bug fix regarding how samples are rounded.

This change changes:  scientific output

This change:  follows style guidelines (See e.g. [PEP8](https://peps.python.org/pep-0008/)), has been proposed using the [contribution guidelines](https://github.com/gwastro/pycbc/blob/master/CONTRIBUTING.md)

This change will: it should not affect anything other than slightly changing the weights of different bins in the PTA histograms.

## Motivation

The current version, truncates samples resulting in all samples between -1 and 1 being sorted into the zero bin. This causes an over-density in this bin compared to others.

## Contents

Rather than truncating the samples, this simply rounds the samples to the nearest integer.

## Testing performed

I have recreated the histogram files with this change and it works the same just without this over density at 0.

## Additional notes

### DIFF
--- a/bin/all_sky_search/pycbc_dtphase
+++ b/bin/all_sky_search/pycbc_dtphase
@@ -187,9 +187,9 @@ for ifo0 in args.ifos:
             dt = (data[ifo0]['t'] - data[ifo1]['t'])
             dp = (data[ifo0]['p'] - data[ifo1]['p']) % (2. * np.pi)
             sr = (data[ifo1]['s'] / data[ifo0]['s'])
-            dtbin = (dt / twidth).astype(int)
-            dpbin = (dp / pwidth).astype(int)
-            srbin = (sr / swidth).astype(int)
+            dtbin = np.rint(dt / twidth).astype(int)
+            dpbin = np.rint(dp / pwidth).astype(int)
+            srbin = np.rint(sr / swidth).astype(int)
 
             # We'll only store a limited range of ratios
             if keep is None:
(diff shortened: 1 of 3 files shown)


titodalcanton:
Thanks @SamInsley! Could you please post plots of the histograms before/after the fix?

SamInsley:
@titodalcanton 
PTA_histogram

ahnitz:
@titodalcanton @SamInsley If you make this change, the application should naturally also have a symmetric change, no? 

e.g. Do you not also need to test the application here https://github.com/gwastro/pycbc/blob/master/pycbc/events/eventmgr_cython.pyx#L39?

ahnitz (changes requested):
See comment, I suspect you need to test a symmetric change on the application side.

SamInsley:
@ahnitz Thanks! The fix also did need to be applied to where we bin the triggers. I have done this and it seems to be working fine in my personal tests, but these automatic tests are still failing and I am not sure why. I don't know if I am understanding it right but do these tests check if the stat values are consistent across both versions, as this fix will definitely have an affect on the measured stat values in some parts of the parameter space.

SamInsley:
@titodalcanton @ahnitz. I believe this is failing the tests because of `test_live_coinc_compare.py`. Since it will use pre-generated histogram files, the output density values will not be the same with this change.

PR #5319, opened 2026-04-08
Remove temporary Git dependency from companion.txt

GarethCabournDavies:
Remove temporary Git dependency for PyPMC - test revert #5292 now that PyPMC has had a release


titodalcanton (approved):
The CI failures look unrelated to this change since I have seen those in other PRs as well. Approving.

PR #5321, opened 2026-04-08
Redshift waveforms: check the epoch for both td and fd unit test

cdcapano:

PR #5307 introduced a utility to redshift waveforms, and associated unit tests. In the unit test a BBH waveform is generated in the source frame and redshifted to the detector frame, then compared to a one generated in the detector frame using the redshifted masses. This is done for both frequency domain and time domain waveforms.

When doing the comparison, the epochs of the two can differ slightly due to numerical precision. This can randomly cause the unit test to fail. Currently, the frequency domain test accounts for this by letting the epochs differ by up to 0.001/sample_rate. The same should have been done for the time domain test. This PR fixes that, to stop the time domain test from randomly failing.

### DIFF
--- a/test/test_waveform_utils.py
+++ b/test/test_waveform_utils.py
@@ -139,6 +139,16 @@ class TestRedshiftWaveform(unittest.TestCase):
         return numpy.linalg.norm(test - ref) / numpy.linalg.norm(ref)
 
 
+    def _check_epochs(self, redshifted_hp, det_hp, err=0.001):
+        """Checks that the epochs of the two waveforms are close enough.
+
+        Small differences in the epochs may arise due to floating point
+        errors. That may cause a failure when we compute the relative L2
+        error, so we'll check that the epochs are close enough.
+        """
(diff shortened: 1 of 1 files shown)


ahnitz:
@cdcapano Why does this happen at all? If the epochs actually should be the same, shouldn't that be explicitly true? Is this because you are switching between time / frequency domain a few times?

cdcapano:
@ahnitz No, it's unrelated to switching between domains. I think it's because the two methods are taking vastly different routes to produce the waveform and start time. The redshift code is determining the epoch by generating the source frame waveform, applying the redshift, then correcting the epoch of the shifted waveform by the amount the start time has been shifted backward (see [here](https://github.com/gwastro/pycbc/blob/master/pycbc/waveform/utils.py#L518-L521)). The detector-frame waveform's epoch is just whatever the waveform generator returns given the redshifted masses. It's not surprising that those two would not be exactly the same down to floating point precision.

cdcapano:
@titodalcanton @ahnitz Could one of you review? Probably good to get this merged soon so as not to break others' PRs.

ahnitz (approved):

PR #5322, opened 2026-04-09
Updated gracedb instances to use kafka

johannferns:
This PR modifies gracedb uploads to use kafka. It modifies all scripts that use the gracedb instance for event uploads and primarily  affects the live search

Changes include the use of a producer that is fed to the create_event module for uploads and an additional key to the GraceDb instance referring to the api_version. These kafka changes require the latest version of [ligo-gracedb](https://pypi.org/project/ligo-gracedb/#history) which is v2.15.7 as of the creation of this PR. Additionally, the kafka extension needs to be installed as well, which can be done using `pip install "ligo-gracedb[kafka]"`

The producer script requires a `bootstrap_servers` argument that is hardcoded to be `kafka-dev.ligo.org:9092` as stated in https://ligo-gracedb.readthedocs.io/en/gracedb-2.15.7/kafka.html

Tests were made by uploading xml files to the gracedb-test server.  [G1637568](https://gracedb-test.ligo.org/events/G1637568/view/) was uploaded using the pycbc_upload_single_event_to_gracedb executable, while [G1637613](https://gracedb-test.ligo.org/events/G1637613/view/) was uploaded using the upload snippet from `pycbc.io.gracedb` (as a result some log messages may not be present since the upload script was the only thing that was run).  Note that by default, these scripts upload events to gracedb-playground if they are not using the production server. The gracedb-playground server currently does not support kafka based uploads.


bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -68,28 +70,42 @@ pycbc.init_logging(args.verbose, default_level=1)
...
+    gracedb = GraceDb(api_version='v2')
+
 else:
-    gracedb = GraceDb(service_url='https://gracedb-playground.ligo.org/api/')
+    gracedb = GraceDb(service_url='https://gracedb-playground.ligo.org/api/', api_version='v2')
 
 labels = [l.upper() for l in (args.labels or [])]
 allowed_labels = gracedb.allowed_labels
 
+print(gracedb._service_url)
```
titodalcanton, line 81: Remove debug prints please.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -68,28 +70,42 @@ pycbc.init_logging(args.verbose, default_level=1)
...
     err_msg += f"{','.join(allowed_labels)}."
     raise RuntimeError(err_msg)
 
 group_tag = 'Test' if args.testing else 'CBC'
+
+
+logging.info("Creating Kafka producer...")
+ca_bundle = certifi.where()
+producer = GraceDbKafkaProducer(
+    bootstrap_servers='kafka-dev.ligo.org:9092',
```
titodalcanton, line 95: I do not like hardcoding URLs like this (I have the same comment above but that predates this PR). This must be a user choice via the CLI.

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -113,9 +115,18 @@ args = parser.parse_args()
...
-    gracedb = GraceDb(service_url='https://gracedb-playground.ligo.org/api/')
+    gracedb = GraceDb(service_url='https://gracedb-playground.ligo.org/api/', api_version='v2')
+
+logging.info("Creating Kafka producer...")
+ca_bundle = certifi.where()
+producer = GraceDbKafkaProducer(
+    bootstrap_servers='kafka-dev.ligo.org:9092',
+    service_url=gracedb._service_url,
+    ca_cert_path=ca_bundle,
+)
```
titodalcanton, line 128: This is adding quite a bit of duplicated code. Please try to create a function or class in pycbc.io.gracedb instead that encapsulates this.

pycbc/io/gracedb.py
```diff
@@ -405,12 +407,22 @@ def upload(self, fname, gracedb_server=None, testing=True,
...
         try:
+            from ligo.gracedb.kafka import GraceDbKafkaProducer
+            logger.info("Creating Kafka producer...")
+            ca_bundle = certifi.where()
+        
+            producer = GraceDbKafkaProducer(
+                bootstrap_servers='kafka-dev.ligo.org:9092',
+                service_url=self.gracedb._service_url,
+                ca_cert_path=ca_bundle,
+            )
```
titodalcanton, line 418: Would this impact latency? How long would this call take? Is there a way to create this at the start, and reuse the `producer` objects for every upload?

titodalcanton (changes requested):
Thanks @johannferns. I have some concerns, see below.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
johannferns: I have made changes allowing the use to choose the `bootstrap_servers` argument and I have set 'kafka-dev.ligo.org:9092' as the default value in case the argument is not provided.

bin/pycbc_upload_xml_to_gracedb
johannferns: The scripts of `bin/pycbc_upload_xml_to_gracedb` and `bin/all_sky_search/pycbc_upload_single_event_to_gracedb` are independent of any modules in `pycbc.io.gracedb` and thus I have opted to place the producer script in those files itself.

pycbc/io/gracedb.py
johannferns: I tested this using timeit and it takes a second to create the producer. However, I refactored the pycbc_live script to create the producer in the create_gdb model so that the producer can be used through the full session. I also tested if the producer was robust when uploaded using threading and it indeed is.

titodalcanton:
@johannferns can you please rebase on master, so we have a fresh run of the tests?

bin/pycbc_live
```diff
@@ -466,18 +468,30 @@ class LiveEventManager(object):
...
             gdbargs['service_url'] = self.gracedb_server
         self.gracedb = GraceDb(**gdbargs)
 
+        logging.info('Creating Kafka producer')
+        from ligo.gracedb.kafka import GraceDbKafkaProducer
+
+        ca_bundle = certifi.where()
+        self.producer = GraceDbKafkaProducer(
+             bootstrap_servers=self.bootstrap_servers,
+             service_url=gracedb._service_url,
```
titodalcanton: Should this be `self.gracedb._service_url` instead?

bin/pycbc_live
```diff
@@ -466,18 +468,30 @@ class LiveEventManager(object):
...
         # Because we do not change any of the request session values when running the
         # code, it should remain thread safe.
-        gdbargs = {'reload_certificate': True, 'reload_buffer': 300}
+        gdbargs = {'reload_certificate': True, 'reload_buffer': 300, 'api_version': 'v2'}
         if self.gracedb_server:
             gdbargs['service_url'] = self.gracedb_server
         self.gracedb = GraceDb(**gdbargs)
 
+        logging.info('Creating Kafka producer')
+        from ligo.gracedb.kafka import GraceDbKafkaProducer
```
titodalcanton: Move this import at the beginning of the method, next to the existing GraceDb import.

bin/pycbc_live
```diff
@@ -993,6 +1009,8 @@ parser.add_argument('--enable-single-detector-upload', action='store_true', defa
 parser.add_argument('--gracedb-server', metavar='URL',
                     help='URL of GraceDB server API for uploading events. '
                          'If not provided, the default URL is used.')
+parser.add_argument('--bootstrap-servers', default='kafka-dev.ligo.org:9092',
```
titodalcanton: Do not use a default value here. This must be given in the configuration.

bin/pycbc_make_skymap
```diff
@@ -565,6 +566,8 @@ if __name__ == '__main__':
                              'production analyses!')
     parser.add_argument('--gracedb-server', metavar='URL',
                         help='URL of GraceDB server API for uploading events.')
+    parser.add_argument('--bootstrap-servers', default='kafka-dev.ligo.org:9092',
```
titodalcanton: Remove default.

bin/all_sky_search/pycbc_upload_single_event_to_gracedb
```diff
@@ -40,6 +42,8 @@ parser.add_argument('--testing', action="store_true", default=False,
 parser.add_argument('--production-server', action="store_true", default=False,
                     help="Upload event to production graceDB. If not given "
                          "events will be uploaded to playground server.")
+parser.add_argument('--bootstrap-servers', default='kafka-dev.ligo.org:9092',
```
titodalcanton: Remove the default value, this has to be indicated in the configuration.

bin/pycbc_optimize_snr
```diff
@@ -52,6 +52,8 @@ parser.add_argument('--chirp-time-window', type=float, default=2.,
 parser.add_argument('--gracedb-server', metavar='URL',
                     help='URL of GraceDB server API for uploading events. '
                          'If not provided, the default URL is used.')
+parser.add_argument('--bootstrap-servers', default='kafka-dev.ligo.org:9092',
```
titodalcanton: Remove default.

bin/pycbc_upload_xml_to_gracedb
```diff
@@ -82,6 +84,8 @@ parser.add_argument('--min-ifar', type=float, metavar='YEARS',
 parser.add_argument('--production-server', action="store_true", default=False,
                     help="Upload event to production graceDB. If not given "
                          "events will be uploaded to playground server.")
+parser.add_argument('--bootstrap-servers', default='kafka-dev.ligo.org:9092',
```
titodalcanton: Remove default.

pycbc/io/.ipynb_checkpoints/gracedb-checkpoint.py
titodalcanton, line 1: Remove this file.

pycbc/io/gracedb.py
```diff
@@ -343,7 +346,7 @@ def save(self, fname):
                           pastrof)
             logger.info('P_astro file saved as %s', self.pastro_file)
 
-    def upload(self, fname, gracedb_server=None, testing=True,
+    def upload(self, fname, gracedb_server=None, bootstrap_servers='kafka-dev.ligo.org:9092', testing=True,
```
titodalcanton: Set the default to None here.

titodalcanton (changes requested):
New round of comments. I remain concerned by the amount of code duplication, which now makes it very hard to maintain how we interface to GraceDB (notably the code for adding the `--bootstrap-servers` option, which has to be repeated 5 times). I really think it should be possible to make this tidier by leveraging the pycbc.io.gracedb module more, at least by having a `insert_gracedb_options()` function like we do in other modules.

bin/pycbc_live
```diff
@@ -990,9 +1004,6 @@ parser.add_argument('--enable-production-gracedb-upload', action='store_true', d
                          'production analyses!')
 parser.add_argument('--enable-single-detector-upload', action='store_true', default=False,
                     help='Upload single ifo events to GraceDB')
-parser.add_argument('--gracedb-server', metavar='URL',
-                    help='URL of GraceDB server API for uploading events. '
-                         'If not provided, the default URL is used.')
```
titodalcanton, line 995: This option is being removed, but it is not being added back in `insert_gracedb_option_group()`. Is this intentional? I see some of the executable scripts hardcode the GraceDB service URL, so maybe that is the reason. It may be a good idea to simply avoid hardcoding URLs anywhere, and having this option in `insert_gracedb_option_group()`, but maybe that is a bigger discussion than this PR. For now, just put this back here.

pycbc/io/gracedb.py
```diff
@@ -35,6 +36,13 @@ def _single_value(value):
     return numpy.asarray(value).item()
 
 
+def insert_gracedb_option_group(parser):
+    """Adds the options to choose config options for uploading to a gracedb server"""
```
titodalcanton, line 40:
```suggestion
    """Adds command-line options to configure access to a GraceDB server."""
```

titodalcanton (changes requested):
This looks much cleaner, thanks. I quickly went over the GraceDB documentation regarding the new Kafka interface, and this looks mostly ok now, apart for what looks like an oversight to fix (see below).

Still, going through all the Kafka cerimony for offline uploads (`pycbc_upload_single_event_to_gracedb`, `pycbc_make_skymap`, `pycbc_upload_xml_to_gracedb`) seems overkill, and I would like to confirm that is really necessary. The [docs](https://ligo-gracedb.readthedocs.io/en/latest/api.html#ligo.gracedb.rest.GraceDb.create_event) seem to imply it is not.

Also, having to import certifi, call `where()`, and indicate the API version in multiple places is a lot of annoying boilerplate. But in the interest of getting Live ready soon, I propose we clean that up in a later PR.

titodalcanton:
Also, I think a rebase on master would be good to pick up recent changes to the tests.

titodalcanton:
Genuine failures from the tests. One seems trivial:
```
  File "<frozen runpy>", line 203, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/runner/work/pycbc/pycbc/.pixi/envs/search/lib/python3.13/site-packages/mpi4py/__main__.py", line 8, in <module>
    main()
    ~~~~^^
  File "/home/runner/work/pycbc/pycbc/.pixi/envs/search/lib/python3.13/site-packages/mpi4py/run.py", line 320, in main
    run_command_line(args)
    ~~~~~~~~~~~~~~~~^^^^^^
  File "/home/runner/work/pycbc/pycbc/.pixi/envs/search/lib/python3.13/site-packages/mpi4py/run.py", line 53, in run_command_line
    run_path(sys.argv[0], run_name="__main__")
    ~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen runpy>", line 294, in run_path
  File "<frozen runpy>", line 98, in _run_module_code
  File "<frozen runpy>", line 88, in _run_code
  File "/home/runner/work/pycbc/pycbc/.pixi/envs/search/bin/pycbc_live", line 1392, in <module>
    evnt.check_coincs(list(results.keys()), best_coinc, psds)
    ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/work/pycbc/pycbc/.pixi/envs/search/bin/pycbc_live", line 542, in check_coincs
    producer=self.producer,
             ^^^^^^^^^^^^^
AttributeError: 'LiveEventManager' object has no attribute 'producer'
```
The other not so much:
```
ValueError: Cannot parse hostname from URL: 'localhost/'
```
Basically the new Kafka client does not seem to like the fact that we are pretending we have a GraceDB server listening on localhost. Not sure how to solve this… Maybe we just have to stop doing that?

